PR #8, opened 2016-10-15
Development requirements and tox automation

narzeja:
This PR adds a requirements-dev.txt for easier development. It adds missing dependencies for running tests.
It also adds tox configurations so you can simply write `tox` in the project root and it will run tests and generate a coverage report.


narzeja:
Closed because I had to make a PR from a feature branch at my own repo instead of master.

PR #18, opened 2016-10-15
Speedup response

jpiasetz:
I took a crack at seeing if I could speed up the response. I think using printf style string formatting should help by saving the join and some encoding.


channelcat:
This is awesome!  Just got 30,709 req/sec!

PR #23, opened 2016-10-15
refactored response object model

sm5art:
I want to change how the Response model works by creating an object that allows intermediary header changing, and overall response changing. 

For the future of this framework, the response should be modified so that more complicated responses like cookies can be used.

i.e
```python
class Response:
    pass
    
    def set_cookie(self, name, value):
         self.add_headers("Cookie":"{}={}".format(name,value))
```
which will be used like this

```python
@app.route("/")
def handler(req, res):
    res.set_cookie("hello", "world")
    return res.json({"hello":"world")
```
where res is instantiated into the handler by the main sanic.py.

This type of structure will allow better response formulation rather than using the bare response types and pushing in the headers manually. We can write stuff like cookies that will be reused.

My next PR will be implementing cookies. 

This is experimental, I just want to know what you think

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,15 +20,16 @@ All tests were run on a AWS medium instance running ubuntu, using 1 process.  Ea
 
 ```python
 from sanic import Sanic
-from sanic.response import json
 
 app = Sanic(__name__)
 
 @app.route("/")
-async def test(request):
-    return json({ "hello": "world" })
+async def test(req, res):
+    res.add_headers({"hello":"world"})
+    return res.json({ "hello": "world" })
 
 app.run(host="0.0.0.0", port=8000)
+
 ```
 
 ## Installation
@@ -58,7 +59,7 @@ app.run(host="0.0.0.0", port=8000)
                      ▄▄▄▄▄
             ▀▀▀██████▄▄▄       _______________
           ▄▄▄▄▄  █████████▄  /                 \
-         ▀▀▀▀█████▌ ▀▐▄ ▀▐█ |   Gotta go fast!  | 
+         ▀▀▀▀█████▌ ▀▐▄ ▀▐█ |   Gotta go fast!  |
        ▀▀█████▄▄ ▀██████▄██ | _________________/
        ▀▄▄▄▄▄  ▀▀█▄▀█════█▀ |/
             ▀▀▀▄  ▀▀███ ▀       ▄▄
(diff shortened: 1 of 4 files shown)


sm5art:
bump

channelcat:
As much as I dislike adding more arguments to the request handlers, this makes sense and makes middleware much more useful.  I'll merge it in :)

sm5art:
ill resolve merge conflicts

channelcat:
I can get them - I'm adding in response as an argument for middleware and exceptions, and adding tests

channelcat:
Alright, I put the changes in the sm5art-cookies branch.  Overall the performance dipped by 1kreq/sec.  I'd like to hold off until we can figure out why it slowed down and fix it.

from sm5art-cookies branch:
Test main:
python3 tests/performance/sanic/simple_server_pip.py 8000
Test new:
python3 tests/performance/sanic/simple_server.py 8000

Benchmark:
wrk -c 100 -t 1 http://127.0.0.1:8000/
Main Requests/sec:  28805.95
New Requests/sec:  27756.66

channelcat:
Thinking of an alternative pattern, just in case:

```python
@app.middleware('request')
async def some_auth_middleware(request):
    # Some auth logic
    request._set_the_cookies = True

@app.middleware('response')
async def some_auth_middleware(request, response):
    if request._set_the_cookies:
        response.cookies['my-cookie'] = 'my-value'
```

sm5art:
ill take a look into it

sm5art:
i like your idea of using middleware to use the new response object.
I think that the performance issues comes from creating an extra response object with each handler

davidbgk:
@sm5art maybe using `__slots__` on your `Response` class might speed up the creation?

sm5art:
i think using python classes will be too slow. will look into different methods of structuring this.

davidbgk:
I use to have good perfs inheriting from namedtuples, not sure if it's pertinent here though.

sm5art:
hmm this is weird, i ran the benchmark: the sm5art-cookies branched had Requests/sec:  36208.95

and the master branch had Requests/sec:   3219.30. 

Something is broken in the master branch

sm5art:
22 ms average latency, on master compared to 3 ms average latency on sm5art-cookies, let me diff the two branches and see what changed in the server logic.

sm5art:
the blueprint feature merge broke latency, tried a benchmark on this commit f2757bc0743f83ea9c12bd936644c936ca61090f, i got Requests/sec:  38994.25

sm5art:
im going to close this pr, and work on making a middleware version

PR #39, opened 2016-10-16
Fix flake8 errors

seemethere:
Makes it so the files are actually under 80 char line length.

Fixes some other areas where exception handling was not being logged where it would be useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,13 @@
 language: python
 python:
-- '3.5'
+  - '3.5'
 install:
-- pip install -r requirements.txt
-- pip install -r requirements-dev.txt
-- python setup.py install
-- pip install flake8
-- pip install pytest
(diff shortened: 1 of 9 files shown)


channelcat:
I'm new to the open source community so I'm not sure if this is a hard rule, but I feel like the code is more readable if function calls are allowed up to 120 characters.  Is 80 characters something the python open source community still agrees on?

I'm sure you spent a chunk of time on this, but I ask since pep8 is 15 years old and monitors are much wider now.

seemethere:
@channelcat Every professional project I've worked on has used pep8 with a line length of 80 but most open source projects don't end up there because they don't start with it. I think it makes it easier when editing especially if you have 2 files open on a single 1080p screen. 

Just my two cents I think it could be good for a young project like this to adopt it earlier rather than later.

channelcat:
Alright, the math checks out.  Merging :).

Issue #50, opened 2016-10-17
Consider allowing a string response to be inferred as `text(..)`


blakev:
Instead of saying `return text(..)` in a route, allow `return 'my text'` by inferring the response type in `write_response`.

**sanic/server.py**:

``` python
class HttpProtocol
    ...
    ...
    def write_response(self, response):
        try:
            keep_alive = self.parser.should_keep_alive() and not self.signal.stopped

            try:
                self.transport.write(response.output(self.request.version, keep_alive, self.request_timeout))

            except AttributeError:
                # attempt to send the response as a text response
                if isinstance(response, str):
                    response = text(response)
                    self.transport.write(response.output(self.request.version, keep_alive, self.request_timeout))
                else:
                    raise

            if not keep_alive:
                self.transport.close()
            else:
                self.cleanup()
        except Exception as e:
            self.bail_out("Writing request failed, connection closed {}".format(e))
```

seemethere:
I like this, on another note we should also be allowed to return a dict and it automatically converts to json

eseglem:
I think explicit is probably better the implicit. Does this change any of the benchmarks? If it slows things at all, may not be worth it.

rzwiefel:
@eseglem From an outside opinion, I would say that you are still allowed to do it by explicitly using a HTTPResponse. And the terseness of not having to put redundant information about the response is a huge usability feature.

Also while I don't believe the two isinstance checks will slow performance, if it is a concern there could be an initial isinstance check for HTTPResponse to short circuit the other checks 
(regarding https://github.com/channelcat/sanic/pull/52)
```if isinstance(response, HTTPResponse):
    pass```

eseglem:
Personally I think it should only be explicit. It should be on the user to know what they are returning. They should know if you are returning text() or json(), and put it on the return. I do not thing it detracts from the usability at all. Even putting in the `isinstance(response, HTTPResponse)` first adds an additional call to isinstance which should not be needed at all. It gets added to every single return, meaning (based on the benchmarks) 30,000 calls to isinstance / second, which just seems like unnecessary overhead for something that should already be known by the user. And if they do not know what type it is, they can do an if before the return. Just personally it seems like an unnecessary feature when the number one purpose of the system (to me at least) is to go fast. Anything that slows the system down, even by a tiny fraction, isn't worth it to me.

[...]

PR #52, opened 2016-10-17
Adds smart handling of response objects

seemethere:
Returning dicts in handlers will now convert them to json and returning anything other than an HTTPResponse will return its `str` representation. 

In response to #50 

# Old
```python
from sanic import Sanic
app = Sanic('test_json')

@app.route('/')
async def handler(request):
    return json({"test": True})
```

# New
```python
from sanic import Sanic
app = Sanic('test_json')

@app.route('/')
async def handler(request):
    return {"test": True}
```

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -11,6 +11,7 @@ except ImportError:
 
 from .log import log
 from .request import Request
+from .response import HTTPResponse, json, text
 
 
 class Signal:
@@ -124,6 +125,10 @@ class HttpProtocol(asyncio.Protocol):
         try:
             keep_alive = all(
                 [self.parser.should_keep_alive(), self.signal.stopped])
+            if isinstance(response, dict):
+                response = json(response)
+            elif not isinstance(response, HTTPResponse):
+                response = text(response)
             self.transport.write(
                 response.output(
                     self.request.version, keep_alive, self.request_timeout))
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -131,10 +131,11 @@ class HttpProtocol(asyncio.Protocol):
         try:
             keep_alive = all(
                 [self.parser.should_keep_alive(), self.signal.stopped])
-            # Handling for returning dicts and everything not an HTTPResponse
-            if isinstance(response, dict):
+            if isinstance(response, HTTPResponse):  # Easy pass first
+                pass
+            elif isinstance(response, (dict, list)):
                 response = json(response)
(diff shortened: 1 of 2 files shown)


sanic/server.py
```diff
@@ -121,9 +122,20 @@ def on_message_complete(self):
     # -------------------------------------------- #
 
     def write_response(self, response):
+        """Attempts to write the resposne to the transport
+
+        This will attempt to smartly handle any type of python object with
```
rzwiefel: Might sound a little better if you say something like ...`attempt to intelligently handle`...

rzwiefel:
Sorry for the pedantic little comments. They just caught my eye 🙂

seemethere:
@rzwiefel No problem! I agree with the comments!

channelcat:
I really like this.  If it doesn't slow the benchmarks I'd like to merge it :).  A few questions though:
1. How do we want to handle a response of None?
2. Should a string response be html or text by default?
3. IMO response type conversion should be the concern of the framework and in sanic.py.  If the http server were ever distributed on its own, this would impose the type conversion on other frameworks.  What are your thoughts?

seemethere:
@channelcat would you be talking about here? 
https://github.com/channelcat/sanic/blob/master/sanic/sanic.py#L140

channelcat:
@seemethere yep!  It did slow things down a little bit so I'd like to keep discussing - gonna post in https://github.com/channelcat/sanic/issues/50

seemethere:
@channelcat 

I'm not seeing a remarkable slow down when I run `wrk`

I'll close this pull request for now while we discuss the feature in #50.

Results:
```shell
bash-3.2$ wrk -c 100 http://0.0.0.0:8000  # With smart handling
Running 10s test @ http://0.0.0.0:8000
  2 threads and 100 connections
    Latency    34.79ms   30.90ms 282.59ms   93.58%
    Req/Sec     1.28k   753.04     4.29k    79.84%
  16434 requests in 10.10s, 1.85MB read
Requests/sec:   1626.61
Transfer/sec:    187.44KB
bash-3.2$ wrk -c 100 http://0.0.0.0:8000  # Without smart handling
Running 10s test @ http://0.0.0.0:8000
  2 threads and 100 connections
    Latency    33.09ms   31.34ms 301.59ms   94.33%
    Req/Sec     1.29k   724.03     4.08k    80.16%
  16444 requests in 10.04s, 1.85MB read
Requests/sec:   1637.13
Transfer/sec:    188.65KB
```

PR #54, opened 2016-10-17
Removes module-level logging override

mykolasmith:
In general, I think it is bad practice for a module to override a user's logging.basicConfig(). It's not common to see libraries wrapping the python `logging` module. Use the builtin module and allow the user to define their own `format`, `level`, etc.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -1,10 +1,9 @@
+import logging
 from cgi import parse_header
 from collections import namedtuple
 from httptools import parse_url
 from urllib.parse import parse_qs
 from ujson import loads as json_loads
-
-from .log import log
 
 
 class RequestParameters(dict):
@@ -82,7 +81,7 @@ class Request:
                     self.parsed_form, self.parsed_files = (
                         parse_multipart_form(self.body, boundary))
             except Exception as e:
-                log.exception(e)
+                logging.exception(e)
                 pass
 
         return self.parsed_form
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -81,7 +81,7 @@ class Request:
                     self.parsed_form, self.parsed_files = (
                         parse_multipart_form(self.body, boundary))
             except Exception as e:
-                logging.exception(e)
+                logging.exception("An error occurred while parsing the request's form body")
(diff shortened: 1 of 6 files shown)


sanic/request.py
```diff
@@ -82,7 +81,7 @@ def form(self):
                     self.parsed_form, self.parsed_files = (
                         parse_multipart_form(self.body, boundary))
             except Exception as e:
-                log.exception(e)
+                logging.exception(e)
```
psykzz: This would actually print the exception twice, should ideally have a more descriptive reason as to what is happening in the exception.

channelcat:
I feel it's pretty important to print out info by default so that users can see the server started, but I don't want to override their log settings.  Is there a way to do that without setting the basic config or loggers?

santagada:
I found out that waitress does the same as sanic and even have a comment saying that if logging has already being configured it is ok:
https://github.com/Pylons/waitress/blob/7e807c9ecefb2388d1e696c3253085d674ac4590/waitress/__init__.py#L9

flask also does something similar, removing all handlers:
https://github.com/pallets/flask/blob/0664a3f2ddfc227797b5a8f0654e9959b15d9e64/flask/logging.py

I think for a web server it makes sense to be configuring logging specially if you can override it by configuring logging before.

mykolasmith:
@channelcat The problem is that you have the logging.basicConfig() in the global block scope of sanic.log, so every time sanic.log gets imported in the application runtime, the basicConfig() gets called again (as all code on the global scope does on import) and overwrites whatever you had in place.

mykolasmith:
If you always want to display that the server was started, just make it `logging.critical()`. I don't understand why you would want to force logging.INFO to display a single message. I am currently testing Sanic as a high-throughput IoT message ingestion API, which needs to make HTTP requests to several backends. I am processing thousands of messages/sec and so this results in way too many logging.INFO messages from the `requests` library, e.g.: "INFO: Starting new HTTP connection (1): <url>`, rendering my message logs unreadable (and actually drastically reducing the webserver's overall performance). For now, I cannot use this in production.

PR #57, opened 2016-10-18
Added multiprocessing

channelcat:


sanic/sanic.py
```diff
@@ -1,5 +1,9 @@
-import asyncio
+from argparse import ArgumentParser
+from asyncio import get_event_loop
 from inspect import isawaitable
+from multiprocessing import Process, Event
```
salamer: I think the multiprocess function should not contain in the program, just let user to decide use multi process or not

salamer: I think the multiprocess function should not contain in the program, just let user to decide use multi process or not

channelcat: If run is executed with 1 worker, it won't use multiprocessing.  Should we not import it by default?

PR #58, opened 2016-10-18
release 0.1.4 - multiprocessing

channelcat:


notedit:
i don't like this multiprocessing,  the subprocess can not auto start and must to disable before_start and after_start.

maybe  we could took some idea from https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/worker.py

channelcat:
@notedit At first I didn't understand what you meant, because I forgot to set those back.  https://github.com/channelcat/sanic/pull/79 should allow them for multiprocessing.

PR #59, opened 2016-10-18
comment fixed

yishibashi:
fixed comment in sanic.py.

(my first pull request 👏 )

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -47,9 +47,8 @@ class Sanic:
     # Decorator
     def exception(self, *exceptions):
         """
-        Decorates a function to be registered as a route
-        :param uri: path of the URL
-        :param methods: list or tuple of methods allowed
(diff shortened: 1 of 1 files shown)


sanic/sanic.py
```diff
@@ -47,9 +47,8 @@ def response(handler):
     # Decorator
     def exception(self, *exceptions):
         """
-        Decorates a function to be registered as a route
-        :param uri: path of the URL
-        :param methods: list or tuple of methods allowed
+        Decorates a function to be registered as a handler for exceptions
+        :param *exceptions: exceptions
```
mikoim: In this method, exceptions parameter is tuple.
I think that the line should be changed as following.

```
:param *exceptions: tuple of exceptions
```

yishibashi: exception(variable in this method) is tuple, but this method (or *exception) does not receive tuple. So I think "exceptions" is better.

PR #79, opened 2016-10-19
Enable after_start and before_stop callbacks for multiprocess

GMLudo:
Hi,

I need to initiate DB socket pools (pgsql...) with multiprocessing setup.
This PR seems to fix that.

Thanks for your response.


channelcat:
OH, I realize now I never set that back.  I was originally considering running on_start after all processes have been completed, but half-way through realized you may want to initialize things per-process, and forgot to change it back.  This also probably fixes the disabled multiprocessing test :D.

PR #80, opened 2016-10-20
Faster router

jpiasetz:
I've done a few different things here.

1. Removed the simple router since I think the regular route with a large enough cache should be very close in terms of speed
2.  Refactor the existing code
3. Added a lru_cache to the get request
4. Made a really simple hash lookup so it doesn't need to loop over all routes

If this lands because of 4. regexes with / in them won't work which is a bit of a trade off. Right now though regexes with : in them will have issues because of the split so I'm not sure it's worse? Might be worse thinking about limiting even further like flask does since that will improve lookup performance.

### DIFF
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -22,3 +22,4 @@ class Config:
 """
     REQUEST_MAX_SIZE = 100000000  # 100 megababies
     REQUEST_TIMEOUT = 60  # 60 seconds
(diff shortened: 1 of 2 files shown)


channelcat:
Excellent work!  I added support for routes with / regexes, and added it to the tests.  I'm seeing 34,676 req/sec against cached routes now! :D

PR #87, opened 2016-10-21
Blueprint start/stop listeners + ordering

channelcat:
Added blueprint start/stop listeners
Added sanic as the first argument in listeners triggered when the server starts/stops
reversed the order of response middleware

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -1,4 +1,5 @@
 from asyncio import get_event_loop
+from functools import partial
 from inspect import isawaitable
 from multiprocessing import Process, Event
 from signal import signal, SIGTERM, SIGINT
@@ -24,6 +25,8 @@ class Sanic:
         self.response_middleware = []
         self.blueprints = {}
         self._blueprint_order = []
+        self.loop = None
+        self.debug = None
 
     # -------------------------------------------------------------------- #
     # Registration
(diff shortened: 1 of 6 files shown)


sanic/sanic.py
```diff
@@ -71,7 +74,7 @@ def register_middleware(middleware):
             if attach_to == 'request':
                 self.request_middleware.append(middleware)
             if attach_to == 'response':
-                self.response_middleware.append(middleware)
+                self.response_middleware.insert(0, middleware)
```
jpiasetz: Why not use a [deque](https://docs.python.org/2/library/collections.html?highlight=collections#collections.deque)?

sanic/server.py
```diff
@@ -157,7 +157,22 @@ def close_if_idle(self):
...
 
 
-def serve(host, port, request_handler, after_start=None, before_stop=None,
+def trigger_events(events, loop):
+    """
+    :param events: one or more sync or async functions to execute
+    :param loop: event loop
+    """
+    if events:
+        if type(events) is not list:
```
jpiasetz: isinstance of iterable might be better.

channelcat: Ah yes, will do :)

sanic/sanic.py
jpiasetz: It's just an iterable. You can iterate over it as many times as you like provided you don't pop i.e.

```python
>>> from collections import deque
>>> my_list = deque(['b'])
>>> my_list.append('c')
>>> my_list.appendleft('a')
>>> for i in my_list:
...     print(i)
... 
a
b
c
>>> for i in my_list:
...     print(i)
... 
a
b
c
>>> 
```

channelcat: oh awesome, I did not know :)

channelcat: Added in https://github.com/channelcat/sanic/pull/105

PR #89, opened 2016-10-21
Refactor server a bit

jpiasetz:


sanic/request.py
```diff
@@ -105,6 +105,18 @@ def args(self):
 
         return self.parsed_args
 
+    @property
```
channelcat: I'm not sure this will speed things up.  URL and method are checked by the router on every request, and version is used by the response.

sanic/server.py
```diff
@@ -1,47 +1,47 @@
...
 try:
     import uvloop as async_loop
 except ImportError:
     async_loop = asyncio
 
+from .config import Config
 from .log import log
 from .request import Request
 
+CONNECTIONS = set()
```
channelcat: I like using the set, but I think capitals implies it is a constant.  Also we will need to make this a member variable again if we want to support running multiple servers on the same event loop.

sanic/server.py
```diff
@@ -122,44 +126,26 @@ def on_message_complete(self):
...
-        """
-        if not self.parser:
-            self.transport.close()
-            return True
-        return False
+            log.error('Writing request failed, connection closed %s', e)
+            self.close()
 
 
 def serve(host, port, request_handler, after_start=None, before_stop=None,
```
channelcat: I kept the configuration settings as part of the arguments so other projects could re-purpose the server in their own frameworks without it being coupled with other parts of Sanic if they wanted.  But, that does come at the cost of passing lots of arguments.  What are your thoughts?

sanic/request.py
jpiasetz: Being lazy is helpful though in case the connection doesn't live long enough i.e. if you get the headers but when you start receiving the body it is too long. Then there's no point calculating the version, URL, etc.

sanic/server.py
jpiasetz: It does imply a constant (it's not enforced though). It might be faster to have a dict of sets rather then member, objects are pretty slow in python. Once the low hanging stuff is cleaned up it will probably be worth it to pass over all the code and remove all the dots on the hot path.

seemethere:
Too many merge conflicts to keep this open, will close, please resubmit when the merge conflicts are sorted out.

PR #94, opened 2016-10-22
Simple blueprint was missing the 'request' parameter

### LINKED ISSUE #90: No request param in blueprint docs
```
@bp.route('/')
async def bp_root():
    return json({'my': 'blueprint'})
```

should be 
```
@bp.route('/')
async def bp_root(request):
    return json({'my': 'blueprint'})
```

narzeja:
Fixes #90.


channelcat:
Thank you! :)

Issue #77, opened 2016-10-19
Examples - Peewee Async


channelcat:
Create a new example of how to use peewee-async with Sanic. It can be a simple python file or an organized project in its own folder, whichever is easiest to understand and gets someone started on the right track.

Peewee-async docs: 
https://peewee-async.readthedocs.io/en/latest/

narzeja:
PR is ready, check it out.

channelcat:
Merged, closing :)

PR #95, opened 2016-10-22
Provide example of using peewee_async with Sanic

narzeja:
This PR adds a simple example of using Peewee_async with Sanic with a PostgreSQL backend.

I have not added additional dependencies into the requirements-dev.txt, but instead the example file lists what is required for running it.

Solves #77


channelcat:
Thank you! :D

PR #101, opened 2016-10-23
Static middleware

imbolc:
It made as middleware for possibility to serve files from root url, e.g. `/robots.txt`

### DIFF
--- a/sanic/static.py
+++ b/sanic/static.py
@@ -1,0 +1,48 @@
+import re
+import os
+from zlib import adler32
+import mimetypes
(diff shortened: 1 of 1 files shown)


channelcat:
This is really awesome :D.  I really like the idea of serving static content from Sanic, so I did some work to integrate this directly into sanic and blueprints.  However, Github desktop ended up making this another pull request, and I'm not pro enough to know how to fix that.  New PR is https://github.com/channelcat/sanic/pull/104.

PR #104, opened 2016-10-24
Static file support

channelcat:
Added sanic.static, blueprint.static, documentation, and testing.

Based on https://github.com/channelcat/sanic/pull/101


imbolc:
So you don't want to overlap dynamic routes with static files? In the current realization it's impossible to serve `/robots.txt` if previous route handles any request `app.route('*')`.

channelcat:
The current route system will pick the most specific routes first, though I'm not sure if we want it this way.  If we were going off of which routes were registered first, you'd need to register robots.txt first.  I could see people wanting to override static files with app routes, and app routes with static files.

Maybe we could allow you to specify a route priority.  What do you think?

channelcat:
@imbolc Looking at your comment in the issue, I think I see what you mean now.  I'm going to think more about how to solve the issue, but at the moment I think your solution of using middleware when you want to serve static content and fallback to other routes is best.

imbolc:
Maybe we can have static route and middleware together, just like options?
But for me personally it's not so important, because it's easy to just add
overriding as custom middleware, it not need to be part of sanic core.

PR #106, opened 2016-10-24
basic logging

akshbn:
Logging can now be done in a way similar to flask.
Eg:
app = Sanic(**name**)
app.log("INFO","this is an info")

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -315,13 +315,14 @@ class Sanic:
         for process in processes:
             process.join()
 
-    def log(self,level,msg):
+    def log(self, level, msg):
         '''
         Logs the message at the specified level.
         Eg: app.log("INFO","this logs an info")
         '''
-        from logging import getLogger,basicConfig,INFO
-        basicConfig(level = INFO,format = "%(asctime)s: %(levelname)s: %(message)s")
+        from logging import getLogger, basicConfig, INFO
+        basicConfig(
+            level=INFO, format="%(asctime)s: %(levelname)s: %(message)s")
         log = getLogger(__name__)
         if level == "DEBUG":
             log.debug(msg)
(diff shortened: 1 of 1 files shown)


sanic/sanic.py
```diff
@@ -314,3 +314,23 @@ def serve_multiple(server_settings, workers, stop_event=None):
             process.terminate()
         for process in processes:
             process.join()
+
+    def log(self, level, msg):
+        '''
```
seemethere: Docstrings should be done with triple double quotes like so:

```python
def log(bar):
    """Logs the message at the specified level
    >>> app.log('INFO', 'this logs an info')
    """
    pass
```

sanic/sanic.py
```diff
@@ -314,3 +314,23 @@ def serve_multiple(server_settings, workers, stop_event=None):
...
+        Logs the message at the specified level.
+        Eg: app.log("INFO","this logs an info")
+        '''
+        from logging import getLogger, basicConfig, INFO
+        basicConfig(
+            level=INFO, format="%(asctime)s: %(levelname)s: %(message)s")
+        log = getLogger(__name__)
+        if level == "DEBUG":
```
seemethere: This can also be done with a dictionary. and maybe be more lenient so that that the level is not case sensitive?

sanic/sanic.py
```diff
@@ -314,3 +314,23 @@ def serve_multiple(server_settings, workers, stop_event=None):
             process.terminate()
         for process in processes:
             process.join()
+
+    def log(self, level, msg):
+        '''
+        Logs the message at the specified level.
+        Eg: app.log("INFO","this logs an info")
+        '''
+        from logging import getLogger, basicConfig, INFO
```
seemethere: Why do imports here? Why not just add these to the top of the file?

pahaz:
I think this is not a part of the Sanic core logic. If you want you can use it in you code.

seemethere:
Agree with @pahaz, closing this for now.

PR #109, opened 2016-10-25
Renamed function file() to send_file() because the name `file` is reserved

pcdinh:
`file` is a reserved built-in symbol. Redefining `file` can caused unpredictable issues

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -153,7 +153,7 @@ def html(body, status=200, headers=None):
                         content_type="text/html; charset=utf-8")
 
 
-async def file(location, mime_type=None, headers=None):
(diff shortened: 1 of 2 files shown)


channelcat:
I took a look and I think 'file' may have been freed up with Python 3.

based on https://docs.python.org/3/library/keyword.html:
```python
>>> import keyword
>>> keyword.iskeyword('file')
False
```

I'd like to keep it simply "file" if we can, just to match the scheme we have going.  I'm open to changing it if it causes more trouble than it's worth though.

pcdinh:
You are right. I forgot that in Python 3, file object was replaced

PR #112, opened 2016-10-25
Allowed force download a file without rendering its content in the browser

pcdinh:


rzwiefel:
This feels strange to have in the core logic of the web server. This is something that a user can very easily specify in their own request if they wish. If we add every possible headers option in order to make the browser do something then this code would become very bloated, very quickly.

sanic/response.py
```diff
@@ -153,13 +153,26 @@ def html(body, status=200, headers=None):
                         content_type="text/html; charset=utf-8")
 
 
-async def file(location, mime_type=None, headers=None):
+async def file(location, mime_type=None, headers=None, force_download=False):
```
pahaz: you can just pass the `'application/octet-stream'` as mime_type and pass the `"Content-Disposition"` headers.

seemethere:
I'd agree with @rzwiefel here. Closing this for now.

PR #114, opened 2016-10-25
Add transport to Request

amsb:
Add transport to the request object (like aiohttp) so that extra information such as peername can be queried as well as to provide the foundation for upgrading the connection to a websocket if someone wanted to that with the help of an external websocket library. While I included a simple test, this is more of a proposal since I am new to the code base and do not understand all the potential ramifications.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -31,12 +31,12 @@ class Request:
     Properties of an HTTP request such as URL, headers, etc.
     """
     __slots__ = (
-        'url', 'headers', 'version', 'method', '_cookies',
+        'url', 'headers', 'version', 'method', 'transport', '_cookies',
         'query_string', 'body',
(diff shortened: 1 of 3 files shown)


pcdinh:
What is the purpose of this? As far as I know, an `transport` instance is assigned automatically when `connection_made` is called. By default it is `TCPTransport` in uvloop.

amsb:
How do I access transport in a request handler?

amsb:
@channelcat, any thoughts on this PR? Setting aside the potential of upgrading the connection to a websocket for the moment, do you have a suggestion on how to grab the remote address/peername from inside a request handler?

amsb:
@seemethere, any thoughts on this suggestion or guidance about accessing the transport from inside a request handler? Thanks!

seemethere:
closed by #147

Issue #116, opened 2016-10-26
Unparseable JSON should not be ignored


jasonab:
request.py, line 64

Returning a None here causes breakage in application code further down the line. Generate a 400 error here on malformed JSON to protect the server.

seemethere:
Should it just raise an Exception that then generates a 400 error?

jasonab:
Yes, that would be my preference.

seemethere:
The actual line is here now: https://github.com/channelcat/sanic/blob/master/sanic/request.py#L69

jasonab:
Correct, that line should not log anything (as I don't want my logs littered with client-side problems), and should raise an InvalidUsage exception.

r0fls:
@jasonab I am not able to reproduce this, am I misunderstanding the bug? I added this test and although it fails, there is a 500 internal server error (not the silent, ignoring you're referring to):

```python
def test_post_invalid_json():
    app = Sanic('test_post_json')

    @app.route('/')
    async def handler(request):
        data = request.json()
        return json(data)

    payload = None
    headers = {'content-type': 'application/json'}

    request, response = sanic_endpoint_test(app, 
                                            data=json_dumps(payload), 
                                           headers=headers)

    assert response.status == '400'
```

jasonab:
You're saying Sanic now throws a 500 error on malformed JSON? That doesn't sound right.

The initial issue was that if JSON parsing fails and throws an exception, Sanic catches that exception and returns None from the json() method. Any code of mine that assumes that we have a non-None value from json() breaks.

What stack trace are you now seeing?

r0fls:
Ah, I was misunderstanding then. That test is malformed post data. I have observed the issue you're referring to as well.

r0fls:
This looks like it's a result of the `dump` function in `ujson`:

```python
>>> import ujson
>>> ujson.dumps(None)
'null'
>>> ujson.dumps("I am not valid json")
'"i am not valid json"'
```

See:
https://github.com/channelcat/sanic/blob/master/sanic/response.py#L5

and:
https://github.com/channelcat/sanic/blob/master/sanic/response.py#L135

r0fls:
I don't think that request.py actually has anything to do with this, since we're talking about invalid json in the response.

jasonab:
The line of code I cited (now at request.py line 69) is for incoming data, not response data. I'm sorry if there's been confusion.

[...]

PR #117, opened 2016-10-26
Solved issue 116

ruchir594:
Hello,

I hope I solved the issue 116 the right way. Happy to receive feedback.


seemethere:
#116 <-- So that the issue gets linked to this PR

sanic/request.py
```diff
@@ -61,7 +62,10 @@ def json(self):
             try:
                 self.parsed_json = json_loads(self.body)
             except Exception:
-                pass
+                return HTTPResponse(json_dumps(self.body),
```
seemethere: I don't think it makes sense to return an HTTPResponse from a function in the request. What do you think @channelcat?

pahaz: there is a really bad idea. I don't expect to get the response in case to access the `request.json`

EFF: @pahaz agreed, makes no sens at all.

seemethere:
Closing for now, we may open up back later.

PR #118, opened 2016-10-26
Allowed redirect a user to another endpoint, use the redirect() function

pcdinh:


pahaz:
Why you use `_request` style? (with underline)

pahaz:
You redirect function doesn't support `https` urls.

pcdinh:
@pahaz Underscore _ is to hide unused variable warnings by Pylint

pcdinh:
@pahaz I just made redirection logic simpler because I think browser will do more work at client side. Tornado or Web2py redirect implementation is very simple too. What do you think?

sanic/response.py
```diff
@@ -158,3 +158,24 @@ def html(body, status=200, headers=None):
...
+    if not content_type:
+        content_type = "text/html; charset=utf-8"
+    if not status:
+        if request.method == "POST":
+            # See: https://en.wikipedia.org/wiki/HTTP_303
+            status = 303
+        else:
+            status = 302
+    # According to RFC 7231, a relative URI is now permitted.
+    headers['Location'] = to_url
```
pahaz: If the `headers` is None it will raise the error

pahaz:
@pcdinh

In my opinion:
 - Refactoring with underscores make sense. But it is not a part of this PR. It would be better if you save current style. Or just make another PR with refactoring.
 - Pass the `request` object to the redirect function looks little bit strange. 

Some examples:
 - Django
     - HttpResponseBase https://github.com/django/django/blob/master/django/http/response.py#L41 
     - HttpResponseRedirectBase https://github.com/django/django/blob/master/django/http/response.py#L419
     - HttpResponseRedirect and HttpResponsePermanentRedirect https://github.com/django/django/blob/master/django/http/response.py#L440-L445

 - Flask
     - `flask.redirect` https://github.com/pallets/werkzeug/blob/master/werkzeug/utils.py#L344
 
 - aiohttp
     - `HTTPMove(HTTPRedirection)` https://github.com/KeepSafe/aiohttp/blob/e505a4bea80f2edd96fdc26c60d753c39f5037a8/aiohttp/web_exceptions.py#L130

You can compare the implementations and rethink your solution.
You also have some conflicts with master now. (could you please fix it)

seemethere:
I would agree with @pahaz on this one

seemethere:
@pcdinh Can you fix the merge conflict and fix the issue where the headers being equal to `None` would raise an exception?

sanic/response.py
zhyc9de: just check
```
    if not isinstance(headers, dict):
        headers = {}
```

r0fls:
This was handled in #294

Issue #96, opened 2016-10-22
Example of using async requests


peterbe:
I actually don't know if it requires a custom requests library or if it's something that can be done "natively" with sanic, but being able to do requests queries async would be awesome. Then you don't have to worry about blocking when doing a sloooow `requests.post(...)`.
Also, something I'm personally interested in, is the ability to use elasticsearch in an async matter.

argaen:
For elasticsearch you can use https://github.com/aio-libs/aioes.

For async http requests, `aiohttp` support this, until sanic supports it maybe you can use its http client (I'm not aware if there is another library doing this).

jackfischer:
Just sent a pull request! #119

EFF:
Seems like this issue should be closed now.

* async request question is answered
* es related question is answered
* example has been provided and merged

seemethere:
closed with #119

PR #119, opened 2016-10-26
Add example with async http requests (Issue #96)

jackfischer:
Adds a simple example using aiohttp to fetch a JSON object with Sanic.


channelcat:
Thank you :)!

PR #122, opened 2016-10-27
Resolved #76: Add sqlalchemy_example.py for sqlalchemy integration example

### LINKED ISSUE #76: Examples - SQLAlchemy
Create a new example of how to use SQLalchemy with Sanic.  It can be a simple python file or an organized project in its own folder, whichever is easiest to understand and gets someone started on the right track.

Example of sqlalchemy + aiopg: 
https://aiopg.readthedocs.io/en/stable/#sqlalchemy-and-aiopg

More research needs to be done on integration with other popular DBs - I'm not sure if this will up the difficulty so I tagged it as both beginner and intermediate.

chhsiao90:
Resolved #76  to implement sqlalchemy example.
Please help review and advise if any, thanks!

### DIFF
--- a/examples/sqlalchemy_example.py
+++ b/examples/sqlalchemy_example.py
@@ -1,0 +1,89 @@
+""" SQLAlchemy integration demo
+Use sqlalchemy to connect a in-memory sqlite database
+Defined a table User contains id, name, fullname, password as its columns
+Provider RESTFul service
+GET /users to get all of users in User table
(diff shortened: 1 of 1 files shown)


chhsiao90:
Sorry, I think what you want is async-support of sqlalchemy example.
The block when doing query would cause request slow....
I will just close the PR and find out other ways!

PR #128, opened 2016-10-28
Keep alive timeout fix

channelcat:
Currently requests time out and forcibly close 60 seconds (configurable) after the connection opens.  This can cause healthy keep-alive requests to spontaneously die mid-request.  I've added in logic to update the request timeout after each response.  With caching the system time every second, the impact seems to be a slowdown of ~300-500 requests/sec.

I looked into setting socket options but I can't seem to find a keep-alive timeout.

Since this isn't a critical fix, I'm leaving this open for a bit in case anyone else has a clever or less impactful solution.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -18,6 +18,10 @@ from .request import Request
 class Signal:
     stopped = False
 
+
+current_time = None
+
+
 class HttpProtocol(asyncio.Protocol):
     __slots__ = (
         # event loop, connection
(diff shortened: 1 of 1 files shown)


sanic/server.py
```diff
@@ -133,13 +147,15 @@ def write_response(self, response):
...
             else:
+                # Record that we received data
+                self._last_request_time = current_time
                 self.cleanup()
         except Exception as e:
             self.bail_out(
                 "Writing request failed, connection closed {}".format(e))
 
     def bail_out(self, message):
-        log.error(message)
```
seemethere: Don't you still want to have this bail out log to error?

Issue #129, opened 2016-10-28
Storage for shared data


imbolc:
Storing of shared data like db connection pool or jinja environment in global variables is not so good idea for well known reasons.

Aiohttp has a nice idea to make app and request as dict-like objects. That allows us to use these as containers for shared data. E.g. you can store db connection as `app['db']` during initialization and then use it in handlers as `request.app['db']`. Dict-interface makes clear distinction between framework data (witch available as attributes) and user data (as dict items). Why wouldn't we borrow this idea?

atribecalledkwest:
This would be super useful, perhaps a simple app.shared/app.globals dict would good enough.

PR #132, opened 2016-10-29
A very basic implementation of a shared resource, app.shared (#129)

atribecalledkwest:
I'm sure there is a major flaw in these 6 lines of changes to make #129 a thing, but it seems to work.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -34,6 +34,7 @@ class Request:
         'url', 'headers', 'version', 'method', '_cookies',
         'query_string', 'body',
         'parsed_json', 'parsed_args', 'parsed_form', 'parsed_files',
+        'shared',
     )
 
     def __init__(self, url_bytes, headers, version, method):
@@ -54,6 +55,7 @@ class Request:
         self.parsed_files = None
         self.parsed_args = None
         self._cookies = None
+        self.shared = None
 
     @property
     def json(self):
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -29,6 +29,7 @@ class Sanic:
         self._blueprint_order = []
         self.loop = None
         self.debug = None
+        self.shared = {}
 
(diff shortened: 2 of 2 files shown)


pahaz:
In my opinion the `app` object represent the application state and shared resource storage.
And if we change line:
https://github.com/channelcat/sanic/pull/132/commits/78707608fb6a5a34330f81cf3222ac9829146e9c#diff-ac7b7152eadc8343e3f10d4780d9a678R177
```
request.shared = self.shared
```
to 
```
request.app = self
```

It would be better and more clear approach. (But it may have effect to performance)

guidiego:
Any update on that issue?

guidiego:
@pahaz I need that cuz we are trying to use Sanic on prod! I do the modifications [here](https://github.com/iclinic/sanic), if you want and if @atribecalledkwest not appear I can do the PR :)

pahaz:
@guidiego . Yes, I understand your problem. I am also use it! This is really a problem.

What do you think about the `request.app`? Is it solve your problem?

guidiego:
Hey @pahaz, so after try some implementations I'm not sure about this will solve my problem. It's totally useful to some contexts, but I ask myself if this `shared` it's the correct thing to do! 

I think that we need this implementation in the most part of cases to transport db contexts and configuration variables to the rest of application, I had to use `python-decouple` and import the file inside the places that I need some "shared contexts" because the Sanic is not totally environment agnostic.

We can discuss about some Flask implementations, like, `from_file`, `from_json` and `from_folder` to set APP shared contexts :) And how we can Implement this most useful with a kind of good pratices like `create_app` and `manages funcs` :D

Let me know if I can help you in something, I think that we can create a gitter for improve the contact :D

seemethere:
Closing, I think the implementation in #163 is a bit better for this.

Issue #133, opened 2016-11-01
raise TypeError when try access request.form and request.method is 'GET'


oxbambooxo:
when a request is GET method, request headers not contain 'Content-Type' value
but in the [request.py](https://github.com/channelcat/sanic/blob/0.1.7/sanic/request.py#L74)

    content_type, parameters = parse_header(
        self.headers.get('Content-Type'))

parse_header need a string value to parse, but now get None, in parse_header

    def parse_header(line):
        parts = _parseparam(';' + line)

it conbine ';' + None, raise the TypeError: Can't convert 'NoneType' object to str implicitly.

in werkzeug request handle, they use default value '' when content-type not exist([werkzeug request wrappers](https://github.com/pallets/werkzeug/blob/0.11/werkzeug/wrappers.py#L356)).

    if self.want_form_data_parsed:
        content_type = self.environ.get('CONTENT_TYPE', '')

i think back off to empty dict may be better, :)

seemethere:
Closed by #137

PR #137, opened 2016-11-03
Fix upload without content-type

imbolc:
Fix this error:
```
  File "/home/imbolc/0/psy/sanic/request.py", line 74, in form
    self.headers.get('Content-Type'))
  File "/home/imbolc/.pyenv/versions/3.5.2/lib/python3.5/cgi.py", line 319, in parse_header
    parts = _parseparam(';' + line)
TypeError: Can't convert 'NoneType' object to str implicitly
```


seemethere:
Closes #133

PR #139, opened 2016-11-04
fix(blueprints): @middleware IndexError

pahaz:
I think it`s just a mistake here (https://github.com/channelcat/sanic/blob/74ae0007d3a7bfea562e8b4a29880184815f9155/sanic/blueprints.py#L112-L113)

```
            args = []
            return register_middleware(args[0])
```

We just set args to empty list and then call the `args[0]`

My code:

```
bp = Blueprint('sessions', '/sessions')

@bp.middleware
async def session_middleware(request):
    ...

```

Produce the error:

```
  File "main.py", line 7, in <module>
    from sessions.blueprint import bp
  File "/Users/pahaz/cross-social-science/sessions/blueprint.py", line 25, in <module>
    @bp.middleware
  File "/Users/pahaz/.virtualenvs/cross-social-science/lib/python3.5/site-packages/sanic/blueprints.py", line 113, in middleware
    return register_middleware(args[0])
IndexError: list index out of range
```

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -109,8 +109,9 @@ class Blueprint:
 
         # Detect which way this was called, @middleware or @middleware('AT')
         if len(args) == 1 and len(kwargs) == 0 and callable(args[0]):
+            middleware = args[0]
             args = []
-            return register_middleware(args[0])
+            return register_middleware(middleware)
         else:
             return register_middleware
 


pahaz:
Please rebuild the master.

CI status: `tests/test_requests.py::test_query_string FAILED` 
My changes don't affect the requests logic.

It's looks like problem in test:

https://travis-ci.org/channelcat/sanic/builds/173357477#L451
```
request, response = sanic_endpoint_test(app, params=[("test1", 1), ("test2", "false"), ("test2", "true")])
```

look at `("test1", 1)`

It may produce the error.

https://travis-ci.org/channelcat/sanic/builds/173357477#L485
```
ValueError: Exception during request: [TypeError('Argument should be str',)]
```

seemethere:
@pahaz I'm not getting the failure locally but I'll look into it.

seemethere:
Well I did a bit of digging, the fix @pahaz identified will work but the problem comes from a seemingly insignificant dependency.

```  File "yarl/_quoting.pyx", line 46, in yarl._quoting._quote (yarl/_quoting.c:1384)
TypeError: Argument should be str```

https://github.com/aio-libs/yarl/blob/master/yarl/_quoting.pyx#L45

For now I'll submit a PR to fix the tests but in the future we need to identify a way to make it a bit less fragile.

seemethere:
@pahaz  the error should be fixed now, it's safe to remerge.

Issue #136, opened 2016-11-03
is there any redis module support for this framwork? 


cctse:
as title

seemethere:
I don't believe at the moment that it does.

argaen:
I believe you could use https://github.com/aio-libs/aioredis anyway.

You could also give a try to https://github.com/argaen/aiocache which is a wrapper for different backends (disclaimer: I'm the maintainer).

argaen:
Hi, I've written a basic example using `aiocache` with Redis:

```python
import asyncio
import aiocache

from sanic import Sanic
from sanic.response import json
from sanic.log import log
from aiocache import RedisCache, cached
from aiocache.serializers import JsonSerializer

app = Sanic(__name__)
aiocache.set_defaults(cache=RedisCache)


@cached(key="my_custom_key", serializer=JsonSerializer())
async def expensive_call():
    log.info("Expensive has been called")
    await asyncio.sleep(3)
    return {"test": True}


@app.route("/")
async def test(request):
    log.info("Received GET /")
    return json(await expensive_call())


app.run(host="0.0.0.0", port=8000, loop=asyncio.get_event_loop())
```

The logs:

```
11:56 $ python cache_example.py 
cPickle module not found, using pickle
2016-11-05 11:57:07,875: INFO: Goin' Fast @ http://0.0.0.0:8000

2016-11-05 11:57:20,399: INFO: Received GET /
2016-11-05 11:57:20,401: INFO: Expensive has been called
2016-11-05 11:57:25,700: INFO: Received GET /
2016-11-05 11:57:26,406: INFO: Received GET /
2016-11-05 11:58:20,421: ERROR: Request timed out, connection closed
```

And the keys in Redis after the call to the / endpoint:

```
127.0.0.1:6379> keys *
1) "my_custom_key"
127.0.0.1:6379> get "my_custom_key"
"{\"test\":true}"
```

seemethere:
Closed by #140

cctse:
@argaen thanks

argaen:
You're welcome @cctse. As said, I'm the maintainer of the aiocache package, in case you have any feedback or problems do no hesitate to open issues there :).

PR #140, opened 2016-11-05
Caching example using aiocache

argaen:
Example of caching using aiocache. #136


seemethere:
I don't know why this change would make a test fail. Can you pull and merge the latest changes in master and see if that fixes this?

I'd like to merge but I want to see a fast forward attempted and the PR retested before I do that.

seemethere:
@argaen the error should be fixed now, it's safe to remerge.

argaen:
I merged right now you should be okay to merge right?

PR #145, opened 2016-11-08
Fix request parse multipart form

pahaz:
I have this code:

```
form = FileForm(request.files)
```

and it raise error because the `request.files` is `dict` but `RequestParameters` is expected =/

This PR is a continue of the unification of `request.files` and `request.form`.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -132,7 +132,7 @@ def parse_multipart_form(body, boundary):
     Parses a request body and returns fields and files
     :param body: Bytes request body
     :param boundary: Bytes multipart boundary
-    :return: fields (dict), files (dict)
+    :return: fields (RequestParameters), files (RequestParameters)
     """
     files = RequestParameters()
     fields = RequestParameters()
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -134,8 +134,8 @@ def parse_multipart_form(body, boundary):
     :param boundary: Bytes multipart boundary
     :return: fields (dict), files (dict)
     """
-    files = {}
-    fields = {}
+    files = RequestParameters()
+    fields = RequestParameters()
 
     form_parts = body.split(boundary)
     for form_part in form_parts[1:-1]:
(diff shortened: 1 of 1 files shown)


pahaz:
There is the third my PR without any tests =/
Sorry for that!

sanic/request.py
```diff
@@ -166,9 +166,16 @@ def parse_multipart_form(body, boundary):
 
         post_data = form_part[line_index:-4]
         if file_name or file_type:
-            files[field_name] = File(
-                type=file_type, name=file_name, body=post_data)
+            file = File(type=file_type, name=file_name, body=post_data)
```
seemethere: I'd suggest naming this parameter something different because naming it this squashes the base `file` type

sanic/request.py
```diff
@@ -166,9 +166,16 @@ def parse_multipart_form(body, boundary):
...
+            if field_name in files:
```
seemethere: Using a `defaultdict` in this case would make more sense. then you wouldn't have to do the check

sanic/request.py
```diff
@@ -166,9 +166,16 @@ def parse_multipart_form(body, boundary):
...
+                files[field_name].append(file)
+            else:
+                files[field_name] = [file]
         else:
-            fields[field_name] = post_data.decode('utf-8')
+            value = post_data.decode('utf-8')
+            if field_name in fields:
```
seemethere: The same idea regarding `defaultdict` goes here as well.

pahaz: Yes, but the purpose of this PR is to change the type of `files` and `fields` to `RequestParameters`.

The `files` and `fields` is not a instance of `defaultdict`.
I agree with you. It would be better to use the `defaultdict`. 
But in this case we need to convert `defaultdict` object to `RequestParameters`.

Are you sure that it would be better? (+ 1 type conversion)

sanic/request.py
seemethere:
```shell
Python 2.7.10 (default, Jul 30 2016, 18:31:42)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> type(file)
<type 'type'>
>>>
```

seemethere: Oh you are right, I was using a `2.7.10` interpreter, didn't realize that was a change.

Issue #146, opened 2016-11-10
how to get request IP address?


xinke2411:
how to get request IP address?

seemethere:
closed by #147

Issue #12, opened 2016-10-15
Add web*socket* support


jesuejunior:
Maybe a plugin or a builtin support for websocket will be great.

channelcat:
I do not have plans to add web socket support any time soon, but from the looks of it it would be possible to integrate without negative affecting the performance of Sanic as an http server (just some simple header checks).  I've added tags for people to pick this up if interested :).

PR #147, opened 2016-11-10
Add the client address to the request header

webtic:
I expected the remote_addr in the Request object but it wasn't there. 
We use the remote ip address extensively for access control and logging.
This PR adds it to the header so we can migrate some services to Sanic.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -116,7 +116,7 @@ class HttpProtocol(asyncio.Protocol):
     def on_headers_complete(self):
         ra = self.transport.get_extra_info('peername')
         if ra:
-             self.headers.append(('Remote-Addr','%s:%s' % ra))
+            self.headers.append(('Remote-Addr', '%s:%s' % ra))
 
         self.request = Request(
             url_bytes=self.url,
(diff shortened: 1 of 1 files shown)


sanic/server.py
```diff
@@ -114,6 +114,10 @@ def on_header(self, name, value):
         self.headers.append((name.decode(), value.decode('utf-8')))
 
     def on_headers_complete(self):
+        ra = self.transport.get_extra_info('peername')
```
seemethere: can we change this to a more explicit name like `remote_addr`?

amsb:
Great! This would close #146 and also address half of my use case for #114. The other half of my use case is captured by #12, so we could close #114 as well.

sanic/server.py
tistaharahap: will this work if sanic is behind a reverse proxy?

jasonab: The WSGI frameworks all pull from X-Forwarded-For, which Sanic should do as well.

tistaharahap: yeay

seemethere:
closes #146 and #114

PR #149, opened 2016-11-11
aiohttp is slightly faster actually

asvetlov:
Disabling access log increases RPS a lot.

aiohttp shows 6K RPS with this change.

It's still much slower than sanic but faster than, say, flask.

P.S. no aiohttp changes in HTTP parsing like switching to httptools, just access log disabling.


asvetlov:
Thanks.
Would you update README with actual results?

seemethere:
@asvetlov Sure, eventually

Issue #153, opened 2016-11-15
Exception monitoring?


jackfischer:
Do we have any way to monitor exceptions in production? Exceptions never seem to hit `sys.excepthook` which makes it hard to integrate services like Airbrake

jackfischer:
Nevermind, I'm an idiot. PR #154

seemethere:
Closed by #154

PR #154, opened 2016-11-16
Example for using error_handler

jackfischer:
I did not realize Sanic had this functionality built it and assumed that it would be a major roadblock to generalized exception monitoring. This example would have saved a lot of time! Eventually hopefully we can get plugins for Sentry and the like.


jackfischer:
@channelcat any feedback? I'm using this in a project and definitely want to do it right

seemethere:
Closes #153

PR #159, opened 2016-11-19
provide default app name

r0fls:
This allows instantiating the Sanic object without providing an app name:

```
from sanic import Sanic

app = Sanic()

@app.route("/")
async def test(request):

```

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -1,7 +1,7 @@
 from asyncio import get_event_loop
 from collections import deque
 from functools import partial
-from inspect import isawaitable
+from inspect import isawaitable, stack, getmodulename
 from multiprocessing import Process, Event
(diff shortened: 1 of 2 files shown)


sanic/sanic.py
```diff
@@ -18,7 +18,10 @@
 
 
 class Sanic:
-    def __init__(self, name, router=None, error_handler=None):
+    def __init__(self, name=None, router=None, error_handler=None):
+        if not name:
```
seemethere: I think it would be better to use `if name is None` or `if isinstance(name, NoneType)` to account for falsey like values like `''`

r0fls:
@seemethere I can see an argument for that, yes. If someone wants the instance name to be an empty string, who am I to blame them. Updated.

Issue #69, opened 2016-10-19
Request argument as a global variable


gaojiuli:
I think it's a little long-winded.
in sanic:

``` python
@app.route("/")
async def test(request):
    return json({"test": True})
```

in flask:

``` python
@app.route("/")
def hello():
    return "Hello World!"
```

Since each function takes a **request** argument, why not set it as a global variable.
Could you tell me the reason why you don't do this，thank you

aborsu:
Because the request argument changes with every request and they are executed asynchronously. So there would be no way to know which function is accessing it.

argaen:
@gaojiuli @aborsu Flask is setting it as a variable for the current Thread, a similar approach can be done using the `asyncio.Task.current_task()` call. Have a look at this repo: https://github.com/Skyscanner/aiotask-context. It's something I coded for storing variables through all the request/response cycle for aiohttp but it's framework agnostic (only depends on asyncio ofc). It's as easy as doing a `context.set('request', request_object)` at the beginning and then in any part of the code you can do a `context.get('request')` to retrieve it.

The problem is that the Task is only propagated when `await` calls are used but it works for the use case we have in our project.

aborsu:
I actually firmly believe that the global request object in flask is a bad design.
Explicit is better than implicit.

argaen:
Well global variables is a bad design. Setting variables in a context because its something you need in most of the calls of your functions its not.

Imagine you want to log the `request_id` to all your log calls. What would you do then, propagate this variable through ALL your code and function calls so you can use it every time you call `logger.info`? This indeed is really bad practice.

I totally agree with the explicit is better than implicit but principles are there to solve problems not to create them. If we wanted to follow it 100% why would you set a class variable when you can pass the value every time you call a function of class which is more explicit???

imbolc:
I agree with @aborsu. Pseudo-global request of flask is not only looks surprisingly for new users. It has significant performance issues (not sure it's the same for `asyncio.Task.current_task()`) and creates additional problems with testing etc.

pcdinh:
Mutable global variables are unsafe and unacceptable in asynchronous request handling. I don't think we have an option here.

guyskk:
Global variables is hard for testing, I prefer explict request argument, then I may create a request object explict and pass it to view function.

See http://wiki.c2.com/?GlobalVariablesAreBad for more.

seemethere:
[...]

PR #161, opened 2016-11-19
provide initialized class for convenience

r0fls:
This makes usage more compact and still supports the old style of importing and instantiating the `Sanic` class directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,16 +30,14 @@ All tests were run on an AWS medium instance running ubuntu, using 1 process.  E
 ## Hello World
 
 ```python
-from sanic import Sanic
+from sanic import sanic
 from sanic.response import json
 
-app = Sanic()
-
-@app.route("/")
+@sanic.route("/")
 async def test(request):
     return json({"hello": "world"})
 
-app.run(host="0.0.0.0", port=8000)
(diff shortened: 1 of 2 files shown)


jiajunhuang:
I don't think it's a good design for a web framework. As a framework, we should provide classes, but not instances(of course, it's just my opinion).

seemethere:
Yeah I'm not a fan of this either. Care to chime in @channelcat?

r0fls:
Personally I think the Sanic community is taking the whole "explicit is better than implicit" thing a little too far. Remember the first line of PEP20 is "Beautiful is better than ugly". So, while there is a tradeoff to providing these types of conveniences, I think preserving and allowing the more expressive underlying methods, while still improving usability via removing unnecessary code is the right approach. The other issue where I feel this way is #69.

r0fls:
I admit passing an instantiated class isn't beautiful, so there is that. However the upside is an improved API for interacting with Sanic, which I think is a win. I don't see a lot of use cases where it makes sense to customize the Sanic class instance. For example, when would a user want to supply their own, alternative router or error handling? None of the examples do this and I don't imagine many end users will. So, why not reduce code complexity, from a user's perspective?

PR #166, opened 2016-11-24
Added console logging statement to response handling block

jbcurtin:
I'd like to add this line so that exceptions will be logged to the console. If there is a better place to put this, please push back and I'll alter my PR accordingly. Performance wise, I'd like to adopt the naive viewpoint that this would be a non issue considering this logic executes only when an exception has been raised.

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -202,6 +202,8 @@ class Sanic:
                 response = self.error_handler.response(request, e)
                 if isawaitable(response):
                     response = await response
+
+                log.exception(e)
             except Exception as e:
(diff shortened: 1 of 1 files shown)


Tim-Erwin:
I think this line should be two lines below. Besides that the (really) correct place to put this is probably in  `exceptions.py:Handler:default`.

Tim-Erwin:
Ups, forget the first sentence, that was a bit premature.

seemethere:
I'd agree with @Tim-Erwin here

seemethere:
This was done in #306

PR #167, opened 2016-11-25
Class based views

AntonDnepr:
I just discovered sanic and was really impressed of its speed, so decide try to build API on it. But I was really lack of class based views, so write some simple implementation.

Changes:
1) added HTTPMethodView class 
2) added debug parameter for sanic_endpoint_test for easier test debugging
3) added missing dev requirement - aiofiles (at least it fails on my machine without it)
4) Add one more way to register routes: add_route. It already existed for the BlueprintSetup, so I just made such method for Sanic class and make available for Blueprint

### DIFF
--- a/docs/class_based_views.md
+++ b/docs/class_based_views.md
@@ -1,30 +1,30 @@
 # Class based views
 
-Sanic has simple class based implementation. You should implement methods(get, post, put, patch, delete) for the class to every HTTP method you want to support. If someone try to use not implemented method, there will be 405 response.
+Sanic has simple class based implementation. You should implement methods(get, post, put, patch, delete) for the class to every HTTP method you want to support. If someone tries to use a method that has not been implemented, there will be 405 response.
 
 ## Examples
 ```python
 from sanic import Sanic
-from sanic.views import MethodView
+from sanic.views import HTTPMethodView
 
 app = Sanic('some_name')
 
 
-class SimpleView(MethodView):
+class SimpleView(HTTPMethodView):
 
-  def get(self, request, *args, **kwargs):
+  def get(self, request):
       return text('I am get method')
 
-  def post(self, request, *args, **kwargs):
+  def post(self, request):
       return text('I am post method')
 
-  def put(self, request, *args, **kwargs):
+  def put(self, request):
       return text('I am put method')
 
-  def patch(self, request, *args, **kwargs):
+  def patch(self, request):
       return text('I am patch method')
 
-  def delete(self, request, *args, **kwargs):
+  def delete(self, request):
       return text('I am delete method')
 
 app.add_route(SimpleView(), '/')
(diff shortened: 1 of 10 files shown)


rednebmas:
As a newcomer to python webdev, the name 'MethodView' is a bit meaningless. Especially because the word 'method' is synonymous with function. After a brief google search it seems that the name came from the flask project. It took me a bit to realize that the name comes from 'HTTP method', but, I was thinking there might be a more intuitive name. What about RESTView (or RestView, not sure which is more pythonic)? Also, it's nice that RESTView is shorter. Just some food for thought!

AntonDnepr:
@rednebmas I don't think that RESTView will be better for this case, because `get`, `post` etc are parts of the HTTP protocol, not REST-specific and one could use them not only for REST architecture. You're right I use MethodView from Flask, because `Sanic is a Flask-like` :). Django use just generic `View` name for such case, so for shorter name just general `View` could be used.
What do you think?

rednebmas:
Oh, yup, that makes sense. I do like `View`, but after thinking about it for a bit I think I prefer `HTTPMethodView`. It is much longer, but it is also a lot clearer, and you will likely only use the name infrequently. It also might indicate that there are other types of views, although ErrorView is the only other thing I can think of at the moment. Thanks for considering my input!

docs/class_based_views.md
```diff
@@ -0,0 +1,44 @@
+# Class based views
+
+Sanic has simple class based implementation. You should implement methods(get, post, put, patch, delete) for the class to every HTTP method you want to support. If someone try to use not implemented method, there will be 405 response.
```
rednebmas: 'If someone tries to use a method that has not been implemented, ...'

docs/routing.md
```diff
@@ -29,4 +29,16 @@ async def person_handler(request, name):
...
+	return text('OK')
+app.add_route(handler1, '/test')
+
+async def handler(request, name):
+	return text('Folder - {}'.format(name))
+app.add_route(handler, '/folder/<name>')
+
+async def person_handler(request, name):
+	return text('Person - {}'.format(name))
+app.add_route(handler, '/person/<name:[A-z]>')
```
rednebmas: I don't understand why these changes are a part of the pull request.

docs/class_based_views.md
```diff
@@ -0,0 +1,44 @@
...
+app.add_route(SimpleView(), '/')
+
+```
+
+If you need any url params just mention them in method definition:
+
+```python
+class NameView(MethodView):
+
+  def get(self, request, name, *args, **kwargs):
```
rednebmas: Does using MethodView require *args, **kwargs to be parameters?

AntonDnepr: They are not required, you're right.

seemethere:
This was a great pull request I really appreciate the effort @AntonDnepr

wojunnihou:
我觉得应该像tornaodo一样，调用handle之前应该调用一个prepare，调用之后调用finish。（I am poor of English）like tornado ,before call the handler you should call prepare ,after call the handle you should call on_finish

PR #173, opened 2016-11-26
fix for cookie header capitalization bug

jackfischer:
Just spent a week with cloudflare support trying to figure out why cookies seem to vanish under some conditions with sanic - it seems like an [issue](http://httpwg.org/specs/rfc7540.html#CompressCookie) with HTTP2. 

There might be a more elegant way to do this or a way the maintainers would prefer. I wanted to get this to your attention right away as this could definitely be a really hard to diagnose problem for some people.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -114,11 +114,10 @@ class Request(dict):
     @property
     def cookies(self):
         if self._cookies is None:
-            if 'cookie' in self.headers: #HTTP2 cookie header
-                self.headers['Cookie'] = self.headers.pop('cookie')
-            if 'Cookie' in self.headers:
+            cookie = self.headers.get('Cookie') or self.headers.get('cookie')
+            if cookie is not None:
                 cookies = SimpleCookie()
-                cookies.load(self.headers['Cookie'])
+                cookies.load(cookie)
                 self._cookies = {name: cookie.value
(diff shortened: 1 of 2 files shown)


sanic/request.py
```diff
@@ -114,6 +114,8 @@ def args(self):
     @property
     def cookies(self):
         if self._cookies is None:
+            if 'cookie' in self.headers: #HTTP2 cookie header
```
AntonDnepr: I think two `if` here are redundant, this could be a bit simplified&rewrited
`cookie = self.headers.get('Cookie') or self.headers.get('cookie')
if cookie:
    cookies = SimpleCookie()
    cookies.load(cookie)
`

jackfischer:
Good call on the review, I was trying to think of a nicer way to do that.

jackfischer:
@AntonDnepr do you think we can merge this?

AntonDnepr:
@jackfischer on my opinion it would be nice to have also tests for situation you described, just to make sure that it wouldn't be never broken in future.

jackfischer:
I'll get on that asap.

jackfischer:
seemethere I have not done this before but I believe the test is set up correctly.

seemethere:
@jackfischer Looks great

PR #182, opened 2016-12-03
Fix request timeout response #181

### LINKED ISSUE #181: NoneType object has no attribute 'should_keep_alive'
Hi,

I have setup a small sanic server locally and I get the following error message from time to time... 

    2016-12-03 11:44:04 - Writing response failed, connection closed 'NoneType' object has no attribute 'should_keep_alive'

I could be wrong, but I does not appear to be related to any of my requests... it just happens from time to time. Any idea why this happens?

38elements:
If Request Timeout occurs, it writes response and closes transport without executing write_response().


38elements:
I am sorry.
I was misunderstanding.

Issue #168, opened 2016-11-25
Live reload server in develop phase


marianis123:
Hello, I'm trying to develop an api for my project using your framework but every time I save the changes I need to close and open server again, and it does it everytime, is there any way to make livereload in sanic?.

webtic:
Did some experimenting on auto-reloading on changes of the code to ease the development cycle. I can correctly detect a change in the project-files (using python watchdog) but how can I restart the whole server? Calling get_event_loop().stop() from the file-has-changed handler just gives a runtime error: RuntimeError: There is no current event loop in thread 'Thread-1'.

istommao:
I think livereload is useful in develop!

AoiKuiyuyou:
I have made a live reload library [AoikLiveReload](https://github.com/AoiKuiyuyou/AoikLiveReload) that is applicable to most of Python programs (and to crash a few of them). Here is the [demo for reloading Sanic server](https://github.com/AoiKuiyuyou/AoikLiveReload/blob/0.1.0/src/aoiklivereload/demo/sanic_demo.py).

Install the library with:
```
pip install AoikLiveReload
```

Add the 3 lines to your code:
```
from aoiklivereload import LiveReloader
reloader = LiveReloader()
reloader.start_watcher_thread()
```

Now when there is a module file change, the program will be reloaded.

istommao:
@AoiKuiyuyou 😄 禅师的库名 不敢恭维啊 啊哈哈哈哈哈

istommao:
Add livereload PR https://github.com/channelcat/sanic/pull/202

seemethere:
Rolling into 0.2.0

istommao:
@seemethere When to release 0.2.0

seemethere:
@istommao There's no official time-table but it's going to be focused on this change as well as the configuration and logging changes.

seemethere:
Going to cut a release for 0.2.0 today, will move this to the next milestone.

atbentley:
Just adding some thoughts on live reload.

Can we do this without adding watchdog to the requirements or setup file? Code running in production should have as little dependencies as possible and a development dependency should never go into production. Perhaps at the start of a debug server a check can be made to see if watchdog is supported, if so the dev server gets livereload otherwise a message along the lines of `to enable live reload run "pip install watchdog"`.

[...]

PR #187, opened 2016-12-08
Add Auto restart

38elements:
Change:
* When the debug flag is enable and the source code is changed,  the server will automatically restart.

This is related #168.

### DIFF
--- a/sanic/__main__.py
+++ b/sanic/__main__.py
@@ -36,6 +36,7 @@ class RestartHandler(PatternMatchingEventHandler):
         time.sleep(0.5)
         self._start()
 
+
 if __name__ == "__main__":
     parser = ArgumentParser(prog='sanic')
     parser.add_argument('--host', dest='host', type=str, default='127.0.0.1')
--- a/sanic/__main__.py
+++ b/sanic/__main__.py
@@ -1,8 +1,40 @@
+import os
+import sys
+import time
 from argparse import ArgumentParser
 from importlib import import_module
+from subprocess import Popen
+from signal import SIGTERM
+from watchdog.observers import Observer
+from watchdog.events import PatternMatchingEventHandler
 
 from .log import log
 from .sanic import Sanic
+
+
+class RestartHandler(PatternMatchingEventHandler):
+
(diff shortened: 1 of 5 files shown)


sanic/__main__.py
```diff
@@ -10,23 +10,70 @@
...
-        app = getattr(module, app_name, None)
-        if type(app) is not Sanic:
-            raise ValueError("Module is not a Sanic app, it is a {}.  "
-                             "Perhaps you meant {}.app?"
-                             .format(type(app).__name__, args.module))
-
-        app.run(host=args.host, port=args.port,
-                workers=args.workers, debug=args.debug)
+        if args.debug and not args.no_restart:
+            import os
```
seemethere: Can we break the imports out and put them at the top?

sanic/__main__.py
```diff
@@ -10,23 +10,70 @@
...
+            import sys
+            import time
+            from subprocess import Popen
+            from signal import SIGTERM
+            from watchdog.observers import Observer
+            from watchdog.events import PatternMatchingEventHandler
+
+            class RestartHandler(PatternMatchingEventHandler):
```
seemethere: Can we also break this class declaration out of the conditional?

38elements: @seemethere 
Thank you for pointing it out.
Import and class declaration moved to the top.

sanic/__main__.py
drinksober: in the class LiveLoadingHandler, can we add a method on_modified, on which we can listen the modified event and get the src_path, and the just reload the module of src_path

just like that 

    def on_modified(self, event):
        module = path and _find(event.src_path)
        if hasattr(module, '__spec__'):
            from importlib import reload
            app.router.reload()
            reload(module)
            reload(sanic)

def _find(path):
    try:
        return import_module(path)
    except (ImportError, TypeError):
        pass
    path = '.'.join(path.split('.')[:-1])
    if '.' in path:
        return _find(path)
    return path

istommao:
Does this implementation (watchdog) merge into the master branch?

PR #188, opened 2016-12-10
Find URL encoded filenames on the fs by decoding them first

webtic:
As-is sanic returns a 404 for files which are requested with a space in their filename.
This due to the browser url-encoding them first. By decoding the filename we make sure the file is actually found.


AntonDnepr:
@webtic may I ask you to leave a comment in code and, if possible, write a test for this one? I think bug should be documented&tested, because the situation you describe can't be easily spotted from the code.

webtic:
added a comment; when needed I will look into a test later

seemethere:
Just rewrote the tests for static files, will include it in a separate pull request.

Issue #178, opened 2016-12-01
Add documentation for response headers


blakev:
Just spent a long time trying to track down a bug related to headers where a key's value was an int.

The header dictionary must be in the format `[str, str]` or the response will silently fail.

```python
    headers = {
        'Age': str(int(time.time() - store_time)),
    }
```

wojunnihou:
可能还会有编码的问题,若能在设置中文的时候，智能识别编码就更好了

seemethere:
Maybe it would be useful to have it try to convert it to a str on it's own?

seemethere:
Closed per #197

PR #197, opened 2016-12-14
convert header values to strings

r0fls:
addresses idea raised in #178

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -1,8 +1,10 @@
 from aiofiles import open as open_async
-from .cookies import CookieJar, Cookie
 from mimetypes import guess_type
 from os import path
+
 from ujson import dumps as json_dumps
+
+from .cookies import CookieJar
 
 COMMON_STATUS_CODES = {
     200: b'OK',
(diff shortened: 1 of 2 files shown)


r0fls:
Alternatively we could raise an exception instead of failing silently.

sanic/response.py
```diff
@@ -97,6 +97,8 @@ def output(self, version="1.1", keep_alive=False, keep_alive_timeout=None):
         headers = b''
         if self.headers:
             headers = b''.join(
+                b'%b: %b\r\n' % (name.encode(), str(value).encode('utf-8'))
+                if type(value) != Cookie else
```
AntonDnepr: May I ask you to avoid use `type` here? I think `isinstance` should work (but please check this I could be wrong). And one more thing: the only difference here is in `str(value)`, so could you please just create additional variable and move `if ... else..` block there to avoid code duplication?

seemethere:
I'd agree on raising an Exception instead of silently failing. Doing the explicit str conversion makes too big of a performance hit, especially when you factor in the type checking as well.

sanic/response.py
seemethere: Yeah I'd like this changed to isinstance as well

seemethere:
In order to be performant we can pull the generator comprehension out and do a simple try/except on here that will catch if a header is not a str and then convert it, similar to what was done in: #215.

What do you think @r0fls?

r0fls:
@seemethere yes definitely, much smarter.

PR #199, opened 2016-12-16
added methods to load config from a file

Tim-Erwin:
Sanic is lacking an easy way to configure an application. This adds basically the a simplified version of Flasks configuration. I added `__getattr__` and `__setattr__` in order to preserve the possibility to access the values with dot-notation.

As a side-effect the `ROUTER_CACHE_SIZE` got moved to `router.py`. First of all it doesn't really work with the new config class. Second, it was misleading in the first place, because it suggested that it was configurable - but it wasn't (except by changing the code in config.py). When changing the config, the `lru_cache` decorator is already fixed and the change has no effect.


AntonDnepr:
Hi @Tim-Erwin !
Thank you for your work.
I'm not really experienced in the topic of configuration from file, but may I ask you why do you inherit from `dict` at the `Config`? I see that you have to declare `__getattr__` and `__setattr__` because of that. Is there any real advantage for using `dict`?
And also may I ask you to write tests for this feature?

sanic/config.py
```diff
@@ -1,5 +1,11 @@
-class Config:
-    LOGO = """
+import os
+import types
+
+
+class Config(dict):
+    def __init__(self, defaults=None):
+        dict.__init__(self, defaults or {})
```
AntonDnepr: Maybe it's better to use `super` here?

sanic/config.py
```diff
@@ -20,6 +26,63 @@ class Config:
...
 """
-    REQUEST_MAX_SIZE = 100000000  # 100 megababies
-    REQUEST_TIMEOUT = 60  # 60 seconds
-    ROUTER_CACHE_SIZE = 1024
+        self.REQUEST_MAX_SIZE = 100000000  # 100 megababies
+        self.REQUEST_TIMEOUT = 60  # 60 seconds
+        self.ROUTER_CACHE_SIZE = 1024
+
+    def __getattr__(self, attr):
+        return self[attr]
```
AntonDnepr: according to https://docs.python.org/3/reference/datamodel.html#object.__getattr__ this method should raise `AttributeError`, but in this case it will raise `KeyError` instead

Tim-Erwin:
Wow, that's a fast review, thanks. I let Config inherit because it gives a lot of flexibility for free, such as using `update()` in order to merge configurations. I wouldn't insist on that, however, I could remove that.
Sure, I'll add tests for that. Usually I just want to be sure that a feature is wanted before I make the effort to write tests :)

sanic/config.py
```diff
@@ -20,6 +26,63 @@ class Config:
...
+
+        You should not use this function to load the actual configuration but
+        rather configuration defaults. The actual config should be loaded
+        with :meth:`from_pyfile` and ideally from a location not within the
+        package because the package might be installed system wide.
+
+        :param obj: an object
+        """
+        for key in dir(obj):
+            if key.isupper():
```
AntonDnepr: can't we change this `for` cicle with built-in `map`? As I remember it also faster then `for`

sanic/config.py
```diff
@@ -20,6 +26,63 @@ class Config:
...
+    def from_pyfile(self, filename):
+        """Updates the values in the config from a Python file. Only the uppercase
+        varibales in that module are stored in the config.
+        :param filename: an absolute path to the config file
+        """
+        module = types.ModuleType('config')
+        module.__file__ = filename
+        try:
+            with open(filename) as config_file:
+                exec(compile(config_file.read(), filename, 'exec'),
```
Tim-Erwin: Well, it's not really third-party, you're including your own file. It's quite common to use a python module as configuration. So I'd call that a well-proven procedure :)

sanic/config.py
Tim-Erwin: By the way, speed should really not be of any concern here, this is only executed at startup. Readability is far more important in this area.

sanic/config.py
AntonDnepr: ok, thank you :)

sanic/config.py
AntonDnepr: I just made quick performance test and this really doesn't matter in this case. +1 about readability, so sorry about disturbing :)


```

File: sanic/config.py [hunk 5]
```diff
@@ -20,6 +26,63 @@ class Config:
...
+            app.config.from_object(default_config)
+
+        You should not use this function to load the actual configuration but
+        rather configuration defaults. The actual config should be loaded
+        with :meth:`from_pyfile` and ideally from a location not within the
+        package because the package might be installed system wide.
+
+        :param obj: an object
+        """
+        for key in dir(obj):
```
AntonDnepr: sorry about disturbing, doesn't really matter, existing code is ok

Tim-Erwin:
So, what do you think, @AntonDnepr , leaving as `dict` ok?

AntonDnepr:
@Tim-Erwin thank you for explaining. As I said I'm not really experienced in this topic, so I was curious about this.
As for me it would be nice to have this feature, but we need to summon @seemethere here

sanic/config.py
AntonDnepr: I don't think that built-in dict will inherit a lot of other stuff, but please go with `super` here just in case

sanic/config.py
seemethere: you could just dynamically import it with importlib if you want to

Tim-Erwin: I tried using importlib, however it fails when the config file does not have the .py suffix. Anyway, whether we are loading with importlib or with compile/exec does not really matter, in both cases we're executing python code. Or is there any benefit I'm missing?

Tim-Erwin:
Ok, I added some documentation. I think I've change everything you asked me to. If there's nothing more to change, I think we're good to go.

Tim-Erwin:
@channelcat , this is marked for release tomorrow. Any thoughts on this? Would love to have this merged.

seemethere:
@Tim-Erwin I'm going to cut a release for 0.2.0 today, I'll move this to the next release.

Tim-Erwin:
I'll just keep resolving conflicts :(

sanic/router.py
```diff
@@ -15,6 +14,8 @@
     'alpha': (str, r'[A-Za-z]+'),
 }
 
+ROUTER_CACHE_SIZE = 1024
```
r0fls: Why not leave this in as a built in configuration value?

Tim-Erwin: Because it actually cannot be modified. The value is read and applied only once when the sanic module is imported, any further changes don't have any effect.

Tim-Erwin:
@seemethere , why is that removed from the milestone? This is super easy and backwards compatible and not in a performance critical path and the code is (as mentioned before) borrowed from Flask to quite some extent, so it can be considered tested in production environments. There is documentation and tests and nobody seems to have serious problems with it. If something is missing or not conforming to Sanic standards, I'd be happy to change or add missing parts. Please let me know how to get this done.

seemethere:
Rolling into `0.3.1`

Tim-Erwin:
Thanks

PR #201, opened 2016-12-18
The HTTPMethodView is Inconvenient

wojunnihou:
The HTTPMethodView is Inconvenient.so I desgin this
HTTPMethodView 用起来不方便，所以我仿照tornaodo重新设计了下.
业务逻辑较为复杂的时候，需要在不同的方法函数中处理，但请求的上下文，需要一个实例来存储。这样的的有点在于：
1，请求上下文统一存储，不必在方法或者函数中显式传递参数；
2，任何请求处理过程都可以修改http请求的结果，不必统一显式传参直到最后make_response的时候进行处理。

### DIFF
--- a/examples/base_handle.py
+++ b/examples/base_handle.py
@@ -9,7 +9,8 @@ bp = Blueprint('my_bluprint')
 class MyHandle(BaseHandle):
     '''
     封装成类，请求单上下文，可以方便的存取
-    Package into a class, request a single context, you can easily access(Baidu Translate)
+    Package into a class, request a single context,
+     you can easily access(Baidu Translate)
     '''
 
     async def post(self, request, *args, **kwargs):


wojunnihou:
我改了，pep8

wojunnihou:
通常习惯用pycharm 格式化，但发现这个代码里面，很多代码木有格式话．我都不好意思自己去格式化下代码．哈哈

AntonDnepr:
Hello @wojunnihou
It's a bit hard to understand the description for me (I tried google translate, but some parts still not really clear).
So as far as I understand this is some kind of syntax sugar + some extra stuff like `make_xml`+`make_image`+`make_redirect` (does sanic really don't have redirect yet? I can't find it with a quick view)
But apart from pros I see several cons in this PR:
1) Really lack of any docs&comments. Example in doc don't introduce new methods
2) Lack of any benchmark data. It's fully new `Response` class - is there any benchmarks?
3) `route_handle` - it's added only for Blueprints. Why only for them? And no mention about them in routing docs.

And here is just some of my concerns about this at all:
1) Do we really need new `Response` class? I mean it have some cool stuff like make_xml, make_image, make_redirect - but they could be added to existing one. I understand that it's done to make `self.response` possible but this leads to next concern.
2) Now we have several ways to make one thing. I really do not like this approach - two ways to build class-based views with really big difference in syntax - as for me it is not the good thing to go. I don't mind if we remove `HTTPMethodView` and replace it with `BaseHandle`, but in my opinion there shouldn't be two approached with really different syntax - this would be confusing and hard to maintain in projects.
3) And on my opinion taking in mind second paragraph: there could be a problem with KISS+DRY principles. This PR create it's own `Response` class with it's own implementation of already implemented methods.
4) `make_` style is a bit different from sanic default way to built a responses, but I think this could be easily fixes if need

wojunnihou:
你说的问题，我基本清楚了，我尽快改善。
这个框架（sanic）本身response真的不好用。另外跟flask不一样的是这个没有多线程，所以每次请求不能存线程变量中，我认为为每个请求创建一个实例来存储请求生命周期的变量较为好统一管理。因此我写了route_handle一个装饰器，来为每次请求创建一个实例。

luizfb:
I don't see what benefits does adding another Response object on top of `HTTPResponse` bring to the table. Sure there are a ton of formats Sanic could support (and there really should be an effort to add redirects if Sanic really still doesn't have them), but ultimately focusing on the "most used formats" is more than enough, specially since Sanic makes it trivial to add our own custom replies by creating wrappers that return `HTTPResponse` objects, much like the `json` and `text` functions work.

I think it's remarkable that Sanic managed to simplify even something as simple as Flask, not having to deal with crossed imports from werkzeug, templating engines (which people who use Flask for backend and API development never end up using), functions and objects with the same interfaces that do the same thing (`Response` and `make_response`). Sanic looks like what Flask would be after some serious refactoring, I think it would be smart to avoid adding unnecessary surface especially since these things can be trivially added by whoever is using Sanic (or someone can just make another project that can be plugged in and adds support to a multitude of other response formats, it's also easy to do and avoids pulling in extra stuff into Sanic).

examples/base_handle.py
```diff
@@ -0,0 +1,27 @@
...
+from sanic.sanic import Sanic
+from sanic.views import BaseHandle
+
+bp = Blueprint('my_bluprint')
+
+
+@bp.route_handle('/test')
+class MyHandle(BaseHandle):
+    '''
+    封装成类，请求单上下文，可以方便的存取
```
pahaz: Please use the international style in docstrings

wojunnihou:
刚腾出点时间，准备修改。但发现这个框架风格跟我想的有较大的区别。因此我关掉了这个pr

PR #202, opened 2016-12-19
Add livereload feature

istommao:


sanic/sanic.py
```diff
@@ -313,6 +299,60 @@ def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
 
         log.info("Server Stopped")
 
+    def run_with_reload(self, **dataset):
+        # Run reloader thread
+        event_loop = start_reloader_thread(self, interval=1)
+
```
seemethere: Can we remove the spaces in between the statements here? More of a style thing than anything else but I feel as though they're not needed.

livereload_example.py
```diff
@@ -0,0 +1,12 @@
+from sanic import Sanic
```
seemethere: I don't think this is necessary at the top level.

istommao:
@seemethere agree with you, i think watchdog maybee the better way!

Issue #208, opened 2016-12-22
Streaming support


alexsv:
Hi,
I'm impressed by the framework, but want to check one thing:
I want to read big file and send a portions of it like 1KB by 1 second

So I expect, that HTTPResponse would accept some iterator.
Any ideas about this?

seemethere:
You could implement a custom protocol based on the base sanic HTTPProtocol.

Closed with #209

alexsv:
Am I right custom protocol is used for the whole application and I can't make one endpoint streaming based?

seemethere:
@alexsv that would be correct. But I would think that there could be a better implementation out there that would not require changing the entire protocol just to enable it for one endpoint.

alexsv:
possible on blueprint level?

seemethere:
Possibly but it would need to be investigated.

38elements:
This is related #697.

Issue #115, opened 2016-10-26
Failure to write request swallows problem


jasonab:
server.py, line 137-139:
If the response is malformed, Sanic will log an error (thankfully), but will not return any kind of response (simply killing the connection).

Sanic should generate an exception that can be handled by the error handler, and ultimately return a 500.

Note that this can be generated by attempting to return a Flask-type tuple instead of a Sanic response object.

r0fls:
@seemethere the natural thing to do here in my opinion is to import the exceptions and raise one if the response from the server is invalid. However that would cause a circular import because exceptions uses response.py. Should we refactor, and include exceptions in response.py? That's tye typical way to deal with circular imports, since if they both rely on each other there is no reason to keep them separate. 

Or, can you think of another way of doing this?

jasonab:
Method write_response:
 except Exception as e:
            self.bail_out(
"Writing response failed, connection closed {}".format(e))

seemethere:
Closed with #223

PR #209, opened 2016-12-22
Customizable protocol

38elements:
In the current implementation, the developer can not change the behavior of the protocol.
By implementing a subclass of HttpProtocol and making it an argument of the `run` function, 
the developer can change the behavior when data is received and the behavior when an error occurs.
This is related #115 and #208 .


sanic/server.py
```diff
@@ -221,12 +221,13 @@ def trigger_events(events, loop):
                 loop.run_until_complete(result)
 
 
-def serve(host, port, request_handler, error_handler, before_start=None,
-          after_start=None, before_stop=None, after_stop=None,
-          debug=False, request_timeout=60, sock=None,
+def serve(protocol, host, port, request_handler, error_handler,
```
seemethere: Would it be smart to put this at the front of the arguments list and possibly break functionality for anyone who uses this? I think putting this at the end and defaulting it to our own HttpProtocol would be better in this case.

38elements:
@seemethere 
Thank you for reviewing.
Your pointing out is correct.
I fixed this code.

38elements:
@seemethere 
Of course I will make document and test.

38elements:
@seemethere 
I made document, example and test.

Issue #176, opened 2016-11-29
Make python's log-format editable by others using sanic


x10an14:
Hiya,

Sanic uses python's logging (in `sanic/log.py`) in a fashion which overrides the ability of others to set it they way they want to.

Could this be refactored into making it configurable by others? I'm not that into python logging/adept at it, but I'll try to make a PR this weekend if nobody else beat me to it!

Example code which gets overwritten in my script:
```python
(...)
from src import app # -> from sanic import Sanic; app = Sanic("my-module")
(...)

if __name__ == '__main__':
    debug = False

    logging_format = "[%(asctime)s] %(process)d-%(levelname)s "
    logging_format += "%(module)s::%(funcName)s():l%(lineno)d: "
    logging_format += "%(message)s"
    logging.basicConfig(
        format=logging_format,
        datefmt=default_parse_fmt + "%f",
        level=logging.DEBUG if debug else logging.INFO, )

    app.run(host="127.0.0.1", port=8000, debug=debug)
```

`sanic/log.py`'s config (below) overwrites mine, even though I only invoke it by importing sanic in my own `__init__.py` file:
```python
import logging

logging.basicConfig(
    level=logging.INFO, format="%(asctime)s: %(levelname)s: %(message)s")
log = logging.getLogger(__name__)
```

The `log.py` module in sanic is barely used anywhere. I easily managed to remove it and replace it with the use specified [here](https://docs.python.org/3.5/howto/logging.html#logging-from-multiple-modules). Everything still worked!

Proof:
```
[29-11-2016 20:04:07%f] 19034-DEBUG sanic::run():l281 >
                 ▄▄▄▄▄
        ▀▀▀██████▄▄▄       _______________
      ▄▄▄▄▄  █████████▄  /                 \
     ▀▀▀▀█████▌ ▀▐▄ ▀▐█ |   Gotta go fast!  |
   ▀▀█████▄▄ ▀██████▄██ | _________________/
   ▀▄▄▄▄▄  ▀▀█▄▀█════█▀ |/
        ▀▀▀▄  ▀▀███ ▀       ▄▄
     ▄███▀▀██▄████████▄ ▄▀▀▀▀▀▀█▌
   ██▀▄▄▄██▀▄███▀ ▀▀████      ▄██
▄▀▀▀▄██▄▀▀▌████▒▒▒▒▒▒███     ▌▄▄▀
▌    ▐▀████▐███▒▒▒▒▒▐██▌
[...]

PR #212, opened 2016-12-23
allow overriding logging basicConfig settings

r0fls:
Addresses #176. Defaults to current behavior.

### DIFF
--- a/examples/override_logging.py
+++ b/examples/override_logging.py
@@ -1,0 +1,23 @@
+from sanic import Sanic
+from sanic.response import text
+import json
+import logging
+
+logging_format = "[%(asctime)s] %(process)d-%(levelname)s "
+logging_format += "%(module)s::%(funcName)s():l%(lineno)d: "
+logging_format += "%(message)s"
+
+logging.basicConfig(
+    format=logging_format,
+    level=logging.DEBUG
+)
+log = logging.getLogger()
+
+# Set logger to override default basicConfig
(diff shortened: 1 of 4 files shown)


r0fls:
Yes, good points. I'm working on a test case but for some annoying reason this test fails when run as part of the group, but passes when run individually (with `python3 -m pytest -vv tests/test_logging.py`).

```python
import asyncio
from sanic import Sanic
from io import StringIO
from sanic.utils import sanic_endpoint_test
import logging

logging_format = '''module: %(module)s; \
function: %(funcName)s(); \
message: %(message)s'''

def test_log():

    log_stream = StringIO()
    logging.basicConfig(
        format=logging_format,
        level=logging.DEBUG,
        stream=log_stream
    )   
    log = logging.getLogger()
    app = Sanic('test_logging', logger=True)
    @app.route('/')
    def handler(request):
        log.info('hello world')
        return text('hello')

    request, response = sanic_endpoint_test(app)
    log_text = log_stream.getvalue().strip().split('\n')[-3]
    assert log_text == "module: test_logging; function: handler(); message: hello world"
```

r0fls:
@seemethere looks like basicConfig settings are overlapping between separate tests. I'll probably open an issue with pytest since that seems like a bug. Fixed here by clearing existing log handlers in the unit test. Note that is not required in the example.

seemethere:
I'll merge once we do a release of 0.1.9 👍

PR #213, opened 2016-12-23
Make it possible to disable the logo by subclassing Config

sfstpala:
Just a small thing. I have seen it too often. :)

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -294,7 +294,8 @@ class Sanic:
 
         if debug:
             log.setLevel(logging.DEBUG)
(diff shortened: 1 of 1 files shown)


sfstpala:
I do it something like this:

    class Sanic(sanic.Sanic):
        def __init__(self, ...):
            ...
            super().__init___(...)
            self.config = MyConfig()

I would also be happy to pass config as a keyword arg to `Sanic.__init__` of course if you prefer that.

rednebmas:
Disabling the logo can also be done as follows

```python
import sanic.config
sanic.config.Config.LOGO = ""
```

PR #214, opened 2016-12-23
Added features to do Sanic more plugable

cr0hn:
This PR try to add some features that already exists in other frameworks (like aiohttp) and that are useful for the plugin development:
- add te "app" property into each request, with te Sanic instances 
- add the property "handler" in each request, that contains the view handed where the request was originated 
- new method to add a middleware: add_middelware

The features do not affect to the performance (I was tested) and the could be very useful try to grow the framework.

request.py
----------

add: new property: 'app'. This property contains an instance of Sanic application.
add: new property: 'handler'. This property will be setted before a request is sent to the view or middleware and contains the handler of view. This is useful to know all information of view, for example: if any plugin decorates the function, only in runtime the middleware or view will know this information. The most common case is when we want to decorate a view o use a cache system. The middleware must know who view are marked as a cached.

sanic.py
--------

add: new method to append a middleware. The idea is to set middleware easily, thinking in possible plugins. Using this way, a plugin could use this method to add their custom middleware. For example: An authentication system or cache.
modified: reordered the the call of "self.router.get(...)", to add the "handler" property to the request before any middleware is called.

server.py
---------

add: app instance in constructors

### DIFF
--- a/sanic/__init__.py
+++ b/sanic/__init__.py
@@ -1,6 +1,6 @@
 from .sanic import Sanic
 from .blueprints import Blueprint
 
-__version__ = '0.1.8'
+__version__ = '0.1.9'
 
 __all__ = ['Sanic', 'Blueprint']
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -39,13 +39,14 @@ class Request(dict):
     """
     __slots__ = (
         'url', 'headers', 'version', 'method', '_cookies',
-        'query_string', 'body',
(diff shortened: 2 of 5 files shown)


cr0hn:
If with featured could be added, I have some plugins that I can contribute: a cache system. a swagger plugin, a hardening plugin and some more that I would need to adapt.

cr0hn:
@seemethere what do you think about that? If I review the patch to try to pass all test, I'll consider to add these features? Thx

PR #215, opened 2016-12-23
Improvement: avoid to encoding in each HTTP Response

cr0hn:
improvement: support fo binary data as a input. This do that the response process has more performance because not encoding needed.

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -1,8 +1,10 @@
 from aiofiles import open as open_async
-from .cookies import CookieJar
 from mimetypes import guess_type
 from os import path
+
 from ujson import dumps as json_dumps
+
+from .cookies import CookieJar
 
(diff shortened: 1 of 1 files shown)


sanic/response.py
```diff
@@ -79,7 +79,9 @@ def __init__(self, body=None, status=200, headers=None,
         self.content_type = content_type
 
         if body is not None:
-            self.body = body.encode('utf-8')
+            self.body = body
+            if type(body) is str:
```
seemethere: You're going to want to use `isinstance` for this one.

cr0hn: isinstance was not used because in many test performance seem slower than type(...)

seemethere:
👍 A test will be added later on that will test this scenario.

PR #218, opened 2016-12-23
Improvement: Set custom backlog

cr0hn:
By default, asyncio runs the server with a backlog of 128 concurrent users, but this value can be customized. This simply path will allow that.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -224,7 +224,7 @@ def trigger_events(events, loop):
 def serve(host, port, request_handler, error_handler, before_start=None,
           after_start=None, before_stop=None, after_stop=None,
           debug=False, request_timeout=60, sock=None,
-          request_max_size=None, reuse_port=False, loop=None):
+          request_max_size=None, reuse_port=False, loop=None, backlog=128):
     """
     Starts asynchronous HTTP Server on an individual process.
     :param host: Address to host on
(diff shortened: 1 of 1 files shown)


sanic/server.py
```diff
@@ -240,6 +240,7 @@ def serve(host, port, request_handler, error_handler, before_start=None,
     :param request_max_size: size in bytes, `None` for no limit
     :param reuse_port: `True` for multiple workers
     :param loop: asyncio compatible event loop
+    :param backlog: maxinum number of concurrent clients
```
seemethere: Typo here on `maximum`.

38elements:
I think that putting `backlog` parameter in [run()](https://github.com/channelcat/sanic/blob/5afae986a0aeae6191a5532a11244cb7cc405f94/sanic/sanic.py#L233-L235) is necessary.

seemethere:
Also @cr0hn is there a way we can write a test for this scenario?

cr0hn:
I don't know how test this kind of features as a unit testing. Any idea?

seemethere:
I'd also agree with @38elements here as well this needs to be include in both run and server_settings

r0fls:
Whoops, I didn't notice this when I implemented this here:
https://github.com/channelcat/sanic/pull/269

This would be nice to have a unit test for though. So if you managed to write one, please make another PR.

Issue #41, opened 2016-10-16
Add some more documentation in some of the source code?


abhishek7:
Not too sure if this is planned, but figured it would help in terms of further contributions as the codebase continues to grow.

akshbn:
I will take this up!

seemethere:
Closed per #228

PR #228, opened 2016-12-25
Adding more docstrings

seanpar203:
I plan on adding more docstrings but I think this is a good start.


seemethere:
Relates to Issue #41

Issue #205, opened 2016-12-22
Convert documentation to Sphinx?


sirex:
Having Sphinx documentation you will be able to upload docs to https://readthedocs.org/

seemethere:
I'm a fan :+1:

wsbmilk:
+10086, alternatively gitbook

kdelwat:
I'm willing to implement this in a PR if there's a consensus 🎉

wilcoschoneveld:
This issue is resolved and can be closed, publishing discussed in #321.

PR #229, opened 2016-12-25
Use Sphinx for documentation

kdelwat:
Following on from #205, I have integrated Sphinx documentation into the project.

This PR creates the necessary config and index files for the change. The [recommonmark](https://readthedocs.org/projects/recommonmark/) extension is included to allow for Markdown documentation, such as the current guides. `sphinx-apidoc` is used to automatically generate additional documentation from the current docstrings.

Upon initially running the documentation build, I found a number of docstring syntax errors that I have fixed in the second commit 😄.

The build process for the documentation is now:

```bash
m -f docs/sanic.rst
rm -f docs/modules.rst
sphinx-apidoc -o docs/ sanic
sphinx-build -b html docs docs/_build
```

HTML is generated in the `docs/_build` directory.

It should now be possible to integrate the repository with ReadTheDocs. This would hopefully make Sanic more accessible to new users and increase adoption 🎉.


.gitignore
```diff
@@ -9,3 +9,6 @@ settings.py
 *.pyc
 .idea/*
 .cache/*
+docs/_build/
+docs/sanic.rst
+docs/modules.rst
```
wilcoschoneveld: I think it would be less confusing if the apidoc files are generated in a _* folder

kdelwat:
@wilcoschoneveld Thanks for the detailed feedback! I'll outline some of my reasoning below:

# Gitignore changes, apidoc build location 
I agree with you on these ones - looks good. 

# Separate index.rst and README.md

This one really annoyed me when I was setting it up! I'd love to just include the current README.md in the index, but unfortunately the Sphinx `include` directive only supports RestructuredText. Are you aware of another way around this issue :smile:? 

# Makefile
I debated a Makefile but decided against it because I didn't want to add infrastructure like that for only one part of the build process - it seemed a bit overkill to me. However, I think it would be great to have one that covers docs, testing, cleaning etc. Maybe one of us could make a separate PR?

# sphinx-apidoc

Another issue for me when I was setting up. Generally I'm a fan of manual documentation as you suggested. However,  for a couple of reasons I decided to include it:

+ For a pre-1.0 release I think it's helpful to have the docs adapt automatically to API changes to make sure they don't get out of sync, although this could be alleviated if we made it a requirement for all PRs to contain the corresponding doc changes. 
+ Currently the module is missing some docstrings, and this could be a good way to encourage greater coverage. 

Please let me know what you think and I'll make the changes later today!

wilcoschoneveld:
Maybe then convert README.md to .rst? GitHub would also render it. Using sphinx-apidocs, at least until the api is stable, sounds like a good idea. Right now there are some conflicts that need to be resolved before a merge can happen. @seemethere, what do you think?

kdelwat:
OK, I've converted the README and moved the autogenerated documentation to the `docs/_api` directory.

seemethere:
I can work on fixing the merge conflicts, I do like the direction though.

wilcoschoneveld:
Looks good to me in the current state! @kdelwat awesome work! @seemethere can you merge this?

seemethere:
Going to cut a release to 0.2.0 will move this to the next milestone

seemethere:
@kdelwat I can no longer fix the merge conflicts on the web client. Can you please resolve them locally?

kdelwat:
@seemethere Sure, I'll get onto it today.

kdelwat:
@seemethere Do you have a preference between a merge commit or rebase?

seemethere:
@kdelwat I think a merge commit would be better in this case and I'll just a squash and merge when we do eventually merge this

kdelwat:
@seemethere All done! I've also added the new guides to the documentation index.

docs/contributing.md
```diff
@@ -6,5 +6,19 @@ Thank you for your interest!
...
 
+## Documentation
+
+Sanic's documentation is built using [sphinx](http://www.sphinx-doc.org/en/1.5.1/). Guides are written in Markdown and can be found in the `docs` folder, while the module reference is automatically generated using `sphinx-apidoc`.
+
+To generate the documentation from scratch:
+
+```bash
+rm -f docs/_api/*
+sphinx-apidoc -o docs/_api/ sanic
```
wilcoschoneveld: you can also build apidoc with -f to overwrite

kdelwat:
More merge conflicts fixed (I'm fighting a losing battle!) and @wilcoschoneveld's note added - I didn't know about the `-f` flag, thanks!

PR #230, opened 2016-12-25
Refactor write_error()

38elements:
When an error occurs in write_error(), process cycles bail_out() and write_error().
This behavior  is not appropriate.

Change:
This closes transport and writes log when an error occurs in write_error().


seemethere:
Does this change anything from the user-perspective as far as being able to see the error?

38elements:
I am sorry.
I was misunderstanding.

PR #231, opened 2016-12-25
Add smart response handling

seemethere:
Adds in smart response handling as seen in issue #50.

From what I tested doesn't seem to negatively affect performance too much if you are using the already built-in response objects.

I'm a little wary of merging this in though as it makes the process of returning a response a little ambiguous and makes things more of a "more than one way" to do it kind of style. There will be a another PR relating to this that addresses these issues but this one and that one cannot coexist.

I'll leave this up for a bit for some discussion.

NOTE:
- This will fail the `test_invalid_response` test. If we do end up merging this I will rewrite that test.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -164,12 +164,9 @@ class HttpProtocol(asyncio.Protocol):
                 # as fast as ever
                 attempt_write(response)
             except AttributeError:
-                try:
-                    # A performant way to check if we have a list or dict
-                    # Both list and dict contain a `clear` function
-                    response.clear
+                if isinstance(response, (list, dict)):
                     attempt_write(json(response))
-                except AttributeError:
+                else:
                     attempt_write(text(str(response)))
 
             if not keep_alive:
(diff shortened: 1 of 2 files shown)


sanic/server.py
```diff
@@ -147,12 +149,29 @@ def on_message_complete(self):
...
+            try:
+                # If we already have an HTTPResponse object this should be
+                # as fast as ever
+                attempt_write(response)
+            except AttributeError:
+                try:
+                    # A performant way to check if we have a list or dict
+                    # Both list and dict contain a `clear` function
+                    response.clear
+                    attempt_write(json(response))
```
r0fls: This would be cleaner and theoretically more performant if it used a type map instead of the doubly nested try/except:

```python
TYPE_MAP = {
    str: text,
    dict: json
     }

try:
    attempt_write(response)
except AttributeError:
    try:
        attempt_write(TYPE_MAP(response))
    except Exception as e:
        self.bail_out("Writing response failed,\
                      connection closed {}".format(e))
```

38elements:
I think that it is better to make some protocols according to the purpose,
Because developer can select a protocol.
For example, fast protocol and feature rich protocol and base protocol for inheritance are.

subyraman:
I think this might end up confusing, especially since there is no handling for `html` types. The current `text`, `json`, and `html` methods seem like a solid API.

PR #233, opened 2016-12-25
Add token property to request

seanpar203:
I added a `token` property to the request object which makes it easier to access the 'Authorization' header via `request.token`

``` py
    @property
    def token(self):
        """
        Attempts to return the auth header token.
        """
        if auth_header is not None:
            return auth_header.split()[1]
        return auth_header
```

The way it's set up is to work with common `Authorization` headers which normally put some text before the actual token like `Bearer`, `Basic` or `Token`. Regardless of what's in front of the actual token, it returns the token.

I added a test to confirm that it's working as intended.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -8,8 +8,9 @@ from sanic.exceptions import InvalidUsage
 
 from .log import log
 
-
 DEFAULT_HTTP_CONTENT_TYPE = "application/octet-stream"
+
+
 # HTTP/1.1: https://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html#sec7.2.1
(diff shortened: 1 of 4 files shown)


sanic/request.py
```diff
@@ -73,6 +74,17 @@ def json(self):
         return self.parsed_json
 
     @property
+    def token(self):
+        """
+        Attempts to return the auth header token.
+        :return: token related to request
+        """
+        auth_header = self.headers.get('Authorization', None)
```
AntonDnepr: you don't need to set `None` as default for `get()`

seanpar203: I just pushed a commit addressing that, thanks!

PR #236, opened 2016-12-26
Add token property to request

seanpar203:
I added a token property to the request object which makes it easier to access the 'Authorization' header via request.token

 ``` py
   @property
    def token(self):
        """
        Attempts to return the auth header token.
        """
        if auth_header is not None:
            return auth_header.split()[1]
        return auth_header
```

The way it's set up is to work with common Authorization headers which normally put some text before the actual token like `Bearer`, `Basic` or `Token`. Regardless of what's in front of the actual token, it returns the token.

I added a test to confirm that it's working as intended.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -73,6 +73,17 @@ class Request(dict):
         return self.parsed_json
 
     @property
+    def token(self):
+        """
+        Attempts to return the auth header token.
+        :return: token related to request
+        """
+        auth_header = self.headers.get('Authorization')
+        if auth_header is not None:
+            return auth_header.split()[1]
+        return auth_header
+
+    @property
     def form(self):
         if self.parsed_form is None:
             self.parsed_form = RequestParameters()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by seanpar203** [comment]
This doesn't have to be a property on the `Request` so I created another solution which is a decorator function. regardless, I think accessing the token value from the `Authorization` header should be simplified.

``` py
def auth_token(func):
    """ Creates a decorator for routes expecting an auth token.

    In the event that a proper token exists it will attach it to the
    `request` object under the `token` key.
    """

    async def decorator(self, request):
        auth_header = request.headers.get('Authorization')

        if auth_header is not None:
            try:
                prefix, token = auth_header.split()
            except ValueError as e:
                error_msg = 'Invalid Authorization header {}'.format(e.args)
                raise ValueError(error_msg)
            else:
                request['token'] = token
                return await func(self, request)
        else:
            raise ValueError('No Authorization Header Value.')

    return decorator
```

I haven't added tests for the decorator function, I've only used it locally and got the expected results.

seemethere:
l like the idea of the property 👍

PR #239, opened 2016-12-27
Added a uri property to the Request class (like uri_template of Falco…

dutradda:
…n Request class)


seemethere:
How does this actually get passed into the request?

seemethere:
Closing due to inactivity.

PR #240, opened 2016-12-27
fix class<Router> doc string typo

neo1218:


38elements:
I think that `<my_paramter:type>` rather than `<my_paramter>:type` is correct.
And Following error occurred.

```
sanic/router.py:34:80: E501 line too long (80 > 79 characters)
```

How about the following?

```
    Usage:
        @sanic.route('/my/url/<parameter>', methods=['GET', 'POST', ...])
        def my_route(request, parameter):
            do stuff...
    or
        @sanic.route('/my/url/<paramter:type>', methods=['GET', 'POST', ...])
        def my_route_with_type(request, parameter):
            do stuff...
```

```
the parameter. Given parameter must be able to be type-casted to this.
```

seemethere:
Was taken care of as of https://github.com/channelcat/sanic/pull/259

PR #242, opened 2016-12-27
Add livereload feature, use watchdog

istommao:
WatchdogReloaderLoop  is refer to werkzeug  WatchdogReloaderLoop


setup.py
```diff
@@ -31,6 +31,7 @@
         'ujson>=1.35',
         'aiofiles>=0.3.0',
         'multidict>=2.0',
+        'watchdog>=0.8.3'
```
JordanP: watchdog is only required if in debug and development phases.If you add it here in that list, you would only complicate work of downstream distros who would want to package Sanic. More generally is a bad design to mix your test requirements with your runtime requirements.

PR #243, opened 2016-12-27
created methods to remove a route from api/router

dutradda:


tests/test_routes.py
```diff
@@ -356,3 +356,81 @@ def test_add_route_method_not_allowed():
...
+    app = Sanic('test_remove_static_route')
+
+    async def handler1(request):
+        return text('OK1')
+
+    async def handler2(request):
+        return text('OK2')
+
+    app.add_route(handler1, '/test')
+    app.add_route(handler2, '/test2')
```
seemethere: Should we also assert that the route was added successfully? Or else this test could return a false positive if a route isn't added successfully.

sanic/router.py
```diff
@@ -23,6 +23,10 @@ class RouteExists(Exception):
     pass
 
 
+class RouteDontExists(Exception):
```
seemethere: Can we change this to `RouteDoesNotExist`? I think it works better in telling what has happened. Or even a `RouteNotFound` exception would suffice in this case.

dutradda:
changes made. I will wait for the 0.2.0 release

PR #244, opened 2016-12-27
Added features to do Sanic more pluggable (example included)

cr0hn:
This PR try to add some features that already exists in other frameworks (like aiohttp) and that are useful for the plugin development:
- add te "app" property into each request, with te Sanic instances 
- add the property "handler" in each request, that contains the view handed where the request was originated 
- new method to add a middleware: add_middelware

The features do not affect to the performance (I was tested) and the could be very useful try to grow the framework and create plugins like:

Example of a A cache system
---------------------------

```python

import asyncio
import uvloop

from aiocache import RedisCache, MemcachedCache, SimpleMemoryCache
from aiocache.serializers import JsonSerializer

#
# Decorator
#
class cache(object):
    def __init__(self, expires: int = 3600, unless: bool = False, use_post_data = False):
        self.expires = expires
        self.unless = unless
        self.use_post_data = use_post_data
    
    def __call__(self, f):
        f.cache_enable = True
        f.cache_expires = self.expires
        f.cache_unless = self.unless
        f.cache_user_post_data = self.use_post_data
        
        return f

#
# Middleware
#
async def cache_middleware(request):
    _handler = request.handler
    try:
        if getattr(_handler, "cache_enable", False):
            #
            # Cache is disabled?
            #
            if getattr(_handler, "cache_unless", False) is True:
                return await _handler(request)
        
            cache_backend = request.app.config.cache
        
            #
            # Make key
            #
            key = "{method}#{path}#{post_data}".format(
                method=request.method,
                path=request.url,
                post_data=request.body if getattr(_handler, "cache_user_post_data") else "")
        
            cached_response = await cache_backend.get(key)
            if cached_response:
                return HTTPResponse(**cached_response)
        
            #
            # Generate cache
            #
            original_response = await _handler(request)
        
            data = dict(status=original_response.status,
                        content_type=original_response.content_type,
                        headers=original_response.headers,
                        body=original_response.body)
        
            expires = getattr(_handler, "cache_expires", 300)
        
            await cache_backend.set(key, data, ttl=expires)
        
            return original_response
    except Exception as e:
        return text(str(e))

    # Not cached
    return await _handler(request)


#
# Cache configuration
#
def setup_cache(app: Sanic,
                cache_type: str = "memory",
                loop=None):
    _cache_backend = None
    if cache_type.lower() == "memory":
        _cache_backend = SimpleMemoryCache(serializer=JsonSerializer())
    
    elif cache_type.lower() == "redis":
        _cache_backend = RedisCache(serializer=JsonSerializer(), loop=loop)
    
    elif cache_type.lower() == "memcached":
        _cache_backend = MemcachedCache(serializer=JsonSerializer(), loop=loop)
    
    else:
        raise HTTPResponse("Invalid cache type selected")
    
    app.config.cache = _cache_backend
    app.add_middleware(cache_middleware)

#
# Main
#   

app = Sanic()
    
@app.route("/")
@cache()
async def test(request):


setup_cache(app, loop=loop, cache_type="memory")
# setup_cache(app, loop=loop, cache_type="redis")
    
app.run(host='127.0.0.1',
        port=8000,
        loop=loop)    
```

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -116,19 +116,21 @@ class Sanic:
         else:
             attach_to = args[0]
             return register_middleware
-    
+
     def add_middleware(self, middleware, attach_to="request"):
         """
-        register a middleware and configure to be called before a request of after
-        
+        register a middleware and configure to be called before a
+        request of after
+
         :param middleware: a callable function (coroutine)
         :type middleware: callable
-        
-        :param attach_to: where to attach the middleware: before a request of after
+
+        :param attach_to: where to attach the middleware: before a
+                          request of after
         :type attach_to: str
         """
         assert attach_to in ("request", "response")
-        
+
         if attach_to == 'request':
             self.request_middleware.append(middleware)
         if attach_to == 'response':
@@ -185,10 +187,10 @@ class Sanic:
         :return: Nothing
         """
         try:
-            
+
             # Add view handler
             request.handler = self.router.routes_static.get(request.url)
-            
+
             # -------------------------------------------- #
             # Request Middleware
             # -------------------------------------------- #
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -116,6 +116,23 @@ class Sanic:
         else:
             attach_to = args[0]
             return register_middleware
+    
+    def add_middleware(self, middleware, attach_to="request"):
+        """
+        register a middleware and configure to be called before a request of after
+        
+        :param middleware: a callable function (coroutine)
+        :type middleware: callable
+        
+        :param attach_to: where to attach the middleware: before a request of after
+        :type attach_to: str
+        """
+        assert attach_to in ("request", "response")
+        
+        if attach_to == 'request':
+            self.request_middleware.append(middleware)
+        if attach_to == 'response':
+            self.response_middleware.appendleft(middleware)
 
     # Static Files
     def static(self, uri, file_or_directory, pattern='.+',
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -185,6 +185,10 @@ class Sanic:
         :return: Nothing
         """
         try:
+            
+            # Add view handler
+            request.handler = self.router.routes_static.get(request.url)
+            
             # -------------------------------------------- #
             # Request Middleware
             # -------------------------------------------- #
@@ -291,6 +295,7 @@ class Sanic:
             'request_timeout': self.config.REQUEST_TIMEOUT,
             'request_max_size': self.config.REQUEST_MAX_SIZE,
             'loop': loop
+            # 'app': self
         }
 
         # -------------------------------------------- #
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -294,8 +294,8 @@ class Sanic:
             'error_handler': self.error_handler,
             'request_timeout': self.config.REQUEST_TIMEOUT,
             'request_max_size': self.config.REQUEST_MAX_SIZE,
-            'loop': loop
-            # 'app': self
+            'loop': loop,
+            'app': self
         }
 
         # -------------------------------------------- #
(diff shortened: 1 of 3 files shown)


sanic/sanic.py
```diff
@@ -168,6 +187,10 @@ def converted_response_type(self, response):
         :return: Nothing
         """
         try:
+
+            # Add view handler
+            request.handler = self.router.routes_static.get(request.url)
```
seemethere: Would this still work with a dynamic route as well?

seemethere:
Also why add the app to the server? What's the point of it?

sanic/sanic.py
cr0hn: I tried with some end-points with dynamic routes and it works fine.

cr0hn:
The point to add the app to the server is really to add to the request object. The reasons are some:

- Access to the shared event loop object in views / middlewares.
- Access to configuration in views.

To access to the above information we need to has access to the `Sanic` instance object. But, probability we want to create a medium-big app using Sanic, and we need to use a scalable structure, like: http://exploreflask.com/en/latest/organizing.html

This organisation was made for Flask and has not in count some things, because the framework provide other methods to obtain this information:

1- Access to the app object in any place: `Flask` provide `current_app` object. But Sanic has not way to do that. If we can't access to the app object we can't access to the global configuration and plugins will have not do things like:

```python
from sanic import Sanic

app = Sanic()

def setup_db(app: Sanic,
             connection_string='postgresql://postgres:postgres@127.0.0.1/test',
             loop=None):
   # STORE DATABASE POOL HANDLER
    app.config.db_engine = loop.run_until_complete(create_engine(connection_string,
                                                                 minsize=10,
                                                                 maxsize=20,
                                                                 timeout=10,
                                                                 loop=loop))

@app.route("/q")
async def query(request):
        results = []
        
        async with request.app.config.db_engine.acquire() as conn:  # ACCESS IN ALL END-POINTS TO DB-POOL.
            async for row in conn.execute("SELECT 1;"):
                results.append(dict(id=row.id))
        
        return json(results)

app.run()
```

This way made each view / middleware self-contained, without the need of external shared objects.

2 - Some times we need to access to the event loop object in the views / middlewares. To do that also need to access to the `Sanic` instance. 

This is especially useful when we run Sanic with more than 1 worker. Sanic executes one instance in different processes. Each process has their own event loop. So, if we do something like that, it break:

```python

import asyncio
import uvloop

from sanic import Sanic

app = Sanic()
    
@app.route("/")
async def test(request):
    url = "https://api.github.com/repos/channelcat/sanic"

   # USE EVENT-LOOP GENERATED IN ANOTHER PROCESS -> BREAK THE APP
    async with aiohttp.ClientSession(loop=loop) as session:
        response = await fetch(session, url)
        return json(response)

app.run(host='127.0.0.1',
            port=8000,
            workers=4)
```

So we probably need to access to the event loop object in some view / middleware, so we need the loop reference used in the current instance:

```python
import aiohttp

from sanic import Sanic

app = Sanic()
    
@app.route("/")
async def test(request):
    url = "https://api.github.com/repos/channelcat/sanic"

   # SELF-CONTAINED VIEW. IT RUNS WITH MULTIPROCESS APPROACH
    async with aiohttp.ClientSession(loop=request.app.loop) as session:
        response = await fetch(session, url)
        return json(response)

app.run(host='127.0.0.1',
            port=8000,
            workers=4)
```

Applying this patch we can allow to do that, without performance impact. Also made more easy the development of extensions.

gpip:
I miss the context local objects available in flask, some of the use cases would be handled by introducing a way to call a function after app.run starts and before requests are served (i.e. a middleware where request doesn't exist yet and is called only once).

But others I'm not sure yet how to handle other than working around them with uglier solutions, for instance I had to drop custom logging filters since those depended on the context local objects. Do you think it would be possible to add something equivalent to `flask.current_app`?

cr0hn:
@seemethere What's about these new features? As time goes by there're more inconsistencies with the main trunk

subyraman:
+1 to the ability to get the `current_app` via the loop. It will make accessing app configuration/routes much easier. I'd love to see this get in.

PR #249, opened 2016-12-29
Print original exception in debug mode.

youknowone:
Sanic is collecting the information of original exceptions but does not print it on console. Because it is very useful to debug user code, it is better to be printed in debug mode.

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -1,5 +1,6 @@
 from asyncio import get_event_loop
 from collections import deque
+from contextlib import suppress
 from functools import partial
 from inspect import isawaitable, stack, getmodulename
 from multiprocessing import Process, Event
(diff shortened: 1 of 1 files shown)


sanic/sanic.py
```diff
@@ -231,6 +232,17 @@ def converted_response_type(self, response):
                     response = HTTPResponse(
                         "An error occured while handling an error")
 
+            if self.debug:
+                error_body = response.body
+                if isinstance(error_body, bytes):
+                    try:
+                        error_body = error_body.decode()
+                    except:
+                        print(
```
seemethere: Would it be better to log this this as an error instead of printing it to stderr?

youknowone: Thanks for a nice advice. I updated it.

seemethere:
So to be clear this is just for when error handlers error out correct?

seemethere:
Nevermind, gonna close this. I think I may have a better way of doing this, will link it to this PR.

PR #253, opened 2016-12-29
add a flag to skip SIGINT and SIGTERM signals registration

dutradda:


seemethere:
What's the reasoning behind this change?

dutradda:
@seemethere, because I need the app running in a second thread, and this signals can just be listen from the main thread. (If you try to run the app in a second thread you got a error).

tests/test_sanic.py
```diff
@@ -0,0 +1,41 @@
+from sanic import Sanic
+from sanic.response import HTTPResponse
+from sanic.utils import HOST, PORT
+from unittest.mock import MagicMock
+import pytest
+import asyncio
+
+
+async def stop(app):
+    await asyncio.sleep(0.2)
```
r0fls: Can we make this a shorter interval, or is sleep required for the demonstration? If we add too many sleep intervals to the tests they're going to start taking forever to run :) So, I think it makes sense to only use them when needed.

r0fls:
Seems like a fine idea -- can you fix the merge conflicts?

r0fls:
Also adding a complete example to the `examples/` would help illustrate the usefulness.

PR #261, opened 2017-01-02
Cache request.json even when it's empty.

yoloseem:
In case of request body is set to `{}`, `[]` <strike>or `null`</strike>, despite it's already processed, `parsed_json` cache won't be used due to its boolean evaluation. This PR resolves that so now it caches whatever it's inside<strike>, and with using `...`(ellipsis), we can cache `null`(`None`) as well</strike>.


seemethere:
Will merge on successful run of tests

Issue #263, opened 2017-01-02
need listen backlog


zii:

r0fls:
@zii are you referring to an access log?

38elements:
I think that the `backlog` is a parameter of [AbstractEventLoop.create_server()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.create_server).

seemethere:
Closed with #269

PR #269, opened 2017-01-04
add configurable backlog

r0fls:
Addresses #263

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -20,7 +20,7 @@ from .exceptions import ServerError
 
 class Sanic:
     def __init__(self, name=None, router=None,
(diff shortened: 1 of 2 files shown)


sanic/sanic.py
```diff
@@ -20,7 +20,7 @@
 
 class Sanic:
     def __init__(self, name=None, router=None,
-                 error_handler=None, logger=None):
+                 error_handler=None, logger=None, backlog=100):
```
38elements: I think that it is better that `backlog` is an argument of `sanic.run()`,
like `host` and `port`.

seemethere:
Closes #263

Issue #203, opened 2016-12-21
workers problem


wsbmilk:
if i use custom event loop below:
loop = uvloop.new_event_loop()
and then:
app.run(workers=4, loop=loop)

log will be:
2016-12-21 20:34:56,993: INFO: Spinning up 4 workers...
python: src/unix/linux-core.c:336: uv__io_poll: Assertion `(unsigned) fd < loop->nwatchers' failed.

and actually just 1 cpu core is working...

but when use default loop.
workers=4 will  work.

os: ubuntu16.04 x64 and python 3.5.2
why?

edevil:
I actually have a similar problem and I'm not even changing the event loop. I tried the example from the documentation and set the workers to 4:
```
from sanic import Sanic
from sanic.response import json

app = Sanic()

@app.route("/")
async def test(request):
    return json({"hello": "world"})

if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8000, workers=4)
```

When I start it, it logs that it is spawning 4 workers, which it does:
```
2016-12-21 14:27:41,936: INFO: Goin' Fast @ http://0.0.0.0:8000
2016-12-21 14:27:41,936: INFO: Spinning up 4 workers...
```

But when I use wrk specifying 100 connections and 100 threads, still only one of the processes is handling the requests:
```
~ ps xau |grep sanic
andre             6351  93.6  0.1  2463520  15224 s002  R+    2:27PM   1:00.47 python hello_sanic.py
andre             6353   0.0  0.1  2463264   8488 s002  S+    2:27PM   0:00.02 python hello_sanic.py
andre             6352   0.0  0.0  2463264   8336 s002  S+    2:27PM   0:00.02 python hello_sanic.py
andre             6350   0.0  0.1  2463264   8540 s002  S+    2:27PM   0:00.02 python hello_sanic.py
andre             6349   0.0  0.1  2431400  17248 s002  S+    2:27PM   0:00.22 python hello_sanic.py
```

Sanic version 0.1.8.

wsbmilk:
here is mine:
code:
```
from sanic import Sanic
from sanic.response import json

app = Sanic(__name__)


@app.route("/")
async def test(request):
    return json({"test": True})


app.run(host="0.0.0.0", port=8000, debug=True, workers=4)
```
result:
```
➜  ~ ps aux|grep simple_server.py
troy     20963  0.0  0.2  78908 21272 ?        S    Dec21   0:00
[...]

PR #276, opened 2017-01-07
fix multiple worker problem

r0fls:
Addresses issue #203. Note that the creator of uvloop advises against conforming to wsgi for asyncio web servers. See https://github.com/python/asyncio/issues/481#issuecomment-271098847

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -16,6 +16,8 @@ from .router import Router
 from .server import serve, HttpProtocol
 from .static import register as static_register
 from .exceptions import ServerError
+from socket import socket
+from os import set_inheritable
 
 
 class Sanic:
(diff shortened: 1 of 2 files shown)


r0fls:
This change also removes some unnecessary code and incorrect logging messages.

r0fls:
@seemethere I'm still working on the tests -- when you use serve_multiple directly it's causing an issue when rebinding to the port. Give me a couple hours before merging please, as I'll be adding a test case for this and fixing the issue with using serve_multiple directly.

seemethere:
@r0fls Will do, thanks for all the good work!

r0fls:
Well, that didn't go as planned... the utils for testing don't appear to be working with this, which will require investigation. So, I've disabled the multiprocessing tests for now. Since MP wasn't really working before, this is still a significant improvement. The below example shows this is working:

```python
import asyncio
import uvloop

import os
from sanic import Sanic

app = Sanic(__name__)

@app.route('/')
async def hello(request):
    print(os.getpid())
    return text("OK")

if __name__ == '__main__':
    app.run(workers=4)
```

Then you can see multiple processes responding:

```
wrk -t50 -c100 http://localhost:8000
3704
3703
3704
3701
3703
3701
3704
3702
```

You should also see improvements from `wrk`. I clocked in at this with 4 workers:

```
$ wrk -t50 -c100 http://localhost:8000
    Latency     2.70ms    1.01ms  10.75ms   72.88%
    Req/Sec   746.49    181.38     6.51k    82.84%
  372222 requests in 10.10s, 45.79MB read
Requests/sec:  36856.63
Transfer/sec:      4.53MB
```

vs this with 1

```
$ wrk -t50 -c100 http://localhost:8000
    Latency     5.18ms    1.81ms  15.60ms   60.86%
    Req/Sec   388.54     75.63   808.00     67.54%
  193713 requests in 10.03s, 23.83MB read
Requests/sec:  19317.26
Transfer/sec:      2.38MB
```

Issue #277, opened 2017-01-08
Add support for vhosts/subdomains


martmists:

r0fls:
@martmists would you say this is good to close based on the above MR?

seemethere:
Closing per #278

PR #278, opened 2017-01-09
add support for virtual hosts

r0fls:
Addresses issue #277. Includes a test and defaults to the current behavior if no hosts are specified.

Here is an example as well:

```python
from sanic import Sanic

app = Sanic()

@app.route('/', host="example.com")
async def hello(request):
    return text("Answer")
@app.route('/', host="sub.example.com")
async def hello(request):
    return text("42")

if __name__ == '__main__':
```

Then:

```
curl -H "Host: example.com" localhost:8000
Answer
curl -H "Host: sub.example.com" localhost:8000
42
```

Note that if you use the `host` key with a route then you must use it for all of them. This is a result of the implementation and should be fixed, to allow default hosts. However since this doesn't impact current behavior and provides basic support, I would see that as a new separate issue. Testing with `wrk` I saw no noticeable performance impact, even when routing with a host. So, I think this should be merged and we can address allowing defaults with vhosts later.

### DIFF
--- a/examples/vhosts.py
+++ b/examples/vhosts.py
@@ -1,0 +1,18 @@
+from sanic.response import text
+from sanic import Sanic
+
+# Usage
+# curl -H "Host: example.com" localhost:8000
+# curl -H "Host: sub.example.com" localhost:8000
+
+app = Sanic()
+
+@app.route('/', host="example.com")
+async def hello(request):
+    return text("Answer")
+@app.route('/', host="sub.example.com")
+async def hello(request):
+    return text("42")
+
+if __name__ == '__main__':
+    app.run(host="0.0.0.0", port=8000)
--- a/sanic/router.py
+++ b/sanic/router.py
@@ -55,8 +55,9 @@ class Router:
         self.routes_static = {}
         self.routes_dynamic = defaultdict(list)
         self.routes_always_check = []
+        self.hosts = None
 
-    def add(self, uri, methods, handler):
+    def add(self, uri, methods, handler, host=None):
         """
         Adds a handler to the route list
         :param uri: Path to match
@@ -66,6 +67,17 @@ class Router:
         When executed, it should provide a response object.
         :return: Nothing
         """
+
+        if host is not None:
+            # we want to track if there are any
+            # vhosts on the Router instance so that we can
+            # default to the behavior without vhosts
+            if self.hosts is None:
+                self.hosts = set(host)
+            else:
+                self.hosts.add(host)
+            uri = host + uri
+
         if uri in self.routes_all:
             raise RouteExists("Route already registered: {}".format(uri))
 
@@ -113,7 +125,9 @@ class Router:
         else:
             self.routes_static[uri] = route
 
-    def remove(self, uri, clean_cache=True):
+    def remove(self, uri, clean_cache=True, host=None):
+        if host is not None:
+            uri = host + uri
         try:
             route = self.routes_all.pop(uri)
         except KeyError:
@@ -137,10 +151,14 @@ class Router:
         :param request: Request object
         :return: handler, arguments, keyword arguments
         """
-        return self._get(request.url, request.method)
+        if self.hosts is None:
+            return self._get(request.url, request.method, '')
+        else:
+            return self._get(request.url, request.method,
+                             request.headers.get("Host", ''))
 
     @lru_cache(maxsize=Config.ROUTER_CACHE_SIZE)
-    def _get(self, url, method):
+    def _get(self, url, method, host=None):
         """
         Gets a request handler based on the URL of the request, or raises an
         error.  Internal method for caching.
@@ -148,6 +166,7 @@ class Router:
         :param method: Request method
         :return: handler, arguments, keyword arguments
         """
+        url = host + url
         # Check against known static routes
         route = self.routes_static.get(url)
         if route:
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -51,7 +51,7 @@ class Sanic:
     # -------------------------------------------------------------------- #
 
     # Decorator
-    def route(self, uri, methods=None):
+    def route(self, uri, methods=None, host=None):
         """
         Decorates a function to be registered as a route
         :param uri: path of the URL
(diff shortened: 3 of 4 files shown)


odarbelaeze:
This is usually take care off in underlying infraestructure like nginx or a load balancer, am I right? Can you elaborate on why is this useful?

r0fls:
@odarbelaeze sanic is probably fastest when exposed directly:

**nginx using proxy pass, server name, and the current master branch**
```
root@circular:~# wrk -t50 -c100 -H "Host: example.com" http://localhost:8000
    Latency    82.03ms   27.95ms 297.58ms   74.96%
    Req/Sec    24.76      9.33    60.00     78.13%
  12348 requests in 10.10s, 10.01MB read
Requests/sec:   1222.90
Transfer/sec:      0.99MB
root@circular:~# ls
root@circular:~# wrk -t50 -c100 -H "Host: example.com" http://localhost:8000
    Latency    53.36ms   64.57ms 680.41ms   92.61%
    Req/Sec    49.99     36.42   222.00     74.60%
  24610 requests in 10.10s, 19.95MB read
Requests/sec:   2437.61
Transfer/sec:      1.98MB
root@circular:~# wrk -t50 -c100 -H "Host: example.com" http://localhost:8000
    Latency    92.43ms  173.15ms   1.91s    94.83%
    Req/Sec    34.73     26.86   222.00     89.86%
  16810 requests in 10.10s, 13.63MB read
Requests/sec:   1664.49
Transfer/sec:      1.35MB
```

**sanic using the proposed host header routing**
```
root@circular:~# wrk -t50 -c100 -H "Host: example.com" http://localhost:8000
    Latency    27.09ms   12.02ms 188.64ms   91.63%
    Req/Sec    75.75     18.56   220.00     80.71%
  38194 requests in 10.10s, 6.48MB read
  Non-2xx or 3xx responses: 38194
Requests/sec:   3781.62
Transfer/sec:    657.35KB
root@circular:~# wrk -t50 -c100 -H "Host: example.com" http://localhost:8000
    Latency    29.16ms   14.66ms 221.17ms   87.83%
    Req/Sec    70.68     23.66   141.00     67.91%
  35493 requests in 10.10s, 6.03MB read
  Non-2xx or 3xx responses: 35493
Requests/sec:   3514.61
Transfer/sec:    610.94KB
root@circular:~# wrk -t50 -c100 -H "Host: example.com" http://localhost:8000
    Latency    28.58ms   12.86ms 200.81ms   80.27%
    Req/Sec    71.62     22.19   280.00     75.69%
  35953 requests in 10.10s, 6.10MB read
  Non-2xx or 3xx responses: 35953
Requests/sec:   3559.85
Transfer/sec:    618.80KB
```


So for those that wish to do so, this change makes sense. These were tested on the smallest DO droplets with a single worker (hence the slow speeds in general). Note that the nginx container was routing to sanic, so feel free to do your own tests. However, this change doesn't impact performance, and it adds support for this feature, should it be desired.

r0fls:
@odarbelaeze I've also added the docker containers I used for testing here:
https://github.com/r0fls/nginx-vs-sanic

In case you would like to report your own findings or tweak them to your liking, for comparison purposes. I would be interested to see something using haproxy, but I think that scenario would still make sense to use in front of multiple sanic servers utilizing the host header.

I do think this discussion should be happening on the issue thread #277 , since I would see the PR thread as more of an opportunity for code review than justification for the feature, if there is an existing issue.

seanpar203:
@r0fls Looks good, very powerful and readable with a python kwarg as well.

odarbelaeze:
@r0fls thanks for the data all around 🎉

yoloseem:
From @r0fls 's above comment https://github.com/channelcat/sanic/pull/278#issuecomment-271222435, I just found that responses from Sanic were not entirely fine:
> Non-2xx or 3xx responses: 35953

Is that really reliable result?

r0fls:
@yoloseem I'll rerun the tests for you.

r0fls:
@yoloseem good catch by the way 👍

r0fls:
@yoloseem haven't reran the test yet because I'm having some docker issues (running remote docker-machines and I don't really know how to manage them very well with my local env :), but here is something to consider in the meantime (including instructions in case anyone else would like to run the tests):

Make sure you're on the current master branch (since these changes are merged). Then, create the below file:

**hello.py**
```python
from sanic import Sanic

app = Sanic()

@app.route('/')
async def hello(request):
    return text("OK")

if __name__ == '__main__':
    app.run(host="0.0.0.0",
            port=8000,
            workers=4)
```

Next, benchmark (I don't actually recommend running with `&` in this case, but for example's sake):

```
$ python3 hello.py &
$ wrk -t50 -c100 -d30 http://localhost:8000
Running 30s test @ http://localhost:8000
    Latency     1.52ms  444.65us  13.04ms   73.72%
    Req/Sec     1.32k   240.78     9.01k    70.47%
  1976116 requests in 30.10s, 243.11MB read
Requests/sec:  65650.12
Transfer/sec:      8.08MB
```

Finally, checkout the previous commit (before this change was merged)

```
$ git checkout b0bf989056c
$ python3 hello.py &
$ wrk -t50 -c100 -d30 http://localhost:8000
Running 30s test @ http://localhost:8000
    Latency     1.57ms  470.62us  18.41ms   82.43%
    Req/Sec     1.28k   215.09    10.13k    80.51%
  1916345 requests in 30.10s, 235.76MB read
Requests/sec:  63661.71
Transfer/sec:      7.83MB
```

Obviously this change didn't make things faster, but it's not slowing them down either. So, it's not a harmful commit. I'll try to get the tests with nginx working again as well. The docker container setup is here https://github.com/r0fls/nginx-vs-sanic

if you'd like to try to fix/run them yourself.

yoloseem:
@r0fls , thanks for taking your time. I just wanted to make sure that the result was fine.

Host-based routing is good-to-have feature of course I think! With that, even if someone want/need to use nginx in front of Sanic, he/she can configure nginx only for the proxying purpose and manage all routings in the sanic-based code rather than dealing with nginx rules. 👍

r0fls:
@yoloseem ya, that nginx test I ran was sketchy, but I agree with your statement 💯  so I think it's unnecessary anyway :)

r0fls:
Also, for the record, here's one using the host header:


**hello.py**

```python
from sanic import Sanic

app = Sanic()

@app.route('/', host="example.com")
async def hello(request):
    return text("OK")

if __name__ == '__main__':
    app.run(host="0.0.0.0",
            port=8000,
            workers=4)
```

```
$ wrk -H "Host: example.com" -t50 -c100 -d30 http://localhost:8000
Running 30s test @ http://localhost:8000
    Latency     1.51ms  415.58us   7.45ms   76.02%
    Req/Sec     1.33k   211.58     7.78k    74.80%
  1986272 requests in 30.10s, 244.36MB read
Requests/sec:  65982.61
Transfer/sec:      8.12MB
```

Not too shabby for a Mac!

PR #281, opened 2017-01-10
Fix doc “Class based views”.

Mellcap:
Hi, I learned Sanic by reading the doc.
At "Class based views", I got this:
```python
  File "class_based_views.py", line 26, in <module>
    app.add_route(SimpleView.as_view(), '/')
AttributeError: type object 'SimpleView' has no attribute 'as_view'
```
Then I check the `HTTPMethodView.__doc__`, I find that we don't need to use `SimpleView.as_view()`, just `SimpleView()` is fine.

So, hope you can approve this pull request.

### DIFF
--- a/docs/class_based_views.md
+++ b/docs/class_based_views.md
@@ -28,7 +28,7 @@ class SimpleView(HTTPMethodView):
   def delete(self, request):
       return text('I am delete method')
 
-app.add_route(SimpleView.as_view(), '/')
+app.add_route(SimpleView(), '/')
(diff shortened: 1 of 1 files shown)


38elements:
The interface of `Class based views` was changed by #274.

seemethere:
Sorry about that! It'll all make sense once we roll out 0.2.0 at the end of this week!

Mellcap:
I am glad to do something if helpful!

Issue #102, opened 2016-10-23
Subdomains in Blueprints


rohitkrai03:
Can i use subdomain routing with blueprints? Basically i have more than two or three modules that i want to be on different subdomains using Blueprints. Below is the code for doing it in flask, but can i do something like this in sanic?

``` python
# -*- coding: utf-8 -*-

from flask import Flask, Blueprint

app = Flask(__name__)
app.config.DEBUG = True

www = Blueprint('www', 'www', subdomain='www')
sub = Blueprint('sub', 'sub', subdomain='sub')

@www.route('/')
def www_hello():
    return 'www.www_hello'


@sub.route('/')
def sub_hello():
    return 'sub.sub_hello'


app.register_blueprint(www)
app.register_blueprint(sub)

if __name__ == '__main__':
    app.run(host='127.0.0.1', port=8888, debug=True) 
```

seemethere:
Closed per #285

PR #285, opened 2017-01-11
add vhosts to blueprints

r0fls:
This partially addresses #102, as you can do this after this change:

```python
from sanic import Sanic

# Usage
# curl -H "Host: bp1.example.com" localhost:8000/question
# curl -H "Host: bp2.example.com" localhost:8000/answer

app = Sanic()
bp = Blueprint("bp", host="bp1.example.com")
bp2 = Blueprint("bp2", host="bp2.example.com")

@bp1.route("/question")
async def hello(request):
    return text("What is the meaning of life?")

@bp2.route("/answer")
async def hello(request):
    return text("42")

app.register_blueprint(bp1)
app.register_blueprint(bp2)

if __name__ == '__main__':
```

### DIFF
--- a/examples/vhosts.py
+++ b/examples/vhosts.py
@@ -1,11 +1,15 @@
 from sanic.response import text
 from sanic import Sanic
+from sanic.blueprints import Blueprint
 
 # Usage
 # curl -H "Host: example.com" localhost:8000
 # curl -H "Host: sub.example.com" localhost:8000
+# curl -H "Host: bp.example.com" localhost:8000/question
+# curl -H "Host: bp.example.com" localhost:8000/answer
 
 app = Sanic()
+bp = Blueprint("bp", host="bp.example.com")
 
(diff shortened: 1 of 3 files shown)


odarbelaeze:
Should you add a test combining the route level `host` parameter and the blueprint level one.

r0fls:
@odarbelaeze I wouldn't see much reason behind mixing them, since you could use a different blueprint for different domains or a separate route without a blueprint. In fact I didn't even add support for the host argument in the blueprint `route` function. Maybe I should at least add the host header as an argument to the blueprint `route` for consistency though. What do you think?

r0fls:
Thinking more I think that makes sense, having the ability to override the host set by the blueprint. So, the blueprint route will just be the default.

r0fls:
@odarbelaeze ok I added that feature and a test for it

PR #286, opened 2017-01-11
Signaling support and middleware behaviour

yoloseem:
Summary:
* Added signaling support with two built-in signals: `request_started`, `request_finished`.
* Fixed request middlewares to be executed only when there's handler so 404 can properly work.
* `sanic_endpoint_test()` uses signal to gather request rather than using middleware.

---

First catch: 
Request middlewares are applied to all requests even if there's no registered routes.
For example, if I write a request-validating middleware, the application will respond 400(or whatever) against any bad requests. And... the 404 will disappear. That was completely unexpected behaviour to me. (Test added: https://github.com/channelcat/sanic/commit/d603bcc3d05e630acd179367b05b35627e0018ef#diff-da5a345e8216a8e47d28218a4f5b9b0f)

To fix the first catch:
Before processing request, the application should determine whether the request is able to be handled or not- looking up the handler from the router. If there's no handler, abort the processing and we can respond with 404.

Trouble:
`sanic_endpoint_test()` was using middleware to collect request.
In case of the situation that middlewares are not executed(testing with 404 paths), tester function can't gather the request so a lot of tests will broken.

New motivation from trouble:
We need to have *pre*-process-handling layer.
I thought we can have two choices here: introducing 'before_before_request' middleware so it can come before the 'before_request' middleware(`@middleware('request')`). Or, having signaling system.
Considering the current need-recording the request- and our current middleware's simplicity, I carefully decided that having signaling system would be better than extend/expand/ing the middlewares.

### DIFF
--- a/docs/signals.md
+++ b/docs/signals.md
@@ -1,0 +1,27 @@
+# Signals
+
+Signals allow you to send notifications to multiple subscribers when the specific event occured.
+Signaling is provided by [Blinker](http://pythonhosted.org/blinker/) library and
+Sanic has few built-in signals:
+
+* `request_started`: Sent when the request processing has started.
+* `request_finished`: Sent when the request processing has finished.
+
+## Examples
+
+```python
+from sanic.signals import request_started
+import my_realtime_stat
+
+app = Sanic(__name__)
+
+@request_started.connect
+def collect_stats(request):
+    my_realtime_stat.incoming_requests.incr(request.url)
+
+
+@request_finished.connect
+def count_errors(response):
+    if response.status != 200:
+        my_realtime_stat.error_responses.incr(response.status)
+```
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -13,6 +13,7 @@ from .log import log
 from .response import HTTPResponse
 from .router import Router
 from .server import serve, HttpProtocol
+from .signals import request_started, request_finished
 from .static import register as static_register
 from .exceptions import ServerError
 from socket import socket, SOL_SOCKET, SO_REUSEADDR
@@ -176,6 +177,8 @@ class Sanic:
         :return: Nothing
         """
         try:
+            request_started.send(request)
+
             # -------------------------------------------- #
             # Request Middleware
             # -------------------------------------------- #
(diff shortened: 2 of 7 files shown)


subyraman:
Do we have the ability to send `request_started` and `request_finished` on a per request basis? I'm thinking of a use case like websockets where it is important that we know when a client has connected and disconnected on a specific endpoint.

yoloseem:
I resolved merge conflicts. Any views on this? @r0fls @channelcat @seemethere @youknowone

docs/signals.md
```diff
@@ -0,0 +1,27 @@
...
+```python
+from sanic.signals import request_started
+import my_realtime_stat
+
+app = Sanic(__name__)
+
+@request_started.connect
+def collect_stats(request):
+    my_realtime_stat.incoming_requests.incr(request.url)
+
```
r0fls: Are you able to incorporate something like this example into a unit test?

sanic/sanic.py
```diff
@@ -176,6 +177,15 @@ def converted_response_type(self, response):
         :return: Nothing
         """
         try:
+            request_started.send(request)
```
r0fls: Does it perform this on every request, even if there are no signal handlers for `request_started`? How does that impact performance? Results from running `wrk` before and after on the simple_server.py would be good to see.

Tim-Erwin: Blinker says it has been optimized for that case. (That doesn't replace a good measurement, of course :) )

Tim-Erwin:
Blinker is nice (although kinda dead), however it doesn't allow receivers to be async, right? That would be a requirement for me in an async framework like Sanic

r0fls:
@yoloseem should we close for now this since Blinker was ruled out due to be synchronous only?

yoloseem:
@r0fls OK! I've been being far from this work recently... T.T

seemethere:
Yeah I think the blinker requirement kind of kills this PR. Will close for now.

dnikolayev:
Hello,

What about re-opening?
The reason of pausing was in sync mode of blinker.

However, there is [AsyncBlink](https://github.com/jucacrispim/asyncblink)

People use this for Tornado. And possibly this will simplify porting Flask apps into Sanic.

Thanks!

PR #291, opened 2017-01-13
Add rich HTML traceback in debug mode, add HTML 500 page in prod

subyraman:
Adds in an HTML traceback page similar to Flask/Django when the user hits on an unhandled exception in debug mode.

screen shot 2017-01-12 at 7 20 27 pm

Adds in Internal Server Error html output also similar to Flask/Django on an unhandled exception in prod mode:

screen shot 2017-01-12 at 7 19 13 pm

If the user has enabled CSP headers the styling won't work, though I suppose we could disable those headers before the response is sent. Otherwise we should consider serving a separate CSS file.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,3 +12,4 @@ kyoukai
 falcon
 tornado
 aiofiles
+beautifulsoup4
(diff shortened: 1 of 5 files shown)


seemethere:
I like this a lot but I'm going to be moving the html and css to a base folder (instead of having them as long `str` constants) in a subsequent PR.

PR #293, opened 2017-01-13
Extensions docs

r0fls:
Since there have been a few extensions discussed in open issues, it makes sense to start keeping track of them.


seemethere:
👍

PR #294, opened 2017-01-14
Add redirect method from @pcdinh

subyraman:
@pcdinh seems to have abandoned #118 so I've picked it up and added some more tests around the `None` bug that was mentioned.


seemethere:
Yay for pytest fixtures!

PR #296, opened 2017-01-14
Add SSL context to server

mdaue:
Currently the Sanic server appears to only support HTTP. This PR will allow the setup of a SSL connection to the server, to upgrade sockets to HTTPS. The motivation behind this PR is to maintain the sanic/uvloop speed gains, without having to rely on other HTTP servers (Apache, Nginx) to provide a reverse proxy, essentially slowing down the connection.

Added an example in the README that details how to create the SSLContext, which can then be passed in. This works for single and multiple workers.

Add ssl variable passthrough to following:
-- sanic.run
-- server.serve
Add ssl variable to loop.create_server to enable built-in async context socket wrapper
Update documentation
Tested with worker = 1, and worker = 2.
Unit tests are passing

Signed-off-by: Matt Daue <mattdaue@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,18 @@ if __name__ == "__main__":
 ## Installation
  * `python -m pip install sanic`
 
+## Use SSL
+ * Optionally pass in an SSLContext:
+```
+import ssl
+certificate = "/path/to/certificate"
+keyfile = "/path/to/keyfile"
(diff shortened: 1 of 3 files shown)


r0fls:
@mdaue just to clarify, when you tested this you received valid SSL?

mdaue:
Yep

mdaue:
Going to provide certificate data shortly

PR #300, opened 2017-01-16
Moved Remote-Addr header to request.ip

channelcat:
Sadly, pulling IP is costly.  Moving it to a property that pulls on-demand should bump Sanic's performance back up by about 10%.  Also Ubuntu stole my commit credz :(.


r0fls:
Would it make sense to cache this? https://github.com/channelcat/sanic/pull/301

PR #301, opened 2017-01-16
cache the remote IP property

r0fls:


r0fls:
Whoops, i thought these were passing for me. I'll have to look into that in a bit.

PR #302, opened 2017-01-17
Trimmed down features of CIMultiDict

channelcat:
Should result in a 8% speed boost

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,3 @@ httptools
 ujson
 uvloop
 aiofiles
-multidict
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -106,11 +106,11 @@ class HTTPResponse:
             for name, value in self.headers.items():
                 try:
                     headers += (
-                        b'%b: %b\r\n' % (name.encode(), value.encode('utf-8')))
+                        b'%b: %b\r\n' % (name.title().encode(), value.encode('utf-8')))
                 except AttributeError:
                     headers += (
                         b'%b: %b\r\n' % (
-                            str(name).encode(), str(value).encode('utf-8')))
+                            str(name).title().encode(), str(value).encode('utf-8')))
 
         # Try to pull from the common codes first
         # Speeds up response rate 6% over pulling from all
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -106,11 +106,11 @@ class HTTPResponse:
             for name, value in self.headers.items():
                 try:
                     headers += (
-                        b'%b: %b\r\n' % (name.encode(), value.encode()))
+                        b'%b: %b\r\n' % (name.encode(), value.encode('utf-8')))
                 except AttributeError:
                     headers += (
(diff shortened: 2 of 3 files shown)


jasonab:
Just so I'm clear, since I submitting the original request: some user agents pass the UA string as "user-agent" and some as "User-Agent."

With CIMultiDict, I could use either spelling. Do I have to use the lower case version after this change?

seemethere:
No `casefold` is case-insensitive. @jasonab

awiddersheim:
Doesn't this break the case where multiple headers of the same name get sent? This is probably a pretty rare scenario but it is possible and seems to be the point of using something like `MultiDict` in the first place as it can handle those types of values. Relevant [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) section.

Here is an example of sending multiple headers of the same name with `curl`:

```
$ curl localhost:8000 -H 'Foo:bar' -H 'Foo:baz' --verbose
* Rebuilt URL to: localhost:8000/
*   Trying ::1...
* connect to ::1 port 8000 failed: Connection refused
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET / HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.43.0
> Accept: */*
> Foo:bar
> Foo:baz
>
< HTTP/1.1 200 OK
< Content-Type: application/json
< Content-Length: 17
< Connection: keep-alive
< Keep-Alive: timeout=60
<
* Connection #0 to host localhost left intact
{"hello":"world"}
```

If the view for this I have the following:

```
print(request.headers.getall('foo'))
```

which results in the following getting displayed:

```
['bar', 'baz']
```

There is no way to do this now after this change.

ashleysommer:
I would like to reiterate the concern raised by @awiddersheim above.
This is no longer a CIMultiDict, it is just a CIDict. It breaks any case where the server should be able to receive multiple of the same header, or send a reply with multiple of the same header.

ashleysommer:
I just found another issue with this implementation.
Calling `my_headers.update()` or `my_headers.setdefault()` both do not convert the key to a case invariant form.
For example:
```
my_headers = CIDict()
my_headers['Foo'] = 'Bar' # <-- This converts 'Foo' to 'foo' on `__setitem__`
my_headers.update({'Foo': 'Baz'}) <-- This does not call `__setitem__` so key remains 'Foo'
```
Expected actual contents of `my_headers` is:
```
{'foo': 'Baz'}
```
However actual contents is:
```
{'Foo': 'Baz', 'foo': 'Bar'}
```
This is because both `update()` and `setdefault()` methods on a `dict` call `PyDict_SetItem()` directly in C. 
See this stackoverflow for more information http://stackoverflow.com/questions/2060972/subclassing-python-dictionary-to-override-setitem

You need to provide override implementations of `update()` and `setdefault()` which either call `__setitem__` _or_ do the casefold itself.

PR #306, opened 2017-01-17
Always log if there's an exception occurred.

zkanda:
Hi,

I'm trying to run sanic in one of my server. My problem is that when there's an exception, I cannot see what happened because there's no log to see. The solution to this is to enable debug mode, the problem with that is the user can see the exception logs as well.

I'm not sure this is the best approach but I propose we always log the exception even though we are not running in debug mode.

### DIFF
--- a/sanic/exceptions.py
+++ b/sanic/exceptions.py
@@ -173,6 +173,7 @@ class Handler:
         try:
             response = handler(request=request, exception=exception)
         except:
+            log.error(format_exc())
             if self.sanic.debug:
                 response_message = (
                     'Exception raised in exception handler "{}" '
@@ -185,6 +186,7 @@ class Handler:
         return response
 
     def default(self, request, exception):
+        log.error(format_exc())
         if issubclass(type(exception), SanicException):
             return text(
                 'Error: {}'.format(exception),


awiddersheim:
I agree with this as well. Thought it was a bit weird when I found out sanic was swallowing all of the exceptions unless in debug.

odarbelaeze:
Isn't it good to shallow those exceptions for security reasons, maybe you could trow these into a log if not in debug mode and then check the log directly on the server?

r0fls:
I see no noticeable degradation from these changes:

*After with no exception*
```
$ wrk -t50 -c100 -d30 http://localhost:8000
Running 30s test @ http://localhost:8000
    Latency     1.45ms  510.97us  15.91ms   82.72%
    Req/Sec     1.39k   272.20     9.94k    73.65%
  2080812 requests in 30.10s, 255.99MB read
Requests/sec:  69129.72
Transfer/sec:      8.50MB
```

*Before with no exception*
```
$ wrk -t50 -c100 -d30 http://localhost:8000
Running 30s test @ http://localhost:8000
    Latency     1.48ms  519.05us  28.79ms   84.82%
    Req/Sec     1.37k   240.10    16.64k    77.67%
  2049662 requests in 30.10s, 252.16MB read
Requests/sec:  68094.88
Transfer/sec:      8.38MB
```

*After with exception*
```
$ wrk -t50 -c100 -d30 http://localhost:8000
Running 30s test @ http://localhost:8000
    Latency     1.65ms  618.78us  26.86ms   77.83%
    Req/Sec     1.22k   316.54    14.46k    74.39%
  1827891 requests in 30.10s, 338.18MB read
  Non-2xx or 3xx responses: 1827891
Requests/sec:  60734.48
Transfer/sec:     11.24MB
```

*Before with exception*
```
$ wrk -t50 -c100 -d30 http://localhost:8000
Running 30s test @ http://localhost:8000
    Latency     1.64ms  509.15us  11.68ms   74.98%
    Req/Sec     1.23k   218.65     2.02k    69.49%
  1832898 requests in 30.02s, 339.11MB read
  Non-2xx or 3xx responses: 1832898
Requests/sec:  61046.20
Transfer/sec:     11.29MB
```

r0fls:
@zkanda is it possible to write a unit test to make sure it's doing what you expect, or is that too difficult? I had some troubles when testing with logging, since there are some oddities with the pytest logging treatment.

r0fls:
I ask because I implemented your change and did not see an exception logged in non debug mode when I used an invalid route:

```python
from sanic import Sanic
import asyncio
import time

app = Sanic()

@app.route('/')
async def hello(request):
    x == 3
    return text("OK")

if __name__ == '__main__':
    app.run(host="0.0.0.0",
            port=8000,
            workers=4)
```

r0fls:
@zkanda that's ok if that's a different bug, but I would need to see an example where this does something before approving it. Can you provide one?

GenericError:
This would bring it into line with similar programs - errors are generally at the default level of output

zkanda:
Seems like something is wrong with the test..

@r0fls how are you running the test? I'm using tox but I would prefer to use py.test directly.

* Here's how I did it, install all requirements-dev.txt
* Run `pytest`

I get a lot of failed test. 90 in total

e.g.

```
tests/test_views.py:194:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

app = <sanic.sanic.Sanic object at 0x7f18d426a828>, method = 'get', uri = '/', gather_request = True, loop = None, debug = False, server_kwargs = {}, request_args = (), request_kwargs = {}
_collect_request = <function sanic_endpoint_test.<locals>._collect_request at 0x7f18d6713950>, _collect_response = <function sanic_endpoint_test.<locals>._collect_response at 0x7f18d42da048>

    def sanic_endpoint_test(app, method='get', uri='/', gather_request=True,
                            loop=None, debug=False, server_kwargs={},
                            *request_args, **request_kwargs):
        results = []
        exceptions = []

        if gather_request:
            def _collect_request(request):
                results.append(request)
            app.request_middleware.appendleft(_collect_request)

        async def _collect_response(sanic, loop):
            try:
                response = await local_request(method, uri, *request_args,
                                               **request_kwargs)
                results.append(response)
            except Exception as e:
                exceptions.append(e)
            app.stop()

        app.run(host=HOST, debug=debug, port=PORT,
                after_start=_collect_response, loop=loop, **server_kwargs)

        if exceptions:
            raise ValueError("Exception during request: {}".format(exceptions))

        if gather_request:
            try:
                request, response = results
                return request, response
            except:
                raise ValueError(
                    "Request and response object expected, got ({})".format(
>                       results))
E               ValueError: Request and response object expected, got ([])

../../../.virtualenvs/sanic/lib/python3.6/site-packages/Sanic-0.2.0-py3.6.egg/sanic/utils.py:51: ValueError
================================================================================== 90 failed, 8 passed, 2 skipped in 6.87 seconds ===================================================================================
```

r0fls:
@zkanda I'm able to run the tests with `python3 -m pytest tests` from within the directory. Is that the command you're using? Anyway I'm fine merging this as is, because I agree it's the right behavior.

zkanda:
hmmm the reason why it's failing on test is because the `request` object is `None` when we call `request.url`. I tried debugging why it's not passing that object properly on testing but I'm not so familiar with the code base yet(and currently at work).

With that said, if we just log without the request then obviously the test would pass but there is redundancy in logging if we are in debug mode.

r0fls:
@zkanda let's start here, if you want to improve the functionality further just open another PR. This is an improvement already since if there is an exception, it should probably be logged.

Issue #287, opened 2017-01-11
Logger shouldn't be initialized until run


cdunklau:
This is kind of a drive-by bug report, so if I've misunderstood something, please disregard.

I noticed this worrisome thing while perusing the code:

https://github.com/channelcat/sanic/blob/4f832ac9afdba11f276b42ccbfbf7aa83c60dc47/sanic/sanic.py#L25-L29

I would think logging really shouldn't be configured until at least `Sanic.run` is called, to provide users the ability to configure it after instantiating but before running it.

tuukkamustonen:
Why does it even have `logger` argument? It is not used for anything except as boolean for checking whether or not to run `logging.basicConfig()`. To me it seems this block and argument should be just removed. What's the rationale?

r0fls:
Well, for the history: logging was previously initialized in that location in the same way, but without the option to disable it, and some users were confused as to how they should customize or override it.

r0fls:
@cdunklau regarding your concern, this is currently possible:

```
from sanic import Sanic
from sanic.response import text
import json
import logging


# Set logger to override default basicConfig
sanic = Sanic(logger=True)
@sanic.route("/")
def test(request):
    return text("hey")

logging_format = "[%(asctime)s] %(process)d-%(levelname)s "
logging_format += "%(module)s::%(funcName)s():l%(lineno)d: "
logging_format += "%(message)s"

logging.basicConfig(
    format=logging_format,
    level=logging.DEBUG
)
log = logging.getLogger()

sanic.run(host="0.0.0.0", port=8000)
```

r0fls:
I do agree that configuring `logging` in the `run` and `serve_multiple` functions probably makes more sense though. There I don't find the need for the boolean option to disable it, which is indeed silly.

cdunklau:
Another thing I noticed...
https://github.com/channelcat/sanic/blob/4f832ac9afdba11f276b42ccbfbf7aa83c60dc47/sanic/sanic.py#L156-L157

That should probably be `warnings.warn(thatmessage, DeprecationWarning)`

r0fls:
Closing per #308

Issue #73, opened 2016-10-19
Why not just wait() for the stop event to happen?


santagada:
in: https://github.com/channelcat/sanic/blob/master/sanic/sanic.py#L257 you spin every 0.3 seconds checking if the event has been set, why not just wait on the event?

channelcat:
I uh... I have no excuse <<;;.  I'll patch it :P

channelcat:
Very strange, when I use wait() it registers the SIGTERM, but doesn't unblock after stop_event.set() is called.

santagada:
I can give it a try.

r0fls:
@channelcat in case you're curious:
>Unix: wait(object_list, timeout) almost equivalent select.select(object_list, [], [], timeout). The difference is that, if select.select() is interrupted by a signal, it can raise OSError with an error number of EINTR, whereas wait() will not.

From here: https://docs.python.org/3/library/multiprocessing.html

r0fls:
@seemethere this is probably good to close I think.

seemethere:
Would agree, closed per #238

r0fls:
Actually this is still an issue but it's been very elusive.

seemethere:
Is this still an issue? Will reopen if necessary.

PR #308, opened 2017-01-17
update logging placement

r0fls:
Implements ideas from #287

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -21,12 +21,7 @@ from os import set_inheritable
 
 class Sanic:
     def __init__(self, name=None, router=None,
-                 error_handler=None, logger=None):
-        if logger is None:
-            logging.basicConfig(
-                level=logging.INFO,
-                format="%(asctime)s: %(levelname)s: %(message)s"
-            )
+                 error_handler=None):
         if name is None:
             frame_records = stack()[1]
             name = getmodulename(frame_records[1])
@@ -154,7 +149,8 @@ class Sanic:
     def register_blueprint(self, *args, **kwargs):
         # TODO: deprecate 1.0
         log.warning("Use of register_blueprint will be deprecated in "
-                    "version 1.0.  Please use the blueprint method instead")
+                    "version 1.0.  Please use the blueprint method instead",
+                    DeprecationWarning)
         return self.blueprint(*args, **kwargs)
 
     # -------------------------------------------------------------------- #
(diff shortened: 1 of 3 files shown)


sanic/sanic.py
```diff
@@ -247,7 +243,7 @@ def converted_response_type(self, response):
     def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
             after_start=None, before_stop=None, after_stop=None, sock=None,
             workers=1, loop=None, protocol=HttpProtocol, backlog=100,
-            stop_event=None):
+            stop_event=None, logger=None):
```
awiddersheim: @r0fls this doesn't seem to get used anywhere.

r0fls: @awiddersheim yes, I know it's irking me. It's a related to an open bug here: https://github.com/channelcat/sanic/issues/73. It could be removed until that's fixed, but it's critical, especially to testing multiprocessing -- as far as I know.

awiddersheim: I guess I'm confused about this change overall as well. Without the boolean doing anything how is someone supposed ot override logging settings?

r0fls: @awiddersheim in my test it was not necessary to have that boolean flag (as well as the current unit test we have for logging). The example still works as well. Essentially, the configuration to basicConfig gets applied. By delaying the configuration to `run()`, we give the user a time to override it. Maybe it should just be in the `sanic` initialization though? In theory that would also work, though I haven't tested it. Or maybe it should still be added as a property of the app, at least, during runtime? I don't see much reasoning for that, but I could be convinced....

awiddersheim: Maybe copy something similar to this:

https://github.com/pallets/werkzeug/blob/cbd8536b78b187766c67d21db3b81d284563a794/werkzeug/_internal.py#L75-L87

awiddersheim: Also, thanks for your explanation.

gpip:
I'm not sure if this affects anyone but if someone is using their own logging settings and configuring a handler named 'sanic' as well other custom loggers, it will be necessary to set 'propagate': False in the logging dict config after this change.

sanic/sanic.py
r0fls: @awiddersheim oh I thought you were talking about stop_event :( derp, I forgot to remove the logger boolean here and you're right that was not needed

Issue #307, opened 2017-01-17
Crash when sending simultaneous requests to long-running routes


ctlaltdefeat:
The following code that runs blocking code in a thread pool causes Sanic to crash for me with a "Fatal Python error: Cannot recover from stack overflow error." when sending 10 or so simultaneous requests to the server.
Commenting out the line that sets the event loop policy to uvloop (the default) makes the situation a bit better, with occasional "Writing response failed, connection closed 'NoneType' object has no attribute 'should_keep_alive'" errors but no crash.
Thoughts?

```
from sanic import Sanic
from sanic.response import text
import asyncio
import uvloop
import time
from functools import partial
from concurrent.futures import ThreadPoolExecutor

asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
loop = asyncio.get_event_loop()
thread_pool = ThreadPoolExecutor(max_workers=20)

app = Sanic()

@app.route('/')
async def test(request):
	return text(await loop.run_in_executor(thread_pool, partial(time.sleep,5)))

app.run(host="0.0.0.0", port=8000, loop=loop, workers=1)

```

ctlaltdefeat:
Running 0.2 by the way.

gpip:
I get the same when running it with asyncpg and asyncpgsa (haven't tried only asyncpg).

```
Fatal Python error: Cannot recover from stack overflow.

Current thread 0x00007fff94a223c0 (most recent call first):
  File ".../lib/python3.6/site-packages/Sanic-0.2.0-py3.6.egg/sanic/response.py", line 152 in text
  File ".../lib/python3.6/site-packages/Sanic-0.2.0-py3.6.egg/sanic/exceptions.py", line 71 in response
  File ".../lib/python3.6/site-packages/Sanic-0.2.0-py3.6.egg/sanic/server.py", line 183 in write_error
  File ".../lib/python3.6/site-packages/Sanic-0.2.0-py3.6.egg/sanic/server.py", line 193 in bail_out
```

It loops between write_error and bail_out until crashing.

ctlaltdefeat:
As @gpip mentioned it seems to happen when using regular asynchronous coroutines too (i.e same error with ```asyncio.sleep(5)``` without the threadpool), so changed title accordingly.

ctlaltdefeat:
Also, seems to happen only when canceling requests.

For example,


```
yes "curl http://localhost:8000" | head -n 100 > commands.txt
parallel --jobs 100 < commands.txt```

[...]

PR #309, opened 2017-01-18
Fixes write_error loop from bail_out function

seemethere:
Fixes stack-overflow found in #307

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -1,4 +1,5 @@
 import asyncio
+import traceback
 from functools import partial
 from inspect import isawaitable
 from signal import SIGINT, SIGTERM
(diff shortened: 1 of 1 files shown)


sanic/server.py
```diff
@@ -189,9 +190,15 @@ def write_error(self, exception):
                 "Writing error failed, connection closed {}".format(e))
 
     def bail_out(self, message):
-        exception = ServerError(message)
-        self.write_error(exception)
-        log.error(message)
+        if self.transport.is_closing():
+            log.error(
+                "Connection closed before error was sent to user @ {}".format(
+                    self.transport.get_extra_info('peername')))
```
r0fls: I think you can use the property here`self.request.ip`

seemethere: Well in the example that we had seen from the issue self.parser was still None so I'd be wary of if self.request was filled out correctly as well. It's just a safe way of doing it.

r0fls:
@seemethere I was going to use another try/except here. Are there any other exceptions that could occur during `self_write` within `bail_out` and lead to a circular exception? If there are, the try/except makes more sense imo.

r0fls:
I would also think a try/except would be faster than checking the `transport.is_closing()` method each time, but I don't have a working test case for this. The issue didn't actually throw any exception for me ... maybe I did something wrong though.

seemethere:
You might need to run it through wrk to replicate the issue.

r0fls:
I saw almost no benefit to using try/except instead, so let's use this for now.

sh0uw4ng:
[241](https://github.com/channelcat/sanic/issues/241)

Issue #267, opened 2017-01-03
The Implement Of HTTPMethodView Is Not A Good  Practice.


TaylorHere:
sorry,I'm lack of English.
```
def __call__(self, request, *args, **kwargs):
        handler = getattr(self, request.method.lower(), None)
        if handler:
            return handler(request, *args, **kwargs)
        raise InvalidUsage(
            'Method {} not allowed for URL {}'.format(
                request.method, request.url), status_code=405)
```
In those code ,when I handle the request, it's not the HTTPMethodView itself to  be instantiation,but the handler.
the handler just carry the request,but not the instance,so that I can't use the instance to store some message.
So, I think it's not the right practice of a "View based Class".
In View based Class we should treat the request as the instance itself,so that I can write the code base on view itself,and do not think about the request.

seemethere:
@AntonDnepr?

AntonDnepr:
@seemethere @TaylorHere class-based views was written in minimalistic style just to provide possibility to write class-based style code. I agree, instance should be added, because we need to have a possibility to call additional class methods in views. I'm not agree that we shouldn't have separate request as a parameter, but this is question to discuss if need.
We have an open PR which tries to implement this kind of approach
https://github.com/channelcat/sanic/pull/201/files
but there are several problems with it and on my opinion it's an overkill for now.

I think this should be mark as thing to do and I'll try to implement this (but I can't start earlier then next week, so if someone could do this faster - it would be cool)

yoshiya0503:
If you don't mind, please reference Specific example in the following mini app.

`
from asyncio import sleep
from sanic import Sanic
from sanic.response import json
from sanic.views import HTTPMethodView

app = Sanic()

class Test(HTTPMethodView):

    def __init__(self):
        self.g = 0

    async def get(self, req):
        self.g += 10
        print(self.g)
        await sleep(0.1)
        print(self.g)
        # !!!  we expect self.g == 10 each context
        return json({'test': 'class test'})

app.add_route(Test(), '/class')

if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8000)
`

If we send some requests **concurrency** (using apache bench).

`
ab -c 10 -n 100 http://127.0.0.1:8000/class
`

g.self is equal to 1000, but actually, we expect g.self is equal to 10.

Oh gotcha!

We expect that **instance variables are always context local(like thread safe)**.
In the case of single thread non-blocking model (same as node.js), because of its feature,
[...]

PR #311, opened 2017-01-18
Feature: Routing overload

youknowone:
When user specifies HTTP methods to function handlers, it automatically
will be overloaded unless they duplicate.

Example:
```
    # This is a new route. It works as before.
    @app.route('/overload', methods=['GET'])
    async def handler1(request):
        return text('OK1')

    # This is the exiting route but a new method. They are merged and
    # work as combined. The route will serve all of GET, POST and PUT.
    @app.route('/overload', methods=['POST', 'PUT'])
    async def handler2(request):
        return text('OK2')

    # This is the existing route and PUT method is the duplicated method.
    # It raises RouteExists.
    @app.route('/overload', methods=['PUT', 'DELETE'])
    async def handler3(request):
	return text('Duplicated')
```

### DIFF
--- a/sanic/router.py
+++ b/sanic/router.py
@@ -3,6 +3,7 @@ from collections import defaultdict, namedtuple
 from functools import lru_cache
 from .config import Config
 from .exceptions import NotFound, InvalidUsage
+from .views import CompositionView
 
 Route = namedtuple('Route', ['handler', 'methods', 'pattern', 'parameters'])
 Parameter = namedtuple('Parameter', ['name', 'cast'])
@@ -78,9 +79,6 @@ class Router:
                 self.hosts.add(host)
             uri = host + uri
 
-        if uri in self.routes_all:
-            raise RouteExists("Route already registered: {}".format(uri))
-
         # Dict for faster lookups of if method allowed
         if methods:
             methods = frozenset(methods)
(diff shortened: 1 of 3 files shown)


seemethere:
Thanks! Great job @youknowone!

youknowone:
By the way, writing this patch, I had a question:
When adding a function to route, why its default methods is not just `['GET']` but all methods?
It looks weird that users must add methods=['GET'] to limit the user access only to get method.
URL endpoints are (practically) never designed to work for all of the HTTP methods. They only work for only few method.

seemethere:
@youknowone That's definitely something we should look into. Maybe it would be beneficial to just default it to `['GET']`

AntonDnepr:
Not tested this, but the implementation of this could cause this error as far as I can see:
https://github.com/channelcat/sanic/issues/267#issuecomment-270414710
maybe I'm wrong, but it looks the very similar way as HttpMethodView before.

youknowone:
@AntonDnepr I agree. I never thought about it. I think the mechanism itself is not harmful but its naming makes confusion. How about the class name does not end with 'view' but ends with 'route' that means it is persistent? Like "HTTPMethodRoute" and "MethodRedirectionRoute".

PR #314, opened 2017-01-19
Add exception handling for closed transports

seemethere:
Adds handling for closed transports in the server for `write_response`
as well as `write_error`, letting it all flow to `bail_out` seemed to be
a little light handed in terms of telling the logs where the
error actually occured.

Handling to fix the infinite `write_error` loop is still there and those
exceptions will get reported on in the debug logs.

Also removed some pesky backslashes along the way.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -168,12 +168,6 @@ class HttpProtocol(asyncio.Protocol):
             self.transport.write(
                 response.output(
                     self.request.version, keep_alive, self.request_timeout))
-            if not keep_alive:
-                self.transport.close()
-            else:
-                # Record that we received data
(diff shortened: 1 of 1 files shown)


sanic/server.py
```diff
@@ -185,16 +188,23 @@ def write_error(self, exception):
             version = self.request.version if self.request else '1.1'
             self.transport.write(response.output(version))
             self.transport.close()
+        except RuntimeError:
```
38elements: In order to reliably execute `self.transport.close()`,
I think it is better to execute `self.transport.close()` in finally clause.

seemethere: Moved

PR #319, opened 2017-01-20
Fixes doc link extensions from .html to .md

seemethere:
Whoops! Totally missed that all the links pointed to `.html` files
instead of `.md` files

@kdelwat


kdelwat:
Unfortunately sphinx won't convert the links to HTML, so they had to have the HTML extension to enable linking in the generated documents.

seemethere:
Oh no I get that, it's just that reading them on github themselves didn't work out so well on github proper. Will fix though once we get on Read the Docs.

kdelwat:
That makes sense, good catch

PR #320, opened 2017-01-20
add method shorthands

r0fls:


tests/test_routes.py
```diff
@@ -10,6 +10,73 @@
 #  UTF-8
 # ------------------------------------------------------------ #
 
+def test_shorthand_routes():
+    app = Sanic('test_shorhand_routes')
+
+    @app.get('')
```
seemethere: Could we split these into separate tests? Perhaps parameterize them so that we just write one test to do this?

r0fls: I couldn't get this to work with parameterization since it didn't like `@app.variable`.

PR #325, opened 2017-01-20
Simplify RequestParameters

seemethere:
Simplifies request parameters, it defined a bit more than it had too,
added some docstrings and made the code simpler as well. Should now
raise a KeyError on `__getitem__` as @amsb had noted on commit 9dd954b


amsb:
Thanks!

seemethere:
No problem @amsb!

PR #327, opened 2017-01-20
Make it easier to override logging

awiddersheim:
Take influence from how Werkzeug configures logging by only configuring
a handler if no root handlers were previously configured by the end
user.

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -22,6 +22,15 @@ from os import set_inheritable
 class Sanic:
     def __init__(self, name=None, router=None,
                  error_handler=None):
+        # Only set up a default log handler if the
+        # end-user application didn't set anything up.
+        if not logging.root.handlers and log.level == logging.NOTSET:
+            formatter = logging.Formatter(
+                "%(asctime)s: %(levelname)s: %(message)s")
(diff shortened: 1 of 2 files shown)


awiddersheim:
Related somewhat to #308.

r0fls:
@awiddersheim this looks great, thanks! Can you also add the same in the `serve_multiple` function, and fix the pyflake errors?

sanic/sanic.py
```diff
@@ -273,10 +273,15 @@ def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
         :param protocol: Subclass of asyncio protocol class
         :return: Nothing
         """
-        logging.basicConfig(
-            level=logging.INFO,
-            format="%(asctime)s: %(levelname)s: %(message)s"
-        )
+        # Only set up a default log handler if the
+        # end-user application didn't set anything up.
+        if not logging.root.handlers and log.level == logging.NOTSET:
```
r0fls: Should this also replace the settings in serve_multiple? https://github.com/awiddersheim/sanic/blob/b459f9c8ed6b459d8c7328bf02746e0f8e6a395b/sanic/sanic.py#L374

PR #330, opened 2017-01-21
update aiohttp example

r0fls:
this example was incorrect -- it should not be required to pass the loop explicitly. We should update the aiopg example as well and possibly remove the loop argument from the multiple worker implementation.

### DIFF
--- a/examples/aiohttp_example.py
+++ b/examples/aiohttp_example.py
@@ -1,11 +1,8 @@
 from sanic import Sanic
 from sanic.response import json
 
-import uvloop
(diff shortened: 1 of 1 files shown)


r0fls:
Maybe this isn't necessary -- it originally uses a single worker. However I think removing some of the examples that pass the loop explicitly and replacing them with their counterparts that get the current event loop within the couroutine makes sense.

awiddersheim:
👍

r0fls:
closing in favor of #335

Issue #305, opened 2017-01-17
`Bad file descriptor` when running multiple workers with explicit loop under macos


gpip:
Hello,

When I run the following using either master's HEAD or the latest release:

```
import asyncio

import uvloop
from sanic import Sanic

asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
loop = asyncio.get_event_loop()

app = Sanic()
app.run(loop=loop, workers=2)
```

The following traceback is generated:

```
2017-01-17 03:17:51,090: INFO: Goin' Fast @ http://127.0.0.1:8000
2017-01-17 03:17:51,090: INFO: Spinning up 2 workers...
2017-01-17 03:17:51,096: ERROR: Unable to start server
Traceback (most recent call last):
  File "uvloop/loop.pyx", line 568, in uvloop.loop.Loop._add_reader (uvloop/loop.c:14392)
KeyError: 16

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".../lib/python3.6/site-packages/Sanic-0.2.0-py3.6.egg/sanic/server.py", line 303, in serve
    http_server = loop.run_until_complete(server_coroutine)
  File "uvloop/loop.pyx", line 1198, in uvloop.loop.Loop.run_until_complete (uvloop/loop.c:25480)
  File "uvloop/loop.pyx", line 1191, in uvloop.loop.Loop.run_until_complete (uvloop/loop.c:25324)
  File "uvloop/loop.pyx", line 1100, in uvloop.loop.Loop.run_forever (uvloop/loop.c:24056)
  File "uvloop/loop.pyx", line 363, in uvloop.loop.Loop._run (uvloop/loop.c:11491)
  File "uvloop/loop.pyx", line 216, in uvloop.loop.Loop._setup_signals (uvloop/loop.c:9069)
  File "uvloop/loop.pyx", line 570, in uvloop.loop.Loop._add_reader (uvloop/loop.c:14440)
  File "uvloop/handles/poll.pyx", line 30, in uvloop.loop.UVPoll.new (uvloop/loop.c:61297)
  File "uvloop/handles/poll.pyx", line 18, in uvloop.loop.UVPoll._init (uvloop/loop.c:61184)
OSError: [Errno 9] Bad file descriptor
2017-01-17 03:17:51,097: ERROR: Unable to start server
Traceback (most recent call last):
[...]

PR #332, opened 2017-01-21
set loop to None if using multiple workers

r0fls:
Idea from https://github.com/channelcat/sanic/issues/305. Not sure if we should log a warning or an error, or possibly even raise an exception. I think logging is better than raising an error, since the program might work correctly without passing the loop.


r0fls:
closing in favor of https://github.com/channelcat/sanic/pull/335

Issue #152, opened 2016-11-15
Sharing the loop with other libraries


masnun:
If we don't pass a loop to the `run` method, it creates a new loop and sets it as the event loop for asyncio. 

```
loop = loop or async_loop.new_event_loop()
asyncio.set_event_loop(loop)
```

But if any other libraries (for example motor) run any code involving the event loop, before we call the `run` method, they are likely to get the default event loop from asyncio before the above code executes. So they would be running on a separate event loop and my server would be on another event loop. 

The obvious solution is to create an uvloop ourselves and passing it to run method. Or deferring any event loop related codes until the server starts (may `await` from other `async` functions inside views). We can also choose use `get_event_loop` method. If uvloop is available and the `get_event_loop` is not using uvloop, we display a warning. 

Or we can simply add a detailed note somewhere as part of the documentation.

1st1:
My 2 cents on this: starting with Python 3.6 and 3.5.3, it will be recommended to never pass the loop explicitly. `asyncio.get_event_loop` was fixed to **always** return you the correct event loop when called from a coroutine.

masnun:
Instead of `new_event_loop`, calling `get_event_loop` should also fix this issue. But users might forget to use uvloop, so if uvloop is not being used, Sanic could print a warning or a message to let the user know that uvloop is available.

seemethere:
@masnun How should we go about knowing if the user isn't using the `uvloop.EventLoopPolicy` with their asyncio loop?

masnun:
Hi, we should be able to use this function - https://docs.python.org/3/library/asyncio-eventloops.html#asyncio.get_event_loop_policy to do so? 

I didn't try it out, a quick google got me there.

seemethere:
Taken care of in sanic 0.2.0 I think. Closing for now!

masnun:
Don't think it's fixed, but ok.

seemethere:
I'll reopen then @masnun

masnun:
Thanks. I will try to come up with some ideas or perhaps a PR. Let's see.

ubergarm:
I'm having some problems/confusion around mixing sanic `loop`s and `workers` when using `asyncio` or `aiohttp` as well.

Passing in an both explicit loop (asyncio / uvloop) and specifying multiple wokers e.g.: `app.run(host="0.0.0.0", port=8000, loop=loop, workers=4)` gives these errors:
```
[...]

Issue #275, opened 2017-01-07
No non-blocking way to start the server in the same thread/loop


nasfarley88:
Currently, there is no non-blocking way to start the server in the same thread/loop. This could possibly be achieved by having `Sanic.run` and subsequent funtions as a coroutine. Or having another function for running with async (e.g. `Sanic.run_async`)

seemethere:
What's the use-case?

nasfarley88:
In my case, there is a telegram bot (https://telegram.org/) running asynchronously, and it would be helpful for the api to have access to the current state of the bot.  At the moment, I run sanic in a different process which makes passing variables tricky (when it needn't be if it were running in the same asyncio/uvloop loop as the telegram bot).

JendaPlhak:
I couldn't agree more. Currently we have non-blocking adapter bodged together using internal sanic functions. It works, but it would be nice to have native support.

seemethere:
Alright I'll look into this for the next release 👍

nasfarley88:
@JendaPlhak Do you possibly have those internal sanic functions to hand? If so, it would be good to use as a stopgap until the next release.

JendaPlhak:
@nasfarley88 Here is our solution working with version `0.1.7`.

```
async def run_web_app(app, port, *, loop, logger, request_timeout=20):
    connections = {}
    signal = sanic.server.Signal()
    handler_factory = lambda: HttpProtocol(
        loop = loop,
        connections = connections,
        signal = signal,
        request_handler = app.handle_request,
        request_timeout = request_timeout,
        request_max_size = 1024*1024
    )

    server = await loop.create_server(handler_factory, None, port)
    for sock in server.sockets:
        sockname = sock.getsockname()
        logger.info("Listening on %s:%d", sockname[0], sockname[1])

    try:
        await asyncio.Future(loop = loop)
    finally:
        server.close()
        await server.wait_closed()

        # Complete all tasks on the loop
        signal.stopped = True
        for connection in connections.keys():
            connection.close_if_idle()
        while connections:
            await asyncio.sleep(0.1, loop = loop)
```

msychra:
I too would love to find this coroutine in sanic! This is actually the last thing I need in order to switch from aiohttp to sanic.

r0fls:
This is addressed by #348. See the example here: https://github.com/channelcat/sanic/pull/348/files#diff-7f0b2849596f21c41bbf61f1904cc707.

seemethere:
[...]

PR #335, opened 2017-01-23
remove loop as argument and update examples

r0fls:
This entirely removes the `loop` as an argument, from tests, examples, and sanic core. Since it's no longer going to be supported by asyncio (https://github.com/channelcat/sanic/issues/152#issuecomment-261316341) it doesn't make sense to *go against the grain*. All of the example modifications except the peewee one have been tested to work.

We'll probably want to issue a deprecation message for a few weeks before merging this.

### DIFF
--- a/examples/sanic_aiopg_sqlalchemy_example.py
+++ b/examples/sanic_aiopg_sqlalchemy_example.py
@@ -12,8 +12,6 @@ import sqlalchemy as sa
 from sanic import Sanic
 from sanic.response import json
 
-asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
-
 database_name = os.environ['DATABASE_NAME']
 database_host = os.environ['DATABASE_HOST']
 database_user = os.environ['DATABASE_USER']
@@ -23,8 +21,6 @@ connection = 'postgres://{0}:{1}@{2}/{3}'.format(database_user,
                                                  database_password,
                                                  database_host,
                                                  database_name)
-loop = asyncio.get_event_loop()
-
 
 metadata = sa.MetaData()
 
@@ -34,40 +30,37 @@ polls = sa.Table('sanic_polls', metadata,
                  sa.Column("pub_date", sa.DateTime))
 
 
-async def get_engine():
-    return await create_engine(connection)
-
 app = Sanic(name=__name__)
-engine = loop.run_until_complete(get_engine())
 
 
-async def prepare_db():
+async def prepare_db(app, loop):
     """ Let's add some data
 
     """
-    async with engine.acquire() as conn:
-        await conn.execute('DROP TABLE IF EXISTS sanic_polls')
-        await conn.execute("""CREATE TABLE sanic_polls (
-                                    id serial primary key,
-                                    question varchar(50),
-                                    pub_date timestamp
-                                );""")
-        for i in range(0, 100):
-            await conn.execute(
-                polls.insert().values(question=i,
-                                      pub_date=datetime.datetime.now())
-                )
+    async with create_engine(connection) as engine:
+        async with engine.acquire() as conn:
+            await conn.execute('DROP TABLE IF EXISTS sanic_polls')
+            await conn.execute("""CREATE TABLE sanic_polls (
+                                        id serial primary key,
+                                        question varchar(50),
+                                        pub_date timestamp
+                                    );""")
+            for i in range(0, 100):
+                await conn.execute(
+                    polls.insert().values(question=i,
+                                          pub_date=datetime.datetime.now())
+                    )
 
 
 @app.route("/")
 async def handle(request):
-    async with engine.acquire() as conn:
-        result = []
-        async for row in conn.execute(polls.select()):
-            result.append({"question": row.question, "pub_date": row.pub_date})
-        return json({"polls": result})
+    async with create_engine(connection) as engine:
+        async with engine.acquire() as conn:
+            result = []
+            async for row in conn.execute(polls.select()):
+                result.append({"question": row.question, "pub_date": row.pub_date})
+            return json({"polls": result})
 
 
 if __name__ == '__main__':
-    loop.run_until_complete(prepare_db())
-    app.run(host='0.0.0.0', port=8000, loop=loop)
+    app.run(host='0.0.0.0', port=8000, before_start=prepare_db)
--- a/examples/sanic_asyncpg_example.py
+++ b/examples/sanic_asyncpg_example.py
@@ -10,8 +10,6 @@ from asyncpg import create_pool
 from sanic import Sanic
 from sanic.response import json
 
-asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
-
 DB_CONFIG = {
     'host': '<host>',
     'user': '<username>',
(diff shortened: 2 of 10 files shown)


seemethere:
I'm a fan but let's leave this up for at least a week so that people can voice their opinions about it as well. If we don't find a compelling reason not to merge this change in by the end of the week this will be rolled into 0.3.0.

r0fls:
Tagging some of the open issues that are related:
https://github.com/channelcat/sanic/issues/305
https://github.com/channelcat/sanic/issues/152

youknowone:
Please see #338 also.

ubergarm:
It's definitely cleaner and less confusing to explicitly specify the loop everywhere. So :+1: 

It would be good to have another similar example showing how to limit the number of concurrent coroutines running using Semaphore or however is best to do so.

r0fls:
@youknowone that should be addressed by https://github.com/channelcat/sanic/issues/275. I have a half finished PR for that.

r0fls:
@ubergarm your implementation looks like the best idea we have for that, as of now. I'll add it as an example.

examples/sanic_aiopg_sqlalchemy_example.py
```diff
@@ -34,40 +30,37 @@
...
 
 
 @app.route("/")
 async def handle(request):
-    async with engine.acquire() as conn:
-        result = []
-        async for row in conn.execute(polls.select()):
-            result.append({"question": row.question, "pub_date": row.pub_date})
-        return json({"polls": result})
+    async with create_engine(connection) as engine:
```
Tim-Erwin: Where is `create_engine` defined?
Do we really want to recommend to create a new engine in each request?

Tim-Erwin:
While in general this seems to be a very good idea, it breaks applications where libraries need to use the same loop Sanic is running in. Let's assume I'm using an async database client which gets initialized before `run`ning the app, how is it supposed to get hold of the loop which is only created later? Do we need to use `before_start` for that? If so, this should at least be documented somehow.

examples/sanic_asyncpg_example.py
```diff
@@ -21,45 +19,40 @@
...
+                        (id, content, post_date) VALUES ({i}, {i}, now())""")
 
 
 @app.route("/")
 async def handler(request):
-    async with pool.acquire() as connection:
-        async with connection.transaction():
-            results = await connection.fetch('SELECT * FROM sanic_post')
-            return json({'posts': jsonify(results)})
+    async with create_pool(**DB_CONFIG) as pool:
```
r0fls: "Also I think create_pool is not defined." see line 8

examples/sanic_aiopg_sqlalchemy_example.py
r0fls: "Where is create_engine defined?" See line 9

Tim-Erwin: Ah, sorry, missed the top part of the file

r0fls:
@Tim-Erwin thanks for your review and feedback. Regarding:

>it breaks applications where libraries need to use the same loop Sanic is running in

One possibility in that scenario is using this new feature:
https://github.com/channelcat/sanic/pull/348

examples/sanic_aiopg_sqlalchemy_example.py
r0fls: I get caught by that all of the time, but that style of import is the way throughout sanic.

sanic/sanic.py
```diff
@@ -275,8 +274,8 @@ def converted_response_type(self, response):
 
     def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
             after_start=None, before_stop=None, after_stop=None, ssl=None,
-            sock=None, workers=1, loop=None, protocol=HttpProtocol,
-            backlog=100, stop_event=None, register_sys_signals=True):
+            sock=None, workers=1, protocol=HttpProtocol, backlog=100,
```
seemethere: Do you think it would be smart to leave the `loop` param (with a deprecation notice from `warnings.warn`) in for `0.3.0` and then officially remove it in the next release?

r0fls: @seemethere https://github.com/channelcat/sanic/pull/350

sanic/sanic.py
```diff
@@ -296,18 +295,11 @@ def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
         :param sock: Socket for the server to accept connections from
         :param workers: Number of processes
                         received before it is respected
-        :param loop: asyncio compatible event loop
         :param protocol: Subclass of asyncio protocol class
         :return: Nothing
         """
         self.error_handler.debug = debug
         self.debug = debug
-        if loop is not None:
```
seemethere: I think keeping them in for `0.3.0` and then taking them off for the next big version change will be the better choice here. Keep the messages but not the functionality of the loop is what I meant.

r0fls: aaah, I see 👍  that is an even better idea, since I want to see the loop removed as soon as possible -- but wait that also doesn't make sense: if it's already been deprecated, why give a deprecation warning? 🤔

r0fls: I'm confused 🤔  :octocat:

r0fls: Oh, I think I get it -- update this PR to leave the loop as an argument, but don't do anything and also print the message: is that what you have in mind? I can get behind that 😄

seemethere: Basically, with this PR we remove the loop functionality. But we keep the `loop` param but it will do nothing except trigger the warnings message so that anyone who was using the `loop` param won't just get a `TypeError` Exception and not know why. Then at the next big release say `0.4.0` we remove the param since we gave people enough of a heads up.

A lot of people just install sanic with `pip install sanic` so not having enough warning can potentially kill a lot of the apps people have built.

TL;DR

```python
if loop is not None:
    warnings.warn((
        '`loop` parameter will be removed after '
        'v`0.3.0` make sure to remove it before the next release!'))
```

seemethere:
Alright without a compelling argument against we will roll this into `0.3.0` scheduled for release tomorrow!

Tim-Erwin:
Wonderful. However, as mentioned in the before, it doesn't seem to be a good idea to create new connection pools in each request. Hence, we should not suggest to do that in our examples. While examples are just that - examples - they should give a *good* starting point. I think many will struggle with this part.

r0fls:
@Tim-Erwin I'm curious to see some performance statistics to back the claim that this is poor practice. While I agree, if you can avoid opening a connection for each query, that is better. However, It seems that passing the loop is not really supported by Python, so we *need* to move away from that. It's less of a performance concern than a sheer "this doesn't work" situation. You're welcome to create a PR or provide a suggestion to improve the examples. They're not going to be set in stone after this.

Tim-Erwin:
@r0fls The reason to use a connection pool is to avoid creating new connections over and over again, right? So even without any difference in performance creating a pool for just one request contradicts its purpose. I understand that passing the loop is no longer supported and that's a good thing, because IMHO the loop should be invisible as long as you don't intend to tinker with it.

In my current project with Sanic I'm explicitly creating the loop and passing it in because my DB client (which I create before running the app) has to use the same loop. So I'm really not an expert in how to do it differently. If I understand correctly we'd have to pass in a function `before_start` that's then provided the loop which we can use to init other libraries?

go2sh:
You have to remember that every context (process, thread) etc. has its own loop. You cannot share a loop over thread or process borders. So you need create a connection pool for every worker thread or process you have. From this point of view, the approach from @r0fls is correct.

But now the `before_start` function kicks in. Here you can create a connection per worker since it is called in the context of the worker (process). It shouldn't event need the loop argument as you can use `get_event_loop`.

But you are right, for a single request a connection pool is overkill since only one connection can be used.

PR #338, opened 2017-01-24
`serve` closes the loop only when it created its own loop.

youknowone:
`serve` closes the loop only when it created its own loop.

When you were running multiple applications using asyncio, sanic must be closed as the latest order because it closes the loop.

The patch limits the behavior only when sanic created its own loop by argument `loop=None`.
And `sanic_endpoint_test` now uses shared event loop by default.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -290,8 +290,10 @@ def serve(host, port, request_handler, error_handler, before_start=None,
     :param protocol: Subclass of asyncio protocol class
     :return: Nothing
     """
-    loop = loop or async_loop.new_event_loop()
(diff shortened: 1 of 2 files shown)


youknowone:
The test fails but it works on 3.5.3. If you are regarding sanic works for every 3.5.x, I will fix the code to check python version.

Tim-Erwin:
A change like this would also allow to use `sanic_endpoint_test` with passing a loop. On the other hand, passing in the loops is supposed to go away (https://github.com/channelcat/sanic/pull/335) making this obsolete.

Issue #143, opened 2016-11-08
high CPU usage with upload


devvit:
CPU usage in uploading:

flask + gevent: 5%-20%
nodejs: 5%-20%
sanic: 20%-90%

anybody noticed this?

webtic:
While I haven't compared it to other frameworks I did notice a heavy CPU load on upload with sanic..

seemethere:
Can't really say I can reproduce this, closing for now. Will reopen if necessary.

PR #342, opened 2017-01-25
Fix slow upload

channelcat:
Appending to the request body re-builds the byte array every time the body was received.  Updated to use a list and join when complete.


channelcat:
This should resolve #143

PR #343, opened 2017-01-25
Cookie deletion and default path

channelcat:
added the interface

```python
del response.cookies['cookie']
```

to expire client-side cookies

Also set the default cookie path to /

### DIFF
--- a/docs/cookies.md
+++ b/docs/cookies.md
@@ -64,7 +64,7 @@ parameters available:
 
 - `expires` (datetime): The time for the cookie to expire on the
                         client's browser.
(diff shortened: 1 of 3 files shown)


docs/cookies.md
```diff
@@ -32,12 +32,39 @@ async def test(request):
...
+    return response
+```
+
 Response cookies can be set like dictionary values and have the following
 parameters available:
 
 - `expires` (datetime): The time for the cookie to expire on the
                         client's browser.
-- `path` (string): The subset of URLs to which this cookie applies.
+- `path` (string): The subset of URLs to which this cookie applies.  Defaults to 0.
```
JordanP: Wrong line for this code comment addition.

PR #348, opened 2017-01-26
add async run

r0fls:
Addresses #275

### DIFF
--- a/sanic/sanic.py
+++ b/sanic/sanic.py
@@ -348,9 +348,8 @@ class Sanic:
         log.debug(self.config.LOGO)
 
         # Serve
-        if ssl is None:
-            proto = "http"
-        else:
+        proto = "http"
+        if ssl is not None:
             proto = "https"
         log.info('Goin\' Fast @ {}://{}:{}'.format(proto, host, port))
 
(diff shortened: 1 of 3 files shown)


sanic/sanic.py
```diff
@@ -348,9 +348,8 @@ def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
         log.debug(self.config.LOGO)
 
         # Serve
-        if ssl is None:
-            proto = "http"
-        else:
+        proto = "http"
```
JordanP: How is this change related to this PR ? What did you want to fix with this change ?

sanic/sanic.py
```diff
@@ -378,6 +377,63 @@ def stop(self):
             self.sock.close()
         get_event_loop().stop()
 
+    async def create_server(self, host="127.0.0.1", port=8000, debug=False,
+                            before_start=None, after_start=None,
+                            before_stop=None, after_stop=None, ssl=None,
+                            sock=None, loop=None, protocol=HttpProtocol,
+                            backlog=100, stop_event=None):
+        '''
```
JordanP: The rest of the code base use double quote.

sanic/sanic.py
```diff
@@ -378,6 +377,63 @@ def stop(self):
...
+            listeners = [partial(listener, self) for listener in listeners]
+            server_settings[settings_name] = listeners
+
+        server_settings['run_async'] = True
+
+        # Serve
+        proto = "http"
+        if ssl is not None:
+            proto = "https"
+        log.info('Goin\' Fast @ {}://{}:{}'.format(proto, host, port))
```
JordanP: Use lazy string evalutation:
```
log.info("Goin' Fast @ %s://%s:%s", proto, host, port)
```

sanic/sanic.py
seemethere: I'm assuming he wants to get rid of an unnecessary `else` statement.

JordanP: The else statement was not unnecessary, it's a matter of coding style. As there is's no way to enforce one coding style over the other, I am not sure we should encourage such code change, it only makes PR longer to review and sometimes puzzling. Any way, it's a minor details :)

sanic/sanic.py
r0fls: @JordanP this introduced an error where the arguments need to be explicitly converted. Please test any code changes if you're unsure, before you make recommendations.

JordanP: I am only here to make suggestion, in the end the author is in charge of writing the code. Also what I suggested was to use "%s" (read my unedited comment again) but what you commited used '{}' which obviously doesn't work. 

I see you are not sold to using the old formatting syntax, I agree with you. But in the case of logging, if you don't leverage lazy string evaluation, and your program heavily calls let's say log.debug(), then  in the ends it adds up. Maybe 1%, 2%, but isn't the point of Sanic to go fast ?

PR #351, opened 2017-01-26
use string formatting in start message

r0fls:


r0fls:
Reverting changes from review here: https://github.com/channelcat/sanic/pull/348#discussion_r97964602

PR #352, opened 2017-01-27
Added RFC 7233 Range HTTP/1.1 Range Requests

aquacash5:
Hey all,

I am currently rewriting my youtube to podcast web app using sanic. I noticed that the implementation of static files doesn't include a way to deal with range requests [RFC 7233](https://tools.ietf.org/html/rfc7233). So instead of writing my own implementation for my app, I tried to implement it directly into sanic.


sanic/response.py
```diff
@@ -97,40 +97,39 @@ def __init__(self, body=None, status=200, headers=None,
     def output(self, version="1.1", keep_alive=False, keep_alive_timeout=None):
         # This is all returned in a kind-of funky way
         # We tried to make this as fast as possible in pure python
-        timeout_header = b''
         if keep_alive and keep_alive_timeout:
-            timeout_header = b'Keep-Alive: timeout=%d\r\n' % keep_alive_timeout
-
+            if 'Keep-Alive' not in self.headers:
```
seemethere: Couldn't this conditional just be rolled up into the conditional above it?

sanic/response.py
```diff
@@ -97,40 +97,39 @@ def __init__(self, body=None, status=200, headers=None,
...
+                self.headers['Keep-Alive'] = keep_alive_timeout
+        if 'Connection' not in self.headers:
+            if keep_alive:
+                self.headers['Connection'] = 'keep-alive'
+            else:
+                self.headers['Connection'] = 'close'
+        if 'Content-Length' not in self.headers:
```
seemethere: How do these affect performance? I would imagine adding the extra effects could have some negative impacts

sanic/response.py
```diff
@@ -97,40 +97,39 @@ def __init__(self, body=None, status=200, headers=None,
...
+            self.headers['Content-Length'] = len(self.body)
+        if 'Content-Type' not in self.headers:
+            self.headers['Content-Type'] = self.content_type
         headers = b''
         if self.headers:
             for name, value in self.headers.items():
                 try:
-                    headers += (
-                        b'%b: %b\r\n' % (name.encode(), value.encode('utf-8')))
+                    headers += (b'%b: %b\r\n' % (
```
seemethere: I prefer the hanging indent here, it keeps all the logic on the same line. Same goes for the line below it as well

sanic/response.py
```diff
@@ -142,13 +141,62 @@ def cookies(self):
...
+        return self.size > 0
+
+    @staticmethod
+    def parse_range(range_header):
+        unit, _, value = tuple(map(str.strip, range_header.partition('=')))
+        if unit != 'bytes':
+            return None
+        start_b, _, end_b = tuple(map(str.strip, value.partition('-')))
+        try:
+            start = int(start_b) if start_b.strip() else None
```
seemethere: Why not just set it equal to 0 in this case? Then you don't have to do the None checks above

sanic/response.py
```diff
@@ -142,13 +141,62 @@ def cookies(self):
...
+            end = int(end_b) if end_b.strip() else None
+        except ValueError:
+            return None
```
seemethere: Returning None here when you expect to return a tuple could prove problematic and would raise a `TypeError`

sanic/response.py
```diff
@@ -142,13 +141,62 @@ def cookies(self):
...
+
+    def __init__(self, request, stats):
+        self.start = self.size = 0
+        self.end = None
+        self.headers = dict()
+        self.total = stats.st_size
+        _range = request.headers.get('Range')
+        if _range:
+            self.start, self.end = ContentRangeHandler.parse_range(_range)
+            if self.start is not None and self.end is not None:
```
seemethere: Is there a way to simplify this logic? It seems like a lot of conditionals for this?

sanic/response.py
aquacash5: I similified this im my latest commit

sanic/response.py
aquacash5: I could see how this could affect performance, so I changed it to use a default_header dictionary and ChainMap.

sanic/response.py
aquacash5: I have simplified the parsing logic

sanic/response.py
aquacash5: Done in new logic

sanic/response.py
```diff
@@ -142,13 +140,45 @@ def cookies(self):
...
+    __slots__ = ('start', 'end', 'size', 'total', 'headers')
+
+    def __init__(self, request, stats):
+        self.size = self.start = 0
+        self.end = None
+        self.headers = dict()
+        self.total = stats.st_size
+        _range = request.headers.get('Range')
+        if _range is None:
+            return
```
seemethere: For these ones in particular would it be better to raise an exception here instead of returning `None` on an`__init__`? It seems like if anyone were to use this it'd be confusing if something went wrong in the middle.

aquacash5: I think I see what you mean. You want me to raise exceptions for the different fail cases.

seemethere: Yeah i think it would work out better that way instead of returning None in an `__init__`

aquacash5: That would make the outer logic simpler.

seemethere:
Seems like something went wrong. Did you try to do a rebase on this Pull Request?

aquacash5:
Sorry yes. Haven't done many of these yet. Do you want me to just re-submit the Pull request?

seemethere:
Yes I think it'd be better to do so.

PR #357, opened 2017-01-27
Move serve_multiple, fix tests

seemethere:
## Overview
Moves serve_multiple out of the app and fixes
multiprocessing tests so that they don't freeze pytest's runner.

## Other notes:
Also moves around some imports so that they are better optimized as
well.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -1,7 +1,6 @@
 import asyncio
 import os
 import traceback
-import warnings
 from functools import partial
 from inspect import isawaitable
 from multiprocessing import Process, Event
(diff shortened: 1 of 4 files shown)


sanic/server.py
```diff
@@ -362,3 +371,49 @@ def serve(host, port, request_handler, error_handler, before_start=None,
...
+
+    :param server_settings: kw arguments to be passed to the serve function
+    :param workers: number of workers to launch
+    :param stop_event: if provided, is used as a stop signal
+    :return:
+    """
+    if stop_event is not None:
+        warnings.warn((
+            'Passing a stop_event will be removed in the version 0.4.0 of '
+            'sanic, please remove all references to it'))
```
r0fls: Deprecating stop event is a good idea 👍

PR #359, opened 2017-01-28
fix deprecation warnings

r0fls:


r0fls:
@seemethere too late to roll this into 0.3.0? The log.warning('message', DeprecationWarning) is in fact throwing an error, which is :(

seemethere:
@r0fls we can roll it for `0.3.1` tomorrow!

Issue #353, opened 2017-01-27
Consider using werkzeug for routing


Tim-Erwin:
Here's a test case:

```
def test_overload_dynamic_routes():
    app = Sanic('test_dynamic_route')

    @app.route('/overload/<param>', methods=['GET'])
    async def handler1(request, param):
        return text('OK1: ' + str(param))

    @app.route('/overload/<param>', methods=['POST', 'PUT'])
    async def handler2(request):
        return text('OK2: ' + str(param))

    request, response = sanic_endpoint_test(app, 'get', uri='/overload/bla')
    assert response.text == 'OK1: bla'

    request, response = sanic_endpoint_test(app, 'post', uri='/overload/bla')
    assert response.text == 'OK2: bla'

    request, response = sanic_endpoint_test(app, 'put', uri='/overload/bla')
    assert response.text == 'OK2: bla'

    request, response = sanic_endpoint_test(app, 'delete', uri='/overload')
    assert response.status == 405

    with pytest.raises(RouteExists):
        @app.route('/overload', methods=['PUT', 'DELETE'])
        async def handler3(request):
            return text('Duplicated')
```

The reason is, that while a route gets replaced with a CompositionView in `routes_static` and `routes_all`, it still lives in `routes_dynamic` next to the CompositionView.

@youknowone , since you implemented that feature you're probably most familiar with that section of the code. Would you mind fixing this issue?

youknowone:
I found the reason and I tried an ad-hoc fix but It is very unreliable.
I think using werkzeug makes the problem easier, which is used by flask.

Tim-Erwin:
Ya, that's what I thought first when I saw that Sanic implements its own routing. werkzeug routing is very powerful and the library also offers other very helpful and well tested stuff. The only thing i don't like about it is, that it has some focus on WSGI. However, that's not the biggest part.

@seemethere @r0fls @channelcat What do you think about replacing Sanic's routing with werkzeug?

youknowone:
I think sanic should do. And more.

If I have right of voting for future of sanic, I believe sanic should import designs and codes of flask as much as possible. The are maintained and tested. Both projects have compatible licenses.

I suggest the strategy:
1. Set basic design policy of sanic which will be different. (ex: context-local is not sanic)
2. Copy designs of flask unless it conflicts to 1 or it has defects.
3. Implementations can be made in a cheap way. Bring codes or write new codes.
4. For non-async codes, we even can bring codes. like routing.

[...]

PR #360, opened 2017-01-28
Fixes route overloading for dynamic routes

seemethere:
# Overview

Addresses #353, now dynamic routes work alongside our newly minted
overloaded routes! Also fixed an unintended side effect where methods
were still being passed in as None for `Sanic.add_route`.

I'll add in some tests so this shouldn't be merged until I add this in.


seemethere:
Actually, this doesn't fully work with HTTPMethodView, due to the way it gets added. Maybe it would be better to just have an add_view function that dynamically gets what methods the view is allowed for? What do you think @AntonDnepr?

seemethere:
Actually have a fix for HTTPMethodView, but it isn't ideal. Also for some reason calls to `HEAD` method endpoints don't work for me locally. 🤔

r0fls:
@seemethere how're you testing the HEAD method? This works for me:

```python
from sanic import Sanic

app = Sanic()

@app.head("/")
async def test(request):
    return text("OK")

app.run()
```

```bash
$ curl --head localhost:80
00
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Content-Length: 2
Connection: keep-alive
Keep-Alive: timeout=60
```

seemethere:
@r0fls I figured it out, the `HEAD` method doesn't return a body so trying to verify the body would be useless. Instead I just switched the test to verify the headers!

r0fls:
OK, I vote we add this test case:

```python
from sanic import Sanic
from sanic.utils import sanic_endpoint_test
from sanic.router import RouteExists
import pytest

@pytest.mark.parametrize("method,attr, expected", [
        ("get", "text", "OK1 test"),
        ("post", "text", "OK2 test"),
        ("put", "text", "OK2 test"),
        ("delete", "status", 405),
])
def test_overload_dynamic_routes(method, attr, expected):
    app = Sanic('test_dynamic_route')

    @app.route('/overload/<param>', methods=['GET'])
    async def handler1(request, param):
        return text('OK1 ' + param)

    @app.route('/overload/<param>', methods=['POST', 'PUT'])
    async def handler2(request, param):
        return text('OK2 ' + param)


    request, response = sanic_endpoint_test(app,
                                            method,
                                            uri='/overload/test')
    assert getattr(response, attr) == expected
```

Then open an issue when we close the other one, and include this test case:

```python
def test_overload_dynamic_routes_exist():
    app = Sanic('test_dynamic_route')

    @app.route('/overload/<param>', methods=['GET'])
    async def handler1(request, param):
        return text('OK1 ' + param)

    @app.route('/overload/<param>', methods=['POST', 'PUT'])
    async def handler2(request, param):
        return text('OK2 ' + param)


    # if this doesn't raise an error, than at least  assert response.text == 'Duplicated' should happen.
    # but neither do
    with pytest.raises(RouteExists):
        @app.route('/overload', methods=['PUT', 'DELETE'])
        async def handler3(request):
            return text('Duplicated')

    request, response = sanic_endpoint_test(app,
                                            'PUT',
                                            uri='/overload/test')
    assert response.text == 'Duplicated'
```

Unless we can get this to pass the second test before we merge this.

youknowone:
Thanks for a nice fix. Now I learned I tried a stupid way than an easy way.

About routing handlers, I suggest:
- class-based view can be a default view handler
- Add common method handler getter to both HTTPMethodView and CompositionView
- Add a wrapper or attribute to function handlers

Then they will share same interface as router's point of view.

About the get-handler overhead, I think `Router._get` can cache them with uri+method input to specific handler function output.

seemethere:
@youknowone I think we can simplify the views portion as well as the adding routes portion, for this PR though I think it's good to focus on the hot-fix for dynamic route overloading.

youknowone:
I also definitely agree about this PR. sorry for making confusion.

seemethere:
@r0fls I'll add those test cases tomorrow and then push up 0.3.1. Don't merge until I push up the test cases.

seemethere:
Test cases have been put in, will merge and roll into `0.3.1` later today.

PR #361, opened 2017-01-28
fix async run

r0fls:


r0fls:
@seemethere this fixes async run but I'm going to work on a test case.

PR #363, opened 2017-01-29
Run helper

r0fls:
Currently the initial logic is duplicated in `run` and `create_server`. This creates a helper method for that.


sanic/sanic.py
```diff
@@ -382,6 +330,56 @@ def stop(self):
...
+        log.info('Goin\' Fast @ {}://{}:{}'.format(proto, host, port))
+
+        return await serve(**server_settings)
+
+    def _helper(self, host="127.0.0.1", port=8000, debug=False,
+                before_start=None, after_start=None, before_stop=None,
+                after_stop=None, ssl=None, sock=None, workers=1, loop=None,
+                protocol=HttpProtocol, backlog=100, stop_event=None,
+                register_sys_signals=True):
+        """
```
seemethere: I'd say that for these helpers we don't really need to replicate the docstrings.

sanic/sanic.py
```diff
@@ -298,66 +298,14 @@ def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
...
-            log.setLevel(logging.DEBUG)
-        log.debug(self.config.LOGO)
-
-        # Serve
-        proto = "http"
-        if ssl is not None:
-            proto = "https"
-        log.info('Goin\' Fast @ {}://{}:{}'.format(proto, host, port))
-
+        server_settings = \
```
seemethere: What's the use of the backslash here? Just use hanging indent style instead.

sanic/sanic.py
```diff
@@ -382,6 +330,56 @@ def stop(self):
         """
         Asynchronous version of `run`.
         """
+        server_settings = \
```
seemethere: Backslash comment also applies here.

sanic/sanic.py
```diff
@@ -382,6 +330,56 @@ def stop(self):
...
+            self._helper(host=host, port=port, debug=debug,
+                         before_start=before_start, after_start=after_start,
+                         before_stop=before_stop, after_stop=after_stop,
+                         ssl=ssl, sock=sock, loop=loop,
+                         protocol=protocol, backlog=backlog,
+                         stop_event=stop_event)
+
+        server_settings['run_async'] = True
+
+        # Serve
```
seemethere: This is still duplicated in the _helper function.

PR #379, opened 2017-01-31
Let exception handler handle inherited exceptions

youknowone:
Original sanic exception handler only could handle exact matching exceptions. New `lookup` method will provide ability to looking up their ancestors without additional cost of performance.

### DIFF
--- a/sanic/exceptions.py
+++ b/sanic/exceptions.py
@@ -139,9 +139,12 @@ class PayloadTooLarge(SanicException):
 
 class Handler:
     handlers = None
+    cached_handlers = None
+    _missing = object()
 
     def __init__(self):
-        self.handlers = {}
+        self.handlers = []
+        self.cached_handlers = {}
         self.debug = False
 
     def _render_traceback_html(self, exception, request):
@@ -160,7 +163,18 @@ class Handler:
             uri=request.url)
 
     def add(self, exception, handler):
-        self.handlers[exception] = handler
+        self.handlers.append((exception, handler))
+
+    def lookup(self, exception):
+        handler = self.cached_handlers.get(exception, self._missing)
+        if handler is self._missing:
+            for exception_class, handler in self.handlers:
+                if isinstance(exception, exception_class):
+                    self.cached_handlers[type(exception)] = handler
+                    return handler
+            self.cached_handlers[type(exception)] = None
+            handler = None
+        return handler
 
     def response(self, request, exception):
         """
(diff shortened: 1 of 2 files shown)


sanic/exceptions.py
```diff
@@ -160,7 +163,20 @@ def _render_traceback_html(self, exception, request):
...
+        self.handlers.append((exception, handler))
+
+    def lookup(self, exception):
+        handler = self.cached_handlers.get(exception, None)
+        if handler is None:
+            for exception_class, handler in self.handlers:
+                if isinstance(exception, exception_class):
+                    self.cached_handlers[type(exception)] = handler
+                    return handler
+            handler = self._missing
```
r0fls: Why not just set handler to `self.default` here? Then in the next line: `self.cached_handlers[type(exception)] = object()`. Since `_missing` is not used anywhere else it  looks unnecessary.

youknowone: The handler can be `self.default` there, but `_missing` is not replacable by `object()`. In that case, `object()` will be created everytime whenever caching miss.

youknowone: @r0fls I fixed the short returning of self.default but keeping _missing.

youknowone:
I am sorry for changing the design a little bit after your last review.

- I swapped the usage of None and _missing. Now missing is the checker and None is the saved value.
- The error handler treat exception handler as an optional one. If error handler returns None, it will fall back to the default handler.

youknowone:
@r0fls will you review this again?

sanic/exceptions.py
```diff
@@ -170,9 +184,12 @@ def response(self, request, exception):
         :param exception: Exception to handle
         :return: Response object
         """
-        handler = self.handlers.get(type(exception), self.default)
+        handler = self.lookup(exception)
         try:
-            response = handler(request=request, exception=exception)
+            response = handler and \
```
r0fls: Can you remove the `\` and use a hanging indent instead? Example here: https://www.python.org/dev/peps/pep-0008/#indentation

sanic/exceptions.py
```diff
@@ -139,9 +139,12 @@ class PayloadTooLarge(SanicException):
 
 class Handler:
     handlers = None
+    cached_handlers = None
+    _missing = object()
```
r0fls: This is pedantic: but is there another variable name aside from `_missing` that makes sense here? For example, `_default`, `_base`, or `_empty_handler`?

youknowone: It actually means missing value. The value `_missing` means the given key does not exist in the cache (cache miss). For the default handler, the value `None` is saved in the cache.

youknowone:
@seemethere Any reason this code is removed in 579afe012ba6b77979bb776a2fb9983fc819b6ac ?

Issue #381, opened 2017-01-31
Clean up loop handling


Tim-Erwin:
Lately, there have been some changes regarding the loop which resulted in kind of a messy situation. I'd suggest, instead of making a bunch of small uncoordinated changes, let's discuss how we'd actually want to have it.

The issues I see are

1. `app.run()` accepts a `loop` parameter which it says will be deprecated in 0.4.0, however, it already ignores that parameter now. Let's either completely remove it or if we accept it, it should do something.
2. `app.run()` takes the default loop (via `_helper`) and passes it to `serve()`. Hence, the code creating a loop in  `serve()` will never execute.
3. As a consequence, Sanic never uses uvloop, except if one creates a loop beforehand and sets it as default loop. That's not documented at all and IMHO we should strive to using uvloop as a default.

While we're at it, we should fix `_helper` and it's usages. Some params are not used, another is misspelled.

I'd be happy to provide a PR, but let's agree on a bearing way first.

Tim-Erwin:
Let's also talk about testing in that context. I find it really hard to test a Sanic app. My current blocker is a `RuntimeError('Event loop is closed')` that gets raised whenever two `sanic_endpoint_test()`s are executed in the same test case. The problem here is that when Sanic stops, it closes the loop. In the next execution, it just takes the default loop and tries to work with that... closed.

I'm using pytest_async with quite some success, but here it  doesn't help. By the way, you can observe the same failure with current master. Let us try to find a good way for testing (API tests and unit tests) and document that. Any suggestions?

seemethere:
Yeah I'm observing the failure right now, I'm going to roll back the commit that cause it.

seemethere:
Fixed the RuntimeError with 487e3352e4c375aac84f80f502eefa6797bcefd2

seemethere:
No I understand the confusion, I think it stems from us trying to solve #275, but I don't think there's enough of a use-case to actually have a non-blocking way to run the server.

Tim-Erwin:
What I liked about the commit was `test_loop_policy.py`, could you re-add that?

seemethere:
I have an idea for how to handle the loop policy / loop sharing, I'll post it up in another commit but needless to say it'll involve addressing concerns over us messing with asyncio global settings (Which I'm not a fan of anyway.)

I'll have the PR up later this week.

r0fls:
Things are definitely a mess. This is what happens for me when I run simple_server.py example with debug set to `True`:

```
[...]

PR #385, opened 2017-02-01
No asyncio global settings

seemethere:
# Overview
This pull request aims to fix issues like #381, #305, #275, and #152. It adds an unambiguous way to get the loop from sanic and works to remove the need to set global asyncio settings as identified by @JendaPlhak here. Which in my opinion third party libraries, like sanic, should not do.

# Example:
```python
from sanic.policy import get_event_loop

def thing_that_needs_event_loop():
    loop = get_event_loop()
```

# Implementation
Basic implementation was done by storing an instantiated uvloop policy inside of sanic.policy and then using that to set the global loop for the sanic server at that time.

# What this means going forward
1. No more passing the loop
  * Server events before/after_start will be the preferred way of sharing the loop but you can also get it through sanic.policy.get_event_loop
2. There is not going to be a way to have a non-blocking way to start a Sanic application
  * There is no way to get that functionality to work with multiprocessing
  * The use cases could be fulfilled in different ways (like running separate apps that talk to each other through an API)

### DIFF
--- a/sanic/policy.py
+++ b/sanic/policy.py
@@ -1,0 +1,13 @@
+import uvloop
+
+_policy = uvloop.EventLoopPolicy()
+
+def get_event_loop():
+    return _policy.get_event_loop()
+
+def new_event_loop():
+    return _policy.new_event_loop()
+
+def set_event_loop(loop):
+    _policy.set_event_loop(loop)
+
--- a/sanic/__init__.py
+++ b/sanic/__init__.py
@@ -1,5 +1,6 @@
-from .sanic import Sanic
-from .blueprints import Blueprint
+from sanic.app import Sanic
+from sanic.blueprints import Blueprint
+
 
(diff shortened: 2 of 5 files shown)


sanic/policy.py
```diff
@@ -0,0 +1,13 @@
+import uvloop
+
+_policy = uvloop.EventLoopPolicy()
+
+def get_event_loop():
```
r0fls: I think it would make sense to add this as a property of the sanic app itself. E.g., have `sanic.loop` return the results of `get_event_loop()`.  What do you think? Also, does this `get_event_loop` always return the current, running event loop?

seemethere: It could possibly be good, maybe each app has it's own instantiated event policy?

seemethere:
This can be revisited later

PR #387, opened 2017-02-02
Add `url_for` method for simple routes, blueprints, HTTPMethodView

### LINKED ISSUE #336: no handy helper functions url_for
is there will be `url_for` in sanic likes `flask.url_for` in the future? I think it's not a good idea to hard code the url when redirect in code and make url in template.
And, it's better if `flash` and `get_flashed_messages` provided.

subyraman:
Adds in a `url_for` method to generate URLs based on an endpoint's handler name, similar to Flask. Provides the same type of functionality as Flask, but will additionally fail URL building if a supplied parameter does not pass the request parameter pattern. From the docs I wrote:

> 
> Sanic provides a `url_for` method, to generate URLs based on the handler method name. This is useful if you want to avoid hardcoding url paths into your app; instead, you can just reference the handler name. For example:
> 
> ```python
> @app.route('/')
> async def index(request):
>     # generate a URL for the endpoint `post_handler`
>     url = app.url_for('post_handler', post_id=5)
>     # the URL is `/posts/5`, redirect to it
>     return redirect(url)
> 
> 
> @app.route('/posts/<post_id>')
> async def post_handler(request, post_id):
>     return text('Post - {}'.format(post_id))
> ```
> 
> Other things to keep in mind when using `url_for`:
> 
> - Keyword arguments passed to `url_for` that are not request parameters will be included in the URL's query string. For example:
> ```python
> url = app.url_for('post_handler', post_id=5, arg_one='one', arg_two='two')
> # /posts/5?arg_one=one&arg_two=two
> ```
> - All valid parameters must be passed to `url_for` to build a URL. If a parameter is not supplied, or if a parameter does not match the specified type, a `URLBuildError` will be thrown.

Caveats: I did not try to make this work with `static` files or the `CompositionView`. The reason is I found no logical pattern for getting a view name for `static` handlers, since there is no unique "handler" surrounding a static file. Same for `CompositionView` handlers. 

One possible idea would be to force the naming of a class based view, like Flask does:

`app.add_route(SpecialClassView.as_view('special_class_name'), '/special_class_view')`

Another idea would be to make the static file API similar to routing, with unique handlers:

```
@app.static(/static/images/<image_file>)
def image_file():
   ...
```
But these were out of scope for this request.

Fixes #336

### DIFF
--- a/docs/sanic/blueprints.md
+++ b/docs/sanic/blueprints.md
@@ -158,3 +158,22 @@ app.blueprint(blueprint_v2)
 
 app.run(host='0.0.0.0', port=8000, debug=True)
 ```
+
+## URL Building with `url_for`
+
+If you wish to generate a URL for a route inside of a blueprint, remember that the endpoint name
+takes the format `<blueprint_name>.<handler_name>`. For example:
+
+```
+@blueprint_v1.route('/')
+async def root(request):
+    url = app.url_for('v1.post_handler', post_id=5)
+    return redirect(url)
+
+
+@blueprint_v1.route('/post/<post_id>')
+async def post_handler(request, post_id):
+    return text('Post {} in Blueprint V1'.format(post_id))
+```
+
+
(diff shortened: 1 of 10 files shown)


subyraman:
There's some issues with the CompositionView I want to clear up.

subyraman:
Reopening this; the CompositionView testing issues are unrelated to this branch. Going to file another PR to add tests for the CompositionView (it has none, currently)

PR #378, opened 2017-01-31
Added the tests, code formatting changes, and the Range Request feature.

aquacash5:

PR #398, opened 2017-02-09
Fixes errors related to #378

seemethere:


sanic/exceptions.py
```diff
@@ -154,87 +155,3 @@ def __init__(self, message, content_range):
 
 class InvalidRangeType(ContentRangeError):
     pass
-  
-  
-class Handler:
```
r0fls: Where does all of this code go?

seemethere: It's in `handlers.py` now

seemethere:
Fixes errors related to #378

sanic/exceptions.py
seemethere: The name also changed as well but I don't think too many people were subclassing it but we can add back a soft-link if people need it for legacy stuff

PR #399, opened 2017-02-09
improve url_for to support multi values and special options

lixxu:
improve `url_for` to support multi values for one arg, such as `arg1=value1&arg1=value2`, add `_anchor`/`_external`/`_scheme` options

option `_method` can be passed but will be ignored as I have no idea how to implement it.

### DIFF
--- a/tests/test_url_building.py
+++ b/tests/test_url_building.py
@@ -10,6 +10,14 @@ from sanic.exceptions import URLBuildError
 
 import string
 
+URL_FOR_ARGS1 = dict(arg1=['v1', 'v2'])
+URL_FOR_VALUE1 = '/myurl?arg1=v1&arg1=v2'
+URL_FOR_ARGS2 = dict(arg1=['v1', 'v2'], _anchor='anchor')
+URL_FOR_VALUE2 = '/myurl?arg1=v1&arg1=v2#anchor'
+URL_FOR_ARGS3 = dict(arg1='v1', _anchor='anchor', _scheme='http',
+                     _server='localhost:{}'.format(test_port), _external=True)
+URL_FOR_VALUE3 = 'http://localhost:{}/myurl?arg1=v1#anchor'.format(test_port)
+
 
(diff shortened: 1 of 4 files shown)


tests/test_url_building.py
```diff
@@ -179,11 +216,11 @@ def foo_with_param(request, param):
         return text(
             'foo from first : {}'.format(param))
 
-    @second_print.route('/foo') # noqa
+    @second_print.route('/foo')  # noqa
```
seemethere: There's no need for a `noqa` we don't run flake8 on the tests

tests/test_url_building.py
```diff
@@ -39,6 +39,30 @@ def test_simple_url_for_getting(simple_app):
         assert response.text == letter
 
 
+def test_simple_url_for_getting_with_duplicate_params(simple_app):
+    kw = dict(arg1=['value1', 'value2'], _anchor='anchor')
+    for letter in string.ascii_letters:
```
seemethere: Instead of doing this in a for loop why not use `pytest.mark.parameterize`?

lixxu: I just followed original author's style

tests/test_url_building.py
lixxu: `noqa` comment is from original author

lixxu: should I revert it back to be for accept? Please advise.

tests/test_url_building.py
lixxu: should I use `pytest.mark.parameterize` to update it? Please advise.

seemethere: I'd say `pytest.mark.parameterize` would be good for these tests. It makes sense to do it that way so we can have them run as separate tests instead of all at once.

subyraman:
Thanks for catching this, it'd be great if you could update the documentation in `routing.md` too!

lixxu:
updated test and docs, please review. thanks.

PR #401, opened 2017-02-10
Enhance error logging

youknowone:
- Prevent to fall into error logging recursion when it doesn't have
request info (Print 'Unknown' when the request is still `None`
- Print repr(e) rather than str(e)

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -179,7 +179,8 @@ class HttpProtocol(asyncio.Protocol):
                     self.request.ip))
         except Exception as e:
             self.bail_out(
-                "Writing response failed, connection closed {}".format(e))
+                "Writing response failed, connection closed {}".format(
+                    repr(e)))
         finally:
             if not keep_alive:
                 self.transport.close()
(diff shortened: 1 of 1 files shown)


seemethere:
I don't know if I would be entirely comfortable giving the consumer of an API a full stack trace if the server is not in debug mode.

youknowone:
When I met the error in the `write_error`, it was unable to find where it is raised at all. (because it is already inside of the error handling code).
I agree verbose full trace can be uncomfortable for normal `write_response` case. How about changing to the full trace for `write_error` and no trace in `write_response` like before?

seemethere:
@youknowone is there a way to check if we are in debug mode and then only write the full trace in those cases?

youknowone:
Ok, I fonud `bail_out` is already handling the traceback case. Maybe I just needed to add 'Unknown' to solve my problem. Just wrapped errors with repr() to show their class name which also has important information as the error message.

seemethere:
It appears as though you have a pep8 error:
`sanic/server.py:182:80: E501 line too long (80 > 79 characters)`

youknowone:
@seemethere thanks, I updated it

PR #403, opened 2017-02-10
add websocket support

r0fls:
Addresses #12. For temporary websocket support we could do something like this, simply using the websockets module.

### DIFF
--- a/examples/ws.html
+++ b/examples/ws.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>WebSocket demo</title>
+  </head>
+  <body>
+    <script>
+      var ws = new WebSocket("ws://127.0.0.1:3000/"),
+        messages = document.createElement('ul');
+      ws.onmessage = function (event) {
+        var messages = document.getElementsByTagName('ul')[0],
+          message = document.createElement('li'),
+          content = document.createTextNode(event.data);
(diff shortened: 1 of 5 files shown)


r0fls:
I guess one concern would be maintaining backward compatibility once we've implemented our own solution. However I think we can issue a deprecation statement leading up to our eventual transition to a different API.

seemethere:
Would this be better served as an extension?

r0fls:
Well, it seems it already exists as an extension. I just think providing an intuitive interface would help users. As opposed to overloading the `before_start` function with many use cases, as we've been advising people to. That seems sloppy to me. 

However, I'm thinking it would make sense to add a general method, like:

```
app.ensure_future(task)
```
This would allow you to do:

```python
import websockets
import random
import datetime
from sanic import Sanic

async def time(websocket, path):
    while True:
        now = datetime.datetime.utcnow().isoformat() + 'Z'
        await websocket.send(now)
        await asyncio.sleep(random.random() * 3)

start_server = websockets.serve(time, '127.0.0.1', 5678)
app = Sanic()
app.ensure_future(start_server)
@app.get()
def hello(request):
    return text('OK')

app.run()
```

@seemethere would you be able to get behind that instead?

r0fls:
https://github.com/r0fls/sanic-websockets/tree/master

PR #408, opened 2017-02-11
Use app decorator instead of run arguments for before_start

agoose77:
Mirror listener of blueprints instead of passing args to run.


sanic/sanic.py
```diff
@@ -54,6 +56,18 @@ def __init__(self, name=None, router=None,
...
 
     # Decorator
+    def listener(self, event):
+        """
+        Create a listener from a decorated function.
+
+        :param event: Event to listen to.
+        """
+        def decorator(listener):
+            self.listeners[event].append(listener)
```
r0fls: Can you make it so that listener method accepts either `before_server_start` or `before_start` as synonyms, etc...? There are a few ways you could do this (e.g. with a dict like {'before_start': 'before_server_start'}, then get the original thing as a default value, or some logic based on the string, where you insert `server` if it's missing)

sanic/sanic.py
```diff
@@ -458,6 +472,16 @@ def _helper(self, host="127.0.0.1", port=8000, debug=False, ssl=None,
...
                           DeprecationWarning)
 
+        # Deprecate this
+        if any(arg is not None for arg in (after_stop, after_start,
+                                           before_start, before_stop)):
+            if debug:
+                warnings.simplefilter('default')
+            warnings.warn("Passing a before_start, before_stop, after_start or"
+                          "after_stop callback will be deprecated in version"
+                          " 0.4.0",
```
seemethere: It'll probably be the next major version after 0.4.0, can you change it to say `in the next major version after 0.4.0`?

agoose77:
@r0fls It seems to me once we ask people to use new syntax, they should learn it directly - this mirrors the blueprint listener syntax already in place.

@seemethere made changes

seemethere:
@agoose77 Did you push the changes up to your remote?

seemethere:
Will merge pending @r0fls approval

PR #411, opened 2017-02-12
Ensure future

r0fls:
I'm not sure how this would work exactly with channelcat/sanic#408 but I'm convinced it's helpful.


r0fls:
I'll add a test an example @seemethere then you'll see how awesome this is :P

r0fls:
This simplifies integration with another server like websockets down to this:

```python
app = sanic.Sanic()

async def time(websocket, path):
    while True:
        now = datetime.datetime.utcnow().isoformat() + 'Z'
        await websocket.send(now)
        await asyncio.sleep(random.random() * 3)

start_server = websockets.serve(time, 'localhost', 3000)
app.ensure_future(start_server)

@app.route('/')
def hello(request):
    return html(open('ws.html').read())

app.run(port=8000)
```

Instead of how you would probably do this currently:

```python
app = sanic.Sanic()

async def time(websocket, path):
    while True:
        now = datetime.datetime.utcnow().isoformat() + 'Z'
        await websocket.send(now)
        await asyncio.sleep(random.random() * 3)

def init(app, loop):
    start_server = websockets.serve(time, 'localhost', 3000)
    loop.run_until_complete(start_server)

@app.route('/')
def hello(request):
    return html(open('ws.html').read())

app.run(port=8000, before_start=init)
```

It works with any Future, coroutine or awaitable, since it's just calling the `ensure_future` method later.

tests/test_create_task.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @app.route('/late')
+    async def set(request):
+        await asyncio.sleep(0.1)
+        return text(e.is_set())
+
+
+    request, response = sanic_endpoint_test(app, uri='/early')
+    assert response.body == b'False'
+
```
seemethere: Maybe a sleep should be added here just in case?

r0fls: I do the sleep in the route, but yes that would also work or both wouldn't hurt. But, if you think about the chain of events, 0.1 in the `/late` route should always be enough. Since the first call to `sanic_endpoint_test` starts the `coro` which only sleeps for 0.05 seconds. Then this route (`/late`) sleeps for twice that time, and doesn't start until afterwards. So, I think it's sufficient length...

r0fls: I just dislike those tests that pause for a long time because my heart always sinks thinking I broke something :)

r0fls: If you're still not convinced I can add it :)

seemethere: No it's okay I think it should be fine.

PR #412, opened 2017-02-12
Header performance gains

seemethere:
Changes from #378 introduced about a 10k request/sec slowdown. This
tries to rememdy it while keeping the same functionality but it's still
not as fast as 0.3.1

With the normal benchmarking I get about:
```console
  2 threads and 100 connections
    Latency     2.92ms  673.12us   9.52ms   78.07%
    Req/Sec    17.10k   605.41    18.11k    72.50%
  340654 requests in 10.02s, 40.28MB read
Requests/sec:  34010.41
Transfer/sec:      4.02MB
```


seemethere:
NOTE: This does not bring us back to previous speeds but it does bring a speed boost we still need to figure out ways to eek out more speed without squashing default headers that the user wants to set.

PR #417, opened 2017-02-13
Fixed aiocache example according to new loop policy

argaen:


r0fls:
@argaen @seemethere shouldn't this be using the decorator now? https://github.com/channelcat/sanic/pull/408

seemethere:
@r0fls I'd agree, maybe I'm a bit too quick on the trigger.

r0fls:
I thought @agoose77 updated all of the examples but I guess not. @argaen or @agoose77 can you update this to use the `@app.listener('before_server_start')` decorator instead?

agoose77:
Well darn, sure thing, tomorrow?

argaen:
I've tried it now but it's giving an error:

```python
00:07 $ python examples/cache_example.py 
2017-02-14 00:07:20,586 WARNING aiocache.log(13) | cPickle module not found, using pickle
  File "examples/cache_example.py", line 24, in <module>
    @app.listener('before_server_start')
AttributeError: 'Sanic' object has no attribute 'listener'
```

I'll check @agoose77 's PR tomorrow :)

seemethere:
Hmm that's strange. Did you make sure to pull the latest from the upstream repo?

argaen:
Yup I've tried in master with same luck. In read the docs the function comes from the Blueprint rather than the app itself, maybe that's the reason

seemethere:
That definitely could be the reason. @agoose77 would you mind taking a look at it?

argaen:
oh my, seems `python setup.py develop` wasn't working until I uninstalled previous Sanic version. Opened https://github.com/channelcat/sanic/pull/418. Sorry for the noise :)

Issue #337, opened 2017-01-24
Sanic _violates_ HTTP, fails with HTTP 1.1 pipelining


squeaky-pl:
HTTP 1.1 improves performance upon HTTP 1.0 in two main areas:

1. *Keep-alive*. Possibility to send many (request, response) pairs on the same connections
2. *Pipelining*. Possibility to send many requests without waiting for a response after each of them. The responses must be returned in the same order that requests arrived. https://en.wikipedia.org/wiki/HTTP_pipelining

> HTTP/1.1 conforming servers are required to support pipelining. This does not mean that servers are required to pipeline responses, but that they are required not to fail if a client chooses to pipeline requests.

Sanic currently doesnt support pipelining, worse it will fail very badly by dropping responses if the client attempts pipelining. This is a HTTP 1.1 protocol violation. If sanic doesn't support pipelining it should at least not consume new requests on the same connection until generating corresponding responses. This way pipelining clients will not break.

At the moment pointing pipelining client at sanic results in a stream of errors in error log:

```
2017-01-24 01:54:53,266: ERROR: Writing response failed, connection closed 'NoneType' object has no attribute 'should_keep_alive'
2017-01-24 01:54:53,266: ERROR: Writing response failed, connection closed 'NoneType' object has no attribute 'should_keep_alive'
2017-01-24 01:54:53,266: ERROR: Writing response failed, connection closed 'NoneType' object has no attribute 'should_keep_alive'
2017-01-24 01:54:53,267: ERROR: Writing response failed, connection closed 'NoneType' object has no attribute 'should_keep_alive'
...

```

You can reproduce issue with `wrk` passing following script into it:

```
-- pipeline.lua, example script demonstrating HTTP pipelining
-- based on: https://github.com/wg/wrk/blob/master/scripts/pipeline.lua

init = function(args)
   local r = {}
   r[1] = wrk.format(nil, "/")
   r[2] = wrk.format(nil, "/")
   r[3] = wrk.format(nil, "/")
   r[4] = wrk.format(nil, "/")
   r[5] = wrk.format(nil, "/")
   r[6] = wrk.format(nil, "/")   
   r[7] = wrk.format(nil, "/")
   r[8] = wrk.format(nil, "/")

   req = table.concat(r)
end

request = function()
   return req
end
```
and run it so:

```
wrk -t 1 -c 100 -d 10 -s pipeline.lua http://localhost:8000
Running 10s test @ http://localhost:8000
  1 threads and 100 connections
[...]

PR #423, opened 2017-02-14
fix pipelining

r0fls:
Addresses https://github.com/channelcat/sanic/issues/337. 

```
$ wrk -t 10 -c 100 -d 10 -s pipeline.lua http://localhost:8000
  10 threads and 100 connections
    Latency    47.21ms   31.48ms 174.15ms   71.39%
    Req/Sec     1.27k   641.14     4.04k    67.24%
  126848 requests in 10.06s, 15.00MB read
Requests/sec:  12608.15
Transfer/sec:      1.49MB
```

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -188,7 +188,6 @@ class HttpProtocol(asyncio.Protocol):
             else:
                 # Record that we received data
                 self._last_request_time = current_time
-                self.cleanup()
 
     def write_error(self, exception):
         try:
@@ -218,14 +217,6 @@ class HttpProtocol(asyncio.Protocol):
             exception = ServerError(message)
             self.write_error(exception)
             log.error(message)
-
-    def cleanup(self):
-        self.parser = None
-        self.request = None
-        self.url = None
-        self.headers = None
-        self._request_handler_task = None
-        self._total_request_size = 0
 
     def close_if_idle(self):
         """


r0fls:
@kolanos apparently you added the `cleanup` method, which doesn't seem necessary -- was there any reason for this, aside from it _seeming_ like a good idea? Cleaning up the connection is preventing it from being reused, which is in turn violating HTTP 1.1

zkanda:
Haha, just removing cleanup can indeed resolve the pipeline issue.
But it's 78% slower on my machine without it.

With cleanup:

```
➜  sanic git:(master) ✗ wrk -t 1 -c 100 -d 10 http://localhost:8000                        <<<
  1 threads and 100 connections
    Latency     2.29ms  767.36us  12.59ms   61.52%
    Req/Sec    43.41k     2.80k   45.72k    89.00%
  432394 requests in 10.03s, 52.78MB read
Requests/sec:  43125.84
Transfer/sec:      5.26MB
```

Without cleanup:

```
➜  sanic git:(master) ✗ wrk -t 1 -c 100 -d 10 http://localhost:8000                        <<<
  1 threads and 100 connections
    Latency    11.07ms    7.34ms  60.46ms   67.29%
    Req/Sec     9.43k     6.52k   43.01k    89.00%
  93852 requests in 10.03s, 11.46MB read
Requests/sec:   9359.23
Transfer/sec:      1.14MB
```

r0fls:
Derp :) I swear we need more tests...

r0fls:
@zkanda btw thanks, I thought we had just gotten much slower :facepalm:

kolanos:
@r0fls I can't remember the reasoning behind the cleanup method, so remove/alter as necessary.

seemethere:
The cleanup method is so that keep-alive works as intended. After every request the variables that are set for the particular request should be cleaned up.

It's not a perfect implementation but it works for right now.

PR #429, opened 2017-02-14
Pep8 cleanups

agoose77:
Make function docstrings more PEP8 verb like.


seemethere:
If / when we do merge I'd suggest a Squash and Merge for this one @r0fls

PR #431, opened 2017-02-14
Add Flask-like `test_client` to replace `sanic_endpoint_test`

subyraman:
This is a big PR but it is fully backwards compatible with previous code. 

The `sanic_endpoint_test` method was incongruous with the Flask testing API, and also wasn't fully documented. Rework testing so that, similar to flask, you run tests against methods like `app.test_client.get`. The methods are essentially sugar around `sanic_endpoint_test`, so it preserves all the work done before.

From the docs I wrote:

> The `test_client` exposes `get`, `post`, `put`, `delete`, `patch`, `head` and `options` methods
> for you to run against your application. A simple example (using pytest) is like follows:
> 
> ```python
> # Import the Sanic app, usually created with Sanic(__name__)
> from external_server import app
> 
> def test_index_returns_200():
>     request, response = app.test_client.get('/')
>     assert response.status == 200
> 
> def test_index_put_not_allowed():
>     request, response = app.test_client.put('/')
>     assert response.status == 405
> ```
> 

Ripped the bandaid in this PR by rewriting the tests to use the `test_client` as well. 

Put in a deprecation notice that `sanic_endpoint_test` would be removed in version 1.0, can be changed if that's too aggressive.

### DIFF
--- a/sanic/utils.py
+++ b/sanic/utils.py
@@ -7,8 +7,8 @@ def sanic_endpoint_test(app, method='get', uri='/', gather_request=True,
                         *request_args, **request_kwargs):
     warnings.warn(
         "Use of sanic_endpoint_test will be deprecated in"
-        "version 1.0.  Please use the `test_client` available"
-        "on the app object.", DeprecationWarning)
+        "the next major version after 0.4.0.  Please use the `test_client` "
+        "available on the app object.", DeprecationWarning)
 
     test_client = TestClient(app)
     return test_client._sanic_endpoint_test(
--- a/sanic/utils.py
+++ b/sanic/utils.py
@@ -1,64 +1,16 @@
-import aiohttp
-from sanic.log import log
-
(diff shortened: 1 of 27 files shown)


sanic/utils.py
```diff
@@ -1,64 +1,16 @@
...
-        try:
-            return results[-1]
-        except:
-            raise ValueError(
-                "Request object expected, got ({})".format(results))
+                        debug=False, server_kwargs={},
+                        *request_args, **request_kwargs):
+    warnings.warn(
+        "Use of sanic_endpoint_test will be deprecated in"
+        "version 1.0.  Please use the `test_client` available"
```
seemethere: You can probably leave this to say `the next major version after 0.4.0`

JordanP:
I had a quick look at this PR and coming from a Flask background this way of testing app seems more natural to me.

PR #434, opened 2017-02-14
Fix mistake in warning

agoose77:


r0fls:
whoops that was backwards, @agoose77 this was previously correct

r0fls:
needs to be reverted

PR #438, opened 2017-02-15
add coverage to tox

r0fls:
Maybe we can add the coverage to tox, to try and encourage people to continue to write new tests?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,7 @@
 [run]
 branch = True
-source = sanic, tests
-omit = site-packages
+source = sanic
+omit = site-packages, sanic/utils.py
 
 [html]
 directory = coverage
--- a/tox.ini
+++ b/tox.ini
@@ -14,10 +14,13 @@ deps =
     aiohttp
     pytest
     beautifulsoup4
+    coverage
 
 commands =
     pytest tests {posargs}
-
+    coverage erase
+    coverage run -m sanic.sanic
+    coverage report
 
 [testenv:flake8]
 deps =


r0fls:
For the record
```
$ coverage report
Name                  Stmts   Miss Branch BrPart  Cover
-------------------------------------------------------
sanic/__init__.py         4      0      0      0   100%
sanic/__main__.py        25     25      6      0     0%
sanic/blueprints.py      90     66     24      0    21%
sanic/config.py          35     26      6      0    22%
sanic/constants.py        1      0      0      0   100%
sanic/cookies.py         67     45     24      0    26%
sanic/exceptions.py      36      8      2      0    74%
sanic/handlers.py        76     57     20      0    20%
sanic/log.py              2      0      0      0   100%
sanic/request.py        119     95     44      0    15%
sanic/response.py        62     44     12      0    24%
sanic/router.py         153    126     76      0    12%
sanic/sanic.py          243    195     99      0    14%
sanic/server.py         218    169     56      0    18%
sanic/static.py          52     41     22      0    15%
sanic/testing.py         60     46     10      0    20%
sanic/views.py           33     23     10      0    23%
-------------------------------------------------------
TOTAL                  1276    966    411      0    18%
```

zkanda:
Yes, I feel this is very important for any project.

What do think of adding a restriction on PR, when the coverage go down, it cannot be merge?

seemethere:
I like coverage but I don't think it's the end all be all of good test coverage. It's a nice place to get information from but I don't think it should fail the tests if coverage falls below a certain threshold.

r0fls:
@zkanda I thought about that, but I think I have to agree with @seemethere here. We don't want it to limit ingenuity, but it would be helpful to be there as a reminder and we could also have an initiative to improve test coverage at some point.

seemethere:
Well maybe to make it more visible we could use a service like codecov?

It's output is similar to this: https://github.com/codecov/example-python/pull/14

I'd just hate for it to get spammy on the PR's

zkanda:
Awesome guys, I guess we can look into something like codecov after.

I hate spam as well, how does other people/project deal with that?

seemethere:
I'm a fan, we'll add this in now but it won't be a blocking gate

PR #440, opened 2017-02-16
isinstance -> try/except

r0fls:
it's faster to use try/except here when these cookies are present

### DIFF
--- a/sanic/cookies.py
+++ b/sanic/cookies.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 import re
 import string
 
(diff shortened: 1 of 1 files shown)


sanic/cookies.py
```diff
@@ -104,13 +103,19 @@ def __setitem__(self, key, value):
...
         for key, value in self.items():
-            if key == 'max-age' and isinstance(value, int):
-                output.append('%s=%d' % (self._keys[key], value))
-            elif key == 'expires' and isinstance(value, datetime):
-                output.append('%s=%s' % (
-                    self._keys[key],
-                    value.strftime("%a, %d-%b-%Y %T GMT")
-                ))
+            if key == 'max-age':
+                try:
```
JordanP: Why do you want to format max-age as an int and fallback to a string ? You should just do: `output.append('%s=%s' % (self._keys[key], value))` without any try/except.

Also the previous code and this one is not semantically the same. In previous case, if key=='max-age' and value in a str, the `if` block will not be executed.

r0fls: >Why do you want to format max-age as an int and fallback to a string ?

Because it can be passed as an integer. The existing logic does this.

r0fls: >Also the previous code and this one is not semantically the same. In previous case, if key=='max-age' and value in a str, the if block will not be executed.

Yes, but I'm saying it should be. At that point, you know the cookie key is expires or max-age -- why continue to compare it to other things, when you can just format it one way and then fallback if it fails? This results in fewer comparisons in the presence of those keys, and hence approximately %4 speedup.

sh0uw4ng: [241](https://github.com/channelcat/sanic/issues/241)
Have you tested with large illegal requests?  I guess this will cause the _stack overflow_ happen.

r0fls: @sh0uw4ng how large are we talking here? What was preventing a stack overflow before? With a large invalid request it would still make it to https://github.com/r0fls/sanic/blob/6726affa7e5802833bd84e02c7b49dbec89a6829/sanic/cookies.py#L123 logic before this change.

sh0uw4ng: @r0fls I'm sorry. I got it wrong.

Issue #449, opened 2017-02-16
ujson soft requirement


abuckenheimer:
I'm as big a fan as ujson as the next but it would be nice to have json parsing/serialization degrade gracefully if you can't fullfill the requirement. Be happy to submit a PR, would look more or less would like what [bottle does](https://github.com/bottlepy/bottle/pull/933/files)

seemethere:
What's the issue with `ujson`?

abuckenheimer:
it's just slightly less portable than the stdlib verison, for example I hit a
`ImportError: /lib64/libc.so.6: version 'GLIBC_2.14' not found`
on rhel6 but it works fine on rhel7. I'll probably find a way to work around that somehow but be an easy change if here if it makes sense to you.

r0fls:
I'm fine with using that logic but let's leave it in requirements.txt imo.

abuckenheimer:
awesome, thanks all

gsakkis:
`ujson` is great but looks like [it has lost steam](https://github.com/esnme/ultrajson/issues/246), including open issues and PRs about segfaults and memory leaks. Rapidjson looks more healthy these days, though ideally the json codec used by Sanic should be configurable. I can work on a PR if you agree.

SiDChik:
it would be much better to use rapidjson instead ujson.

PR #450, opened 2017-02-16
#449 use stdlib json module if ujson is unavailible

abuckenheimer:
#449

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -3,11 +3,13 @@ from collections import namedtuple
 from http.cookies import SimpleCookie
 from httptools import parse_url
 from urllib.parse import parse_qs
-
-from ujson import loads as json_loads
-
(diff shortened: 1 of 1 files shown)


sanic/request.py
```diff
@@ -3,11 +3,13 @@
...
-from ujson import loads as json_loads
-
 from sanic.exceptions import InvalidUsage
-from sanic.log import log
+try:
+    from ujson import loads as json_loads
+except ImportError:
+    from json import loads as json_loads
+
+from .log import log
```
r0fls: this should be `sanic.log`

abuckenheimer: eww I'm on the behind on the repo somehow

abuckenheimer:
just going to redo this so the history is cleaner

PR #455, opened 2017-02-16
Adds some safety around response types being wrong

seemethere:
Before we didn't check if the response object was actually a response
object. This PR catches the exception should it happen and reports it to
the user.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -167,12 +167,18 @@ class HttpProtocol(asyncio.Protocol):
     # -------------------------------------------- #
 
     def write_response(self, response):
+        keep_alive = (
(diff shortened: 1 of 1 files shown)


sanic/server.py
```diff
@@ -167,12 +167,18 @@ def on_message_complete(self):
     # -------------------------------------------- #
 
     def write_response(self, response):
+        keep_alive = (
```
r0fls: Why not leave this in the try/except? It seems that would be more useful in case the parser object was invalid (e.g. the current pipelining response)

seemethere: I don't want to catch that attribute error in the same try/except. I'm also thinking about bumping the parsing out of the protocol into another separate function where the HTTPRequest basically builds itself.

PR #459, opened 2017-02-17
Add app loop safety

seemethere:
`app.loop` should only be useable when the app is actually running.

This PR solves for that.

Also adds some docstrings surrounding `create_server`

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -10,7 +10,7 @@ from urllib.parse import urlencode, urlunparse
 
 from sanic.config import Config
 from sanic.constants import HTTP_METHODS
-from sanic.exceptions import ServerError, URLBuildError
+from sanic.exceptions import ServerError, URLBuildError, SanicException
 from sanic.handlers import ErrorHandler
 from sanic.log import log
 from sanic.response import HTTPResponse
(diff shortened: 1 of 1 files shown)


seemethere:
This will be a `squash and merge` when it gets merged in

sanic/app.py
```diff
@@ -50,13 +50,18 @@ def __init__(self, name=None, router=None, error_handler=None):
...
         self.listeners = defaultdict(list)
+        self.is_running = False
 
         # Register alternative method names
         self.go_fast = self.run
 
     @property
     def loop(self):
         """Synonymous with asyncio.get_event_loop()."""
+        if not self.is_running:
```
r0fls: Can we add a boolean `self.is_async` or something, then say `if not self.is_running or self.is_async` here? Then when using the run async option users can still call `app.loop` if wanted, which shouldn't be an issue.

seemethere: We could but if a user is running an app async chances are they probably already have a global loop that they want to use instead.

r0fls: Good point, wouldn't `asyncio.get_event_loop()` still return that loop though? In any case, this message is a little unclear: https://github.com/channelcat/sanic/pull/459/files#diff-c4444a35dbd5e37ee480b0e8888e0880R64

I would phrase that (for the current situation): "Loop can only be retrieved after the app has started running. Not supported with run_async"

seemethere: I'd agree, will push that change up

PR #465, opened 2017-02-19
WIP: Live Reload

seemethere:
This will the be the PR to track progress on the live reload feature as highlighted in #168 

# Notes:
### Things that need to be thought about for this feature are:
- How to effectively start / stop the app with the observer
  - Can we do it without having to do a subprocess call? (Probably not)
- What should we watch? (Is it all of `sys.path` like Flask? Or should it just the directory (recursive) of the original caller?)
- Should this be a part of `0.4.0` or should we cut a release for `0.4.0` and just roll this in for a possible `0.5.0`? (Probably `0.5.0`)
- Should watchdog be a soft-dependency? (I vote yes for this one)

### Things that have already been tried:
- Reusing the ReloaderLoop from `werkzeug` (Doesn't work with the signal handlers)

### Things that should be considered:
- Maybe we don't write any live_reload code and just document ways to do live reload like:
  - Using commands similar to http://entrproject.org/, http://pm2.keymetrics.io/ (from @mkorkmaz)

### DIFF
--- a/sanic/_live_reload.py
+++ b/sanic/_live_reload.py
@@ -1,0 +1,48 @@
+import os
+import signal
+import time
+
+from watchdog.events import RegexMatchingEventHandler
+from watchdog.observers import Observer
+
+
+class SanicEventHandler(RegexMatchingEventHandler):
+    def __init__(self, interval=1):
+        super().__init__(regexes=['.*\.py', '.*\.pyc', '.*\.pyo'])
+        self.interval = interval
+        self.last_reload = self.current_time
+
+    def on_created(self, event):
+        self.trigger_reload(event.src_path)
+
+    def on_modified(self, event):
+        self.trigger_reload(event.src_path)
+
+    def on_moved(self, event):
+        self.trigger_reload(event.src_path)
+
+    def on_deleted(self, event):
+        self.trigger_reload(event.src_path)
+
+    @property
+    def current_time(self):
+        return time.time()
+
+    def trigger_reload(self, src_path):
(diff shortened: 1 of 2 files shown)


subyraman:
Awesome. I'd be in favor of making live reload a hard dependency, this is one of those development tools that people can reasonably expect and makes the dev cycle a lot easier.

Watching `sys.path` sounds good so that the entire app isn't reloading when static resources change.

r0fls:
So does this branch work for live reload?

seemethere:
@r0fls Not fully yet

Right now it just has the file system handlers but it'll need a reloader loop to fully function but I think I might just scrap the idea and tell users to use external commands like `entr` or `pm2` if they want live reload.

I don't think the complexity of writing into the main package is worth it.

r0fls:
Ya, if we can just maybe have a section in the docs that explains how to do it and it's not too crazy to, that would probably be better I agree.

seemethere:
Yeah if it's as simple as:

```console
$ ls * | entr -r python my_app.py
```

Then I don't see why implementing live reload would be necessary.

lixxu:
@seemethere better to have if develop under Windows, or someone make an extension

messense:
With Gunicorn worker, reload is simple:

```
gunicorn --reload --bind localhost:8000 --worker-class sanic.worker.GunicornWorker app:app
```

webtic:
Is anyone using entr with MacOS? 
A `ls -d server.py | entr python3 server.py` is picking up any changes for me on 10.11.6

PR #467, opened 2017-02-20
condense methods

r0fls:
can't tell if I like this style better; could also add here if we do: https://github.com/channelcat/sanic/blob/master/sanic/app.py#L122


JordanP:
That would confuse IDE and autocompletion.

PR #469, opened 2017-02-20
websocket support

### LINKED ISSUE #12: Add web*socket* support

miguelgrinberg:
Fixes #12

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -19,7 +19,7 @@ from sanic.server import serve, serve_multiple, HttpProtocol
 from sanic.static import register as static_register
 from sanic.testing import TestClient
 from sanic.views import CompositionView
-from sanic.ws import WebSocketProtocol, ConnectionClosed
+from sanic.websocket import WebSocketProtocol, ConnectionClosed
 
 
 class Sanic:
@@ -52,7 +52,7 @@ class Sanic:
         self.sock = None
         self.listeners = defaultdict(list)
         self.is_running = False
-        self.needs_websocket = False
+        self.websocket_enabled = False
 
         # Register alternative method names
         self.go_fast = self.run
@@ -171,13 +171,13 @@ class Sanic:
         return handler
 
     # Decorator
-    def ws(self, uri, host=None):
+    def websocket(self, uri, host=None):
         """Decorate a function to be registered as a websocket route
         :param uri: path of the URL
         :param host:
         :return: decorated function
         """
-        self.needs_websocket = True
+        self.websocket_enabled = True
 
         # Fix case where the user did not prefix the URL with a /
         # and will probably get confused as to why it's not working
@@ -201,8 +201,17 @@ class Sanic:
 
         return response
 
-    def add_ws_route(self, handler, uri, host=None):
-        return self.ws(uri, host=host)(handler)
+    def add_websocket_route(self, handler, uri, host=None):
+        """A helper method to register a function as a websocket route."""
+        return self.websocket(uri, host=host)(handler)
+
+    def enable_websocket(self, enable=True):
+        """Enable or disable the support for websocket.
+
+        Websocket is enabled automatically if websocket routes are
+        added to the application.
+        """
+        self.websocket_enabled = enable
 
     def remove_route(self, uri, clean_cache=True, host=None):
         self.router.remove(uri, clean_cache, host)
@@ -501,8 +510,8 @@ class Sanic:
         :return: Nothing
         """
         if protocol is None:
-            protocol = WebSocketProtocol if self.needs_websocket \
-                else HttpProtocol
+            protocol = (WebSocketProtocol if self.websocket_enabled
+                        else HttpProtocol)
         server_settings = self._helper(
             host=host, port=port, debug=debug, before_start=before_start,
             after_start=after_start, before_stop=before_stop,
@@ -538,8 +547,8 @@ class Sanic:
               way to run a Sanic application.
         """
         if protocol is None:
-            protocol = WebSocketProtocol if self.needs_websocket \
-                else HttpProtocol
+            protocol = (WebSocketProtocol if self.websocket_enabled
+                        else HttpProtocol)
         server_settings = self._helper(
             host=host, port=port, debug=debug, before_start=before_start,
             after_start=after_start, before_stop=before_stop,
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -19,6 +19,7 @@ from sanic.server import serve, serve_multiple, HttpProtocol
 from sanic.static import register as static_register
 from sanic.testing import TestClient
 from sanic.views import CompositionView
+from sanic.ws import WebSocketProtocol, ConnectionClosed
 
 
 class Sanic:
@@ -51,6 +52,7 @@ class Sanic:
         self.sock = None
         self.listeners = defaultdict(list)
         self.is_running = False
+        self.needs_websocket = False
 
         # Register alternative method names
         self.go_fast = self.run
@@ -167,6 +169,40 @@ class Sanic:
 
         self.route(uri=uri, methods=methods, host=host)(handler)
         return handler
+
+    # Decorator
+    def ws(self, uri, host=None):
+        """Decorate a function to be registered as a websocket route
+        :param uri: path of the URL
+        :param host:
+        :return: decorated function
+        """
+        self.needs_websocket = True
+
+        # Fix case where the user did not prefix the URL with a /
+        # and will probably get confused as to why it's not working
+        if not uri.startswith('/'):
+            uri = '/' + uri
+
+        def response(handler):
+            async def websocket_handler(request, *args, **kwargs):
+                protocol = request.transport.get_protocol()
+                ws = await protocol.websocket_handshake(request)
+                try:
+                    # invoke the application handler
+                    await handler(request, ws, *args, **kwargs)
+                except ConnectionClosed:
+                    pass
+                await ws.close()
+
+            self.router.add(uri=uri, handler=websocket_handler,
+                            methods=frozenset({'GET'}), host=host)
+            return handler
+
+        return response
+
+    def add_ws_route(self, handler, uri, host=None):
+        return self.ws(uri, host=host)(handler)
 
     def remove_route(self, uri, clean_cache=True, host=None):
         self.router.remove(uri, clean_cache, host)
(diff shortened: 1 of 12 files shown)


miguelgrinberg:
Let me know if this is something you are open to incorporate into the project.

If you think this is worthwhile, I can take it to the finish line, which means:

- add unit tests
- add support for ws routes in blueprints

If you have any issues against merging this patch, let me know, and I'll look into releasing this as a separate extension.

Thanks!

sanic/app.py
```diff
@@ -168,6 +170,40 @@ def add_route(self, handler, uri, methods=frozenset({'GET'}), host=None):
         self.route(uri=uri, methods=methods, host=host)(handler)
         return handler
 
+    # Decorator
+    def ws(self, uri, host=None):
```
seemethere: Can we name something other than `ws`I don't think it's necessarily explicit in what it actually is. I think something like `websocket` or `web_socket` would be good here.

sanic/app.py
```diff
@@ -459,6 +495,9 @@ def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
         :param protocol: Subclass of asyncio protocol class
         :return: Nothing
         """
+        if protocol is None:
+            protocol = WebSocketProtocol if self.needs_websocket \
```
seemethere: Prefer parenthesis to backslashes here.

sanic/app.py
```diff
@@ -486,13 +525,16 @@ def stop(self):
...
                             before_stop=None, after_stop=None, ssl=None,
-                            sock=None, loop=None, protocol=HttpProtocol,
+                            sock=None, loop=None, protocol=None,
                             backlog=100, stop_event=None):
         """Asynchronous version of `run`.
 
         NOTE: This does not support multiprocessing and is not the preferred
               way to run a Sanic application.
         """
+        if protocol is None:
```
seemethere: Prefer parenthesis to backslashes here.

sanic/ws.py
```diff
@@ -0,0 +1,53 @@
+from sanic.server import HttpProtocol
+from httptools import HttpParserUpgrade
+from websockets import handshake, WebSocketCommonProtocol
+from websockets import ConnectionClosed  # noqa
+
+
+class WebSocketProtocol(HttpProtocol):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.ws = None
```
seemethere: I would say `ws` in this instance too should be renamed to `websocket` or `web_socket`

r0fls: +1 for websocket vs web_socket

seemethere: I'd say the same 👍

miguelgrinberg:
@seemethere @r0fls added a unit test and addressed your comments. The coverage report in tox is weird, coverage is off while the tests run, so the low coverage numbers you get on all the modules is incorrect. I enabled coverage here and measured 87% for the new websocket.py file, and 85% for the entire sanic package. Let me know if you have any more feedback.

jamesstidard:
I've been playing around with this branch and I'm getting an exception raised after a idle websocket client connection (maybe about a minute).

I created a client directly using [websockets](https://github.com/aaugustin/websockets), [here's the gist](https://gist.github.com/jamesstidard/2a38ffeff8861c708562f94b84b2137f), which didn't have this problem.

[Here's the Sanic version](https://gist.github.com/jamesstidard/eb619fcd598bb6e3eb946f8dca93cf32) and the stack trace I recieve:
```bash
2017-02-25 00:28:17,904: ERROR: NoneType: None

2017-02-25 00:28:17,905: ERROR: Traceback (most recent call last):
  File "/Users/jamesstidard/Development/Miscellany/Hello-Sanic/venv/src/sanic/sanic/app.py", line 431, in handle_request
    response = await response
  File "/usr/local/var/pyenv/versions/3.6.0/lib/python3.6/asyncio/coroutines.py", line 128, in throw
    return self.gen.throw(type, value, traceback)
  File "/Users/jamesstidard/Development/Miscellany/Hello-Sanic/venv/src/sanic/sanic/app.py", line 193, in websocket_handler
    await handler(request, ws, *args, **kwargs)
  File "/usr/local/var/pyenv/versions/3.6.0/lib/python3.6/asyncio/coroutines.py", line 128, in throw
    return self.gen.throw(type, value, traceback)
  File "/Users/jamesstidard/Development/Miscellany/Hello-Sanic/hellosanic/serve.py", line 29, in websocket
    message = await ws.recv()
  File "/Users/jamesstidard/Development/Miscellany/Hello-Sanic/venv/lib/python3.6/site-packages/websockets/protocol.py", line 284, in recv
    loop=self.loop, return_when=asyncio.FIRST_COMPLETED)
  File "/usr/local/var/pyenv/versions/3.6.0/lib/python3.6/asyncio/tasks.py", line 307, in wait
    return (yield from _wait(fs, timeout, return_when, loop))
  File "/usr/local/var/pyenv/versions/3.6.0/lib/python3.6/asyncio/tasks.py", line 390, in _wait
    yield from waiter
concurrent.futures._base.CancelledError

2017-02-25 00:28:17,906: ERROR: Exception occurred while handling uri: "/ws"
  File "/path/to/project/venv/src/sanic/sanic/app.py", line 431, in handle_request
    response = await response
  File "/usr/local/var/pyenv/versions/3.6.0/lib/python3.6/asyncio/coroutines.py", line 128, in throw
    return self.gen.throw(type, value, traceback)
  File "/path/to/project/venv/src/sanic/sanic/app.py", line 193, in websocket_handler
    await handler(request, ws, *args, **kwargs)
  File "/usr/local/var/pyenv/versions/3.6.0/lib/python3.6/asyncio/coroutines.py", line 128, in throw
    return self.gen.throw(type, value, traceback)
  File "/path/to/project/hellosanic/serve.py", line 29, in websocket
    message = await ws.recv()
  File "/path/to/project/venv/lib/python3.6/site-packages/websockets/protocol.py", line 284, in recv
    loop=self.loop, return_when=asyncio.FIRST_COMPLETED)
  File "/usr/local/var/pyenv/versions/3.6.0/lib/python3.6/asyncio/tasks.py", line 307, in wait
    return (yield from _wait(fs, timeout, return_when, loop))
  File "/usr/local/var/pyenv/versions/3.6.0/lib/python3.6/asyncio/tasks.py", line 390, in _wait
    yield from waiter
concurrent.futures._base.CancelledError
```

Let me know if there's anymore details I can give you, or if I'm just doing something silly.

miguelgrinberg:
@jamesstidard I missed that, thanks. Sanic imposes a 60 second timeout on handlers, I need to circumvent that for websocket routes.

jamesstidard:
@miguelgrinberg No problem. Thank you for taking the time to implement WebSockets - I got the impression it was a low priority.

miguelgrinberg:
@jamesstidard yeah, probably not high priority for the project admins, but I personally would like to have it so that I can have sanic support in my [socketio](https://github.com/miguelgrinberg/python-socketio) server.

miguelgrinberg:
@jamesstidard Let me know if your websocket client is now happy.

jamesstidard:
Seems to be behaving, Thank you.

r0fls:
I wouldn't say that it's not high priority. Just haven't had the time to implement it. I am curious though, from an ops perspective it might make more sense to have the websocket application be it's own process (outside of the sanic app). Curious what logic there is for having them run in the same loop, aside from convenience.

r0fls:
Not that I don't support this, since it at least makes sense on a small scale. On a large scale though, is it really the ideal way to run in production?

miguelgrinberg:
@r0fls As I mentioned above, I'm particularly interested in extending my Socket.IO server on to sanic (I already have aiohttp support). The Socket.IO protocol requires a combined HTTP + WebSocket interface on the same host and port, so it is really inconvenient if the servers are running separately, even if they share the same process as you currently recommend.

As far as plain websocket, I think there are use cases that benefit from a separate servers approach, like you say, but there are cases where that is an inconvenience. If you are into microservices, for example, you are probably well set up to have a separate websocket server that can share access to storage and other services. If you have a monolithic server, then there's going to be all sorts of problems if you are forced to host just the websocket service separately.

messense:
@miguelgrinberg 
Please take a look at my comment : https://github.com/channelcat/sanic/pull/545#issuecomment-286952147

dfee:
@miguelgrinberg I'm confused by your [statement](https://github.com/channelcat/sanic/blob/master/sanic/websocket.py#L14) that "timeouts make no sense for websocket routes".

I agree that makes sense, but the code doesn't seem to keep the connection alive. Was that intended, or was that more of a *TODO*?

I.e. currently, it seems that we need to currently manage the ping/pong to keep sanic from disconnecting.

miguelgrinberg:
@dfee That statement I made was in relation to the bug reported in https://github.com/channelcat/sanic/pull/545#issuecomment-286952147. The timeouts I was talking about were the ones sanic imposes on all routes, which default to 60 seconds if I remember correctly. A WebSocket route will last for as long as the two parties are connected, so you can't apply that kind of timeout.

> but the code doesn't seem to keep the connection alive.

Can you give me more details, or a way to reproduce this?

dfee:
@miguelgrinberg gotcha.

Here's what I've done to work around that auto-disconnect (I manually intervene):
```
        while True:
            try:
                data = await asyncio.wait_for(self.ws.recv(), 30)
            except asyncio.TimeoutError:
                await self.ws.ping()
                continue

            handle(data) # …
```

miguelgrinberg:
@dfee this "auto-disconnect" that you mention, is it the client or the server doing it? Sounds like if you need to send pings to the client to prevent a disconnection it might be the client that has that requirement. I guess we can add a background ping to prevent this, but I'm not sure this is a problem in general, it might be specific to the client that you are using.

dfee:
I'm using the latest version of Chrome. Does it not time out for you?

miguelgrinberg:
@dfee I don't recall getting timeouts, but maybe I did not leave the connection idle long enough. As far as I know, there isn't a required ping/pong frequency in the standard, this is left to handle by the client and server on their own. I'll look at how other ws servers do this, I honestly have no idea what the right frequency to ping might be. Consider that if you have 10K clients, you'll be running as many pings every 30 seconds.

dfee:
Oh man. I tried for a couple hours this morning to get to the root of it. I finally did. This is not a problem with `websockets` or `sanic`.

I needed to add the following to my nginx config:
```
proxy_connect_timeout 7d;
proxy_send_timeout 7d;
proxy_read_timeout 7d;
```

Thanks for your help @miguelgrinberg

PR #470, opened 2017-02-21
route logic for trailing slash

r0fls:


r0fls:
As captured by the new unit test, this should pass but currently fails:

```python

 @app.get('/get')
 def handler(request):
         return text('OK')
 
 request, response = app.test_client.get('/get')
 assert response.text == 'OK'
 
 request, response = app.test_client.get('/get/')
 assert response.text == 'OK'
```

r0fls:
I also considered putting the logic in the `_get` method but then it would be performed on every request.

PR #472, opened 2017-02-21
Pipelining and basic http2

channelcat:
Adds support for pipelining and basic http2.

Leaving this here for comments.

### DIFF
--- a/sanic/protocols/http2.py
+++ b/sanic/protocols/http2.py
@@ -22,7 +22,7 @@ def HTTP2Protocol(ssl, parent_class):
     ssl.set_alpn_protocols(["h2", "http/1.1"])
 
     _HTTP2Protocol = type("HTTP2Protocol", (HTTP2Checker, parent_class), {})
-    _HTTP2Protocol.poo = type("H2Protocol", (H2Protocol, parent_class), {})
+    _HTTP2Protocol.protocol_class = type("H2Protocol", (H2Protocol, parent_class), {})
 
     return _HTTP2Protocol
 
@@ -33,7 +33,7 @@ class HTTP2Checker:
         protocol = ssl_object.selected_alpn_protocol()
 
         if protocol == 'h2':
-            self.__class__ = self.poo
+            self.__class__ = self.protocol_class
             self.connection_made(transport)
         else:
             super().connection_made(transport)
--- a/sanic/protocols/http2.py
+++ b/sanic/protocols/http2.py
@@ -17,12 +17,12 @@ from sanic.log import log
 RequestData = collections.namedtuple('RequestData', ['headers', 'data'])
 
 
-def HTTP2Protocol(ssl, parent_class):
+def generate_http2_protocol(ssl, parent_class):
     ssl.set_ciphers("ECDHE+AESGCM")
     ssl.set_alpn_protocols(["h2", "http/1.1"])
 
     _HTTP2Protocol = type("HTTP2Protocol", (HTTP2Checker, parent_class), {})
-    _HTTP2Protocol.protocol_class = type("H2Protocol", (H2Protocol, parent_class), {})
+    _HTTP2Protocol.protocol_class = type("H2Protocol", (Http2Protocol, parent_class), {})
 
     return _HTTP2Protocol
 
@@ -39,7 +39,7 @@ class HTTP2Checker:
             super().connection_made(transport)
 
 
-class H2Protocol(asyncio.Protocol):
+class Http2Protocol(asyncio.Protocol):
 
     def connection_made(self, transport):
         # TODO: Request timeouts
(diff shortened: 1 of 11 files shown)


sanic/protocols/http2.py
```diff
@@ -0,0 +1,134 @@
...
+from h2.exceptions import ProtocolError
+
+from sanic.request import Request
+from sanic.log import log
+
+RequestData = collections.namedtuple('RequestData', ['headers', 'data'])
+
+
+def generate_http2_protocol(ssl, parent_class):
+    ssl.set_ciphers("ECDHE+AESGCM")
```
Lukasa: It may be worth adding ChaCha20 here as well: it's safe to use with HTTP/2 and is deployable with OpenSSL 1.1 and higher as well as with LibreSSL.

sanic/protocols/http2.py
```diff
@@ -0,0 +1,134 @@
...
+            self.connection_made(transport)
+        else:
+            super().connection_made(transport)
+
+
+class Http2Protocol(asyncio.Protocol):
+
+    def connection_made(self, transport):
+        # TODO: Request timeouts
+        self.h2_connection = H2Connection(client_side=False)
```
Lukasa: While the explicit kwarg here is supported now it's deprecated as of v2.5.0 in favour of an [explicit configuration object](https://python-hyper.org/h2/en/stable/api.html#configuration). This may provide a few other useful switches as well.

sanic/protocols/http2.py
```diff
@@ -0,0 +1,134 @@
...
+            self.transport.write(self.h2_connection.data_to_send())
+            for event in events:
+                if isinstance(event, RequestReceived):
+                    self.receive_headers(event.headers, event.stream_id)
+                elif isinstance(event, DataReceived):
+                    self.receive_data(event.data, event.stream_id)
+                elif isinstance(event, StreamEnded):
+                    self.stream_complete(event.stream_id)
+                elif isinstance(event, ConnectionTerminated):
+                    self.transport.close()
```
Lukasa: This code doesn't appear to handle flow control. It's definitely the most subtle part of the whole procedure, and you'll need it to get the server to be really stable. I appreciate this is an early draft of the patch so it's a natural thing to leave out, but for when you get around to it I wrote some [introductory documentation](https://python-hyper.org/h2/en/stable/advanced-usage.html#flow-control) that might help.

r0fls:
@seemethere are you waiting for @channelcat to address @Lukasa's concerns before merging this, or just still reviewing? Besides that it has some conflicts now...

Superman132:
We should merge this are remove it from the 0.5.0 milestone

sanic/server.py
```diff
@@ -298,6 +352,13 @@ def serve(host, port, request_handler, error_handler, before_start=None,
     :param protocol: subclass of asyncio protocol class
     :return: Nothing
     """
+
+    if http2:
+        if not ssl:
```
pickfire: Isn't SSL optional in HTTP2 even though most of the browser doesn't do http2 without ssl?

manisenkov:
Any progress on this?

sanic/server.py
navidnabavi: for upstreams and micro-services usages it is better to support http2 without ssl too. As it happened in grpc.

swift1911:
Any progress on this?

erm:
I have started on a PR to `uvicorn` that will implement `h2` support with flow control https://github.com/encode/uvicorn/pull/73 (work in progress).

It relies on ASGI for the request handling, but the protocol implementation is similar enough to Sanic that the flow control logic could be reused or serve as a guide for this PR once completed.

sayoun:
Sanic is a great framework and we will be using it in production in a couple of months, but we need HTTP/2 support, does anyone why this PR is stalled since last year ? @channelcat ? @seemethere ?

vlad1777d:
@r0fls , can http2 suppor be merged to master ?

ahopkins:
Hey everyone.... sorry for the stall out in this one. We're gearing up for a more minor release with cleanups for December, and will certainly work to tackle this in 2019. I think it is reasonable to aim for the 19.3 release given that the code is mostly complete, just needs to be updated and fully tested.

ahopkins:
Everyone following this, there is a discussion on [the community forums](https://community.sanicframework.org/t/considering-asgi-support/70) regarding adopting ASGI support, which would handle this using a third party package instead of with the internal server. Please join the conversation and make your opinions known.

ahopkins:
I am placing this PR on hold. The likely outcome is that we will just close this one altogether as we add these features into the code base for the 19.03 release in an alternative fashion.

sjsadowski:
This was an accidental close on my part, but I don't think I'm going to re-open it. There's a lot of work that would need to be done to bring it in line with the current codebase, and the branch is basically stale at this point.

ahopkins:
And, we will be addressing this in a different manner going forward.

PR #473, opened 2017-02-21
Add `stream` method for streaming content, add docs and examples

subyraman:
This PR adds in a new Response type, `StreamingHTTPResponse`, which can be asynchronously written to by a coroutine callback and streamed to the client. 

I made a purposeful attempt to avoid interfering with the main `write_response` path. This adds a new `HttpProtocol.stream_response` method that is passed as a callback to `Sanic.handle_request`. Inside `Sanic.handle_request` an instance check is made to determine if the response should be streamed or synchronously written to the transport, so performance remains the same.

Before this PR:

```
(sanic) sraman@sraman: sanic$ wrk http://127.0.0.1:8000/
Running 10s test @ http://127.0.0.1:8000/
  2 threads and 10 connections
    Latency     2.30ms  527.77us  10.43ms   91.70%
    Req/Sec     2.18k   125.53     2.36k    78.50%
  43427 requests in 10.01s, 5.30MB read
Requests/sec:   4337.96
Transfer/sec:    542.25KB
```

After this PR:

```
(sanic) sraman@sraman: sanic$ wrk http://127.0.0.1:8000/
Running 10s test @ http://127.0.0.1:8000/
  2 threads and 10 connections
    Latency     2.30ms  427.62us  10.79ms   90.26%
    Req/Sec     2.18k   111.73     2.33k    73.50%
  43397 requests in 10.01s, 5.30MB read
Requests/sec:   4334.05
Transfer/sec:    541.76KB
```

From the docs I wrote:

> # Streaming
> 
> Sanic allows you to stream content to the client with the `stream` method. This method accepts a coroutine callback which is passed a `StreamingHTTPResponse` object that is written to. A simple example is like follows:
> 
> ```python
> from sanic import Sanic
> from sanic.response import stream
> 
> app = Sanic(__name__)
> 
> @app.route("/")
> async def test(request):
>     async def sample_streaming_fn(response):
>         response.write('foo,')
>         response.write('bar')
> 
>     return stream(sample_streaming_fn, content_type='text/csv')
> ```
> 
> This is useful in situations where you want to stream content to the client that originates in an external service, like a database. For example, you can stream database records to the client with the asynchronous cursor that `asyncpg` provides:
> 
> ```python
> @app.route("/")
> async def index(request):
>     async def stream_from_db(response):
>         conn = await asyncpg.connect(database='test')
>         async with conn.transaction():
>             async for record in conn.cursor('SELECT generate_series(0, 10)'):
>                 response.write(record[0])
> 
>     return stream(stream_from_db)
> ```

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -1,7 +1,5 @@
-import asyncio
 from mimetypes import guess_type
 from os import path
-from inspect import isawaitable
 from ujson import dumps as json_dumps
 
 from aiofiles import open as open_async
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -1,5 +1,7 @@
+import asyncio
 from mimetypes import guess_type
 from os import path
+from inspect import isawaitable
 from ujson import dumps as json_dumps
 
 from aiofiles import open as open_async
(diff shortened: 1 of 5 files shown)


sanic/response.py
```diff
@@ -73,28 +73,121 @@
 }
 
 
-class HTTPResponse:
+class BaseHTTPResponse:
+    def _encode_body(self, data):
+        try:
+            # Try to encode it regularly
+            return data.encode()
+        except AttributeError:
```
r0fls: Can we check if it's already bytes here? Then we can avoid the round trip from bytes -> str -> bytes in that case.

seemethere: No this is for the case where it's a str in the first place, else it could be something like an int hence the AttributeError

subyraman: @seemethere I get what @r0fls is saying, the body being passed into the encoder might already be in bytes form (like if it is jsonified) so a type check is faster than attempting to re-encode it. I'll add it in.

subyraman: Actually looking at it, I don't think there's a situation where the body passed would be in bytes form, since `ujson.dumps` returns a `str`. Bytes should be passed into the response as `body_bytes`, so I think we're okay as it is.

r0fls: ok that makes sense then

PR #475, opened 2017-02-22
Make Sanic app instance accessible to blueprint

messense:
Blueprints are supposed to be some kind of pluggable components, it should be able to access current app instance to read configurations and do other useful things.

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -18,6 +18,7 @@ class Blueprint:
         :param name: unique name of the blueprint
         :param url_prefix: URL to be prefixed before all route URLs
         """
+        self.app = None
         self.name = name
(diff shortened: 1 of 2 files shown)


subyraman:
Hey @messense I think this should be closed in favor of #476, which will add a reference to `app` on every `request` object.

messense:
@subyraman Wouldn't that be overkill? For simple usage, direct access global `app` varialbe should good enough.

seemethere:
@messense I still think having the currently running app in the request handler is valuable. If not for the sole fact that it allows you to be explicit in that you do want the running app and not just an app that sits globally.

messense:
I am ok with having currently running app in the request handler.

PR #476, opened 2017-02-22
Inject app into request object

seemethere:
Allows for injection of the app object into the request object through
the request handler.

This allows for users to setup things like database connections etc. in
listeners and then utilize them throughout the their various route
handlers.

Usage is fairly simple like so:
```python
@app.get('/')
async def handler(request):
    request.app.anything
```

Name is up in the air but I'll leave this up for a few days and I'll
change it if we get a consensus

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -360,6 +360,8 @@ class Sanic:
             # Request Middleware
             # -------------------------------------------- #
 
+            request.app = self
+
             response = False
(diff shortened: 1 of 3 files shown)


jamesstidard:
Personally, I like `app`. More intuitive then and `config`, `settings` or `context`. You know you are getting the application instance you created in your `app = Sanic(__name__)` ritual.

PR #477, opened 2017-02-23
Fix test cases for requests post

messense:


messense:
Found another bug: #478

PR #481, opened 2017-02-23
url_for has something wrong with //

lixxu:
`url_for` not works on below code:
```python

bp = Blueprint('home', url_prefix='/')

@bp.route('/')
async def index(request):
    pass

@bp.route('/login')
async def login(request):
    pass

app.blueprint(bp)

# url_for('home.login') will get '//login'
# @bp.route('/login') must be @bp.route('login') to make it work
# after the merge, both ways will be work
```

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -41,7 +41,7 @@ class Blueprint:
             # Prepend the blueprint URI prefix if available
             uri = url_prefix + future.uri if url_prefix else future.uri
             app.route(
-                uri=uri,
+                uri=uri[1:] if uri.startswith('//') else uri,
                 methods=future.methods,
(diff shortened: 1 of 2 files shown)


sanic/blueprints.py
```diff
@@ -41,7 +41,7 @@ def register(self, app, options):
             # Prepend the blueprint URI prefix if available
             uri = url_prefix + future.uri if url_prefix else future.uri
             app.route(
-                uri=uri,
+                uri=uri[1:] if uri.startswith('//') else uri,
```
r0fls: why would the uri start with `//`?

r0fls: oh derp i see now :)

PR #482, opened 2017-02-24
Rework logging

messense:
Refs:

https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/

http://docs.python-guide.org/en/latest/writing/logging/#logging-in-a-library

Currently use the `sanic.log.log` is misleading, seeing some logs like:

```
2017-02-24 06:43:14,827 ERROR Process-15055 MainThread sanic.log:71    NoneType: None
```

logging configuration:

```python
LOGGING = {
    'version': 1,
    'disable_existing_loggers': False,
    'root': {
        'handlers': ['console'],
        'level': 'DEBUG'
    },
    'handlers': {
        'console': {
            'level': 'DEBUG',
            'class': 'logging.StreamHandler',
            'formatter': 'default',
            'stream': sys.stderr,
        }
    },
    'formatters': {
        'default': {
            'format': '%(asctime)s %(levelname)-2s Process-%(process)d %(threadName)s %(name)s.%(funcName)s:%(lineno)-5d %(message)s',  # NOQA
        },
    },
}
logging.config.dictConfig(LOGGING)
```

I don't really know what is going on, `sanic.log` module has only 3 lines of codes, makes it hard to debug.

### DIFF
--- a/examples/cache_example.py
+++ b/examples/cache_example.py
@@ -9,16 +9,17 @@ If you want more info about the package check
 https://github.com/argaen/aiocache
 """
 
+import logging
 import asyncio
 import aiocache
 
 from sanic import Sanic
 from sanic.response import json
-from sanic.log import log
 from aiocache import cached
 from aiocache.serializers import JsonSerializer
 
 app = Sanic(__name__)
+log = logging.getLogger(__name__)
 
 
 @app.listener('before_server_start')
--- a/examples/try_everything.py
+++ b/examples/try_everything.py
@@ -1,11 +1,12 @@
 import os
+import logging
 
 from sanic import Sanic
-from sanic.log import log
 from sanic.response import json, text, file
 from sanic.exceptions import ServerError
 
 app = Sanic(__name__)
+log = logging.getLogger(__name__)
 
 
 @app.route("/")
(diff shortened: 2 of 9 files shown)


messense:
WIP.

I'd like to know your opinions.

messense:
After this patch:

```
2017-02-24 07:22:11,425 ERROR Process-20767 MainThread sanic.handlers.log:73    NoneType: None
```

Now I know I should take a look at `sanic.handlers` module.

r0fls:
I'm all for the idea of reworking logging, but I think this leaves quite a bit to be wanted. I'm curious what people think about using an async logging approach e.g. http://pyzmq.readthedocs.io/en/latest/logging.html

messense:
@r0fls I think async logging approach can simply be configured by adding a handler?

```python
pub = context.socket(zmq.PUB)
pub.bind('tcp://*:12345')
handler = PUBHandler(pub)
logger = logging.getLogger()
logger.addHandler(handler)
```

messense:
@r0fls FYI:

[Logging HOWTO Basic Logging Tutorial](https://docs.python.org/3/howto/logging.html#handlers)

PR #486, opened 2017-02-24
The loop argument of serve()

mlyko:
I suppose that a new loop should be created only if the loop argument is not provided

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -298,7 +298,7 @@ def serve(host, port, request_handler, error_handler, before_start=None,
     :param protocol: subclass of asyncio protocol class
     :return: Nothing
     """
-    if not run_async:
+    if not loop:
         loop = async_loop.new_event_loop()
         asyncio.set_event_loop(loop)
 


r0fls:
@mlyko no we're actually not supporting passing the loop except in the async run, which is also not the recommended way to run except if absolutely necessary. We will be removing it as an argument soon. @seemethere when are we planning to remove it?

What are you hoping to pass the loop in for, or did you just notice that as weird?

seemethere:
@r0fls Well if the `create_server` function stays around there's actually no way to take out the loop as an argument.

r0fls:
@seemethere yes, from serve, but we can remove it from `run`

seemethere:
@r0fls Oh yes I'd agree to remove it from run altogether.

mlyko:
@r0fls thanks for the immediate feedback :) As already @seemethere discussed, the run() creates a loop and passes it to the serve() and next serve() creates another loop. That's way I thought it might be a mistake...

mlyko:
@r0fls By the way, do you plan to implement support for the HTTP client side (sendig HTTP requests)?

r0fls:
>run() creates a loop and passes it to the serve() and next serve() creates another loop.

I don't see that happening anywhere, feel free to point me to the line of code but I only see it being created in `serve`.

>@r0fls By the way, do you plan to implement support for the HTTP client side (sendig HTTP requests)?

That is not on the agenda. What would you need that for? Can you send them with something like the `requests` library?

mlyko:
> I don't see that happening anywhere, feel free to point me to the line of code but I only see it being created in serve.
Yes, you're right it has been removed in the repo. I see it in the sanic 0.3.1. But I still see it in the create_server().

> That is not on the agenda. What would you need that for? Can you send them with something like the requests library?
Well, it would be nice to have one lib for all HTTP jobs - asynchronous and faster than e.g. aiohttp :)

r0fls:
The goal is to be more of a modern flask then a full http swiss army knife. If there is room for a better aiohttp though, that is interesting. If you think it's slow, I wonder if that library is any faster when used with uvloop for the event loop? 

Anyway I'm going to close this issue since we don't want the loop being used there, as it causes issues. Cheers.

mlyko:
> I wonder if that library is any faster when used with uvloop for the event loop?

Yes, of course it is faster with uvloop, but still much slower then Sanic (in case the server side) that why I'm asking about client side implementation in Sanic :)
Regards.

PR #491, opened 2017-02-27
remove stop_event

r0fls:
This is doing precisely nothing so we should remove it unless someone voices a strong argument for keeping it.

To witness it doing nothing, use the below example:

```python
from sanic import Sanic
from sanic.response import json, text
from multiprocessing import Event

stop_event = Event()

app = Sanic(__name__)

@app.route("/")
async def test(request):
    stop_event.set()
    return text('ok')

app.run(host="0.0.0.0", port=8000, debug=True, workers=2, stop_event=stop_event)
```

This should stop after handling the first request, but does not.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -554,19 +554,24 @@ class Sanic:
         if protocol is None:
             protocol = (WebSocketProtocol if self.websocket_enabled
                         else HttpProtocol)
+        if stop_event is not None:
+            if debug:
+                warnings.simplefilter('default')
+            warnings.warn("stop_event will be removed from future versions.",
+                          DeprecationWarning)
         server_settings = self._helper(
             host=host, port=port, debug=debug, before_start=before_start,
             after_start=after_start, before_stop=before_stop,
             after_stop=after_stop, ssl=ssl, sock=sock, workers=workers,
             loop=loop, protocol=protocol, backlog=backlog,
-            stop_event=stop_event, register_sys_signals=register_sys_signals)
+            register_sys_signals=register_sys_signals)
 
         try:
             self.is_running = True
             if workers == 1:
                 serve(**server_settings)
             else:
-                serve_multiple(server_settings, workers, stop_event)
+                serve_multiple(server_settings, workers)
         except:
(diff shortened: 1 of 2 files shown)


sanic/server.py
```diff
@@ -404,11 +407,12 @@ def serve_multiple(server_settings, workers, stop_event=None):
     server_settings['host'] = None
     server_settings['port'] = None
 
-    if stop_event is None:
-        stop_event = Event()
+    def sig_handler(signal, frame):
+        log.info("Received signal {}. Shutting down.".format(
```
seemethere: If I remember correctly, last time I tried to do this it would raise an exception and spew a garbage exception in the logger when you tried to stop the server.

The specific commit for reference is ca9845988f24943dddc6b4351a3448f86a7402a3.

And then the re-adding commit is 4ca2e1d8e4dd21d9dcf2f120975bafdc13cd081d

r0fls: I think that is because you didn't replace the signal handlers...

seemethere: That's possible.

r0fls: Without doing something here https://github.com/channelcat/sanic/pull/491/files#diff-bd84a6e523017a7f3d6bd72ee689dce2R414 you will get errors.

sanic/app.py
```diff
@@ -514,7 +523,11 @@ def _helper(self, host="127.0.0.1", port=8000, debug=False,
                 protocol=HttpProtocol, backlog=100, stop_event=None,
                 register_sys_signals=True, run_async=False):
         """Helper function used by `run` and `create_server`."""
-
+        if stop_event is not None:
```
seemethere: Would this make it so that it would log the warning twice?

r0fls: Actually no, since i don't pass it through even if it's included as an argument. I could pass it through to `_helper` and then only log it once there though...

r0fls: or we could just not log anything or accept stop_event in `_helper`, since that shouldn't be used directly anyway

PR #507, opened 2017-03-02
Default to json if ujson isn't installed. Provide install extras

jkbbwr:
This allows for sanic to be installed on platforms that either don't have or cannot support uvloop and ujson. 

To install sanic with all the toys one can use `pip install sanic["all"]` but the base install should leave uvloop and ujson out as they are extras.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import codecs
 import os
 import re
 from setuptools import setup
-
+from distutils.util import strtobool
 
 with codecs.open(os.path.join(os.path.abspath(os.path.dirname(
         __file__)), 'sanic', '__init__.py'), 'r', 'latin1') as fp:
@@ -15,41 +15,49 @@ with codecs.open(os.path.join(os.path.abspath(os.path.dirname(
     except IndexError:
         raise RuntimeError('Unable to determine version.')
 
+uvloop = 'uvloop>=0.5.3'
+ujson = 'ujson>=1.35'
+
 install_requires = [
     'httptools>=0.0.9',
     'aiofiles>=0.3.0',
+    ujson,
+    uvloop
 ]
 
-ujson = [
-    'ujson>=1.35',
-]
+if strtobool(os.environ.get("SANIC_NO_UVLOOP", "false")):
+    install_requires.remove(uvloop)
 
-uvloop = [
-    'uvloop>=0.5.3',
-]
+if strtobool(os.environ.get("SANIC_NO_UJSON", "false")):
+    install_requires.remove(ujson)
 
-setup(
-    name='sanic',
-    version=version,
-    url='http://github.com/channelcat/sanic/',
-    license='MIT',
-    author='Channel Cat',
-    author_email='channelcat@gmail.com',
-    description=(
-        'A microframework based on uvloop, httptools, and learnings of flask'),
-    packages=['sanic'],
-    platforms='any',
-    install_requires=install_requires,
-    extras_require={
-        "all": ujson + uvloop,
-        "ujson": ujson,
-        "uvloop": uvloop,
-    },
-    classifiers=[
-        'Development Status :: 2 - Pre-Alpha',
-        'Environment :: Web Environment',
-        'License :: OSI Approved :: MIT License',
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: 3.6',
-    ],
-)
+
+def run_setup():
+    setup(
+        name='sanic',
+        version=version,
+        url='http://github.com/channelcat/sanic/',
+        license='MIT',
+        author='Channel Cat',
+        author_email='channelcat@gmail.com',
+        description=(
+            'A microframework based on uvloop, httptools, and learnings of flask'),
+        packages=['sanic'],
+        platforms='any',
+        install_requires=install_requires,
+        classifiers=[
+            'Development Status :: 2 - Pre-Alpha',
+            'Environment :: Web Environment',
+            'License :: OSI Approved :: MIT License',
+            'Programming Language :: Python :: 3.5',
+            'Programming Language :: Python :: 3.6',
+        ],
+    )
+
+
+try:
+    run_setup()
+except:
+    install_requires.remove(uvloop)
+    install_requires.remove(ujson)
+    run_setup()


jkbbwr:
Looks like the CI doesn't install flake8 which is why the CI is failing.

renstrom:
Funny, I was actually just about to suggest (and provide a solution) for this myself.

jkbbwr:
It got partially done for one of them but not the other.

sanic/response.py
```diff
@@ -1,6 +1,10 @@
 from mimetypes import guess_type
 from os import path
-from ujson import dumps as json_dumps
+
+try:
+    from ujson import dumps as json_dumps
+except ImportError:
+    from json import dumps as json_dumps
```
renstrom: This just a personal preference, by not just just do this? I guess it increase the the amount of changes needed but...

```python
try:
     import ujson as json
except ImportError:
    import json
```

jkbbwr: Tried to stay with the original function naming scheme.

sanic/request.py
```diff
@@ -7,7 +7,11 @@
 try:
     from ujson import loads as json_loads
 except ImportError:
-    from json import loads as json_loads
+    from json import loads as _json_loads
+
+    def json_loads(i):
+        i = i.decode() if isinstance(i, bytes) else i
+        return _json_loads(i)
```
renstrom: This just an FYI. But Python 3.6 added support for `json.loads()` on `bytes`. Not that it helps much since we also support 3.5.

```python
>>> import json
>>> json.loads(b'{"foo": "bar"}')
{'foo': 'bar'}
```

Anyway, this got me thinking a bit. Is the `isinstance` check really necessary? Since it's only used to parse the HTTP body, which should always be in `bytes`, right?

jkbbwr: The problem is python3.5 crashes on doing `json.loads(b"{}")` and this project needs to support Python 3.5

renstrom: Yes, but I mean it's safe to always run `i.decode()` since `i` will always be `bytes`.

```python
def json_loads(i):
    return _json_loads(i.decode())
```

jkbbwr: Is it always going to be bytes?

renstrom: Well as far as I can see it's only used by `Request.json` to parse the HTTP body, i.e. `self.body`, which should always be `bytes`. Or maybe I'm missing something? I'll leave that up to @channelcat.

jkbbwr: It passes tests without the instance check. So I will wait for their comment then push that commit if its needed.

sanic/request.py
```diff
@@ -7,7 +7,11 @@
 try:
     from ujson import loads as json_loads
 except ImportError:
-    from json import loads as json_loads
+    from json import loads as _json_loads
+
+    def json_loads(i):
+        i = i.decode() if isinstance(i, bytes) else i
```
seemethere: I think in the interest of speed here it'd be better off just doing a try, except. Try to decode if bytes, catch an `AttributeError` if the `decode` method does not exist. Similar to what we do in other places as well.

sanic/response.py
seemethere: I'm fine with the original naming scheme.

setup.py
```diff
@@ -36,6 +40,11 @@
     packages=['sanic'],
     platforms='any',
     install_requires=install_requires,
+    extras_require={
+        "all": install_requires + ujson + uvloop,
+        "ujson": install_requires + ujson,
+        "uvloop": install_requires + uvloop,
```
renstrom: Is `install_requires` really supposed to be here? Shouldn't this just contain the *extra* packages, i.e. whatever package besides what's already defined in `install_requires`.

renstrom: For reference: https://github.com/kennethreitz/requests/blob/master/setup.py#L97-L100

jkbbwr:
@seemethere Updated following your feedback let me know if I missed anything :)

subyraman:
Do we want to remove uvloop and ujson entirely as dependencies? Without uvloop+sanic I get nearly half the performance. I get the feeling that this will confuse users. 

Removing them from the initial install might confuse users about performance. Can't we check for OSes in the install process instead, and offer warnings at runtime that there is degraded performance without uvloop?

jkbbwr:
If you could make pip do subtractive installs It wouldn't be a problem, but it only does additive.

jkbbwr:
Plus uJSON will build fine on lots of systems then fail at runtime, and uvloop fails to compile on some systems.

subyraman:
uvloop is [close-ish](https://github.com/MagicStack/uvloop/issues/14#issuecomment-280288339) to being buildable on Windows. 

Without the uvloop the main performance benefits of Sanic really collapse. It'd be worth thinking about this before adding uvloop as a simple "extra".

jkbbwr:
uvloop is not currently buildable for pypy is another example. 
Any developer can read the documentation and execute the "all" extra's. Otherwise those that need specific components removed can do so.

renstrom:
I have no opinion about uvloop, but I definitely think ujson should be opt-in.

jkbbwr:
The biggest problem I see is there is no way to fall back if uvloop fails to install. If that was the case then I would just let sanic try to install it fail, roll back and install without uvloop

renstrom:
>The biggest problem I see is there is no way to fall back if uvloop fails to install. If that was the case then I would just let sanic try to install it fail, roll back and install without uvloop

Actually, I think there might be. I'm not exactly sure what exceptions to catch, but something like this could possibly work.

```python
install_requires = [...]

def run_setup():
    setup(...)

try:
    run_setup()
except Exception:  # Probably catch something more explicit
    install_requires.remove('uvloop')
    run_setup()
```

This is for example [simplejson](https://github.com/simplejson/simplejson/blob/master/setup.py) conditionally builds its C extension. However since this is a dependency, perhaps it's all handled outside the scope of `setup.py`?

jkbbwr:
It would be nice for @channelcat to weigh in on this once they are around.

jkbbwr:
Its been approved and it has been 5 days. Is there any chance of this going ahead?

r0fls:
I think I have to agree with @subyraman here. Although the project works without these, the framework was originally inspired by uvloop. We're not really targeting Windows as a supported platform. I'm ok with this change, and the code is fine, which is why I approved it. However in general I don't feel great about putting Windows users first, in the sense that we're disrupting current usage patterns to fit their needs. Have you tried doing something like what was suggested here?  https://github.com/channelcat/sanic/pull/507#issuecomment-283900669

r0fls:
@jkbbwr assuming you're on windows, care to test #543? You can run:

```bash
git checkout -b r0fls-windows-setup master
git pull https://github.com/r0fls/sanic.git windows-setup
python3 setup.py install
```

jkbbwr:
@r0fls I am not actually on windows. In my situation I was using a python interpreter that cannot support C extensions.

r0fls:
@jkbbwr what do you mean here? Which import is missing?

>The script in that branch is missing an import which causes the try catch to fail.

Regarding this:

>Adding that import GCC triggers a system exit when it can't properly compile the extension which then doesn't re-execute the setup.

So no exception is raised, it just exits? What system are you using exactly? Can you create a repo or gist with the configuration for a docker container or VM where this is replicable?

Hellowlol:
Its is possible to install sanic on windows. Im using it. I just had to install some c++ compiler, pip even pointed to the website where i could download it. 

A warning about missing speedups would be nice.

jkbbwr:
Perhaps this solution will make everyone happy.

```

**Comment by r0fls** [comment]
@jkbbwr that looks good to me, I say we leave the try/except solution that @renstrom proposed though as well, since it could help with other platforms.

**Comment by jkbbwr** [comment]
I left the try/except in but I don't think it does what we think it does.

```
         if (PyCoro_CheckExact(obj)) {
             ^~~~~~~~~~~~~~~~~
    error: command 'gcc' failed with exit status 1
    
    ----------------------------------------
Command "/home/jakob/.virtualenvs/pypy3.5-tmp/bin/pypy3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-tofyz_4m/uvloop/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-z9uj26r8-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/jakob/.virtualenvs/pypy3.5-tmp/include/site/python3.5/uvloop" failed with error code 1 in /tmp/pip-build-tofyz_4m/uvloop/
```

r0fls:
@jkbbwr is that the case on windows as well? Maybe we just need to catch more exceptions? https://github.com/simplejson/simplejson/blob/master/setup.py#L41-L48

jkbbwr:
@r0fls I don't have windows and can't test. But Catching raw `except:` doesn't help either.

r0fls:
@jkbbwr ok would you be willing to close this and make a new PR with your suggested idea of using the environment variable? I'm all for that.

funkyfuture:
@jkbbwr I think it would be better to provide the options as cli arguments to `setup.py`.

The try/except approach would also leave some packages installed in case of later errors.

jkbbwr:
Re opening as two seperate pull requests

PR #512, opened 2017-03-03
Fix `request.url` and other url properties

### LINKED ISSUE #508: url attributes in request
Assume the following:

```python
from sanic import Sanic

app = Sanic()

@app.route("/foo/bar")
async def test(request):
    print(request.url)

if __name__ == "__main__":
```

`request.url` gives `/foo/bar`, but what I'm actually looking for is something like `0.0.0.0:8000/foo/bar`.
For reference see how it's done in Flask: http://flask.pocoo.org/docs/0.12/api/#flask.Request.path

Is this intended?

subyraman:
`request.url` was just returning the path. Fix #508 and add other Flask-like properties for parsing the URL, like `request.path`, `request.host`, `request.query_string`, `request.scheme`

By lazyloading the URL parsing it appears there is an associated performance boost.


sanic/request.py
```diff
@@ -142,6 +138,40 @@ def ip(self):
...
+        if self.transport.get_extra_info('sslcontext'):
+            return 'https'
+
+        return 'http'
+
+    @property
+    def host(self):
+        # it appears that httptools doesn't return the host
+        # so pull it from the headers
+        return self.headers['Host']
```
r0fls: Can we replace this with `self.get('Host', '')` just to be safe?

subyraman: Done.

PR #515, opened 2017-03-04
Use cgitb instead of DIY traceback

youknowone:
This is a proof-of-concept PR and not done yet.

Will you look into this is useful?
Sanic can remove its own traceback printing code and it is a standard library.
(well, the output html is not very attractive but it is just a debugging tool)


seemethere:
Going to close this for now. Does not seem as though it is active.

PR #527, opened 2017-03-08
update asyncpg example

r0fls:


nszceta:
Why was the connection pool code removed?

r0fls:
@nszceta because it doesn't make sense to open a connection pool for a single connection.

r0fls:
If you can keep the same connection pool open and use that (which is possible by attaching it to the `request.app` attribute), then it would make sense to use it.

nszceta:
@r0fls what are your thoughts regarding https://github.com/channelcat/sanic/issues/540 ?

PR #543, opened 2017-03-12
windows setup

r0fls:
Implements idea from https://github.com/channelcat/sanic/pull/507. Also includes a 🚀

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,23 +16,23 @@ with codecs.open(os.path.join(os.path.abspath(os.path.dirname(
         raise RuntimeError('Unable to determine version.')
 
 setup_kwargs =  {
-        'name': 'sanic',
-        'version': version,
-        'url': 'http://github.com/channelcat/sanic/',
-        'license': 'MIT',
-        'author': 'Channel Cat',
-        'author_email': 'channelcat@gmail.com',
-        'description': (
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -15,33 +15,47 @@
...
         raise RuntimeError('Unable to determine version.')
 
-install_requires = [
-    'httptools>=0.0.9',
-    'ujson>=1.35',
-    'aiofiles>=0.3.0',
-    'websockets>=3.2',
-]
+setup_kwargs =  {
+        'name': 'sanic',
```
messense: The indent should be four spaces.

messense: Please consider setup [Hound CI](https://houndci.com/) for automated code style review.

seemethere: The setup.py isn't covered by the flake8 check that we run.

messense: @seemethere It would be much nicer to let Hound CI submit reviews to pull requests than looking into Travis CI logs.

seemethere: I can understand the reasoning but I don't want to add another dependency in our development pipeline. 

I'll look into Hound CI but chances are it'll stay how it is. Also we could look into something like PEP8 Speaks or something else.

messense: Integrate with PEP8 Speaks would be great too.

PR #544, opened 2017-03-12
Add response documentation

messense:
Feel free to criticize my poor English.

### DIFF
--- a/docs/sanic/response.md
+++ b/docs/sanic/response.md
@@ -1,0 +1,99 @@
+# Response
+
+Use functions in `sanic.response` module to create responses.
+
+- `text` - Plain text response
+
+```python
+from sanic import response
+
+
+@app.route('/text')
+def handle_request(request):
+    return response.text('Hello world!')
+```
+
+- `html` - HTML response
+
+```python
+from sanic import response
+
+
+@app.route('/html')
+def handle_request(request):
+    return response.html('<p>Hello world!</p>')
+```
+
+- `json` - JSON response
+
+
+```python
+from sanic import response
+
+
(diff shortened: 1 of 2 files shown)


docs/sanic/response.md
```diff
@@ -0,0 +1,99 @@
+# Response
+
+You can use functions in `sanic.response` module to create response.
```
r0fls: Sorry to be picky, but how about "Use functions in `sanic.response` module to create responses." instead?

r0fls: Mostly it's just the response vs responses: "You can use functions in `sanic.response` module to create responses" would be great too.

r0fls: I didn't see the original comment, this would be my only complaint.

docs/sanic/response.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `file` - File response
+
+```python
+from sanic import response
+
+
+@app.route('/file')
+def handle_request(request):
+    return response.file('/srv/www/whatever.png')
```
messense: Should this be:

```python
return await response.file('/srv/www/whatever.png')
```
?

cc @r0fls @seemethere

r0fls: @messense good point, it should actually be:

```python
@app.route('/file')
async def handle_request(request):
    return await response.file('/srv/www/whatever.png')
```

Note the `async` as well.

r0fls:
@messense feel like making another PR to make the different response types hyperlinks? It's helpful  for referencing them to other people...

messense:
@r0fls I'll do that later.

Issue #501, opened 2017-03-01
Is there anyway to measure sanic performance?


Gowtham95india:
Hi, 

Is there a way to measure sanic performance with Newrelic or Any other way to do? As per discussion with newrelic, it won't support sanic as of now. Is there any other way to measure it?

r0fls:
Yes it looks like new relic only works with wsgi as of now. Have you looked at data dog? http://docs.datadoghq.com/integrations/python/

Gowtham95india:
No, I checked with netuitive, appinternals riverbed. But that doesn't seems working.

Sniedes722:
@Gowtham95india Any of those could have more readily Gunicorn support, in which case you could consider running it as a [GunicornUVLoopWorker](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/worker.py#L249) as referenced in #61 . I'm not sure if it has a considerable effect on performance, however as @r0fls said, data dog looks worth it to integrate for your needs.

Gowtham95india:
@Sniedes722 

Yeah.., I want to avoid Gunicorn, just to check pure performance of Sanic. We want to some high load and see the response time. So, I think it is better to test with pure sanic web workers. 

I'm trying to implement datadoghq as metioned by r0fls.

r0fls:
@Gowtham95india any luck? :) I'm curious...

Gowtham95india:
@r0fls as of now, no luck. I tested with almost 5-6 monitoring solutions. But, as per my discussion with datadoghq., they are going to release new version of python agent very soon, in which they are adding support of asynchronous frameworks. 

Newrelic stated the same, but it is like, they didn't even start working on that. 

So, waiting for datadoghq update for python agent. The only problem with all the monitoring solution providers is that Sanic is ASYNC framework. 

I'll update once I get the response from Datadoghq.

r0fls:
Good to know they're working on it anyway...

Gowtham95india:
yea, at last. Hope, will get at least one way to measure it.

abuckenheimer:
Hey folks I actually was working on creating a more formal version of this but a MVP using [prometheus](https://prometheus.io/) looks like

    import prometheus_client as pc
    from time import perf_counter
    rec_counter = pc.Counter('http_requests_total',
        'Total number of HTTP requests made.', ['method', 'url', 'status'])
    timer = pc.Histogram('http_request_duration_seconds',
[...]

Issue #107, opened 2016-10-24
Handling redployments


tomchristie:
Hiya,

One thing that's not clear to me from the documentation is how users should be expected to handle app redeployments (without dropping existing connections). Presumably you'd need to bring up the new servers on a different port, switch the load balancer, and then (gracefully) stop the existing server.
- Does sanic handle graceful redeployment without dropping connections?
- Have I misunderstood how app redeployment should be handled with sanic?
- Would this be worth documenting in the deployment section.

(Also related to https://github.com/channelcat/sanic/issues/61 since graceful reloads are one of the benefits of running an application framework underneath a server such as gunicorn.)

channelcat:
Hmm, Sanic is set up to properly drain connections on close, but does not support reloading, so you would have to do what you outlined.  Looking at Gunicorn, it looks like it creates an arbiter before loading any application code.  To create workers, the arbiter forks the process first, then initializes the workers to allow graceful reloading.  Pretty neat.

We should be able to support that paradigm by shifting to running apps with `python -m sanic file.app` (currently supported) when using multiple workers.  I think I'd like to add this, but I'm not sure when I will be able to get to it, so I'm tagging in case someone else wants to pick it up first.

Issue #61, opened 2016-10-18
Gunicorn worker


santagada:
Is there any way to mix sanic with gunicorn? Gunicorn has signals to increase/decrease workers and a bunch of other features that are useful in running a service in production.

channelcat:
I believe it would be possible to create a worker similar to how aiohttp does.  I'm currently planning to deploy via docker, so putting it behind gunicorn isn't something I'm looking to do soon.  I tagged this issue in case people want to pick it up.

tomchristie:
Seconded. I'd be very interested to see the core of this built into a gunicorn worker, that'd allow a whole ecosystem of frameworks to make use of it, rather than tying the server implementation to a single app framework implementation.

Currently the latency for a basic 'return json response' aiohttp/gunicorn deployment is way too high, this looks far more promising.

channelcat:
I'm hesitant to put a lot of work into making a Gunicorn worker, because I'm worried the overhead of layering Gunicorn on top of sanic will significantly decreases performance and wouldn't end up being our recommended way of deploying (based on the project's goals).  But, I also don't want to spend forever implementing Gunicorn's feature set.  

What are the features of Gunicorn you find useful?  Also, what do you mean by allowing an ecosystem of frameworks?  I'd like to know what possibilities this could open up :).  Thanks!

tomchristie:
> What are the features of Gunicorn you find useful?

1. Graceful redeployments.
2. Process monitoring/timeouts/restarts.

> what do you mean by allowing an ecosystem of frameworks?

If there's a minimal server that simply exposes a request/response API, then many different frameworks can be built on top of that.

If the uvloop server that sanic comes with really is faster than other python http server alternatives then it'd be great to see that implemented in the form of an agnostic python web server that any framework author can then build on top of.

santagada:
ok so there is two different things here, uvloop+httptools worker for gunicorn (maybe with some parts of aiohttp that implements wsgi) that would make all frameworks be able to use most of the stuff that sanic uses to get fast.

the second would be to actually have sanic running on top of gunicorn. @channelcat for the second I don't see how it would be a performance penalty as gunicorn doesn't really do anything related to receiving connections by itself, it is just a manager for a bunch of workers and it keeps the socket open during a reload which seems exactly what you want.

I think maybe sanic future is running outside of gunicorn and I wouldn't put that as a high priority, and also idea 1 could be done in gunicorn without sanic at all.

GMLudo:
[...]

PR #545, opened 2017-03-12
Gunicorn worker

messense:
~This is really just a WIP~, it's working on the `simple_server` and `try_everything` example.

```bash
gunicorn simple_server:app --bind localhost:8000 --worker-class sanic.worker.GunicornWorker
```

Ping #61 #107 #501

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -578,6 +578,10 @@ class Sanic:
         """This kills the Sanic"""
         get_event_loop().stop()
 
+    def __call__(self):
+        """gunicorn compatibility"""
+        return self
+
     async def create_server(self, host="127.0.0.1", port=8000, debug=False,
                             before_start=None, after_start=None,
                             before_stop=None, after_stop=None, ssl=None,
@@ -686,9 +690,10 @@ class Sanic:
             server_settings['run_async'] = True
 
         # Serve
-        proto = "http"
-        if ssl is not None:
-            proto = "https"
-        log.info('Goin\' Fast @ {}://{}:{}'.format(proto, host, port))
+        if host and port:
+            proto = "http"
+            if ssl is not None:
+                proto = "https"
+            log.info('Goin\' Fast @ {}://{}:{}'.format(proto, host, port))
 
         return server_settings
--- a/examples/simple_server.py
+++ b/examples/simple_server.py
@@ -9,4 +9,5 @@ async def test(request):
     return json({"test": True})
 
 
-app.run(host="0.0.0.0", port=8000)
+if __name__ == '__main__':
+    app.run(host="0.0.0.0", port=8000)
--- a/examples/try_everything.py
+++ b/examples/try_everything.py
@@ -70,6 +70,11 @@ def query_string(request):
 # Run Server
 # ----------------------------------------------- #
 
+@app.listener('before_server_start')
+def before_start(app, loop):
+    log.info("SERVER STARTING")
+
+
 @app.listener('after_server_start')
 def after_start(app, loop):
     log.info("OH OH OH OH OHHHHHHHH")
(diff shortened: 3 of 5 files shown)


messense:
Performance (MacBook Pro (Retina, 15-inch, Mid 2015)):

`simple_server` `app.run` with 4 workers:

```
wrk -c 4 -t 1 -d 30s http://localhost:8000/
Running 30s test @ http://localhost:8000/
  1 threads and 4 connections
    Latency   124.63us  121.46us   6.21ms   99.11%
    Req/Sec    31.78k     5.25k   38.80k    52.82%
  951568 requests in 30.10s, 112.53MB read
Requests/sec:  31612.82
Transfer/sec:      3.74MB
```

`simple_server` Gunicorn with 4 workers:

```
wrk -c 4 -t 1 -d 30s http://localhost:8000/
Running 30s test @ http://localhost:8000/
  1 threads and 4 connections
    Latency   120.28us   93.70us   5.86ms   99.12%
    Req/Sec    32.42k     4.65k   37.49k    68.00%
  967393 requests in 30.00s, 114.40MB read
Requests/sec:  32246.56
Transfer/sec:      3.81MB
```

messense:
The `after_server_start` and other events not working for now.

r0fls:
That's impressive that they're going the same speed. We saw a significant slowdown when @seemethere implemented this. What benchmarks do you see with a single worker?

messense:
@r0fls 

`simple_server` `app.run` with 1 worker:

```
wrk -c 4 -t 1 -d 30s http://localhost:8000/
Running 30s test @ http://localhost:8000/
  1 threads and 4 connections
    Latency   165.06us   45.18us   3.79ms   96.53%
    Req/Sec    23.71k     1.06k   24.76k    90.03%
  710069 requests in 30.10s, 83.97MB read
Requests/sec:  23590.63
Transfer/sec:      2.79MB
```

`simple_server` Gunicorn with 1 worker:

```
wrk -c 4 -t 1 -d 30s http://localhost:8000/
Running 30s test @ http://localhost:8000/
  1 threads and 4 connections
    Latency   185.10us  117.67us   6.28ms   96.74%
    Req/Sec    21.74k     2.71k   23.74k    86.71%
  651122 requests in 30.10s, 77.00MB read
Requests/sec:  21629.70
Transfer/sec:      2.56MB
```

sanic/app.py
```diff
@@ -658,9 +662,10 @@ def _helper(self, host="127.0.0.1", port=8000, debug=False,
             server_settings['run_async'] = True
 
         # Serve
-        proto = "http"
-        if ssl is not None:
-            proto = "https"
-        log.info('Goin\' Fast @ {}://{}:{}'.format(proto, host, port))
+        if host and port:
```
r0fls: This will always be true because they have default values `"127.0.0.1"` and `8000`

messense: In `GunicornWorker` they are set to `None` while pass `sock`.

messense: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.create_server

> sock can optionally be specified in order to use a preexisting socket object. If specified, host and port should be omitted (must be None).

messense:
Mostly working now, websocket works but seems not closing on signal int while client still connected.

messense:
Actually, it won't shutdown in `app.run` mode with client still connected to websocket endpoint too.

miguelgrinberg:
@messense yes, sanic simply waits for all connections to end, assuming they are short lived, but websocket will go on until the client goes away.

I'll think about how the websocket tasks can be signaled to exit.

messense:
I have extracted these changes into a package named [sanic-gunicorn](https://github.com/messense/sanic-gunicorn), feel free to try it.

If you guys don't want `Gunicorn` support in `Sanic` core, I'll close this PR and maintain the `sanic-gunicorn` package instead.

r0fls:
That sounds good. I'm fine with this change but I'm going to leave it up to @seemethere to make the ultimate decision.

messense:
@r0fls Either way, in order to support Gunicorn, Sanic app instance need to be a callable:

```python
class Sanic:
    def __call__(self):
        return self
```

seemethere:
Documentation to go along with this would be amazing

PR #549, opened 2017-03-13
add websocket scheme in request

ai0:


sanic/request.py
```diff
@@ -142,6 +142,10 @@ def ip(self):
 
     @property
     def scheme(self):
+        if self.app.websocket_enabled \
+                and self.headers.get('upgrade') == 'websocket':
+            return 'ws'
```
messense: With `sslcontext` it should be `wss` I think?

go2sh: :+1:

PR #558, opened 2017-03-15
Add a new example by using aiomysql

Zheaoli:
Here's the new example by using aiomysql with a global connection pool.

### DIFF
--- a/examples/sanic_aiomysql_with_global_pool.py
+++ b/examples/sanic_aiomysql_with_global_pool.py
@@ -15,23 +15,23 @@ database_host = os.environ['DATABASE_HOST']
 database_user = os.environ['DATABASE_USER']
 database_password = os.environ['DATABASE_PASSWORD']
 app = Sanic()
-asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 
 
-async def get_pool(*args, **kwargs):
+@app.listener("before_server_start")
+async def get_pool(app, loop):
     """
-    the first param in *args is the global instance ,
+    the first param  is the global instance ,
     so we can store our connection pool in it .
     and it can be used by different request
     :param args:
     :param kwargs:
     :return:
     """
-    args[0].pool = {
+    app.pool = {
         "aiomysql": await aiomysql.create_pool(host=database_host, user=database_user, password=database_password,
                                                db=database_name,
                                                maxsize=5)}
-    async with args[0].pool['aiomysql'].acquire() as conn:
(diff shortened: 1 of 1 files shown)


examples/sanic_aiomysql_with_global_pool.py
```diff
@@ -0,0 +1,63 @@
...
+import uvloop
+from sanic import Sanic
+from sanic.response import json
+
+database_name = os.environ['DATABASE_NAME']
+database_host = os.environ['DATABASE_HOST']
+database_user = os.environ['DATABASE_USER']
+database_password = os.environ['DATABASE_PASSWORD']
+app = Sanic()
+asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
```
r0fls: This doesn't need to be set. It will be set this way here: https://github.com/channelcat/sanic/blob/master/sanic/server.py#L343-L345

r0fls: So you can simply remove line 18, it will have the same effect.

examples/sanic_aiomysql_with_global_pool.py
```diff
@@ -0,0 +1,63 @@
...
+
+
```
r0fls: use `@app.listener('before_server_start')` here instead of passing it to the `run` function. That is no longer supported.

r0fls: So decorate `get_pool` with `app.listener`

examples/sanic_aiomysql_with_global_pool.py
```diff
@@ -0,0 +1,63 @@
...
+async def get_pool(*args, **kwargs):
```
r0fls: this can just take the loop and app (less obfuscated):  `async def get_pool(loop, app):`. Then use `app` instead of `args[0]` on line 34.

examples/sanic_aiomysql_with_global_pool.py
Zheaoli: You mean the Sanic will not support `befor_start` param in `run` method in the future？

r0fls: @Zheaoli yes, it's now implemented as a decorator as of https://github.com/channelcat/sanic/pull/408. What version do you have installed? I'll show you an example if you like. It doesn't seem to exist in the docs...

r0fls: We must be a bit behind in `pip`. Let me look into that...

Issue #557, opened 2017-03-15
allow configurable strict_slashes in router


phongphan:
Looks like sanic automatically add both with and without trailing slash routes to every route declaration since 0.4.0 207ec1e0322776cd795739a8ce1acf390e8b6888.

I think it would be configurable, maybe by adding `strict_slashes` parameter to `Route` constructor. And let the user to pass it manually to the `app`.

Or maybe also allow adding to any individual route just like Flask does.

r0fls:
I could see adding this knob, but I'm curious _why_? When would you want `http://route.com/get` but not `http://route.com/get/`, or vice-versa?

phongphan:
They could have difference meaning sometimes.
Eg. distinction between file and directory while serving file/directory resource metadata.

r0fls:
Closing with: https://github.com/channelcat/sanic/pull/559

PR #559, opened 2017-03-16
accept strict_slash routes

r0fls:
Addresses #557

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -101,7 +101,8 @@ class Sanic:
         return decorator
 
     # Decorator
-    def route(self, uri, methods=frozenset({'GET'}), host=None):
+    def route(self, uri, methods=frozenset({'GET'}), host=None,
+              strict_slashes=False):
         """Decorate a function to be registered as a route
 
         :param uri: path of the URL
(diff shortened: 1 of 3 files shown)


r0fls:
This allows passing the `strict_slash` parameter to the route decorators to prevent it from adding both the version with and without a trailing slash.

sanic/router.py
```diff
@@ -95,7 +95,13 @@ def parse_parameter_string(self, parameter_string):
 
         return name, _type, pattern
 
-    def add(self, uri, methods, handler, host=None):
+    def add(self, uri, methods, handler, host=None, strict_slashes=False):
+        # just add what was given
+        if strict_slashes:
```
seemethere: Here why not just return after adding the regular version if `strict_slashes`?

sanic/app.py
```diff
@@ -101,7 +101,8 @@ def decorator(listener):
         return decorator
 
     # Decorator
-    def route(self, uri, methods=frozenset({'GET'}), host=None):
+    def route(self, uri, methods=frozenset({'GET'}), host=None,
```
seemethere: Just a thought but later on in the future if we want to add more arguments to these (and in particular the shorthand versions) we may just want to start using `*args` and `**kwargs` to bubble them up to the `add_route` function without having to change them in 10 different places

PR #560, opened 2017-03-16
cancel websocket tasks if server is stopped

miguelgrinberg:
I think this addresses the issue raised in #545 regarding stopping websocket tasks when the server is stopped.

Any non-websocket task that is takes a long time to run will also prevent the server from exiting, but for HTTP routes I guess that is not very common.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1,7 +1,7 @@
 import logging
 import re
 import warnings
-from asyncio import get_event_loop
+from asyncio import get_event_loop, ensure_future, CancelledError
 from collections import deque, defaultdict
 from functools import partial
 from inspect import isawaitable, stack, getmodulename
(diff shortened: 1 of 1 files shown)


sanic/app.py
```diff
@@ -191,9 +192,14 @@ def response(handler):
...
                 try:
-                    # invoke the application handler
-                    await handler(request, ws, *args, **kwargs)
-                except ConnectionClosed:
+                    # schedule the application handler
+                    # its future is kept in self.websocket_tasks in case it
+                    # needs to be cancelled due to the server being stopped
+                    fut = ensure_future(handler(request, ws, *args, **kwargs))
+                    self.websocket_tasks.append(fut)
+                    await fut
```
messense: if `ConnectionClosed` raised here, the future will never be removed from `self.websocket_tasks`, is this ok?

miguelgrinberg: Right, missed that.

messense:
Confirmed that this patch works fine.

Issue #566, opened 2017-03-18
Nginx+docker-compose example


itielshwartz:
Hi,

I'm in the middle of a Sanic based project.
I wanted to use docker and Nginx but couldn't find a simple **HOW-TO**,
So I ended up creating a simple skeleton app for using sanic behind Nginx (using docker-compose).

Would love getting some feedback (as I'm new to Sanic), And also wanted to ask if it's OK adding some link for it in the docs - as I think nginx+Sanic would be a common combination.

itielshwartz:
Oops forgot to give a link - [nginx+docker-compose example](https://github.com/itielshwartz/sanic-nginx-docker-example) :)

r0fls:
Awesome! However, this might make more sense as part of our examples (in a directory within examples) then as a separate repo. There isn't really a place for this in the docs right now, since it's not really an extension.

r0fls:
Or, if you prefer to add it as an extension that's fine too...

itielshwartz:
Hi, @r0fls sorry for the late response, thanks for the feedback.
I would prefer to put it as an extension (at least for now), as I plan to rapidly improve the repo and add some more basic docs on starting a sanic project, would be easier as a separate project.

seemethere:
This is awesome 😄 

@r0fls maybe it would be good to include something like [these lines](https://github.com/itielshwartz/sanic-nginx-docker-example/blob/master/app/server.py#L20) in the base configuration that automatically check if there's an environment variable for thing and override if they exist.

PR #573, opened 2017-03-22
allow setting config from individual env variables

r0fls:
Idea from here: https://github.com/channelcat/sanic/issues/566


r0fls:
uh looks like the tests are failing...

PR #577, opened 2017-03-22
Hotfixes tests failing from URL object change

seemethere:
aiohttp decided to use yarl for their new URL objects so that they
aren't plain strings anymore which means that this single test fails.
Not a huge change but this should fix the testing suite.

### DIFF
--- a/tests/test_redirect.py
+++ b/tests/test_redirect.py
@@ -88,4 +88,4 @@ def test_chained_redirect(redirect_app):
     assert request.url.endswith('/1')
     assert response.status == 200
     assert response.text == 'OK'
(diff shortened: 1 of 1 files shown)


seemethere:
Seems like things are still failing will dive deeper into this later tonight.

r0fls:
@seemethere this does fix one test, I vote we merge this anyway.

r0fls:
Unless you want to wait until we fix the other 3

r0fls:
either way

seemethere:
I'm fine with merging this in and then focusing on the next set of failures.

Issue #582, opened 2017-03-23
Testing without response.text()


qwIvan:
The testing will paser response body as text, but in my project, the response body is a large video. So the operation will raise `UnicodeDecodeError`.
https://github.com/channelcat/sanic/blob/master/sanic/testing.py#L25

qwIvan:
Is there any idea to resolve the problem?

r0fls:
@qwIvan this should fix it: https://github.com/channelcat/sanic/pull/580. Feel free to reopen this if you're still having issues.

PR #583, opened 2017-03-23
I don't think we need to load response body in testing

qwIvan:
If you need this, you can call response.text() or response.body() manually.
Is there any idea to resolve the issue #582? 
Just commented the loading operation? 
Or make response.text and response.body as `@property` function to compatible the old versions?


youknowone:
Using `await` is a painful work in various environments. How about adding mime type check or a try-except capture?

qwIvan:
Or using arg `load_body`?
https://github.com/channelcat/sanic/pull/583/commits/1562b81522d827539545a28e592aed1966ce0099
@youknowone

qwIvan:
It fixed by #580

PR #592, opened 2017-03-24
Document synchronous response.write in streaming

weargoggles:
The Streaming section of the docs [was updated](https://github.com/subyraman/sanic/blob/d8a6d7e02f7628ab4b910e31a6860147c3a2134f/docs/sanic/streaming.md) to make clear that a synchronous write should be used in the callback, but this section was not updated.

### DIFF
--- a/docs/sanic/response.md
+++ b/docs/sanic/response.md
@@ -55,8 +55,8 @@ from sanic import response
 @app.route("/streaming")
 async def index(request):
     async def streaming_fn(response):
-        await response.write('foo')
-        await response.write('bar')
+        response.write('foo')
+        response.write('bar')
     return response.stream(streaming_fn, content_type='text/plain')
 ```
 


r0fls:
@subyraman is this correct?

subyraman:
Yes, it is strictly true that right now the `write` method is synchronous (it just basically wraps `transport.write`).

But I've been thinking that it should be changed to be asynchronous after reading[ this article](https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/), if we don't `drain` the transport after each write it could end up causing problems with Sanic. Give me a day to think about it :)

weargoggles:
The intention of this PR is only to synchronise examples on two pages of the documentation.

subyraman:
I know, but I think it'd be best to actually make the `response.write` method asynchronous so each write can be drained from the buffer :) I'm coming up with a PR for that.

subyraman:
@r0fls you can go ahead and approve this. Dealing with backpressure would require a larger refactor since Sanic uses the raw `Protocol` class instead of a stream class.

Issue #594, opened 2017-03-26
Workers don't stop on sigint


r0fls:
To reproduce this use:

**worker_test.py**
```python
from sanic import Sanic
from sanic.response import json

app = Sanic(__name__)


@app.route("/")
async def test(request):
    return json({"test": True})


if __name__ == '__main__':
    app.run(host="0.0.0.0", port=8000, workers=4)
```

Then run `kill -INT <pid of server>`.

Finally you can see they haven't all been killed:

```
ps aux | grep worker_test.py
```

nszceta:
I do not recommend to integrate multiprocessing within Sanic. I know some people want it. It is much safer and potentially more performant to use a process supervisor such as supervisord to run multiple Sanic workers.

Supervisord Example: to start 4 processes, on ports 3000, 3001, 3002, 3003:
```
[program:tile_server]
process_name=tile_server_%(process_num)s
command=/home/ubuntu/.pyenv/versions/3.6.0/bin/python3.6 -OO /home/ubuntu/tiler.py %(process_num)s
user=ubuntu
directory=/home/ubuntu
numprocs=2
numprocs_start=3000
stdout_logfile=/home/ubuntu/tile_server.stdout.log
stderr_logfile=/home/ubuntu/tile_server.stderr.log
```

Nginx:
```
upstream my_server {
  server 127.0.0.1:3000;
  server 127.0.0.1:3001;
  server 127.0.0.1:3002;
  server 127.0.0.1:3003;
}

server {

  # fill me in

  location / {
    proxy_pass                  http://my_server;
    proxy_set_header            Host $host;
  }

}
```

r0fls:
@nszceta I don't know that I would go that far, I'm running it with multiple workers on a project and it hasn't failed or anything. Since you say it's more performant, are you willing to provide benchmark comparison? `wrk -t 10 -c 100 -d 30 http://localhost:8000` would work for that.

r0fls:
Anyway if we have an option there to run multiple workers (which some people are using) it makes sense to have have it respond to sigint properly. This isn't a statement about sanic multiple workers being a better method for production than supervisord.

nszceta:
What happens when one of your processes hangs for whatever reason? Does your app keep working normally?

r0fls:
I haven't experienced that yet. Yes it's just kept working so far...

atungw:
FYI, I don't believe workers are working properly in Sanic (even on the master branch). If you run the following code...

```
[...]

PR #595, opened 2017-03-26
fix multiple worker issue

r0fls:
addresses #594


messense:
How is this gonna affect inflight connections?

seemethere:
What exactly does this do?

r0fls:
@seemethere if you look at the issue #594, as reported by a user in gitter, our implementation doesn't stop correctly on sigint. This fixes that.

r0fls:
@messense this will probably not handle that correctly. However right now it's not even stopping, so still maybe a good idea...

messense:
I don't think people would tend to use it in production if it does not handle graceful shutdown well. Maybe we should consider remove it?

r0fls:
Going to have to rethink this. I don't think we'll be removing the multiple workers implementation entirely but it is starting to show serious weaknesses.

Issue #72, opened 2016-10-19
Document logging


santagada:
Sanic doesn't have a logger like flask, so how to use logging in it

wsbmilk:
+10086

r0fls:
Currently you would just use logging like you would normally in a python app. There is an example here: https://github.com/channelcat/sanic/blob/master/examples/override_logging.py


What kind of documentation would be helpful here?

santagada:
In the future would be nice to have a full manual like flask with a section on logging. But for now just pointing on the readme after the hello world example to a "comprehensive example" with logging and other things a real service need (db access, cookies, post/get requests, maybe ssl, maybe cron tasks)

seemethere:
@santagada I do agree that a full example is really needed that is on the TO-DO list!

SakuraSound:
Does #601 satisfy this issue?

seemethere:
Closing with #601

PR #601, opened 2017-03-27
Detailed example with logging, database access, environment variables, and basic middleware

SakuraSound:
I'm currently learning sanic (am a long-time aiohttp user). I was reading the beginner Issues and ran across #72. So I created this example of accessing (GET, POST) a passworded redis cache using aioredis, and added document logging (where log files are placed in location specified by APP_LOGFILE).

Created detailed example of using sanic. Adds configurations based on various environment variables, handles database access (using aioredis), uses middleware to check for db object and attach it to request object, and logs events to a logfile (which is set using environment variables).

I hope this satisfies #72. If not, I can go back and make more changes.

### DIFF
--- a/examples/detailed_example.py
+++ b/examples/detailed_example.py
@@ -1,6 +1,9 @@
+# This demo requires aioredis and environmental variables established in ENV_VARS
 import json
 import logging
 import os
+
+from datetime import datetime
 
 import aioredis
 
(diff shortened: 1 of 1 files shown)


examples/detailed_example.py
```diff
@@ -0,0 +1,96 @@
...
+app = Sanic(name=__name__)
+
+logger = None
+
+
+@app.middleware("request")
+async def attach_db_connectors(request):
+    if not hasattr(request.app.config, "REDIS"):
+        logger.info("Setting up connection to Redis Cache")
+        request.app.config.REDIS = await aioredis.create_pool((app.config.REDIS_HOST, int(app.config.REDIS_PORT)),
```
go2sh: Hm, is it really necessary to create a pool of connections for each request? On a request basis, you normally use only one. A connection pool in a global scope makes more sense.

SakuraSound: Ah... thanks for that. Was only meaning for that to be made once.
I can fix this.

SakuraSound: After going back to the code, reading the [docs](https://sanic.readthedocs.io/en/latest/sanic/request_data.html) and running some tests, the connection pool is only being made once. Technically, request.app is a reference to the global app that is created.  It is then stored in app.config.REDIS to be reused (pull a connection from the pool) for future requests. 

This could probably be documented a bit better in my example to showcase that. Originally I was going to make this example multiple files. Not sure if that is the best way, but I can change it to be that if it would make this point clearer.

go2sh: Ah sorry. I missed that. A better way would be to use the `@app.listener('before_server_start')` event. See `sanic/examples/sanic_aioredis_example.py`.

SakuraSound: Got it. I'll use that instead.

SakuraSound:
So Ive made some updates to the example based on suggestions by @go2sh

PR #604, opened 2017-03-29
Fixing the unittests

seemethere:
Recently due to some decisions made in aiohttp's recent `0.2.0` release some features in sanic's testing suite have broken down. For now, until we fix what's broken or find an alternative, sanic's testing infrastructure will require aiohttp versions less than `0.2.0`.

This PR also adds support for running our unittests in a Docker container so that we can have consistent environments in terms of running the unittests (Had some differences between my local OSX box and the Travis Jobs so this a nice happy medium).

To actually run them it's now:
`make test`

Other than that this PR also includes some cleaning of requirements files, re-organizing some tests, as well as fixing some random PEP8 errors.

Hopefully this will start turning the builds green... 🤞


Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.6
```
r0fls: What about 3.5? Does this method still test that?

seemethere: Oh no not for now. I guess it could be added in it but installing python3.5 could be a PITA in terms of waiting for it to install.

If anything this is just a nice to have.

jkbbwr:
Can we get these merged and a release made to pypi. This is starting to drag on and on.

r0fls:
@seemethere I agree we should do a pypi release.

PR #619, opened 2017-04-08
Add few ways to load config

kotyara1005:


sanic/config.py
```diff
@@ -97,12 +102,53 @@ def from_object(self, obj):
             if key.isupper():
                 self[key] = getattr(obj, key)
 
+    def from_json(self, filename, silent=False):
+        """Updates the values in the config from a JSON file. This function
+        behaves as if the JSON object was a dictionary and passed to the
```
r0fls: were you still finishing this docstring?

r0fls: Add a newline after `"""` as we do that throughout the rest of the code.

sanic/config.py
```diff
@@ -97,12 +102,53 @@ def from_object(self, obj):
...
+                obj = json.loads(json_file.read())
+        except IOError as e:
+            if silent and e.errno in (errno.ENOENT, errno.EISDIR):
+                return False
+            e.strerror = 'Unable to load configuration file (%s)' % e.strerror
+            raise
+        return self.from_mapping(obj)
+
+    def from_mapping(self, *mapping, **kwargs):
+        """Updates the config like :meth:`update` ignoring items with non-upper
```
r0fls: Add a newline after """

sanic/config.py
```diff
@@ -97,12 +102,53 @@ def from_object(self, obj):
...
+            )
+        mappings.append(kwargs.items())
+        for mapping in mappings:
+            for (key, value) in mapping:
+                if key.isupper():
+                    self[key] = value
+        return True
+
     def load_environment_vars(self):
+        """Looks for any SANIC_ prefixed environment variables and applies
```
r0fls: Leave this as is, put back the newline after """

r0fls:
Can you add docs demonstrating the two new ways to add config?

r0fls:
@kotyara1005 looks great, just a few things regarding updating docs and docstring formatting.

r0fls:
hmm i guess since you went MIA and because other parts are using the `"""docstring...``` notation (rather than a newline) then we can merge this for now... note one of the doctstrings is still apparently incomplete.

38elements:
I think it is necessary to add this to [docs/sanic/config.md](https://github.com/channelcat/sanic/blob/master/docs/sanic/config.md).

38elements:
I think there is no need to make it possible to create configuration files by multiple formats.

38elements:
I think there is little needs to make it possible to create configuration files by multiple formats.

r0fls:
ok let's close this for now then unless we see support to reopen it.

Issue #608, opened 2017-04-01
Access log


messense:
Sanic should have access log. Don't you think so?

seemethere:
How so?

messense:
@seemethere Every other web frameworks have built-in access log I think, it is quite useful.

r0fls:
I think we should definitely add this, though it should be capable of being disabled.

seemethere:
Closed by #625

PR #625, opened 2017-04-11
based on issue #608, create access log

zenixls2:
use predefined 'network' logger in default.yml.
could be disabled using `log_config_path=None`, otherwise if not defined, use the `default.yml` in package, or use user defined path.
default.yml should follow python3 logging dictConfig schema.


messense:
I don't think adding `pyyaml` as a dependency is a good idea. Use `logging.dictConfig` as default would be better.

zenixls2:
@messense you mean to create a default config dict instead of providing default yaml file?

zenixls2:
by the way, it seems configure sanic logger breaks something...
I would like to check it first.

messense:
@zenixls2 Yes. We should avoid adding too much dependencies and prefer standard library modules if possible.

messense:
Ref Django logging: https://docs.djangoproject.com/en/1.11/topics/logging/#configuring-logging

zenixls2:
now predefined param becomes `log_config`.
we have a `DEFAULT_LOG_CONFIG` defined in `config.py`.
pass in a new dict or set it to None to use new config or close access log.

sanic/config.py
```diff
@@ -1,10 +1,75 @@
+from sanic.defaultFilter import DefaultFilter
 import os
-
 import types
 
 SANIC_PREFIX = 'SANIC_'
 
 
+DEFAULT_LOG_CONF = {
```
messense: Why not just named it `LOGGING` like Django does.

sanic/config.py
```diff
@@ -1,10 +1,75 @@
...
+        },
+        'access': {
+            'format': '%(asctime)s - [%(levelname)s][%(host)s]: ' +
+                      '%(request)s %(message)s %(status)d %(byte)d',
+            'datefmt': '%Y-%m-%d %H:%M:%S'
+        }
+    },
+    'handlers': {
+        'internal': {
+            'class': 'logging.handlers.TimedRotatingFileHandler',
```
messense: I think it should log to `sys.stderr` by default. Set `filename` to `access.log` won't help much because most people don't want to write logs to current working directory, they most likely want to write logs to `/var/log/xxx/access.log` or simply log to `syslog`.

sanic/server.py
```diff
@@ -189,6 +193,14 @@ def write_response(self, response):
                 response.output(
                     self.request.version, keep_alive,
                     self.request_timeout))
+            if self.has_log:
```
messense: I don't think this is necessary. If you don't want sanic's logging, just configuration logging and disable sanic related loggers.

messense:
[Configuring Logging for a Library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)

sanic/server.py
zenixls2: actually calling logging functions makes sanic a lot slower, even if it does nothing.
i think this is necessary for performance purpose.

sanic/config.py
zenixls2: Are we going to make sanic a Django-like project? If that's true, I will follow this naming.

sanic/server.py
messense: Agreed.

sanic/config.py
messense: Not really. But it should not be named with prefix `DEFAULT_` as it is a config, not just default.

`LOGGING_CONF` would be fine but it is already in `config` module so `LOGGING` makes more sense.

messense: But when use with `app.config.from_object` method and etc it should be called `LOGGING` instead of `DEFAULT_LOGGING`.

sanic/config.py
zenixls2: yes, while I was writing this configuration, I did have the same idea.
but I had a hard time to cope with the permission problems (/var/log/xxx needs admin to access). Maybe syslog is a choice.

sanic/config.py
zenixls2: Ok, that make sense

zenixls2:
change default config to `LOGGING`, and set default output to `stderr`.
users still could change to use any of the `handlers` defined in `LOGGING`
ex:
`
LOGGING['network']['handlers'] = ['accessTimedRotatingFile', 'errorTimedRotatingFile']
app.run(log_config=LOGGING)
`
Notice that `accessSysLog` and `errorSysLog` handlers cannot be used when system is without syslog support.

r0fls:
@messense is this still in need of changes or otherwise incomplete, in your opinion?

r0fls:
@zenixls2 can you add something to the docs about using this?

zenixls2:
@r0fls Is there any documentation guideline that must follow?
I'll finish this later.

seemethere:
Why didn't the travis job go off for this PR? 🤔

sanic/config.py
```diff
@@ -1,9 +1,119 @@
...
+            'formatter': 'simple',
+            'stream': sys.stderr
+        },
+        # before you use accessSysLog, be sure that log levels
+        # 0, 10, 20 have been enabled in you syslog configuration
+        # otherwise you won't be able to see the output in syslog
+        # logging file.
+        'accessSysLog': {
+            'class': 'logging.handlers.SysLogHandler',
+            'address': _address_dict.get(platform.system(),
```
r0fls: is there a reason you're defaulting to windows here?

r0fls:
@zenixls2 no guidelines for docs right now. I think it would make sense as its own section "logging". It should include details of how to enable/disable it, and any other important configuration you're able to document.

sanic/config.py
zenixls2: @r0fls for you to check. just follow python3's implementation and recommendation on default value.
actually not specified to Windows here
https://docs.python.org/3/library/logging.handlers.html#sysloghandler

seemethere:
I'm a tentative 👍 on this and will merge once we get some documentation on all the features as well. And maybe a test for good measure?

sanic/config.py
r0fls: oh i see, i just noticed it was the same as the windows value

zenixls2:
Documentation added.

PR #627, opened 2017-04-11
Updating examples for 0.5.0

Sniedes722:
Trying to continue to update docs based on the current docs.

### DIFF
--- a/examples/exception_monitoring.py
+++ b/examples/exception_monitoring.py
@@ -1,9 +1,7 @@
 """
 Example intercepting uncaught exceptions using Sanic's error handler framework.
-
 This may be useful for developers wishing to use Sentry, Airbrake, etc.
 or a custom system to log and monitor unexpected errors in production.
-
 First we create our own class inheriting from Handler in sanic.exceptions,
 and pass in an instance of it when we create our Sanic instance. Inside this
 class' default handler, we can do anything including sending exceptions to
@@ -39,7 +37,7 @@ server's error_handler to an instance of our CustomHandler
 """
 
 from sanic import Sanic
-from sanic.response import json
+from sanic import response
 
 app = Sanic(__name__)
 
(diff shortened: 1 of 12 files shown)


Sniedes722:
Travis build is failing due to unrelated tests.

seemethere:
Should be fixed after #626, just make sure to merge those changes in with yours

Sniedes722:
@seemethere does it make sense to keep this open for a bit?

r0fls:
@Sniedes722 are you planning to add more to it?

Sniedes722:
@r0fls will throw in a few more over the next couple days.

Sniedes722:
Updated basic examples to be in-line with the docs. Also added a detailed example, a port of [Plotly's Flask Example](https://github.com/plotly/plotlyjs-flask-example) to Sanic, for browser-based graphs using pandas, numpy, and some Sanic extensions. I'll open a new request whenever I have time to add more.

38elements:
@seemethere @r0fls 
Would you please check this pull request?
I think most of the changes are unnecessary.

Sniedes722:
@38elements it seems like the only qualm you have is that I used responses **exactly** like the docs list them:
```python
import response

@app.route("/")
async def index(request):
    return response.response_type()

```

r0fls:
@Sniedes722 @38elements I personally often use them that way (`from sanic import response`). However if we want to standardize the usage pattern (in either direction) I'm ok with that. Beyond that, I can see @38elements's point that adding examples that don't exhibit sanic functionality aren't necessary. I agree the plotly example might make more sense as its own extension or gist, in order to keep the examples focusing on sanic features.

38elements:
I think changing `text()` to `response.text()` is needless.
This change cause losting consistency.

I thought that implicit standards for  adding examples existed.
I thought that the example is one of the following.

* Example of how to use Sanic

* Example of how to use Sanic extensions

* Example of how to use Sanic and asynchronous library

r0fls:
@38elements yes I think that is a fair assumption. The examples obviously aren't part of the core, so the approval is less strict in that area. However I understand your concern.

Issue #629, opened 2017-04-12
Installation from pypi broken on 0.5.0


adamserafini:
Installation from pypi on Ubuntu works fine for version 0.4.1.

```pip3 install sanic==0.4.1```

It fails for ```0.5.0```:

```pip3 install sanic==0.5.0```

```UnicodeEncodeError: 'ascii' codec can't encode character '\U0001f680' in position 2: ordinal not in range(128)```

For example, the following `Dockerfile` fails to build:

```
FROM ubuntu:16.10
RUN apt-get -y update && apt-get install -y python3 python3-pip
RUN pip3 install sanic==0.5.0
```

PR #631, opened 2017-04-12
Fix installation on Ubuntu 16.10 #629

adamserafini:
Fixes issue #629. 

Printing a unicode string at the end of the `setup.py` script is asking for trouble. It's also redundant: the pip tool itself tells the user whether installation was successful.


r0fls:
As much as I liked the rocket, it looks like it has to go... 🚀

PR #632, opened 2017-04-13
Add path type for router

messense:
Just like Flask's path route rule: http://flask.pocoo.org/docs/0.12/quickstart/#variable-rules

### DIFF
--- a/sanic/router.py
+++ b/sanic/router.py
@@ -16,6 +16,7 @@ REGEX_TYPES = {
     'int': (int, r'\d+'),
     'number': (float, r'[0-9\\.]+'),
     'alpha': (str, r'[A-Za-z]+'),
+    'path': (str, r'[^/].*?'),
 }
 
 ROUTER_CACHE_SIZE = 1024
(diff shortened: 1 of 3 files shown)


sanic/router.py
```diff
@@ -161,10 +163,10 @@ def add_parameter(match):
             parameters.append(parameter)
 
             # Mark the whole route as unhashable if it has the hash key in it
-            if re.search('(^|[^^]){1}/', pattern):
+            if re.search(r'(^|[^^]){1}/', pattern):
                 properties['unhashable'] = True
             # Mark the route as unhashable if it matches the hash key
-            elif re.search(pattern, '/'):
```
messense: I believe this is a bug.

seemethere: Does it not work the other way around?

messense: It was plain wrong. 

Note that the above comment says `Mark the route as unhashable if it matches the hash key`, the hash key is `/`, so `re.search(pattern, '/'):` means search `pattern` in `/` not search `/` in `pattern`.

Issue #637, opened 2017-04-14
Configuration for keep-alive


38elements:
I think that configuration to turn off keep-alive is necessary.

messense:
+1

Issue #639, opened 2017-04-15
this.signal.stopped in __init__() is True when python -m pytest tests is executed


38elements:
`this.signal.stopped` in `__init__()` is `True` when `python -m pytest tests` is executed.
I think that it is appropriate that `this.signal.stopped` in `__init__()` is `False`.

`this.signal.stopped` in `__init__()` is `False` when `python -m pytest tests/test_foo.py` is  is executed.

Would you tell me how to turn `this.signal.stopped` in `__init__()`  into `False` when `python -m pytest tests` is executed.

[sanic/server.py](https://github.com/channelcat/sanic/blob/7fe418d1b7c0d6fd03104f8a641da5c08a114da6/sanic/server.py#L71-L90)

```
    def __init__(self, *, loop, request_handler, error_handler,
                 signal=Signal(), connections=set(), request_timeout=60,
                 request_max_size=None, request_class=None):
        self.loop = loop
        self.transport = None
        self.request = None
        self.parser = None
        self.url = None
        self.headers = None
        self.signal = signal
        self.connections = connections
        self.request_handler = request_handler
        self.error_handler = error_handler
        self.request_timeout = request_timeout
        self.request_max_size = request_max_size
        self.request_class = request_class or Request
        self._total_request_size = 0
        self._timeout_handler = None
        self._last_request_time = None
        self._request_handler_task = None
        # Add the following code
        print(self.signal.stopped)
```

tests/test_foo.py

```
from sanic import Sanic
from sanic.response import text

app = Sanic('test_foo')


@app.route('/1')
async def handler_1(request):
    return text('OK')


def test_server_error_request_timeout():
    request, response = app.test_client.get('/1')
    assert False
```

38elements:
[signal](https://github.com/channelcat/sanic/blob/7fe418d1b7c0d6fd03104f8a641da5c08a114da6/sanic/server.py#L72) are shared each applications.

If you change it like the following code, it solves.

sanic/app.py

```
-from sanic.server import serve, serve_multiple, HttpProtocol
+from sanic.server import Signal, serve, serve_multiple, HttpProtocol

+        signal = Signal()
         server_settings = self._helper(
             host=host, port=port, debug=debug, before_start=before_start,
             after_start=after_start, before_stop=before_stop,
             after_stop=after_stop, ssl=ssl, sock=sock, workers=workers,
             loop=loop, protocol=protocol, backlog=backlog,
-            register_sys_signals=register_sys_signals)
[...]

PR #646, opened 2017-04-17
allow disabling keep alive

r0fls:
Fixes: https://github.com/channelcat/sanic/issues/637

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -71,7 +71,7 @@ class HttpProtocol(asyncio.Protocol):
     def __init__(self, *, loop, request_handler, error_handler,
                  signal=Signal(), connections=set(), request_timeout=60,
                  request_max_size=None, request_class=None,
-                 no_keep_alive=False):
+                 keep_alive=True):
         self.loop = loop
         self.transport = None
         self.request = None
@@ -89,11 +89,11 @@ class HttpProtocol(asyncio.Protocol):
         self._timeout_handler = None
         self._last_request_time = None
         self._request_handler_task = None
-        self._no_keep_alive = no_keep_alive
+        self._keep_alive = keep_alive
 
     @property
     def keep_alive(self):
-        return (not self._no_keep_alive
+        return (self._keep_alive
                 and not self.signal.stopped
                 and self.parser.should_keep_alive())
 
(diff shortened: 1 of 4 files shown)


sanic/server.py
```diff
@@ -88,10 +89,13 @@ def __init__(self, *, loop, request_handler, error_handler,
         self._timeout_handler = None
         self._last_request_time = None
         self._request_handler_task = None
+        self._no_keep_alive = no_keep_alive
 
     @property
     def keep_alive(self):
-        return self.parser.should_keep_alive() and not self.signal.stopped
+        return (self.parser.should_keep_alive()
```
messense:
```python
@property
def keep_alive(self):
    return (not self._no_keep_alive
            and not self.signal.stopped
            and self.parser.should_keep_alive())
```

Would be better?

38elements:
I think it is better to be able to set KEEP_ALIVE by [config](https://github.com/channelcat/sanic/blob/master/sanic/config.py) and there is a problem like #639.

r0fls:
Well this does get loaded from config here: https://github.com/channelcat/sanic/pull/646/files#diff-c4444a35dbd5e37ee480b0e8888e0880R682

How does #639 apply exactly?

38elements:
I am sorry. 
I was misunderstanding for config.

r0fls:
Do we think the variable makes more sense as just `KEEP_ALIVE` vs `NO_KEEP_ALIVE`?

38elements:
No, just  I did not read config.py.

38elements:
About #639, it occurred when I implemented a similar test.

38elements:
I am sorry for changing my opinion.
Because I feel that `KEEP_ALIVE` is more common, I think that `KEEP_ALIVE` is better.

r0fls:
No worries, I think I agree so better late than never :) I'll update it to use that...

38elements:
Thanks.

38elements:
I think that it is  better to add tests for that `keep alive` is valid and invalid.
For #639, I think that the test when `keep alive`  is valid will not work.

38elements:
I understood the cause.
I commented in #639.

38elements:
My English is not good.
I am sorry if I confused you.

r0fls:
No you're English is fine. I'm just traveling. I'll try to add the test soon thank you. If you have an idea for it in mind you can put it here otherwise I'll plan to do it soon...

38elements:
Thanks.

PR #502, opened 2017-03-01
Fixed token parse error

Sniedes722:
Token handler was returning a 'list of out range' Index Error, so I added token as a Request attribute and returned it in a manner identical to the ip handler.

PR #649, opened 2017-04-17
Fixed token attribute

Sniedes722:
Originally referenced in #502, this updates token in line with other parsed request attributes and also updates the test to reflect how token type and value should both be asserted.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -77,7 +77,7 @@ class Request(dict):
 
         :return: token related to request
         """
-        self._token = getattr(self, '_token', 
+        self._token = getattr(self, '_token',
                                     self.headers.get('Authorization'))
         return self._token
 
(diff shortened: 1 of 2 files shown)


Sniedes722:
@r0fls I'm ok with this too. This token attribute has honestly been holding me back from building an Oauth lib. Separating the token type from the token itself (like you've done) is honestly much better.

r0fls:
@Sniedes722 ok that makes sense. Awesome you're working on an oauth library since we need auth extensions. Please make a PR with a link to it once you're ready. 

Regarding this PR, I think this works better for what you need from Sanic: https://github.com/channelcat/sanic/pull/656. Otherwise there isn't much purpose to the token attribute (since you can access the same thing directly with the Authorization header).

r0fls:
You can see from the tests in the other PR that it accepts: Bearer Token <TOKEN>, Token <TOKEN>, and <TOKEN> and produces <TOKEN> in all three cases. I think this is the best of both worlds.

Sniedes722:
@r0fls haven't disagreed yet my friend.

PR #652, opened 2017-04-19
Fix `this.signal.stopped` is `True` #639

38elements:
This solves problem that `this.signal.stopped` is `True` in second or or subsequent tests.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -548,13 +548,12 @@ class Sanic:
                 warnings.simplefilter('default')
             warnings.warn("stop_event will be removed from future versions.",
                           DeprecationWarning)
-        signal = Signal()
         server_settings = self._helper(
             host=host, port=port, debug=debug, before_start=before_start,
             after_start=after_start, before_stop=before_stop,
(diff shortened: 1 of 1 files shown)


sanic/app.py
```diff
@@ -548,12 +548,13 @@ def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
...
             warnings.warn("stop_event will be removed from future versions.",
                           DeprecationWarning)
+        signal = Signal()
         server_settings = self._helper(
             host=host, port=port, debug=debug, before_start=before_start,
             after_start=after_start, before_stop=before_stop,
             after_stop=after_stop, ssl=ssl, sock=sock, workers=workers,
             loop=loop, protocol=protocol, backlog=backlog,
-            register_sys_signals=register_sys_signals)
+            register_sys_signals=register_sys_signals, signal=signal)
```
r0fls: what is the reasoning for adding this as an argument in both cases, why not just do `signal=Signal()` within the helper and do it once?

38elements: Thank you for reviewing.
The both `signal` variables is not necessary.
They was removed.

PR #662, opened 2017-04-25
invariant: body after request is processed must be binary

rsrdesarrollo:
Body must be of type bytes after request is processed, even if there is no data.
If not, code will fail for example in [here](https://github.com/channelcat/sanic/blob/master/sanic/request.py#L97) if the request does not have any parameters in the body but the content-type is application/x-www-form-urlencoded

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -173,8 +173,7 @@ class HttpProtocol(asyncio.Protocol):
         self.request.body.append(body)
 
     def on_message_complete(self):
-        if self.request.body:
-            self.request.body = b''.join(self.request.body)
+        self.request.body = b''.join(self.request.body)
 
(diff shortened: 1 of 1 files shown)


seemethere:
I don't have a milestone to group this with but I will merge once I do.

sanic/server.py
```diff
@@ -173,8 +173,7 @@ def on_body(self, body):
         self.request.body.append(body)
 
     def on_message_complete(self):
-        if self.request.body:
-            self.request.body = b''.join(self.request.body)
+        self.request.body = b''.join(self.request.body)
```
r0fls: This will fail if the request body is `None`. Does that ever happen?

seemethere: No originally `self.request.body` instantiates itself as a list

PR #666, opened 2017-04-27
Fix #665 - ImportError not preserved in __main__.py

### LINKED ISSUE #665: __main__ top-level script shadowing ImportError
Given `main.py`:
```python
import nonexistent  # assuming this is... non-existent


app = Sanic(__name__)


@app.route('/')
async def index(request):
    return response.html('<p>Hello</p>')
```

When we try to import something non-existent, exception `ImportError` will be thrown,
[line 38 of __main__.py](https://github.com/channelcat/sanic/blob/5fd62098bd2f2722876a0873d5856d70046d3889/sanic/__main__.py#L38) does not preserve the exception, reporting preset message (what's provided on commandline) instead.  This is what we will get:

```
python -m sanic main.app
No module named main found.
  Example File: project/sanic_server.py -> app
  Example Module: project.sanic_server.app
```

It is very hard to find the true cause if the import statement that failed is, let's say, buried three levels of modules deep, for example.

pyx:
Now, running with `main.py` (mentioned in #665), will report this:
```
python -m sanic main.app
No module named 'nonexistent' found.
  Example File: project/sanic_server.py -> app
  Example Module: project.sanic_server.app
```

The wording is far from perfect, but at least we now know what failed to be imported.

### DIFF
--- a/sanic/__main__.py
+++ b/sanic/__main__.py
@@ -36,9 +36,9 @@ if __name__ == "__main__":
         app.run(host=args.host, port=args.port,
                 workers=args.workers, debug=args.debug, ssl=ssl)
     except ImportError as e:
-        log.error("{} found.\n"
+        log.error("No module named {} found.\n"
                   "  Example File: project/sanic_server.py -> app\n"
                   "  Example Module: project.sanic_server.app"
-                  .format(e))
+                  .format(e.name))
     except ValueError as e:
         log.error("{}".format(e))
--- a/sanic/__main__.py
+++ b/sanic/__main__.py
@@ -35,10 +35,10 @@ if __name__ == "__main__":
 
         app.run(host=args.host, port=args.port,
                 workers=args.workers, debug=args.debug, ssl=ssl)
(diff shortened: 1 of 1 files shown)


pyx:
Achievement unlocked - magic number 666. :)

sanic/__main__.py
```diff
@@ -35,10 +35,10 @@
 
         app.run(host=args.host, port=args.port,
                 workers=args.workers, debug=args.debug, ssl=ssl)
-    except ImportError:
-        log.error("No module named {} found.\n"
+    except ImportError as e:
+        log.error("{} found.\n"
```
r0fls: This probably shouldn't say `found`?

pyx: I have to admit, it is confusing, the default message of `ImportError` is negative, the final output reads okay.

Anyway, since Sanic only works with Python 3.5+, we can use ImportError.name to report the missing module name, so I've pushed an improved version, please check the latest commit.

Issue #659, opened 2017-04-24
Blueprint routes does not support instance methods


metatoaster:
Given a class that provide some async method, with a registration method of some kind that will add a route to expose through a sanic instance like so:

```python
from sanic import Sanic
from sanic import Blueprint
from sanic import response


class Manager(object):

    async def info(self, request):
        return response.json({'id': id(self)})

    def register(self, app):
        app.add_route(self.info, '/info')


app = Sanic()
manager = Manager()
manager.register(app)
```

The registration will be done fine and the app will be able to serve the registered route (i.e. `/info`) as expected.  However if blueprints were involved, i.e. if the `Manager` class were to provide a blueprint and add the route to it instead, then have the app register the entire blueprint:

```python
class Manager(object):

    def __init__(self):
        self.blueprint = Blueprint(str(id(self)))
        self.blueprint.add_route(self.info, '/info')

    async def info(self, request):
        return response.json({'id': id(self)})

    def register(self, app):
        app.blueprint(self.blueprint)
```

The following exception will happen

```
Traceback (most recent call last):
  File "bpfail.py", line 21, in <module>
    manager.register(app)
  File "bpfail.py", line 16, in register
    app.blueprint(self.blueprint)
  File "lib/python3.6/site-packages/sanic/app.py", line 233, in blueprint
    blueprint.register(self, options)
  File "lib/python3.6/site-packages/sanic/blueprints.py", line 40, in register
    future.handler.__blueprintname__ = self.name
AttributeError: 'method' object has no attribute '__blueprintname__'
```

As Python does not appear to support assignment of arbitrary attributes to instance methods (but are allowed for standard functions and lambdas), this either need to be worked around or documented as unsupported.

38elements:
Is there needs of putting instance method in blueprint?

metatoaster:
The fix could be as simple as catching the `AttributeError: 'method' object has no attribute '__blueprintname__'` and raising it as unsupported, or as I said, put it into the documentation as to why this can't be done, or use a workaround.

Also, when I realized what the intent of the marker was (also what it is trying to do), exposing the instance methods directly as a route is probably not as clean anyway. The better solution is to have a setup function to be the closure for the actual callables that will be the route handlers, which can wrap around the actual raw methods that provide the logic; being encapsulated in a closure also mean that the registration can be done for any number of blueprints.

The goal of this issue is to more document a potential problem, which hopefully will have a resolution that minimise any potential negative impacts that will be caused by this.

[...]

PR #667, opened 2017-04-27
Recovery the issue 659

Zheaoli:
Recovery the issue 659

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -50,7 +50,7 @@ class Blueprint:
                 methods=future.methods,
                 host=future.host or self.host,
                 strict_slashes=future.strict_slashes
-            )(future.handler)
+                )(future.handler)
 
         for future in self.websocket_routes:
             # attach the blueprint name to the handler so that it can be
@@ -62,7 +62,7 @@ class Blueprint:
                 uri=uri,
                 host=future.host or self.host,
                 strict_slashes=future.strict_slashes
-            )(future.handler)
+                )(future.handler)
 
         # Middleware
         for future in self.middlewares:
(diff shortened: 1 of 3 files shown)


Zheaoli:
#659

38elements:
If needs of putting instance method in blueprint is nothing, nobody try to put instance method in blueprint.
I think this is unnecessary if nobody try to put instance method in blueprint.

Zheaoli:
Actually,not nobody
In issue 659 , he want to use class and method to make his code structure more clear. Although it's not right , but I can understand it.
So,as what he say, It's necessary to give me some guide/exception for that blueprint handler shouldn't be a instance method

sanic/blueprints.py
```diff
@@ -93,10 +95,16 @@ def route(self, uri, methods=frozenset({'GET'}), host=None,
         :param uri: endpoint at which the route will be accessible.
         :param methods: list of acceptable HTTP methods.
         """
+
```
r0fls: I'm fine with adding this log mesage, but do the whitespace changes here conform to the rest of the project?

Zheaoli: oh ,it just format by PyCharm , That`s OK, I will recovery it.

38elements:
If it is necessary to add exception for instance method,
I think making it possible like #676 is better than adding exception.

Zheaoli:
@38elements 
Seriously , I dont't think it's better than `raise exception` 
here is detail:
If you have N instance method to add in the blueprint , that means you have to make another N `_handler` to decorate it .
If the N is huge , Python have to use a lot of memory to maintain a call stack , I think is a wasting-memory way.
So, I don't think it's right to allow people add instance method in blueprint route.

38elements:
I think this and #676 are not necessary.
#676 making it possible to put instance method, I think it is better.

PR #668, opened 2017-04-28
Add the killing of children

seemethere:
Kills children processes when parent process receives a signal to
shutdown.

Solves for #594

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -502,6 +502,8 @@ def serve_multiple(server_settings, workers):
     def sig_handler(signal, frame):
         log.info("Received signal {}. Shutting down.".format(
             Signals(signal).name))
+        for process in processes:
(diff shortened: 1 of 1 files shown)


seemethere:
No with this you send a sigint to the parent process as originally stated
in the issue

r0fls:
>the parent process

This is what I meant by the lowest number process listed in `ps aux | grep app.py`

r0fls:
I thought this was working a long time ago 👿

seemethere:
Are you sure you're killing the parent process? When I did it with an `os.getpid()` it seemed to work fine.

r0fls:
derp, i was sending a kill signal not int. `kill -2 parent_pid` works

r0fls:
nice title btw

seemethere:
You know how I do :D

PR #670, opened 2017-04-28
In case of error when starting sanic

ticosax:
Don't exit with code 0

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -570,7 +570,7 @@ class Sanic:
                 serve(**server_settings)
             else:
                 serve_multiple(server_settings, workers)
-        except Exception:
+        except:
             log.exception(
                 'Experienced exception while trying to serve')
             raise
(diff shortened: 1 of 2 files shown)


sanic/app.py
```diff
@@ -570,9 +570,10 @@ def run(self, host="127.0.0.1", port=8000, debug=False, before_start=None,
                 serve(**server_settings)
             else:
                 serve_multiple(server_settings, workers)
-        except:
+        except Exception:
```
r0fls: this behaves the same as a bear `except:`

ticosax: Not really, a bare `except:` captured everything that also inherits `BaseException` including signals like KeyboardInterrupt or SystemExit.
that you probably don't want to capture here. it is considered anti-pattern.
Your call though.

seemethere: We do want it to catch `KeyboardInterrupt` though

ticosax: > We do want it to catch KeyboardInterrupt though

Fair enough, I will revert that change

messense: I would think if you want to catch everything, using

```python
try:
    ...
except BaseException:
    ...
```

is better than bare `except:` as it is much more clear.

PR #676, opened 2017-05-01
Blueprint.add_route() for instance method

38elements:


Zheaoli:
I get some feedback in #667

PR #681, opened 2017-05-02
Remove before_start, before_stop, after_start and after_stop

38elements:
Since current version is 0.5.2, this removes before_start, after_start, before_stop and after_stop.


niteeshm:
Why has this been removed ? Now if I need to use the loop then how do I use it ?

38elements:
[Listeners](http://sanic.readthedocs.io/en/latest/sanic/middleware.html#listeners)  will solve.

PR #682, opened 2017-05-02
fix README.rst -> long_description encoding

graingert:
woops, was going to make this change before making a PR, then forgot!


graingert:
this is ok for now because README.rst happens to be ASCII.

PR #695, opened 2017-05-05
aiopeewee example

kszucs:
Asyncio compatible aiopeewee orm example

required package: `aiopeewee`: https://github.com/kszucs/aiopeewee


kszucs:
Thanks!

PR #696, opened 2017-05-05
Add Request.stream

38elements:
This adds `Request.stream`.


38elements:
I am sorry.
I add stream stream decorator at #697 .

PR #697, opened 2017-05-06
Add Request.stream

38elements:
This adds `Request.stream` and `stream decorator`, as below.

```python
from sanic import Sanic
from sanic.views import CompositionView
from sanic.views import HTTPMethodView
from sanic.views import stream as stream_decorator
from sanic.response import stream, text

bp = Blueprint('blueprint_request_stream')
app = Sanic('request_stream')


class SimpleView(HTTPMethodView):

    @stream_decorator
    async def post(self, request):
        result = ''
        while True:
            if body is None:
                break
            result += body.decode('utf-8')
        return text(result)


@app.post('/stream', stream=True)
async def handler(request):
    async def streaming(response):
        while True:
            if body is None:
                break
            body = body.decode('utf-8').replace('1', 'A')
            response.write(body)
    return stream(streaming)


@bp.put('/bp_stream', stream=True)
async def bp_handler(request):
    result = ''
    while True:
        if body is None:
            break
        result += body.decode('utf-8').replace('1', 'A')
    return text(result)


async def post_handler(request):
    result = ''
    while True:
        if body is None:
            break
        result += body.decode('utf-8')
    return text(result)

app.blueprint(bp)
app.add_route(SimpleView.as_view(), '/method_view')
view = CompositionView()
view.add(['POST'], post_handler, stream=True)
app.add_route(view, '/composition_view')


if __name__ == '__main__':
```

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -60,6 +60,7 @@ class Sanic:
         self.sock = None
         self.listeners = defaultdict(list)
         self.is_running = False
+        self.is_request_stream = False
         self.websocket_enabled = False
         self.websocket_tasks = []
 
@@ -110,12 +111,14 @@ class Sanic:
 
     # Decorator
     def route(self, uri, methods=frozenset({'GET'}), host=None,
-              strict_slashes=False):
+              strict_slashes=False, stream=False):
         """Decorate a function to be registered as a route
 
         :param uri: path of the URL
         :param methods: list or tuple of methods allowed
         :param host:
+        :param strict_slashes:
+        :param stream:
         :return: decorated function
         """
 
@@ -124,7 +127,11 @@ class Sanic:
         if not uri.startswith('/'):
             uri = '/' + uri
 
+        if stream:
+            self.is_request_stream = True
+
         def response(handler):
+            handler.is_stream = stream
             self.router.add(uri=uri, methods=methods, handler=handler,
                             host=host, strict_slashes=strict_slashes)
             return handler
@@ -159,6 +166,13 @@ class Sanic:
     def delete(self, uri, host=None, strict_slashes=False):
         return self.route(uri, methods=frozenset({"DELETE"}), host=host,
                           strict_slashes=strict_slashes)
+
+    def stream(
+            self, uri, methods=frozenset({"POST"}), host=None,
+            strict_slashes=False):
+        return self.route(uri, methods=methods, host=host,
+                          strict_slashes=strict_slashes,
+                          stream=True)
 
     def add_route(self, handler, uri, methods=frozenset({'GET'}), host=None,
                   strict_slashes=False):
@@ -651,6 +665,8 @@ class Sanic:
         server_settings = {
             'protocol': protocol,
             'request_class': self.request_class,
+            'is_request_stream': self.is_request_stream,
+            'router': self.router,
             'host': host,
             'port': port,
             'sock': sock,
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -144,13 +144,13 @@ class Sanic:
         return self.route(uri, methods=frozenset({"GET"}), host=host,
                           strict_slashes=strict_slashes)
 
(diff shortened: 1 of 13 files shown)


r0fls:
does this replace https://github.com/channelcat/sanic/pull/696?

38elements:
I am sorry.
Yes, I do.

tests/test_request_stream.py
```diff
@@ -0,0 +1,54 @@
...
+
+app.blueprint(bp)
+
+
+def test_request_stream():
+    data = ""
+    for i in range(1, 250000):
+        data += str(i)
+    request, response = app.test_client.post('/stream', data=data)
+    text = data.replace('1', 'A')
```
r0fls: why is this needed? 🤔

38elements: This is used [here](https://github.com/38elements/sanic/blob/4197d148e4550002fb2d351a1bb577788ba627d6/tests/test_request_stream.py#L54).

38elements: `request handler` replaces '1' with 'A'.
This makes the correct result.

38elements: I removed replacing '1' with 'A'.

sanic/app.py
```diff
@@ -27,7 +27,7 @@
 class Sanic:
 
     def __init__(self, name=None, router=None, error_handler=None,
-                 load_env=True, request_class=None,
+                 load_env=True, request_class=None, is_request_stream=False,
```
r0fls: I think it makes more sense to set `self.is_request_stream = True` in the `app.stream` method, if used. This would avoid having to initialize it here.

r0fls: So around line 170

38elements: I think it is better to add `self.is_request_stream = True` in `app.route()` for `blueprint.stream()`. 

```
if stream:
    self.is_request_stream = True
```

38elements:
There is a bug.
This can not use `CompositionView`.

```
AttributeError: 'function' object has no attribute 'is_stream'
```

38elements:
I add `stream` in `composition_view.add()` and test for  `CompositionView` and `HTTPMethodView`.

38elements:
I am sorry.
I added `stream decorator` for HTTPMethodView.

38elements:
I think that it is necessary to put `is_request_stream` in constructor of Sanic for  `CompositionView` and `HTTPMethodView`.

38elements:
There is a bug.
 `CompositionView` and `HTTPMethodView`  can not set `self.is_request_stream = True`.
It is necessary to set True to `self.is_request_stream`  in `app.add_route()` for  `CompositionView` and `HTTPMethodView` .
I will fix it and add tests.

38elements:
I fixed it and added tests.

seemethere:
So instead of it's own decorator wouldn't it make sense to make it as just a parameter that you can pass in through the regular method decorators?

38elements:
Thank you for reviewing.
Since I think it is more consistent, I think it is better `stream parameter` than `stream decorator`. 
I will delete `stream decorator` and add  `stream parameter` and change tests.

38elements:
I replaced stream decorator to stream parameter.
Only post, put and patch decorator have stream argument.

seemethere:
Will merge once I create a milestone for this

38elements:
Thank you.

38elements:
I added a heading in `streaming.md`.

r0fls:
we should also remove this from the todo here: https://github.com/channelcat/sanic/blob/master/README.rst#todo

38elements:
@r0fls 
I think the line should be removed by #733.

gsvijayraajaa:
Trying to execute the code shared by @38elements . Getting the error: from sanic.views import stream as stream_decorator
ImportError: cannot import name 'stream'

Is the code merged to master ?

PR #698, opened 2017-05-06
exception_base decorator

asdfghjBennyHuang:
I added a app.exception_base as an alternative to app.exception, the differences is that exception_base use a real try except syntax to test is a error is a subclass of another error. It can let user take better care of different kinds of exceptions. 
The main reason I add this is that sanic expose the error messages very clearly to the client side, I think we need an approach to handle unexpected error and keep server information away from attackers.

### DIFF
--- a/docs/sanic/exceptions.md
+++ b/docs/sanic/exceptions.md
@@ -43,3 +43,11 @@ Some of the most useful exceptions are presented below:
   usually occurs if there is an exception raised in user code.
 
 See the `sanic.exceptions` module for the full list of exceptions to throw.
+
+You can also use `exceptions_base`, it can catch the exception just like
+normal catching. It can be used for catching all unexpected error and eliminate
+the risk that server information being stolen by attackers/
+
+@app.exception_base(Exception)
+def catch_everything(request, exception):
+	return text( "Server Error!",500)
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -290,7 +290,7 @@ class Sanic:
             return handler
 
         return response
-		
+
     # Decorator
     def middleware(self, middleware_or_request):
         """Decorate and register middleware to be called before a request.
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -273,6 +273,25 @@ class Sanic:
         return response
 
     # Decorator
+    def exception_base(self, *exceptions):
+        """Decorate a function to be registered as a handler for exceptions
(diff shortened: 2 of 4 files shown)


docs/sanic/exceptions.md
```diff
@@ -43,3 +43,11 @@ Some of the most useful exceptions are presented below:
   usually occurs if there is an exception raised in user code.
 
 See the `sanic.exceptions` module for the full list of exceptions to throw.
+
+You can also use `exceptions_base`, it can catch the exception just like
+normal catching. It can be used for catching all unexpected error and eliminate
+the risk that server information being stolen by attackers/
```
r0fls: What information are you preventing from exposing exactly? Sanic exception tracebacks?

asdfghjBennyHuang: Exactly. I am afraid that sometimes tracebacks expose too much. Also when in production, I can use a customized page for unexpected error.
I want to add a handler for all unexpected exceptions instead of wrapping each handlers with a try block. So I add this.

38elements: [CustomErrorHandler](https://github.com/channelcat/sanic/blob/a64c636a336382564bae7a209ea06719639e86a7/examples/exception_monitoring.py) may solve.

asdfghjBennyHuang: yes, inheriting ErrorHandler and overwrite the `lookup` function can achieve the same, but rather than the user to directly touch the ErrorHandler Interface, which may be changed in the future, I hope to have built in support for this important security-related issue and can be directly used by every user. I do handle other minor issue through using CustomHandler. I just think it may be good to let all the user to have this one as nature support.

38elements: @asdfghjBennyHuang 
Do you put `debug=True` in `app.run()`?
The following code like your test code does not display tracing.


```
from sanic import Sanic

app = Sanic(__name__)


@app.route('/')
async def test(request):
    a = [1, 2, 3]
    return text(a[5])


if __name__ == '__main__':
```

Would you post the code which display tracing?

asdfghjBennyHuang: Oh my
I did put `debug=True` in my code.
Sorry for bothering you then...

r0fls:
@asdfghjBennyHuang did you decide this is unnecessary, since you were running in debug mode?

PR #704, opened 2017-05-09
Fix camel case module

### LINKED ISSUE #703: Consistent module naming
I don't want to be the bad guy 😄 , but there is a module file named with camelCase. Disregard me if this is not a problem.

seemethere:
Closes #703


r0fls:
i personally don't see any benefit to keeping this outside of sanic/log.py

docs/sanic/logging.md
```diff
@@ -85,11 +85,11 @@ By default, log_config parameter is set to use sanic.config.LOGGING dictionary f
 
 And `filters`:
 
-- accessFilter (using sanic.default_filter.DefaultFilter)<br>
+- accessFilter (using sanic.logging.DefaultFilter)<br>
```
r0fls: I think this needs to be `using sanic.log.DefaultFilter`

sanic/config.py
```diff
@@ -1,10 +1,11 @@
-from sanic.defaultFilter import DefaultFilter
 import os
 import sys
 import syslog
 import platform
 import types
 
+from sanic.logging import DefaultFilter
```
r0fls: whoops, here too :)

seemethere: I feel like we should have a test for this somewhere...

PR #712, opened 2017-05-11
Fixed plotly_example, now works

stopspazzing:
Old file didn't work because was half finished. Didn't need to have  5+ more dependencies just for a simple example.
Remember: K.I.S.S

PR #714, opened 2017-05-13
Improve example

38elements:
This revert #627.
Since `plotly_example` is inappropriate as an example, this removes it.
Since  the exampe of header exists in [response.md](https://github.com/channelcat/sanic/blob/1787f8617f6afd82434be81d25cdc4548badbf64/docs/sanic/response.md#modify-headers-or-status), this removes `modify_header_example.py` .
Since  the exampe of `url_for` exists in [routing.md](https://github.com/channelcat/sanic/blob/db58bd68f5aca0a475971c3c32e3bb22f8f719b0/docs/sanic/routing.md#url-building-with-url_for),  this removes `url_for_example.py` .
This needs discussion about tests, examples, and document.

This is related to #627, #712, and #725.


38elements:
[tests](https://github.com/channelcat/sanic/tree/master/tests), [document](https://github.com/channelcat/sanic/tree/master/docs), old examples, [README.md](https://github.com/channelcat/sanic#hello-world-example) and [code](https://github.com/channelcat/sanic/blob/master/sanic/handlers.py) are using `text()` and `json()`.
Most code in sanic uses `text()` and `json()`.
Most code in sanic  not use `response.text()` and `response.json()`.

Sniedes722:
@38elements ok!

38elements:
I think changing `text()` to `response.text()` is needless.
This change cause losting consistency.

I thought that implicit standards for  adding examples existed.
I thought that the example is one of the following.

* Example of how to use Sanic

* Example of how to use Sanic extensions

* Example of how to use Sanic and asynchronous library

messense:
I think `from sanic.response import json` is terrible, what I am gonna do to also `import json`?

Always use `from sanic import response` is much better.

38elements:
@messense
Thank you for pointing out.
I think it is not important in this case.
If it is important, the [example](https://github.com/channelcat/sanic/blob/47205136720aa8e701c4aa80302428fa0da13a66/README.rst#hello-world-example) in README.rst was changed.
The example in README.rst was not changed.
I think `json()` can be judged by context and is more consistent.

messense:
In practice, I alway prefer to use `from sanic import response`. As your codebase grows, you can easily change/add response types just by calling `response.type()` and there is no need to change the import. (It's also more friendly to version control systems).

The `READMD.rst` has not changed does not mean it's best practice, it's just a very very simple and naive example that this import thing doesn't really matter.

Sniedes722:
@messense @38elements exactly, it looks much cleaner to me:
```python

@app.route("/")
async def test(request):
    response.json({"Some":"json"})

@app.route("/")
async def test2(request):
    response.text('Some text')
``` 
compared to:

```python
from sanic import Sanic
from sanic.response import json, text

@app.route("/")
async def test(request):
    json({"Some":"json"})

@app.route("/")
async def test2(request):
    text('Some text')

``` 

what if I'm using other packages that use similar calls?? it's definitely not needless, as you say.

38elements:
@messense 
I read your opinion.
But my opinion does not change.
If you think your opinion is correct, it is better that you send pull request of the contents.
There is `All pull requests must be consistent with the existing code.` in `CONTRIBUTING.md`.
I obey the `CONTRIBUTING.md` rather than your opinion.

Sniedes722:
This seems like way too much conflict over one small change. @r0fls @seemethere do you guys have any opinion on this?

r0fls:
Personally I agree with @Sniedes722 and @messense regarding using `from sanic import response` vs `from sanic.response import json`. However I don't think @38elements is really saying not to do that, he's just adhering to the contributing rules. So, this PR makes sense. But since the discussion has come up, maybe we should consider moving everything to the `from sanic import response` style? I'm +1 on that, but it would be nice to get some more agreement there as well... @38elements @seemethere how would you two we feel about switching all examples to that style?

38elements:
I want to confirm the other things.
Do you agree to remove `plotly_example`?
Do you agree to remove `modify_header_example.py`?
Do you agree to remove `url_for_example.py`?

38elements:
@messense  @r0fls 
I think your opinion is subjective rather than objective.
Does your personal opinion take precedence over the contributing rules?

messense:
No, our personal opinion does not take precedence over the contributing rules.  But the contributing rules is very recent thing, we should cleanup codebase before strictly apply that. In the long term I think it would be a mistake to always try to be `be consistent with the existing code` while there is a bad smell in it.

Removing those examples is fine by me.

38elements:
@r0fls 
I am using `from sanic.response import json`, but no problem has occurred.
I think the system design is different.
My `request handler` is small.

38elements:
I think request handler is controller.
I think the logic is not in controller.
In most case, `import json` is not putted in request handler.
I think I should avoid fat controller.

38elements:
If you think that content of `CONTRIBUTING.md` is inappropriate, tring to fix it is right.
I think that if the existing code style is inappropriate, it is better  to replace all the relevant parts.

messense:
If you agree with us that advocating `from sanic import response` is more appropriate as it avoids lots of potential problems `from sanic.response import json` style may lead to, would you mind do the honor to replace all the relevant parts  in this PR ?

38elements:
Do I need to answer that?
My opinion does not change.
As I already wrote, I think the problem you pointed out is not really a problem.
I found a problem. And it is trying to solve it.
In other words, I am acting.
You only repeat the subjective opinion.
I do not plan to discuss with people who only complain.

38elements:
Django uses render() in [document](https://docs.djangoproject.com/en/1.11/topics/http/shortcuts/#example).
Flask uses render_template() in [document](http://flask.pocoo.org/docs/dev/tutorial/views/).
Rails uses render() in [document](http://guides.rubyonrails.org/layouts_and_rendering.html#using-render).

messense:
I couldn't notice but agree the comment in https://github.com/channelcat/sanic/pull/732#issuecomment-302418648 that you are now a little hostile.

We all want to make Sanic better for people to use. That's why we discuss this and contribute to it.
Every opinion is subjective including yours, accusing others being subjective isn't a really a good argument.

Documentation and examples are the first thing users look at while often the last thing developers tend to do. So it's better to make sure it advocating best practices so users won't find it hard to use or don't scale well into larger codebase.

Django/Flask/Rails use the `render*()` in documentation because they don't share the same response namespace structure as Sanic do, so that's OK.

38elements:
I think my leaving this community solve this problem.
I will leave this community.
Good luck!

r0fls:
@38elements I don't think anyone wants you to leave the community. We all value your work and participation. We just want to be welcoming to newcomers as well and try to remain positive.

messense:
@38elements I am sorry if you feel offended. As @r0fls said, we don't want you or anyone to leave the community.

The problem won't be solved by your leaving that's for sure. To solve it we have to reach an agreement. I understand your concerns with contributing rules, we should obey it in general but certainly should not be dictated by it.

BTW, Manish Goregaokar did a very good talk ["Open" as in "Welcoming": Facilitating Contributions](https://www.youtube.com/watch?v=AHprJNUCgQ0) recently on RustFest 2017, I think we can all learn from it.

38elements:
@r0fls @seemethere 

Thanks a lot for your good will.
You are always a great person.
Your contribute is great.

Whether `from sanic import response` or `from sanic.response import json` is a small problem.
There is a big problem behind this and others.
Some people try to overcome by repeating same subjective and illogical opinions.
A person not observing the rules is making a new rule.
They may become majority because they are advantageous.
I think it will have a bad influence.

I may fork and start my project.
It may solve.

Issue #37, opened 2016-10-16
Blueprint middleware applied globally


narzeja:
I was a little too hasty updating the blueprints documentation for middleware/exceptions, and just realized that middlewares and exceptions registered through a blueprint decorator are applied to all routes.

Is this intended behaviour? If so, then the the blueprint documentation must be updated once more.

Alternative behaviour: middleware registered on a blueprint, are only applied to routes for that blueprint.
Some considerations:
- should middleware applied to the app-object (instance of Sanic) also be applied to blueprint middleware?
- if so, how will ordering be handled?

channelcat:
Ah, I missed that in the pull request.  I've updated the documentation for now.  My thoughts on this are:

Applied locally:
 * Is more self-contained, prevents conflicting code.
 * Applying global middleware and exception handling requires registering them separately from the blueprint.

Applied globally:
 * Modularizes middleware and exception handling.
 * Blueprints can facilitate whole modules, allowing a single interface for sharing code between projects.
 * Can use decorators to achieve local middleware and exception handling:
```python
def authenticated(func):
    def handler(request, *args, **kwargs):
        # Middleware goes here
        try:
            return func(request, *args, **kwargs)
        except MyCustomException:
            return text("son, i am dissapoint")
    return handler

@bp.route('/')
@authenticated
def index(request):
    return text("We did it fam")

...
```

In both scenarios you can achieve the same thing, but I think the latter offers more.

As for ordering, I'm thinking they should be applied in the order the blueprint was registered, and in the order they were registered in the blueprint.  I think it would allow blueprints to require other blueprints in a way that's easy to discern in the code.

narzeja:
My concern with using a decorator for local middleware/exceptions, is that it requires a different way of adding them, compared to registering on the application object. My fear is, that it becomes too complicated or too involved to handle middleware/exceptions on blueprints.

I would prefer, that you register middleware and exceptions in the same manner (`@<obj>.middleware`), regardless of applying it on the app-object (affecting all routes) or the blueprint (affecting only those blueprint routes).

eth3lbert:
What is the different between using `@app.middleware` and `@bp.middleware`? Are there any benefits to use `@bp.middleware`?

Tim-Erwin:
[...]

PR #715, opened 2017-05-13
add support for adding middlewares to a specific handler

pandafeeder:
Hi everyone,
Currently when adding a middleware using `@app.middleware`, we're adding them globally.
It seems more flexible if we could add middleware only to interested handlers. This pull request aims to support this feature. It works like below:
`
async def authenticate(request, lock_id):
    if lock_id != request.headers['key_id']:
        return response.json(
            status=403,
            headers={"Authentication": "Failed"}
        )

async def say_goodby(request, respone, lock_id):
    respone.headers.update({"Add-Header": "May the force be with you"})

@app.route('/lock/<lock_id>', middleware={"request": authenticate, "response": say_goodbye})
async def handler(request, lock_id):
    return response.json({"message": "welcome back"})
`
Adding middleware to interested handlers by passing a middleware keyword argument to route decorator.

We can also pass multi-middleware as a list of functions to middleware keyword argument.

It executes a list of functions in order and do a early responding just like global middleware added by `@app.middleware`

All test are clean, you can take a look at [https://travis-ci.org/pandafeeder/sanic](url)

Thanks a lot.

### DIFF
--- a/docs/sanic/middleware.md
+++ b/docs/sanic/middleware.md
@@ -26,6 +26,32 @@ async def print_on_response(request, response):
 	print("I print when a response is returned by the server")
 ```
 
+Besides above two types of middleware declared using the `@app.middleware` decorator,
+you can aslo mount middleware functions to a specific handler as `'request'` or `'response'`
+middleware. Request middleware receives exactly the same parameters as your handler function.
+Response middleware receives an additional `'response'` parameter comparing to Request middleware.
+
+```python
+async def authenticate(request, lock_id):
+    if lock_id != request.headers['key_id']:
+        return response.json(
+            status=403,
+            headers={"Authentication": "Failed"}
+        )
+
+async def say_goodby(request, respone, lock_id):
+    respone.headers.update({"Add-Header": "May the force be with you"})
+
+@app.route('/lock/<lock_id>', middleware={"request": authenticate, "response": say_goodbye})
+async def handler(request, lock_id):
+    return response.json({"message": "welcome back"})
+```
+
+In above example, you can also pass in multi Request or Response middlewares
+as a list of functions to the middleware keyword argument in route decorator.
+They will be excuted in order if no early response happens. Please see below
+`'Responding early'` section for detail.
+
 ## Modifying the request or response
 
 Middleware can modify the request or response parameter it is given, *as long
@@ -68,6 +94,9 @@ async def halt_request(request):
 async def halt_response(request, response):
 	return text('I halted the response')
 ```
+
+Above description also works for middleware functions that mounted on a specific handler
+using the route decorate.
 
 ## Listeners
 
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -128,7 +128,7 @@ class Sanic:
             # attach middleware to specific handler
             if middleware:
                 for r in ('request', 'response'):
-                    mid = middleware[r]
+                    mid = middleware.setdefault(r, None)
                     if mid:
                         if callable(mid):
                             vars(handler)["_"+r+"_middleware"] = mid,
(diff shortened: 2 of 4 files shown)


pyx:
Is this better than regular decorator?

like so:
```python

def before(handler):
    @wraps(handler)
    async def wrapped(request, *args, **kwargs):
        # do something
        return await handler(request, *args, **kwargs)
    return wrapped

def after(handler):
    @wraps(handler)
    async def wrapped(request, *args, **kwargs):
        resp = await handler(request, *args, **kwargs)
        # do something
        return resp
    return wrapped

@app.route('/')
@after
@before
async def index(request):
    return response.text('okay')
```

There is no runtime overhead on routes don't need per-request middleware in this way.
If this is too much boilerplate code, or the order of applied decorators is confusing, some helper functions can help (implementations of helper functions omitted, it's not hard to write):

```python
@request_middleware
async def before(request):
    # just like regular 'request' handler

@response_middleware
async def after(request, response):
    # just like regular 'response' handler

@app.route('/')
@middlewares(before, after)
async def index(request):
    return response.text('okay')
```

pandafeeder:
Hi @pyx ,

Well, I think the only runtime effort added is a `hasattr` check and the benefit is that you don't have to bother writing all those decorators yourself. I think when there're multi middlewares, it will become  a bit more handy and code will be easier to read. 😊

Thanks

r0fls:
@pandafeeder does this fix https://github.com/channelcat/sanic/issues/37?

pyx:
Hi @pandafeeder 

IMHO, with decorators, we pay for what we ask for, while as in this PR, we will be forced to spend extra clock cycle whether we need this feature or not, and IMHO this particular scenario (per-request middlewares), are of small percentage, so it might not be worth it.
As of the burden of all these boilerplate code, like I said, a couple re-usable utility functions should be all we need.

Take, for example, your demonstration of user authentication, if all your routes are like "walled gardens", you basically need globally registered middlewares instead of this, if there is public facing portion and admin/profile part behind user sign-in, then by logic, I would say most of the traffic will be on public facing routes, in either situation, paying for what we need only once in a while (no matter how small the payment is) all the time, might not be a good choice.

I, as a user, have this concern, and am worrying about performance-hit, is because right now there is no clear winner among async web micro frameworks, and there are at least 3 other competing frameworks (I don't want to name names, but it is easy to find out which) emerged, and by some benchmarks, some are already being faster than Sanic.  I think every little bit will help sanic stays longer in this healthy competition.

r0fls:
Can you report the performance before and after this change? Something like `wrk -c 15 -t 100 localhost:8000` before and after.

38elements:
I am sorry.
I think this is not neccesary.
I think this is alias of decorator.
I do not like alias.
I think benefit for adding lines is small.
I think keeping small is better.
I think that  the first method of @pyx is better.

pandafeeder:
Hi @pyx, @38elements 
Got your point, thanks.

Hi @r0fls ,
Sorry, no, it's not a fix for #37 .
For performance comparing:
I made five times of testing for a average count and import sanic directly from repo dir, not install, here is the result:
test command: `wrk -t12 -c400 -d10s http://127.0.0.1:8000`

**before change**:
app with middleware defined using @pyx 's approach:
average Requests/sec:  3599.80

app without middleware:
average Requests/sec:  3922.62

**after change**:
app with middleware:
average Requests/sec: 3511.56

app without middleware:
average Requests/sec: 3753.03

Well, after all, it doesn't seem a good idea. Let me just close this PR.

Thank all of you for your comments.

pyx:
@pandafeeder Thank you for taking your time in writing the PR. I like the way you deal with criticism, nice. Cheers.

PR #717, opened 2017-05-13
Fix "TypeError: not all arguments converted during string formatting"

### LINKED ISSUE #716: IPv6 addresses and netlog: TypeError: not all arguments converted during string formatting
Netlog is trying to log the IP address using request.ip (which is socket.getpeername()), and fails due to differences between returned tuples depending on the address family.

AF_INET returns (host, port), while AF_INET6 a four-tuple: (host, port, flowid, scopeid).
(https://docs.python.org/3/library/socket.html#socket.socket.getpeername)

I've got a PR ready, which simply changes the formatting at https://github.com/channelcat/sanic/blob/0.5.4/sanic/server.py#L204 to use host and port parts of the tuple explicitly.

Did you have any other plans on handling it? Currently it incorrectly logs each request that fails (and that's every IPv6 request) as a 500.

jrocketfingers:
Resolves #716 

socket.getpeername() returns AF_INET6 address family four-tuple, with
flowid and scopeid.

In server's write_response, an exception is raised when an IPv6 client
connects due to four-tuple elements having two unused elements (flowid
and scopeid).

This makes sure that only the first two (host and port) are used in log
string formatting.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -143,7 +143,7 @@ class Request(dict):
     @property
     def ip(self):
         if not hasattr(self, '_ip'):
-            self._ip = self.transport.get_extra_info('peername')
+            self._ip = self.transport.get_extra_info('peername') or (None, None)
         return self._ip
 
     @property
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -143,7 +143,8 @@ class Request(dict):
     @property
     def ip(self):
         if not hasattr(self, '_ip'):
-            self._ip = self.transport.get_extra_info('peername') or (None, None)
+            self._ip = (self.transport.get_extra_info('peername') or
+                        (None, None))
         return self._ip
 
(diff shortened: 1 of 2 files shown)


r0fls:
To fix the above concern you can replace: https://github.com/channelcat/sanic/blob/master/sanic/request.py#L146

with:

```python
self._ip = self.transport.get_extra_info('peername') or (None, None)
```

jrocketfingers:
Missed that, updating right now.

r0fls:
@jrocketfingers sorry looks like we have a pep8 problem now, can you fix that?

jrocketfingers:
On it, just running it through flake (I'm not at my workstation, didn't have the env ready, so it took a while). Thanks for suggesting the preferred split. I was going to opt for a \ and or in the next line.

PR #722, opened 2017-05-15
Add py3*-no-ext test env

messense:


messense:
```
2017-05-15 02:13:39 - (sanic)[ERROR]: Traceback (most recent call last):
  File "/home/travis/build/channelcat/sanic/sanic/app.py", line 471, in handle_request
    response = await response
  File "/home/travis/build/channelcat/sanic/sanic/app.py", line 209, in websocket_handler
    protocol = request.transport.get_protocol()
AttributeError: '_SelectorSocketTransport' object has no attribute 'get_protocol'
```

Interesting.

messense:
@miguelgrinberg Any idea why websocket tests is failing on Python 3.5 without ujson and uvloop?

miguelgrinberg:
@messense Yes. The `Transport` classes in asyncio do not have a `get_protocol()` method. Only the uvloop transports do. Looks like the protocol is available at `transport._protocol`, but this is clearly a private attribute. Maybe use `request.transport.get_protocol()` if it exists, and if not, try `request.transport._protocol`?

messense:
@miguelgrinberg That works, thanks!

sanic/app.py
```diff
@@ -206,7 +206,12 @@ def websocket(self, uri, host=None, strict_slashes=False):
         def response(handler):
             async def websocket_handler(request, *args, **kwargs):
                 request.app = self
-                protocol = request.transport.get_protocol()
+                try:
+                    protocol = request.transport.get_protocol()
```
r0fls: would using `protocol = getattr(request.transport, 'get_protocol()', request.transport._transport)` make more sense then a try/except?

messense: I don't think `getattr(request.transport, 'get_protocol()`)` would work, it will raise `AttributeError: 'Transport' object has no attribute 'get_protocol()'`. It has to be:

```python
get_protocol = getattr(request.transport, 'get_protocol', None)
if get_protocol is not None:
    protocol = get_protocol()
else:
    protocol = request.transport._protocol
```

which is not better.

r0fls: good point...

sanic/request.py
```diff
@@ -64,7 +66,11 @@ def __init__(self, url_bytes, headers, version, method, transport):
     def json(self):
         if self.parsed_json is None:
             try:
-                self.parsed_json = json_loads(self.body)
+                if sys.version_info[:2] == (3, 5) and json_loads is json.loads:
```
r0fls: Instead of doing this, we should define `json_loads` after:

```python
def json_loads(data):
    return json.loads(data.decode())
```

Then we don't have to add the overhead of the version check and the try/except to every json access.

messense:
I think maybe we can drop `tox-travis` and configure multiple test matrix in `.travis.yml` to take advantage of parallel build of speed up tests.

PR #731, opened 2017-05-17
Check that the Authorization headers are actually provided.

jrocketfingers:
`request.token()` fails when the `Authorization` header is not provided. If I understood the intent correctly, token should return None in case no token has been provided. This should cover for that.

~TODO: Add a regression unittest.~

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -79,7 +79,7 @@ class Request(dict):
         :return: token related to request
         """
         auth_header = self.headers.get('Authorization')
-        if 'Token ' in auth_header:
+        if auth_header is not None and 'Token ' in auth_header:
             return auth_header.partition('Token ')[-1]
         else:
             return auth_header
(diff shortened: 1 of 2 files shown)


sanic/request.py
```diff
@@ -79,7 +79,7 @@ def token(self):
         :return: token related to request
         """
         auth_header = self.headers.get('Authorization')
-        if 'Token ' in auth_header:
+        if auth_header is not None and 'Token ' in auth_header:
```
r0fls: Would it make more sense to provide the empty string as the default instead of None? `auth_header = self.headers.get('Authorization', '')`. Then `if 'Token ' in auth_header` will fail rather than raise an exception.

jrocketfingers: I'd prefer it being None, as it's being passed in the else branch. One thing I'd like to check is `request.token is not None`, as it's super clean to read.

The other way to make it return `None`, yet use `''` would be to return an explicit none in the else branch. Not sure how I feel about the readability of the latter, but you guys are the core devs, gonna do it the way you choose.

Issue #730, opened 2017-05-16
Refactoring register_middleware


jrocketfingers:
Is there a particular reason that we're nesting register_middleware within the middleware decorator? https://github.com/channelcat/sanic/blob/master/sanic/app.py#L280

The usecase I have is that I'm doing an `init_auth(app)` which should register all the necessary middlewares, and I have to write things like `app.middleware('request')(middleware)`.

I'd prefer being able to write `app.register_middleware(middleware, attach_to='request')`.

I can make a quick PR if factoring it out makes sense.

38elements:
It is changing the expression method for registering middleware.
Would you tell me other advantages?
Sanic uses many decorator for registering.
I think using the decorator is more consistent.

jrocketfingers:
Can you point out any particular disadvantages? All I'm seeing is gaining a cleaner way to explicitly register middleware.

On the other hand, if we're going to invoke consistency here, I'd like to call upon the ability to explicitly register routes using `app.router.add`, besides the `app.route` decorator.

38elements:
Does `Refactoring register_middleware` mean `replacing app.middleware() to app.register_middleware()`?

jrocketfingers:
No, it means keeping both, just extracting `register_middleware` out one level, to make it available as `app.register_middleware` (currently it's nested within the decorator).

I'll make a quick PR in a couple of minutes so we can discuss on some cold hard code.

38elements:
Thank you for responding.
I thought this means replacing.
But, I think there is not sufficient benefit for adding lines.
I think this is not necessary.
But, others may think there is sufficient benefit.
If you think this is necessary with confident, you send pull request.

jrocketfingers:
I didn't get around to doing it, I've had some issues with making a partial out of the method. I'm hoping to get around to it later this evening.

In any case, I don't think we're even changing the number of lines. It's purely extracting the nested function into a method. Literally everything stays the same. (I guess there's an extra newline to keep it pep8 conforming, if we're to nitpick).

jrocketfingers:
@38elements, could you take a look at #732?

r0fls:
closed with #732

PR #732, opened 2017-05-18
Extract register_middleware into a method.

jrocketfingers:
I finally got to it, so here's an example of the idea suggested in #730.

Handily enough, the example itself is in the blueprints.py.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -272,24 +272,25 @@ class Sanic:
 
         return response
 
+    def register_middleware(self, middleware, attach_to='request'):
+        if attach_to == 'request':
+            self.request_middleware.append(middleware)
+        if attach_to == 'response':
+            self.response_middleware.appendleft(middleware)
+        return middleware
+
     # Decorator
     def middleware(self, middleware_or_request):
         """Decorate and register middleware to be called before a request.
         Can either be called as @app.middleware or @app.middleware('request')
         """
-        def register_middleware(middleware, attach_to='request'):
-            if attach_to == 'request':
-                self.request_middleware.append(middleware)
-            if attach_to == 'response':
-                self.response_middleware.appendleft(middleware)
-            return middleware
 
         # Detect which way this was called, @middleware or @middleware('AT')
         if callable(middleware_or_request):
-            return register_middleware(middleware_or_request)
+            return self.register_middleware(middleware_or_request)
 
         else:
-            return partial(register_middleware,
+            return partial(Sanic.register_middleware, self,
                            attach_to=middleware_or_request)
 
     # Static Files
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -290,7 +290,7 @@ class Sanic:
             return self.register_middleware(middleware_or_request)
 
         else:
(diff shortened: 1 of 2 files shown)


38elements:
I think this change is no effect.
The express only changed.
The change having no benefit should not be merged.

jrocketfingers:
That's exactly why I'm proposing it. It affects nothing, yet gives us a better way to express the intent. Imho, readability is improved, for the price of free.

I can't help but ask, why exactly are you so hostile? Sanic could've been left in the state of it's initial commit*, and just had features plopped on until we've got an organically grown mass.

* haven't read the initial commit, I'm just trying to make a point about not changing anything.

pyx:
https://github.com/channelcat/sanic/pull/732/commits/bf5438d5733c85c72dc092365a9570bdd2531e00#diff-c4444a35dbd5e37ee480b0e8888e0880L292

You could use
```python
partial(self.register_middleware, attach_to=middleware_or_request)
```
instead of `partial(Sanic.register_middleware, self, ...`

By not hard-coding the class name, the former works with subclasses as well.

A quick example:
```python
from functools import partial, wraps


class Deco:
    def log(self, func=None, *, before=True, after=True):
        if func is None:
            return partial(self.log, before=before, after=after)

        @wraps(func)
        def traced(*args, **kwargs):
            if before:
                print('calling <', func.__name__, '> with:', args, kwargs)
            result = func(*args, **kwargs)
            if after:
                print('result:', result)
            return result

        return traced


deco = Deco()


@deco.log
def total(*numbers):
    return sum(numbers)


@deco.log(after=False)
def output(*args):
    print(*args)


@deco.log(before=False, after=False)
def no_trace():
    print('no_trace called')


total(1, 2, 3, 4, 5)
print('-' * 20)
output("Hello,", "world!")
print('-' * 20)
no_trace()
```

output:
```
calling < total > with: (1, 2, 3, 4, 5) {}
result: 15
--------------------
calling < output > with: ('Hello,', 'world!') {}
Hello, world!
--------------------
no_trace called
```

jrocketfingers:
@pyx, not sure how's the example related, since I'm trying to do the opposite [1], but yeah, thanks for pointing out the hardcoded `Sanic.register_middleware`!

That was a cop out early this morning, once the variant you suggested failed the tests. Now that I think about it, I think it was something along the lines of just putting `register_middleware` one indent back, without adding the `self` argument ;p

[1] we already have bound decorators, now I need to make that decorator return a partial to another bound method

pyx:
@jrocketfingers Ha, it's my fault, I was banging out the code in a rush, I should have the inner function `traced` defined as a method of `Deco`, then it could be more in line with you code.

Yes, your latest change is what I mean by "works with subclasses."

r0fls:
For the record, I think it's best to avoid nested functions that provide no benefit. They should be reserved for function factories, etc...

jrocketfingers:
@r0fls, hence the PR, it was clear there was no actual benefit from having it nested, plus the change provides a clean way to register middleware (I utterly hate double parentheses

PR #733, opened 2017-05-18
Add file_stream response handler

ashleysommer:
For streaming large static files
Like `file()` but breaks the file into chunks and sends it with a `StreamingHTTPResponse`
Chunk size is configurable, but defaults to 4k, this seemed to be the sweet spot in my testing.
Also supports ContentRange same as `file()` does.


38elements:
I think it is necessary to add test for `file_stream()`.

ashleysommer:
There is a test in `try_everything.py`
Here: https://github.com/channelcat/sanic/pull/733/commits/181977ad4ed8dac3443f5906a44883e035971f2c#diff-b9137c4a6c19839608462a8314657683R68

It just looked what `file()` does and did the same in regards to tests.

Should he have a better test or more tests?

38elements:
Is `try_everything.py`  example?
A test exists in [tests](https://github.com/channelcat/sanic/tree/master/tests) directory.

ashleysommer:
Oh. I see.
There is no test for `file()` in the `tests/` directory either.
That is why I didn't add a test for `file_stream()` in the `tests/` directory.
Should I add a test for both `file_stream()` and `file()`?

38elements:
IMHO,
I think that it is only necessary to test your deliverables.

ashleysommer:
Added real tests for `file()` and `file_stream()` response helpers.

38elements:
If `file_stream()` is able to substitute `file()`, I think it would be better to replace `file()` to `file_stream()` in future.

r0fls:
I would like to see performance comparisons for smaller files that don't require streaming before we consider removing `file()`.

ashleysommer:
>Can we move all of the tests to test_response.py? They're both response types so it makes sense to consolidate them there.
Ok, Done.

>If file_stream() is able to substitute file(), I think it would be better to replace file() to file_stream() in future.
If file() will be replaced to file_stream(), it is need to post document about replacing.

No, `file_stream()` is _not_ a direct replacement for `file()` and is _not_ intended to replace `file()` in any way.

In my other PR https://github.com/channelcat/sanic/pull/734 the `static()` helper is modified to have the _option_ of using `file_stream()` (but only for large files), but it is _off_ by default.

> I would like to see performance comparisons for smaller files that don't require streaming before we consider removing file().
As would be expected, performance is _decreased_ for small files that do not require streaming.

PR #734, opened 2017-05-18
Add option to static helper to use streaming for large files.

ashleysommer:
Depends on https://github.com/channelcat/sanic/pull/733

Adds an option to the `static()` helper to switch on streaming for large files.
By default uses a 1M threshold.
ie. if the static file to serve is >= 1M it will stream the file.
This threshold value is configurable by passing an int instead of a bool to `stream_large_files` parameter of `static()`.

### DIFF
--- a/sanic/static.py
+++ b/sanic/static.py
@@ -13,11 +13,12 @@ from sanic.exceptions import (
     InvalidUsage,
 )
 from sanic.handlers import ContentRangeHandler
-from sanic.response import file, HTTPResponse
+from sanic.response import file, file_stream, HTTPResponse
 
 
 def register(app, uri, file_or_directory, pattern,
-             use_modified_since, use_content_range):
+             use_modified_since, use_content_range,
+             stream_large_files):
     # TODO: Though sanic is not a file server, I feel like we should at least
     #       make a good effort here.  Modified-since is nice, but we could
     #       also look into etags, expires, and caching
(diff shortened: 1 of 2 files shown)


ashleysommer:
Note, this PR includes the commits from https://github.com/channelcat/sanic/pull/733

Look at just commit https://github.com/channelcat/sanic/pull/734/commits/592364eb94f676b9b85dba68cfbab6d50b3b5a3c for the changes pertaining to this PR.

sanic/static.py
```diff
@@ -93,6 +98,16 @@ def register(app, uri, file_or_directory, pattern,
...
+                if stream_large_files:
+                    if isinstance(stream_large_files, int):
+                        threshold = stream_large_files
+                    else:
+                        threshold = 1024*1000
+
+                    if not stats:
+                        stats = await stat(file_path)
+                    if stats.st_size >= threshold:
+                        return await file_stream(file_path, headers=headers, _range=_range)
```
jrocketfingers: Flake8 fails here, line's too long.

ashleysommer:
Fixed flake8 line length inconsistency.

sanic/static.py
ashleysommer: Thanks, fixed.

ashleysommer:
rebased this PR to master after https://github.com/channelcat/sanic/pull/733 was merged

sanic/static.py
```diff
@@ -93,6 +98,17 @@ def register(app, uri, file_or_directory, pattern,
                     headers=headers,
                     content_type=guess_type(file_path)[0] or 'text/plain')
             else:
+                if stream_large_files:
```
r0fls: Can you provide a benchmark before/after this change, particularly when this feature is disabled? `wrk -t 12 -c 100 -d 20 localhost:8000/some_file.txt` should be sufficient.

seemethere: I think since this is just affecting static files doesn't necessarily this should affect the side of performance we necessarily care about

ashleysommer: Thats correct, plus it is disabled by default so existing static files performance will not have changed at all.
Are simply interested in seeing whether the `if stream_large_files:` conditional is adding any slow-down over the current `static` implementation?
I can still do the benchmark if you think it is required.

PR #738, opened 2017-05-19
Add Code of Conduct

messense:
https://en.wikipedia.org/wiki/Code_of_conduct

Add a code of conduct would help form a friendly, safe and welcoming community.

Others project's code of conduct for reference:

Django: https://www.djangoproject.com/conduct/reporting/
Python: https://www.python.org/community/diversity/
Rust: https://www.rust-lang.org/en-US/conduct.html

### DIFF
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
+nationality, personal appearance, race, religion, or sexual identity and
+orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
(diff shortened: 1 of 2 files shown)


CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at [INSERT EMAIL ADDRESS]. All
```
messense: Not sure how this `INSERT EMAIL ADDRESS` should be. :-)

r0fls: we could consider creating a sanic-maintainers@gmail.com email and forwarding that to interested members?

pyx: If you are using something like google groups, there will be an email for general submission.  Maybe you can use that.

r0fls: Hmm if we are using that i'm not aware... i wonder if there is a way to forward to the github group? Then again, that's probably not the effect we'd want anyway.

r0fls: Then again, I think it should allow for a certain level of anonymity. So it might be best to keep the email list small. Hopefully this wouldn't be used much if at all ✌️ ☮️

messense: So let's go with  sanic-maintainers@gmail.com ?

Issue #739, opened 2017-05-19
Any plan to include 401?


pyx:
I am working on a simple authentication extension, I need to throw an exception of `401 unauthorized`, what is the proper way to do that?

Should I subclass `sanic.exceptions.SanicException` as:
```python
class Unauthorized(SanicException):
    status_code = 401
```
in my package, or is it a good idea to include that seemingly common HTTP status code in Sanic?

Further, to the best of my knowledge, there is no `abort` function like in Flask and bottle in Sanic, should Sanic pre-define all possible common status codes?  Maybe an `abort` function is better than this.

Thanks.

pyx:
I can draft a PR to either implement those status code as individual exceptions, or a generic `abort` function which raises specified status code with an `SanicException`.

In latter case, we will need a pre-define mapping for existing exceptions, to speed up the look up[1] time,  It is actually not going every well with the current implementation.

[1]: we don't want `abort(404)` to raise `SanicException(status_code=404)` instead of `NotFound`.

pyx:
Thinking more about that, concrete exception for each HTTP error code is not as flexible as the way Flask and bottle using, they both can work on the error code as plain old int, especially the error handling part [1] [2],  which allows user to use and catch any (im)possbie codes without restriction.

[1]: Flask's `app.errorhandler`, works on an interger, http://flask.pocoo.org/docs/dev/quickstart/#redirects-and-errors
[2]: Bottle's decorator `error`, also needs only an integer, https://bottlepy.org/docs/dev/tutorial.html#error-pages

r0fls:
@pyx does this seem like an apt solution to you? https://github.com/channelcat/sanic/pull/740/files

Then you can do:

```python
from sanic import Sanic
from sanic import response
from sanic.exceptions import SanicException

app = Sanic(__name__)

@app.route("/")
async def test(request):
    raise SanicException(status_code=401)
    return response.json({"test": True}, status=304, headers=headers)


if __name__ == '__main__':
    app.run(host="0.0.0.0", port=8000)
```

pyx:
@r0fls Thanks for the quick PR.

Raising a `SanicException` with status code is what I am going to use as of right now.

But the part that worries me is style and consistency between Sanic, extensions and end user code.

Let's say that, right now, I am using `SanicException` with 401 in my extension, to catch that, users of my extension will have to use a generic error handler, and tests status code inside that.  To make it more elegant, I can provide an `Unauthorized` exception, so that my users can use:
```python
[...]

PR #740, opened 2017-05-19
add abort function

r0fls:
Address https://github.com/channelcat/sanic/issues/739

### DIFF
--- a/docs/sanic/exceptions.md
+++ b/docs/sanic/exceptions.md
@@ -17,7 +17,7 @@ def i_am_ready_to_die(request):
 	raise ServerError("Something bad happened", status_code=500)
 ```
 
-You can also use the `abort` function with the correct status code:
+You can also use the `abort` function with the appropriate status code:
 
 ```python
 from sanic.exceptions import abort
--- a/docs/sanic/exceptions.md
+++ b/docs/sanic/exceptions.md
@@ -17,6 +17,19 @@ def i_am_ready_to_die(request):
 	raise ServerError("Something bad happened", status_code=500)
 ```
 
+You can also use the `abort` function with the correct status code:
+
+```python
+from sanic.exceptions import abort
+from sanic.response import text
+
+@app.route('/youshallnotpass')
+def no_no(request):
+        abort(401)
+        # this won't happen
+        text("OK")
+```
+
 ## Handling exceptions
 
 To override Sanic's default handling of an exception, the `@app.exception`
--- a/sanic/exceptions.py
+++ b/sanic/exceptions.py
@@ -117,6 +117,20 @@ INTERNAL_SERVER_ERROR_HTML = '''
 '''
 
 
+SANIC_EXCEPTIONS = {}
+
+
+def add_status_code(code):
+    """
+    Decorator used for adding exceptions to SANIC_EXCEPTIONS.
+    """
+    def class_decorator(cls):
+        cls.status_code = code
+        SANIC_EXCEPTIONS[code] = cls
+        return cls
+    return class_decorator
+
+
 class SanicException(Exception):
 
     def __init__(self, message, status_code=None):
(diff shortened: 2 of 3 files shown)


pyx:
Should we have customizable message in `abort` then?

And what should we do if we raise something pre-defined, such as 404:
`abort(404)` does not raise `NotFound`, this is the inconsistency I talked about in #739 

Thanks for the quick action anyway.

pyx:
To address that, we can look up the exception by the number, and use generic one if it is not defined, that's why I mentioned the "fast look up" in #739.
Then, it's actually more maintenance required then integer-based solution.

r0fls:
@pyx I see what you mean. I'm going to update the PR so that it raises that. However can you elaborate more on what you mean by an integer based solution?

pyx:
@r0fls sure, the preferred API is something like (*please be noted, this is my own personal view*):

```python
@app.route('/off-limit')
async def area51(request):
    abort(403)
    # or raise HTTPError(403)

@app.route('/wiki-leak')
async def top_secret(request):
    abort(410, "no longer exist')
    # or raise HTTPError(410, "no longer exist")

@app.error(403)
async def error403(error):
    return response.json({'error_code': error.status_code, 'message': error.message})
```

This way, all we need in sanic, is actual a mapping like so (very similar to what we already have):
```python
class HTTPError(Exception):  # or whatever base Exception
    ...

error_codes = {
...
403: "Forbidden",
404: "Not found",
...
410: "Gone",
}
```

And the default error handling is matching on the number, and look up the error message if not provided in the HTTPError exception instance or `abort` call.  No more defining each individual exceptions, no more testing in `abort` (to address the problem I mentioned above), any future addition of HTTP error code (albeit the situation is rare) will be handled by merely one more line in `error_codes`, etc.

That's why I prefer this design, and that's why I believe both Flask and Bottle were designed this way.

pyx:
Correction:
The signature of error handlers should be:
```python
async def error_handler(request, error):
   ...
```

r0fls:
@pyx yes I agree that would probably be better. However since it's a breaking change I'm just going to do this for now. We should begin a list of breaking changes we would consider introducing with the 0.x -> 1.x update though, e.g. 3.5 support.

pyx:
@r0fls nicely done, thanks.

```python
SANIC_EXCEPTIONS = {}

def status_code(code):
    def class_decorator(cls):
        cls.status_code = code
        SANIC_EXCEPTIONS[code] = cls
        return cls
    return class_decorator

...

@status_code(404)
class NotFound(SanicException):
    pass

@status_code(400)
class InvalidUsage(SanicException):
    pass

...
```

This is my way to reduce duplication (not that it should be written this way).

But in either way, it cannot deal with exceptions sharing status code, such as `NotFound` and `FileNotFound`. I have no idea how to make it work now.

r0fls:
@pyx we have to choose the more common exception in that case. If someone wants to raise a specific exception (e.g. URLBuildError or FileNotFound) they would have to do that explicitly: `raise FileNotFound("File not found")`. I added the suggested refactor, it's more lines of code, but I am happier to have `SANIC_EXCEPTIONS` at the top...

r0fls:
It's really quite annoying that I can't define the `SANIC_EXCEPTIONS = {"foo": Bar}` before class `Bar` is defined 👿

pyx:
@r0fls

Yes, seems it's used locally, maybe a name with lower case that begins with `_` .

The decorator does two things, 1. register the decorated class in SANIC_EXCEPTIONS, 2. add said status code as class attribute.
The first avoid the duplication of class name in defining SANIC_EXCEPTION, the second remove the duplication in defining `status_code` in class body so that you don't have to repeat the number which is error-prone. (so I use `pass` in class body if it only need `status_code`).

You sort of can define `SANIC_EXCEPTIONS` at the beginning. Put `SANIC_EXCEPTIONS` into a function scope, and call that function at the end of the file, assign the returned `SANIC_EXCEPTIONS` as `SANIC_EXCEPTIONS` in global scope. Please, don't try this at home.  :D

pyx:
Excellent, just curious, why do we repeat the status_code in class body...

r0fls:
thanks for the review!

PR #742, opened 2017-05-21
changes required for unix socket support

r0fls:


syrusakbary:
Could be possible to publish a new version to PyPI with this change? (having issues running sanic with sockets now)

PR #744, opened 2017-05-22
Update config file loader docs

algtmatt:
I think this documentation bit is wrong on from_file()

If you look at the 'tests/test_config.py' test_load_from_file() test, it is using 'app.config.from_pyfile(config_file.name)'

Updating my config loader to from_pyfile is working.


r0fls:
Thanks for fixing that!

Issue #746, opened 2017-05-24
Allow Websocket max_size limit


InvaderGir1:
I saw that sanic uses the websockets package meaning this protocol:
[https://github.com/aaugustin/webso)ckets/blob/master/websockets/protocol.py](url)
Will it be possible to expose the max_size limit, queue etc, to sanic's WebSocketCommonProtocol? (I currently did a change allowing arguments to be passed from sanic's to the WebsocketCommonPortocol created in line 65. (adding the max_size I want for my sockets)

I would be more then glad to add it and create a pull request if required.

Thanks!

messense:
I guess you can make some config items like `WEBSOCKET_MAX_SIZE`, `WEBSOCKET_MAX_QUEUES` etc. and pass it into `WebSocketCommonProtocol`, so people can change it by [configuration](http://sanic.readthedocs.io/en/latest/sanic/config.html).

messense:
I created a pull request for this #748

seemethere:
Closed with #748

PR #748, opened 2017-05-26
Add websocket max_size and max_queue configuration

messense:
Addresses #746

### DIFF
--- a/sanic/websocket.py
+++ b/sanic/websocket.py
@@ -6,9 +6,12 @@ from websockets import ConnectionClosed  # noqa
 
 
 class WebSocketProtocol(HttpProtocol):
-    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


sanic/websocket.py
```diff
@@ -6,9 +6,12 @@
 
 
 class WebSocketProtocol(HttpProtocol):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, websocket_max_size=None,
```
r0fls: We could also do the below here instead but I'm fine with this as is

```python
self.websocket_max_size = kwargs.get('websocket_max_size')
self.websocket_max_queue = kwargs.get('websocket_max_queue')
```

messense: I thought about this too but choose to be more consistent with `HttpProtocol`.

PR #756, opened 2017-05-28
fixes #755 fragmented headers

### LINKED ISSUE #755: fragmented headers
I stumbled upon an error where long header values (OAuth2 tokens in my case) may lead to fragmentation of the header names. This error was only reproducible with a remote server, not with a local dev server.

The error happens here: [server.py#L167](https://github.com/channelcat/sanic/blob/master/sanic/server.py#L167). The `value` param seems to be `None` if the header is not fully parsed. An empty header value is represented by `b''`.

Adding a debug log statement `log.debug('on_header: %s %s', name, value)` to the `on_header` function yield the following:

```
on_header: b'Conn' None
on_header: b'ection' b'close'
```

or 

```
on_header: b'Connection' None
on_header: b'' b'close'
```

I fixed this by adding the header only if the `value` is not `None` and saving the `name` values to a new class variable (`_header_fragment`) on the `HttpProtocol` instance.

I will open a pull request.

qwesda:


messense:
Could you please add a test case?

qwesda:
I can not figure out a how to add a test for this. I have the following code that works if I run it directly, but I can't get it working in tox. I get an `BrokenPipeError: [Errno 32] Broken pipe` Exception I guess tox kills the test before it completes. Any idea how I can get this working?

```python
HOST = 'localhost'
PORT = 42199
request_bytes = b"""POST /path HTTP/1.1
Host: localhost:42199
Content-Type: text/plain; charset=utf-8
X-Empty-Header: 
Connection: close
Content-Length: 10

1234567890

"""


app = Sanic('test_fragmented_parsing')
app.config.LOGO = None


@app.route('/path', methods={'POST'})
async def handler(request):
    assert request.headers['host'] == '%s:%s' % (HOST, PORT)
    assert request.headers['connection'] == 'close'
    assert request.headers['content-length'] == '10'
    assert request.headers['content-type'] == 'text/plain; charset=utf-8'
    assert request.headers['x-empty-header'] == ''

    assert request.body == b'1234567890'

    return text('OK', status=200)


def test_fragmented_parsing():
    async def send_fragmented_request(sanic, loop):
        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        sock.connect((HOST, PORT))

        for i in range(1, len(request_bytes)):
            sock.send(request_bytes[i-1:i])

            await asyncio.sleep(.01)

        sock.close()

    app.run(host=HOST, debug=True, port=PORT)
```

qwesda:
Any help with this?

messense:
I am not sure how to fix it.

cc @r0fls @seemethere

qwesda:
Since nobody seems to help with a test for this, can this please just be merged? 

The bug will hit people when they move to production and it is pretty annoying to track down. 
The changes in this PR are pretty simple to check for correctness. 
Also it doesn't seem to break anything.

messense:
I am ok with merging this.

qwesda:
Glad to hear this. 

Somebody from the core team should probably open an issue for introducing tests that can catch bugs like this. #784 seems also related to this and would only be caught by running tests between a server/client and simulating a slow internet connection.

youknowone:
Because this bug is critical to the service reliability, I hope to see next release of Sanic in (very) near future.

adamserafini:
I think we are also experiencing this issue production, symptoms seem similar - we have long headers and unhandled 400s.

We will test latest version of Master.. but would be nice to see an official release with this change merged in.. let me know if anything I can do to help!

PR #762, opened 2017-05-31
only single repetition

stopspazzing:


sanic/router.py
```diff
@@ -163,7 +163,7 @@ def add_parameter(match):
             parameters.append(parameter)
 
             # Mark the whole route as unhashable if it has the hash key in it
-            if re.search(r'(^|[^^]){1}/', pattern):
+            if re.search(r'(^|[^^]){0}/', pattern):
```
r0fls: hmm what's going on here, shouldn't the pattern come first? https://docs.python.org/3/library/re.html#re.search

stopspazzing: LOL, you are right, guess found a bug by accident. Let me update it with correct search code.

messense: I think you might be confused by the `pattern` variable name? The first argument is a pattern, the second is the string which named `pattern`.

Issue #771, opened 2017-06-04
Test client does not support synchronous response.json


jrocketfingers:
As per https://github.com/channelcat/sanic/blob/master/sanic/testing.py#L28 we can only go for `response.text`. For the sake of consistency, should we await json as well/convert to json on the spot?

Since Sanic errs whenever the test client runs in a coroutine, having this awaited for us is only possible within the test client. Of course, we can always do something along the lines of `json.loads(response.text)`, but, as I already said, for the sake of consistency, we might want to expose both.

As far as I've understood, the consensus is that the speed is not crucial for the testing utilities as well.

Can anyone weigh in on this?

yunstanford:
yeah, exactly.

jrocketfingers:
Resolved in #773

PR #773, opened 2017-06-05
Testing: store JSON response in local request

mbatchkarov:
Hi, I am testing a JSON API and I noticed that the provided test client conveniently replaces the `.text` property of the response object with an actual text, but does not do the same for the `.json` property. This makes tests harder to read because `response.json()` needs to be awaited in each test.

I'm submitting a test that demonstrate the issue and the least invasive fix that I could think of. Let me know if that's acceptable. 

PS I've not added any error handling yet- I'm waiting to see if you think this fix has legs before putting any more work into it.

### DIFF
--- a/sanic/testing.py
+++ b/sanic/testing.py
@@ -1,4 +1,5 @@
 import traceback
+from json import JSONDecodeError
 
 from sanic.log import log
 
@@ -26,9 +27,14 @@ class SanicTestClient:
                     session, method.lower())(url, *args, **kwargs) as response:
                 try:
                     response.text = await response.text()
-                    response.json = await response.json()
                 except UnicodeDecodeError as e:
                     response.text = None
+
+                try:
+                    response.json = await response.json()
+                except (JSONDecodeError, UnicodeDecodeError):
+                    response.json = None
+
                 response.body = await response.read()
                 return response
 
--- a/sanic/testing.py
+++ b/sanic/testing.py
@@ -26,6 +26,7 @@ class SanicTestClient:
                     session, method.lower())(url, *args, **kwargs) as response:
                 try:
                     response.text = await response.text()
+                    response.json = await response.json()
                 except UnicodeDecodeError as e:
                     response.text = None
                 response.body = await response.read()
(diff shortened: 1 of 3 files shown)


sanic/testing.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, app):
                     session, method.lower())(url, *args, **kwargs) as response:
                 try:
                     response.text = await response.text()
+                    response.json = await response.json()
```
messense: This will break any code that uses `await response.json()` previously. It's a breaking change.

messense:
To get the json result, I think you can use `json.loads(response.text)` to avoid calling `await`?

sanic/testing.py
seemethere: What kind of code would do that with the response beforehand?

jrocketfingers: @messense - the code can't really use await response.json() when using the test_client since it'll complain that you can't run the client while another loop is running. So I don't think no one is using it at this point.

#771 - related.

jrocketfingers:
I'm doing that right now, and yes, it works, but we might want to consider adding some consistency if we ever wanna grow a stable API. Other than that `json.loads(respons.text)` is the best way for now.

mbatchkarov:
Re getting a consistent API in the future: some libraries add an underscore suffix for values that are derived or cached, e.g. `response.json` is a coroutine and `response.json_` is the the equivalent of `json.loads(await response.json)`. 

I am not clear if the prevailing opinion is to close this PR or to clean up and merge?

jrocketfingers:
I'd say clean up and merge.
I'd also vote for response.json not being a coroutine (so replacing it in the same way response.text is now), unless we can enable running sanic test_client from a coroutine.

sanic/testing.py
seemethere: So you need to have a try catch around this that will catch the `JsonDecodeError` that's causing all of the CI tests to fail.

mbatchkarov:
Great. I'll clean up shortly

jrocketfingers:
Unfortunately there are still problems with this. `aiohttp` raises exceptions that are not covered in that try-except block.

Namely, `ClientResponseError` with a message `Attempt to decode JSON with unexpected mimetype: text/plain; charset=utf-8`. This happens because I'm returning a text response.

We'll need to tweak the tests to cover this.

PR #776, opened 2017-06-06
Add Request.cancel_stream()

38elements:


r0fls:
@38elements is this not needed?

r0fls:
It appears useful to me.

38elements:
@r0fls 
I am sorry.
If you think this is necessary,
Would you please implement this and send a pull request and merge.
I left this community at #714.
I only temporary return.
There is the problem which I refered at #714.
I leave this community.
Good luck!

38elements:
@r0fls 
I am sorry.
Since it is able to use `request.transport.close()`,
this change is unnecessary.

38elements:
@r0fls
I am sorry.
Since it is able to use request.transport.close(),
this change is unnecessary.

```
@app.post('/cancel_stream', stream=True)
async def cancel_stream(request):
    if condition:
        request.transport.close()
    result = ''
    while True:
        if body is None:
            break
        result += body.decode('utf-8')
    return text(result)
```

I left at #714.
I only temporary return.
There is a problem which I refered at #714.
I leave this community.

r0fls:
ok @38elements thanks for reporting your findings

Issue #779, opened 2017-06-06
Sanic could be MutableMapping ?


yunstanford:
The sanic class could be MutableMapping. it's very easy to add any application specific stuffs there.

app["session"] = ClientSession()
app["db"] = MongoDb()

seemethere:
I'm not opposed to it.

People are already doing something similar to this though with attributes:

```python
app.session = ClientSession()
```

yunstanford:
yes, it works with attributes, but it may overwrite default/builtin attributes by mistake.

PR #780, opened 2017-06-07
The sanic class could be MutableMapping.

yunstanford:
The sanic class could be MutableMapping. it's very easy to add any application specific stuffs there.

```
app["session"] = ClientSession()
app["db"] = MongoDb()
```
We can do something similar to this though with attributes, but it may overwrite default/builtin attributes by mistake.


seemethere:
Relates to #779 

I can see how this would be appealing but why not just set an attribute in the base class called `extensions`?

I think abstracting regular objects to act like "special" objects like mappings etc. is an anti-pattern and we really shouldn't be doing that with the base Sanic app.

Example of what could be an alternative:
```
app.extensions = {}
app.extensions['db'] = SomeDBDriver()
```

What do you think @r0fls?

r0fls:
I agree this could easily be accomplished by setting that attribute. The MutableMapping doesn't appear to add any desirable functionality or performance benefits.

seemethere:
I'd agree.

I think we should leave this to users to define and not necessarily something that has to be part of core.

I'm going to close this PR along with it's corresponding issue.

yunstanford:
i see you guys point, but i don't think those are "special" objects, they are just customized objects, application specifically. ideally, it's better to differentiate builtin attributes and "customized attributes", or even don't setattr for base app. 

i know you guys follow a "Flask" like pattern, but we did overwrite builtin attribute before by mistake. that's why i have this opinion.

Also, that's how aiohttp work, and i feel better.

But, if you guys have strong opinion to oppose it, that' fine.

PR #783, opened 2017-06-08
add content_type property in request

yunstanford:
small change but convenient to have content_type property.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -8,7 +8,7 @@ import pytest
 from sanic import Sanic
 from sanic.exceptions import ServerError
 from sanic.response import json, text
-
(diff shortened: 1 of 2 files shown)


tests/test_requests.py
```diff
@@ -192,6 +192,23 @@ def test_token():
     assert request.token is None
 
 
+def test_content_type():
+    app = Sanic('test_content_type')
+
+    @app.route('/')
+    async def handler(request):
+        return text('OK')
```
seemethere: Could you also add an assert here as well? I think it'd be more useful to have the assert in the route handler than it would be outside of it.

yunstanford: sure. i will update

Issue #759, opened 2017-05-30
HttpParserCallbackError on NotFound (404) using request stream=True


ubergarm:
The new request `stream=True` feature is great!

However, it seems to break `sanic.exeptions.NotFound` (404 errors) in my test.

To reproduce (modified from [examples/request_stream](https://github.com/channelcat/sanic/blob/master/examples/request_stream/server.py)):

## Test Code
```python3
from sanic import Sanic
from sanic.response import stream

app = Sanic(__name__)

@app.post('/stream', stream=True)
async def handler(request):
    async def streaming(response):
        while True:
            body = await request.stream.get()
            if body is None:
                break
            body = body.decode('utf-8').replace('1', 'A')
            response.write(body)
    return stream(streaming)

if __name__ == '__main__':
    app.run(host='0.0.0.0', port=8000)
``` 

## Test 404
```bash
http -f POST localhost:8000/invalidpath foo:"testing123"
$ http -f POST localhost:8000/invalidpath foo:"testing123"
HTTP/1.1 400 Bad Request
Connection: close
Content-Length: 18
Content-Type: text/plain; charset=utf-8

Error: Bad Request
```

Log
```
2017-05-30 19:26:22 - (sanic)[ERROR]: Traceback (most recent call last):
  File "httptools/parser/parser.pyx", line 270, in httptools.parser.parser.cb_on_headers_complete (httptools/parser/parser.c:4402)
  File "httptools/parser/parser.pyx", line 123, in httptools.parser.parser.HttpParser._on_headers_complete (httptools/parser/parser.c:2079)
  File "/usr/lib/python3.6/site-packages/sanic/server.py", line 184, in on_headers_complete
    self.request)
  File "/usr/lib/python3.6/site-packages/sanic/router.py", line 354, in is_stream_handler
    handler = self.get(request)[0]
  File "/usr/lib/python3.6/site-packages/sanic/router.py", line 290, in get
    return self._get(request.path, request.method, '')
  File "/usr/lib/python3.6/site-packages/sanic/router.py", line 339, in _get
    raise NotFound('Requested URL {} not found'.format(url))
sanic.exceptions.NotFound: Requested URL /invalidpath not found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/sanic/server.py", line 159, in data_received
    self.parser.feed_data(data)
[...]

PR #786, opened 2017-06-09
Handle stream 404

yunstanford:
This PR is for fixing https://github.com/channelcat/sanic/issues/759

The root cause is that it calls `self.router.is_stream_handler(self.request)` in on_headers_complete, inside is_stream_handler, it tries to get the handler again (https://github.com/channelcat/sanic/blob/master/sanic/router.py#L354), which will raise another exception during handling the NotFound exception, because it cannot find matched handler and will raise NotFound again, and errors out in https://github.com/channelcat/sanic/blob/master/sanic/router.py#L354.

if the exception is NotFound, we should make is_stream_handler returns False directly, instead of raising another NotFound.


yunstanford:
seems test_gunicorn worker is failing, but i didn't make any change around that

yunstanford:
rebuild fixes the test_gunicorn_work test. looks like test_gunicorn_worker test flaky

PR #789, opened 2017-06-11
Coverage report

yunstanford:
Add correct coverage report.

current is 83%. you can see missing lines in the report.


tests/test_worker.py
```diff
@@ -11,7 +11,7 @@
 def gunicorn_worker():
     command = 'gunicorn --bind 127.0.0.1:1337 --worker-class sanic.worker.GunicornWorker examples.simple_server:app'
     worker = subprocess.Popen(shlex.split(command))
-    time.sleep(1)
+    time.sleep(3)
```
r0fls: why is this needed?

PR #795, opened 2017-06-15
Prevent `run` from overriding logging config set in constructor

ekampf:
When creating the `Sanic` instance I provide it with a customized `log_config`.
Calling `run` overrides these settings unless I provide it *again* with the same `log_config`.
This is confusing and error prone. `run` shouldnt override configurations set in the `Sanic` constructor...

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -567,6 +567,7 @@ class Sanic:
             host, port = host or "127.0.0.1", port or 8000
 
         if log_config:
+            self.log_config = log_config
             logging.config.dictConfig(log_config)
         if protocol is None:
             protocol = (WebSocketProtocol if self.websocket_enabled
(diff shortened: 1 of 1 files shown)


sanic/app.py
```diff
@@ -545,7 +545,7 @@ def test_client(self):
     def run(self, host=None, port=None, debug=False, ssl=None,
             sock=None, workers=1, protocol=None,
             backlog=100, stop_event=None, register_sys_signals=True,
-            log_config=LOGGING):
+            log_config=None):
```
r0fls: Why is this needed? This changes the current behavior and no longer use the `LOGGING` config from sanic/config.py by default.

r0fls: actually nevermind, derp, i see what you did there

r0fls:
verified this works and agree it's an improvement 👍

Issue #713, opened 2017-05-12
worker recycling when memory leaks


hsvlz:
Hi,

in the case of small and slow memory leaks, it is usually beneficial to be able to recycling workers. Gunicorn proposes the max_requests flag to set the maximum number of requests a worker will process before restarting. 
Is there something similar that could be put in place within the sanic server, or do you just advise to use gunicorn with the sanic worker instead?

Thank you and keep up the good work!

antinescience:
Gunicorn does it by keeping track of the number of requests on a specific worker and then, when it exceeds that limit, basically kills the worker off and brings it back up, as seen [here](https://github.com/benoitc/gunicorn/blob/2b839ca14437c61780d6eaaa8b24ad31e021f2c6/gunicorn/workers/sync.py#L173).

At least looking at `server.py`, you'd probably have to muck around with running something in `before_stop`. I'm not 100% clear on what that'd look like; might try to take a stab at this this week, as a recycle trigger (either by time, like IIS does, or max requests like Gunicorn) could be incredibly handy here.

yunstanford:
Yeah, i am working on that part. it's very weak if it cannot hook up correct with gunicorn recycling feature. it's very import. I will send a PR out when i am done. i am also thinking a little refactoring work needs to be done for server.py module.

PR #800, opened 2017-06-17
Recycling gunicorn worker

yunstanford:
https://github.com/channelcat/sanic/issues/713

1. add recycling feature, it's important when memory leaks.
2. add unit tests for gunicorn worker
3. trigger_events(before_start, loop) has been called twice when using gunicorn worker. fixed that.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -432,8 +432,6 @@ def serve(host, port, request_handler, error_handler, before_start=None,
     if debug:
         loop.set_debug(debug)
 
-    trigger_events(before_start, loop)
-
     connections = connections if connections is not None else set()
     server = partial(
         protocol,
@@ -463,12 +461,15 @@ def serve(host, port, request_handler, error_handler, before_start=None,
         sock=sock,
         backlog=backlog
     )
+
     # Instead of pulling time at the end of every request,
     # pull it once per minute
     loop.call_soon(partial(update_current_time, loop))
 
     if run_async:
         return server_coroutine
+
+    trigger_events(before_start, loop)
 
     try:
         http_server = loop.run_until_complete(server_coroutine)
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -74,8 +74,8 @@ class HttpProtocol(asyncio.Protocol):
     def __init__(self, *, loop, request_handler, error_handler,
                  signal=Signal(), connections=set(), request_timeout=60,
                  request_max_size=None, request_class=None, has_log=True,
-                 keep_alive=True, is_request_stream=False, router=None, state=None,
-                 **kwargs):
+                 keep_alive=True, is_request_stream=False, router=None,
+                 state=None, **kwargs):
         self.loop = loop
         self.transport = None
         self.request = None
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -157,6 +157,7 @@ class HttpProtocol(asyncio.Protocol):
             self.headers = []
             self.parser = HttpRequestParser(self)
 
+        # requests count
         self.state['requests_count'] = self.state['requests_count'] + 1
 
         # Parse request chunk or close connection
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -74,7 +74,7 @@ class HttpProtocol(asyncio.Protocol):
     def __init__(self, *, loop, request_handler, error_handler,
                  signal=Signal(), connections=set(), request_timeout=60,
                  request_max_size=None, request_class=None, has_log=True,
-                 keep_alive=True, is_request_stream=False, router=None,
+                 keep_alive=True, is_request_stream=False, router=None, state=None,
                  **kwargs):
         self.loop = loop
         self.transport = None
@@ -99,6 +99,9 @@ class HttpProtocol(asyncio.Protocol):
         self._request_handler_task = None
         self._request_stream_task = None
         self._keep_alive = keep_alive
+        self.state = state if state else {}
+        if 'requests_count' not in self.state:
+            self.state['requests_count'] = 0
 
     @property
     def keep_alive(self):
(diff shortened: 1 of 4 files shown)


sanic/server.py
```diff
@@ -154,6 +157,8 @@ def data_received(self, data):
             self.headers = []
             self.parser = HttpRequestParser(self)
 
+        self.state['requests_count'] = self.state['requests_count'] + 1
```
messense: I think it's better to be incremented in `connection_made` method?

sanic/worker.py
```diff
@@ -109,7 +114,13 @@ def run(self):
             while self.alive:
                 self.notify()
 
-                if pid == os.getpid() and self.ppid != os.getppid():
+                req_count = sum(
+                    self.servers[srv]["requests_count"] for srv in self.servers
+                )
+                if self.cfg.max_requests and req_count > self.cfg.max_requests:
```
messense: It's better to use `self.max_requests` to support 'jitter'.

https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base.py#L54

sanic/worker.py
```diff
@@ -109,7 +114,13 @@ def run(self):
...
+                    self.alive = False
+                    self.log.info("Max requests, shutting down: %s", self)
```
messense: `Max requests exceeded, `

sanic/worker.py
yunstanford: ok, make senses.

sanic/worker.py
yunstanford: i will update.

sanic/server.py
yunstanford: hmm, it depends on the needs/design. Both are fine. i'd prefer on_message_complete or data_received when it mainly starts process the request.

messense: I have seen some weird exceptions caused by `HttpRequestParser`, so to be safe I think it should be incremented before parsing the request.

https://docs.python.org/3/library/asyncio-protocol.html#asyncio.Protocol.data_received

> data_received() can be called an arbitrary number of times during a connection

So it's incorrect to incremented it every time in `data_received`

yunstanford: Any thoughts here ? @seemethere @messense @r0fls

messense: 1. Request count should be incremented regardless of http parsing exception
2. I am not sure, but incrementing in `data_received` is also kind of problematic I think.

yunstanford: How about on_message_complete ?

I can update to connection_made if you guys all agree on this.

yunstanford: I tried this line in different places, connection_made, data_received, on_message_complete, etc. 
when i put it in data_received and on_message_complete, looks good to me. However, the request count becomes weird when puts in connection_made.

seems like aiohttp puts it in a similar place. https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_protocol.py#L296

Anyway, both data_received and on_message_complete look good to me.

yunstanford:
Can we get this merge in ? or any feedback ?

r0fls:
@yunstanford final request would be a little snippet in the docs if you have time. Otherwise looks great to me.

seemethere:
If we merge we should squash and merge

sanic/server.py
```diff
@@ -99,6 +99,9 @@ def __init__(self, *, loop, request_handler, error_handler,
         self._request_handler_task = None
         self._request_stream_task = None
         self._keep_alive = keep_alive
+        self.state = state if state else {}
+        if 'requests_count' not in self.state:
+            self.state['requests_count'] = 0
```
messense: The only thing I am concerned now is that storing `requests_count` in a dict maybe overkill? Is it possible to simply that?

But I am not opposed to merge this and optimize it later.

yunstanford:
sounds good. i will squash and add some docs.

seemethere:
No need to squash on your side. We can do that on ours.

sanic/server.py
r0fls: This is also a good point -- why not simply use `self.request_count`?

seemethere:
I don't want to lose the review history.

yunstanford:
ok, i will add a little bit in doc then.

sanic/server.py
yunstanford: We need to get the request_count from the gunicorn worker side.

yunstanford: I'd like to refactor the `serve` and `serve_multiple` to be a Server Class. Then, we could reference the Server itself inside http_protocol.
Another gotcha of `serve` method is the return value/type. it may change depending on the `run_async`. 
a good pattern may just wrap them up as a Server class.

yunstanford: For now, if we wanna optimize it, probably make` request_count` as gunicorn_worker's attribute, and then pass in the worker ?

antinescience:
To be clear, this doesn't implement native recycling in Sanic directly, right? It's specifically for running it behind Gunicorn?

yunstanford:
@antinescience it doesn't make sense to implement the recycling logic inside Sanic. 

Let's take a step back. if you wanna implement recycling logic inside Sanic, and we should not have `downtime` for sure, then you may end up with a very similar model that gunicorn use. Basically, several workers run behind a parent process.

I don't think implement native recycling in Sanic is the right direction.

r0fls:
@yunstanford @antinescience I think it could make sense to put recycling logic in the multiple workers implementation only. However that is probably the most experimental way of running sanic, in that it's probably safer to use something like supervisord to administer the processes for you (as @nszceta has been advocating for a long time :) )

@seemethere if you can squash and merge this when you have a chance, I have no idea how to do from github's UI ¯\_(ツ)_/¯

yunstanford:
@r0fls Yeah, agree, it's much safer. i'd like probably take supervisord/runit as a even higher level monitoring process. something like, supervisord/runit =>application master process => workers. i am thinking the monitoring process doesn't have to know any application specific settings.

PR #811, opened 2017-06-22
Added support for 'Authorization: Bearer <TOKEN>' header...

Frzk:
...in `Request.token` property.

Also added a test case for that kind of header.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -86,11 +86,15 @@ class Request(dict):
 
         :return: token related to request
         """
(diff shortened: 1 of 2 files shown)


sanic/request.py
```diff
@@ -86,11 +86,15 @@ def token(self):
 
         :return: token related to request
         """
+        prefixes = ('Token ', 'Bearer ')
```
allan-simon: micro optimization I know, but Bearer being more common, shouldn't it be first ?

PR #812, opened 2017-06-22
regular expression pattern fix

stopspazzing:
This fixes reversed `string, pattern` to correct `re.search(pattern, string, flags=0)` format.

### DIFF
--- a/sanic/router.py
+++ b/sanic/router.py
@@ -163,10 +163,10 @@ class Router:
             parameters.append(parameter)
 
             # Mark the whole route as unhashable if it has the hash key in it
-            if re.search(r'(^|[^^]){1}/', pattern):
+            if re.search(pattern, r'(^|[^^]){1}/'):
                 properties['unhashable'] = True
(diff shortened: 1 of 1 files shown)


yunstanford:
It'd be great to add unit test as well. The unit test should fail before your change, it should pass with your fix.

sanic/router.py
```diff
@@ -163,10 +163,10 @@ def add_parameter(match):
             parameters.append(parameter)
 
             # Mark the whole route as unhashable if it has the hash key in it
-            if re.search(r'(^|[^^]){1}/', pattern):
+            if re.search(pattern, r'(^|[^^]){1}/'):
```
r0fls: I'm an idiot. `r'(^|[^^]){1}/'` is of course our regex pattern here. The `pattern` variable should be renamed since that is actually the string that to search. Also we definitely need a unit test for this since we shouldn't be able to switch these around.

sjsadowski:
Closing; PR author failed to provide correlating test as requested by reviewer.

Issue #810, opened 2017-06-20
RuntimeError: Event loop stopped before Future completed on Ctrl-C with open websocket connections


intra-au:
What is the proper way to shutdown the server if it's running some sleeping tasks?

I'm trying to stop the server running the following code with Ctrl-C:
```
async def on_server_start():
  while 1:
    with sql_connect() as conn:
      conn.execute("DELETE from connections where disconnected > 0 and disconnected < ?",(time.time()-1800,))
    await asyncio.sleep(300)

if __name__ == '__main__':
    log.info("Server started...")

    app.add_task(on_server_start())
    try:
      app.run(host="127.0.0.1", port=8083,debug=False)
    except KeyboardInterrupt:
      app.stop()
```

However I get only:
^C2017-06-21 02:38:08 - (sanic)[INFO]: Stopping worker [14215]

and when I press Ctrl-C again I get traceback
```
Traceback (most recent call last):
  File "/home/intra/OST/ost3/lib/python3.5/site-packages/sanic/app.py", line 559, in run
    serve(**server_settings)
  File "/home/intra/OST/ost3/lib/python3.5/site-packages/sanic/server.py", line 467, in serve
    loop.run_until_complete(asyncio.sleep(0.1))
  File "uvloop/loop.pyx", line 1201, in uvloop.loop.Loop.run_until_complete (uvloop/loop.c:25597)
RuntimeError: Event loop stopped before Future completed.
2017-06-21 02:26:32,205:ERROR:sanic:Experienced exception while trying to serve
Traceback (most recent call last):
  File "/home/intra/OST/ost3/lib/python3.5/site-packages/sanic/app.py", line 559, in run
    serve(**server_settings)
  File "/home/intra/OST/ost3/lib/python3.5/site-packages/sanic/server.py", line 467, in serve
    loop.run_until_complete(asyncio.sleep(0.1))
  File "uvloop/loop.pyx", line 1201, in uvloop.loop.Loop.run_until_complete (uvloop/loop.c:25597)
RuntimeError: Event loop stopped before Future completed.
```

yunstanford:
app has already registered those system signals.
https://github.com/channelcat/sanic/blob/master/sanic/server.py#L476

it will catch ctrl-c, then stop the server and also close the loop for you. https://github.com/channelcat/sanic/blob/master/sanic/server.py#L507

You should not call app.stop(), which is essentially just close the loop again. That caused your problem.

intra-au:
[...]

PR #815, opened 2017-06-25
add graceful timeout when shutdown

yunstanford:
Fix for https://github.com/channelcat/sanic/issues/810

We should add graceful_shutdown_timeout, in case it hangs forever when we deploy the service.

There is an example in the issue page (https://github.com/channelcat/sanic/issues/810) for reproducing.

This PR includes,
1. fix for `serve`
2. fix for `gunicorn worker`
3. unit test


sanic/server.py
```diff
@@ -507,8 +516,26 @@ def serve(host, port, request_handler, error_handler, before_start=None,
...
+        # Force close non-idle connection after waiting for
+        # graceful_shutdown_timeout
+        coros = []
+        for conn in connections:
+            if hasattr(conn, "websocket") and conn.websocket:
+                coros.append(conn.websocket.close_connection(force=True))
+            else:
+                conn.close()
+
+        _shutdown = asyncio.gather(*coros, loop=loop)
```
messense: What happens when `coros` is empty here?

yunstanford: It doesn't matter. you can have a try. i think next line `loop.run_until_complete(_shutdown)` just finishes immediately.

yunstanford:
Any feedback ? or can we merge this in ?

Issue #778, opened 2017-06-06
Routing falls over if URL has a trailing slash


mbatchkarov:
Here's a self-contained example:

```python
from random import uniform

FLASK = False

if FLASK:
    from flask import Flask as App
    from flask import jsonify as json
else:
    from sanic import Sanic as App
    from sanic.response import json

app = App(__name__)
DATA = dict()

@app.route('/data', methods=['POST'], strict_slashes=False)
def create_data(*args):
    DATA[int(uniform(0, 1e6))] = 1
    return json('post ok')

# this is broken if URL contains a trailing slash, and the handler
# is the second function in the module (works fine for topmost function)
@app.route('/data', methods=['GET'], strict_slashes=False)
def list_data(*args):
    return json(DATA)


app.run(port=8000)

```

Make the following calls (`httpie` syntax):
```
http get localhost:8000/data
http post localhost:8000/data
http post localhost:8000/data/

http get localhost:8000/data/
```

The first three work as expected, the last one raises `sanic.exceptions.InvalidUsage: Method GET not allowed for URL /data/`. The same code works as expected on Flask- change the `FLASK` constant at the top of the provided script.

The issues appears to be with the order in which routes are added to the app. If I swap the order of `create_data` and `list_data`, then `create_data` starts failing with the same exception.

mbatchkarov:
PS I found a workaround. I don't think it is particularly elegant, but it gets the job done until the cause of the issue is resolved.

``` python
from httptools import parse_url

@app.middleware('request')
def strip_trailing_slash(request):
    if request.url.endswith('/'):
        request._parsed_url = parse_url(request.url.rstrip('/').encode('utf8'))
```

sjsadowski:
Closing per #816

PR #816, opened 2017-06-26
Update app.py

kingname:
use Python's build-in function `rstrip('/')` to remove the tail slash instead of judge if the tail char is slash and then remove it.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -382,8 +382,8 @@ class Sanic:
                     'Endpoint with name `{}` was not found'.format(
                         view_name))
 
-        if uri != '/' and uri.endswith('/'):
-            uri = uri[:-1]
+        if uri != '/':
+            uri = uri.rstrip('/')
 
         out = uri
 


lixxu:
```
>>> '//'.rstrip('/')
''
```

kingname:
@lixxu is there any circumstances where two slash together `//` is valid in **router** ?

what's more, 
if a person's code is `//`, then:

```python
if uri != '/' and uri.endswith('/'):
    uri = uri[:-1]
```
will hide this problem, that may cause some  problem.

but if sanic use `rstrip`, then the uri will become empty string, this uri is absolutely invaild and will cause exception in the later code. So the develop can find the problem.

So may be I should move this code before line 380 so that it can be find in `if not uri`?

lixxu:
I encountered `//` long time before when `url_for` was first added into sanic, but I forgot the code to reproduce it.

kingname:
@lixxu And that circumstance must be an abnormal situation, right?

r0fls:
@kingname I think although it's _abnormal_ it's a valid uri. So it probably shouldn't be excluded from the logic.

kingname:
@r0fls OK

r0fls:
@kingname sorry I didn't mean to shut you down or anything. I have experienced // in the wild (not in the `url_for` method but other places). Was shocked, but it is valid.

Anyway, thank you for the PR/discussion.

kingname:
@r0fls I will come back later with new commits and pr

r0fls:
@kingname please do. If you're interested in the routing this a bug that I introduced when I added that functionality, but haven't had the time to fix yet: https://github.com/channelcat/sanic/issues/778

r0fls:
Basically the `strict_slash` logic when adding a route isn't factoring in if it's the same method.

PR #819, opened 2017-06-27
convert environment vars to int if digits

### LINKED ISSUE #817: Configs loaded from environmental variables aren't properly typed
When setting configs using environmental variables `export SANIC_REQUEST_TIMEOUT=30`

```
app = Sanic(__name__)
print(type(app.config.REQUEST_TIMEOUT))  #  <class 'str'>
```

The problem is in this function
```
# .../sanic/config.py
    def load_environment_vars(self):
        """
        Looks for any SANIC_ prefixed environment variables and applies
        them to the configuration if present.
        """
        for k, v in os.environ.items():
            if k.startswith(SANIC_PREFIX):
                _, config_key = k.split(SANIC_PREFIX, 1)
                self[config_key] = v  # os.environ values are always of type str
```

r0fls:
Fix #817

### DIFF
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -201,11 +201,10 @@ class Config(dict):
         for k, v in os.environ.items():
             if k.startswith(SANIC_PREFIX):
                 _, config_key = k.split(SANIC_PREFIX, 1)
-                # This is a float or an int
-                if v.replace('.', '').isdigit():
-                    if '.' in v:
+                try:
+                    self[config_key] = int(v)
+                except ValueError:
+                    try:
                         self[config_key] = float(v)
-                    else:
-                        self[config_key] = int(v)
-                else:
-                    self[config_key] = v
+                    except ValueError:
+                        self[config_key] = v
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -201,4 +201,7 @@ class Config(dict):
         for k, v in os.environ.items():
             if k.startswith(SANIC_PREFIX):
                 _, config_key = k.split(SANIC_PREFIX, 1)
-                self[config_key] = v
+                if v.isdigit():
+                    self[config_key] = int(v)
+                else:
+                    self[config_key] = v
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -201,7 +201,11 @@ class Config(dict):
         for k, v in os.environ.items():
             if k.startswith(SANIC_PREFIX):
                 _, config_key = k.split(SANIC_PREFIX, 1)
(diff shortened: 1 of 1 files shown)


sanic/config.py
```diff
@@ -201,4 +201,7 @@ def load_environment_vars(self):
         for k, v in os.environ.items():
             if k.startswith(SANIC_PREFIX):
                 _, config_key = k.split(SANIC_PREFIX, 1)
-                self[config_key] = v
+                if v.isdigit():
+                    self[config_key] = int(v)
```
seemethere: What if they're float values? o.O

r0fls: @seemethere I was thinking that too at first but we don't have any float values in the config right now... I think we can use float instead of int though and the timeout should be happy.

r0fls: @seemethere actually i'm just going to check for a `.`

seemethere: Maybe something similar to: https://stackoverflow.com/a/379966?

sanic/config.py
```diff
@@ -201,7 +201,11 @@ def load_environment_vars(self):
         for k, v in os.environ.items():
             if k.startswith(SANIC_PREFIX):
                 _, config_key = k.split(SANIC_PREFIX, 1)
-                if v.isdigit():
-                    self[config_key] = int(v)
+                # This is a float or an int
```
seemethere: I think it'd work out better like:

```python
try:
    self[config_key] = int(v)
except ValueError:
    try:
        self[config_key] = float(v)
    except ValueError:
        self[config_key] = v
```

r0fls: \To me, the `isdigit` check is more explicit, but I guess if we want to convert bools to 0/1 as well the try/except will do that.

r0fls: I think try/except would be better if this was a performance intensive piece of the code.

yunstanford: hmm...` try/except` is much more expensive than `if`check. it may be better to use `if` check.

seemethere: This is a method that's called once when the app instantiates, I don't think performance is really a concern here.

r0fls: @yunstanford oh really, I've heard it was the other way... oh well, I'm done changing this back and forth :) if someone wants to modify it further we're happy to have another PR.

yunstanford: nvm, you didn't catch the error, just `pass`, so there are not much differences, then.

It's much more expensive to `catch` the exception.

PR #820, opened 2017-06-27
Protocol configurable gunicorn worker

youknowone:
Users can replace the protocol like

```python
GunicornWorker.http_protocol = ThereOwnProtocol
```

Disclaimer: I am not sure this is a good idea.


youknowone:
I am maintaining local patches of HttpProtocol, not because of http protocol problem but because of Sanic's design.
With this fix, I at least don't need to maintain a gunicorn worker only to replace HttpProtocol.

Does it make sense?

yunstanford:
Customizing protocol is not a good pattern.. `SanicGunicornWorker` should be only for `SanicProtocol`, because it has `Sanic` specific stuff. 

It may cause problem, because `SanicGunicornWorker` may not be consistent for all  customized `protocol`. 

Let's say, 
`Sanic http_protocol` has a request_count attribute (https://github.com/channelcat/sanic/blob/master/sanic/server.py#L104), and it is used in `Sanic GunicornWorker`(https://github.com/channelcat/sanic/blob/master/sanic/worker.py#L118). But if anyone create a new protocol, which doesn't have `request_count` attribute, then it will be broken.
It may be very confusing to the user.

The `Worker` is designed for the `Protocol`.  

In your case, you are confident that you patches are good, then it works. But the pattern is not good, and we are not guarantee the`SanicGunicornWorker` works for all `protocols`.

I am opposed to your change, anyway. i am just thinking the design pattern.

sanic/worker.py
```diff
@@ -45,8 +48,9 @@ def init_process(self):
 
     def run(self):
         is_debug = self.log.loglevel == logging.DEBUG
-        protocol = (WebSocketProtocol if self.app.callable.websocket_enabled
-                    else HttpProtocol)
+        protocol = (
+            self.web_socket_protocol if self.app.callable.websocket_enabled
```
r0fls: Looks like this should be `self.websocket_protocol`

youknowone: @r0fls Thanks, I fixed it

PR #821, opened 2017-06-27
Inverted the order of prefixes in Request.token property.

Frzk:
As suggested by @allan-simon
See: https://github.com/channelcat/sanic/pull/811#pullrequestreview-46144327

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -86,7 +86,7 @@ class Request(dict):
 
         :return: token related to request
         """
-        prefixes = ('Token ', 'Bearer ')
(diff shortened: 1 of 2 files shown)


Frzk:
The test that fails uses "Bearer Token" as authentication scheme in the "Authorization" header (ie. `Authorization: Bearer Token <TOKEN>`
See https://github.com/channelcat/sanic/blob/master/tests/test_requests.py#L178)

I've never seen this scheme. Is it valid ? I guess it should be either `Bearer` or `Token` but not both. Am I wrong ?

r0fls:
Yes we should remove that.

PR #822, opened 2017-06-27
Adding version to blueprints

Typhon66:
I can imagine this could be useful in a lot of cases where people may want to create different versions of their blueprints. This could be expanded on later to do potentially help with routing optionally. For now i just want to include it as an option that i can access through the blueprint object.

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -14,7 +14,7 @@ FutureStatic = namedtuple('Route',
 
 
 class Blueprint:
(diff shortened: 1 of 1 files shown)


seemethere:
I mean but if it's not functional, what's the difference in just adding a new variable after instantiation?

```python
bp = Blueprint()
bp.version = '0.1.0'
```

r0fls:
closing this in favor of https://github.com/channelcat/sanic/pull/850

PR #826, opened 2017-06-29
Pytest sanic

yunstanford:
add pytest-sanic for easy testing.


seemethere:
Can you make sure to add this to the wiki as well?

yunstanford:
Sure.

PR #828, opened 2017-06-30
A buffer which helps httptools not to corrupt requests

### LINKED ISSUE #755: fragmented headers

youknowone:
Roughly fix #755 

This patch will slow Sanic, so I recommend to fix httptools or replace http parser to another one in future.
For now, a slow server is better than a broken server.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -56,6 +56,40 @@ class CIDict(dict):
         return super().__contains__(key.casefold())
 
 
+class RequestBuffer:
+    def __init__(self):
+        self.saved = b''
+        self.passed_header = False
+        self.received = b''
+
+    def put(self, data):
+        if not self.passed_header:
+            self.saved += data
+            if b'\r\n\r\n' in self.saved:
+                data = self.saved
+                self.saved = None
+                self.passed_header = True
+            else:
+                try:
+                    i = self.saved.rindex(b'\r\n')
+                except ValueError:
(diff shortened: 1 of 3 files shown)


lcDotcom:
`httptools` is fast though. Do you have any recommendation parser package to replace `httptools` ?

r0fls:
If we're considering this, @youknowone can you post benchmarks so we understand how much of a slowdown this is?

r0fls:
@youknowone I'm closing this since https://github.com/channelcat/sanic/pull/756 was merged. Please reopen if you feel it is still necessary.

PR #831, opened 2017-07-02
Improve Documentation.

yunstanford:
This PR includes:

1. Add sphinx auto-doc support, doc builds will extract doc_string automatically.
2. Add `API Reference` section. Adding api reference in Documentation is important. It's better to provide that in Documentation instead of letting users find anything they need in codebase directly.
3. Fixes doc_build warnings

I also suggest that we should write doc_string when you are writing sth that will be public to users.
`sphinx auto-doc` will extract out these doc_strings and put them in Documentation, `API references`.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,20 +1,225 @@
-# Minimal makefile for Sphinx documentation
+# Makefile for Sphinx documentation
 #
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
-SPHINXPROJ    = Sanic
-SOURCEDIR     = .
+PAPER         =
 BUILDDIR      = _build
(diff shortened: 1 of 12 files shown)


lcDotcom:
it'd be great to have API reference section.

yunstanford:
Can we also merge this in if no objections?

r0fls:
@yunstanford shamefully I don't know how to review this -- are you able to give me instructions for deploying the docs changes locally? Is it just the below? How do I start the sphinx server?

```sphinx-apidoc -fo docs/_api/ sanic
sphinx-build -b html docs docs/_build```

yunstanford:
If you have installed all dependencies, then simply do `make html` in `/docs`. Then you can see those static pages in `_build` directory.

PR #837, opened 2017-07-06
add release notes to docs

lixxu:


lixxu:
@r0fls if it's not a good idea to add this release notes to doc, I'll close it.

r0fls:
ok if you agree with @lcDotcom then that's fine. I would agree if it's going to require updating two places, let's keep the work required to maintain things minimal unless it adds a lot of good functionality.

r0fls:
However if there is a good reason for this and others are interested in it, please leave it open so it can be further discussed. It's up to you.

r0fls:
I am undecided here. @seemethere what's your opinion? He's been managing the releases and I think he said that he prefers to keep it in one place.

r0fls:
uh oh, i think the rebase went wrong. I guess I vote we close this since there was objection from @lcDotcom and I suspect @seemethere doesn't want any extra work when he's releasing the latest version...

lixxu:
I'll close it.

PR #839, opened 2017-07-09
Drop benchmarks from README

asvetlov:
The reason is: aiohttp with disabled access log shows about 16,000 RPS on sanic's own benchmark.
It's pretty much faster than 3,000 RPS from the table.

I'm not a Sanic dev team member. You should not trust users to update this table but manage periodic updates yourself.
If you don't want to do it --- it's up to you.
Please just drop very incorrect and outdated numbers from README in this case.


yunstanford:
yeah, it should have been updated. I agree.

r0fls:
Ya, I think we should update all of the benchmarks or remove them since it's been quite a long time since they were performed.

r0fls:
@asvetlov I'm ok with merging this, since I agree it's not representative. However like I said I would prefer to drop that table entirely for now until those can be updated. @seemethere @messense @channelcat @yunstanford any opposition? We can update them now, but with sanic's enhanced logging the table is probably going to look a lot less meaningful... so we might just consider dropping it entirely for the time being. We can add a section to the docs about "going fast" which discusses disabling the logging as a solution....?

r0fls:
Again, the table was accurate ~2 years ago when it was created FTR

yunstanford:
Yeah, i agree, either drop the table entirely or update them all the accurate number is fine. it's not good to keep outdated numbers there.

r0fls:
ok let's remove the table altogether for now since it's outdated.

PR #850, opened 2017-07-13
Versioning

r0fls:
add versioning to blueprints and routes

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -4,8 +4,8 @@ from sanic.constants import HTTP_METHODS
 from sanic.views import CompositionView
 
 FutureRoute = namedtuple('Route',
-                         ['handler', 'uri', 'methods',
-                          'host', 'strict_slashes', 'stream'])
+                         ['handler', 'uri', 'methods', 'host',
+                          'strict_slashes', 'stream', 'version'])
 FutureListener = namedtuple('Listener', ['handler', 'uri', 'methods', 'host'])
 FutureMiddleware = namedtuple('Route', ['middleware', 'args', 'kwargs'])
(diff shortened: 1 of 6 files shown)


sanic/blueprints.py
```diff
@@ -43,12 +44,20 @@ def register(self, app, options):
             future.handler.__blueprintname__ = self.name
             # Prepend the blueprint URI prefix if available
             uri = url_prefix + future.uri if url_prefix else future.uri
+
+            # Blueprint global version will override route version
+            if self.version is not None:
+                version = self.version
+            else:
+                version = future.version
```
yunstanford:
```python
version = self.version or future.version
```

r0fls: @yunstanford maybe the future version should take precedence actually? Then I can knock this comment out of the docs: https://github.com/channelcat/sanic/pull/850/files#diff-52c81980ed2f6ad30d4db74d97124d6fR32

r0fls: That is the version passed from the route decorator

sanic/blueprints.py
```diff
@@ -89,21 +95,21 @@ def register(self, app, options):
                 app.listener(event)(listener)
 
     def route(self, uri, methods=frozenset({'GET'}), host=None,
-              strict_slashes=False, stream=False):
+              strict_slashes=False, stream=False, version=None):
         """Create a blueprint route from a decorated function.
 
         :param uri: endpoint at which the route will be accessible.
         :param methods: list of acceptable HTTP methods.
```
yunstanford: maybe add `:param version` in doc_string for auto-doc ?

r0fls:
btw this was inspired by the hug API, for the record.

Issue #801, opened 2017-06-17
sanic behind proxy - option to replace IP of request.ip with X-Forwarded-For value


intra-au:
Dear Devs,
Please consider adding an option to replace request.ip with X-Forwarded-For value, it will make life a lot easier for those of us who running sanic behind nginx or load balancer.

I can get real IP with request.headers.get('X-Forwarded-For') but would be good to have an option to set custom header for sanic to get ip address from so it can get added in sanic logs, console etc.
It's quite common and as an example Flask has ProxyFix to address this.

andreymal:
#853 looks unconfigurable and unsafe, something like [werkzeug ProxyFix](https://github.com/pallets/werkzeug/blob/64d1d2117cc177b9caf18bb571f32133492978c3/werkzeug/contrib/fixers.py#L97) would be better

andreymal:
...but this is still unsafe :/

sjsadowski:
Perhaps, but the original need was addressed, reviewed, and merged. @andreymal perhaps you can submit a new PR to address your concerns?

untitaker:
BTW, the default configuration in Sanic is still completely unsafe.

https://github.com/huge-success/sanic/blob/891f99d71d3bd17501a9762b0843736f4377e3f3/sanic/request.py#L358-L362

The docstring is misleading in the sense that it implies Sanic has the necessary knowledge to decide whether a header can be trusted or not, but per default it just trusts any header that comes its way.

>Perhaps, but the original need was addressed, reviewed, and merged.

BTW, this mindset of "let's just address one need at a time and don't think about security implications" is exactly why IP-based limits in strawpoll can still be bypassed by sending X-Forwarded-For.

andreymal:
>Sanic has the necessary knowledge

Yes, it has. See `PROXIES_COUNT` and `REAL_IP_HEADER` options ([#1539](https://github.com/huge-success/sanic/pull/1539))

untitaker:
@andreymal sanic has the necessary knowledge if somebody took time to configure those values. Per default sanic just trusts `x-forwarded-for` and `x-real-ip` as far as I can tell.

untitaker:
In any case, I can see why you would want to have this as a default, I am just saying that this needs to be made very explicit as it works unlike most other Python web frameworks I've seen, where reading from proxy headers is usually opt-in (Flask, Django at least).

[...]

PR #853, opened 2017-07-15
proxy fix

yunstanford:
request from https://github.com/channelcat/sanic/issues/801

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -45,7 +45,7 @@ class Request(dict):
     __slots__ = (
         'app', 'headers', 'version', 'method', '_cookies', 'transport',
         'body', 'parsed_json', 'parsed_args', 'parsed_form', 'parsed_files',
-        '_ip', '_parsed_url', 'uri_template', 'stream'
+        '_ip', '_parsed_url', 'uri_template', 'stream', '_remote_addr'
     )
 
     def __init__(self, url_bytes, headers, version, method, transport):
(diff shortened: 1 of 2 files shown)


sanic/request.py
```diff
@@ -142,7 +142,7 @@ def raw_args(self):
     @property
     def cookies(self):
         if self._cookies is None:
-            cookie = self.headers.get('Cookie') or self.headers.get('cookie')
+            cookie = self.headers.get('Cookie')
```
r0fls: Is this needed? What does this have to do with the remote address change?

yunstanford: it's not related. this is code cleanup, since `header` is a Case Insensitive Dict.

andreymal:
I have a small question: is it really necessary to return an empty string when `X-Forwarded-For` is not available? IMHO `self._remote_addr = self.ip` would be more convenient

ahopkins:
@andreymal Can you either open a new issue, or put the question on [the forums](https://community.sanicframework.org) with context relating to current? This is a pretty old PR.

andreymal:
@ahopkins discussion of this PR in a place other than this PR doesn't seem like a good idea ¯\\(°_o)/¯

Also, your link is not working (502 Bad Gateway)

ahopkins:
@andreymal Sorry for the `502`, I was updating the discourse install just now.

There really is no discussion to be had on this PR. It is merged two years ago, and itself does not have a lengthy conversation. If there is something in this PR that you think needs to be changed, it warrants its own conversation.

PR #863, opened 2017-07-24
handle keep-alive timeout gracefully

### LINKED ISSUE #760: keepalive message recorded in debug mode
ENVIRONMENT
==========
centos7_x64
chrome: 58.0.3029.110
sanic: 0.5.4
python: 3.5.3 or 3.6.1

CODE
====
```
from sanic import Sanic
from sanic.response import text,json

app = Sanic(__name__)

@app.route('/')
async def test(request):
    return text('hello world')

if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8080, debug=True)
```

ISSUE DESCRIPTION
=============
when issue http request from chrome the first time,  Sanic RequestTimeout exception thrown out in 60s after we get the correct response from the server, I am not sure this is the expected behavior or not, I didn't see any reference about this error in the document:

```
2017-05-31 02:50:34 - (sanic)[INFO]: Goin' Fast @ http://0.0.0.0:8080
2017-05-31 02:50:34 - (sanic)[INFO]: Starting worker [3277]
I print when a request is received by the server
2017-05-31 02:50:38 - (network)[INFO][192.168.31.164:53576]: GET http://192.168.31.227:8080/number/123  200 13

2017-05-31 02:51:38 - (sanic)[ERROR]: NoneType: None
2017-05-31 02:51:38 - (network)[INFO][]: NoneNone  408 22
```

COMMENT
========
After capturing the tcpdump packets, a HTTP 408 message send by Sanic to browser when it reaches SANIC_REQUEST_TIMEOUT(60s), also it can be seen in the codes event_loop will call connection_timeout() after waiting for that time,  so I assume this is the expected behavior of Sanic.  please help confirm it is true or not?  If it is true, then it had better mention this symptom in the document and perhaps add a switch to turn off send back this exception to client.

zyguan:
This should fix #760 

1. `raise RequestTimeout` so that `sys.exc_info` is available for the default error handler
2. do not log annoying keep-alive timeouts

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -139,8 +139,10 @@ class HttpProtocol(asyncio.Protocol):
                 self._request_stream_task.cancel()
             if self._request_handler_task:
                 self._request_handler_task.cancel()
-            exception = RequestTimeout('Request Timeout')
-            self.write_error(exception)
+            try:
+                raise RequestTimeout('Request Timeout')
+            except RequestTimeout as exception:
+                self.write_error(exception)
 
     # -------------------------------------------- #
     # Parsing
@@ -317,6 +319,7 @@ class HttpProtocol(asyncio.Protocol):
                 self.cleanup()
 
     def write_error(self, exception):
+        response = None
         try:
             response = self.error_handler.response(self.request, exception)
             version = self.request.version if self.request else '1.1'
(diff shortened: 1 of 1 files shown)


sanic/server.py
```diff
@@ -139,8 +139,10 @@ def connection_timeout(self):
                 self._request_stream_task.cancel()
             if self._request_handler_task:
                 self._request_handler_task.cancel()
-            exception = RequestTimeout('Request Timeout')
-            self.write_error(exception)
+            try:
```
r0fls: How is this try/except logic different from the current situation? To me it appears identical, since any other exception would get raised with this implementation, and you do the same thing in the except clause.

zyguan: try this
```python
def print_exc_info(exc):
    print(format_exc())

print_exc_info(RuntimeError())

try:
    raise RuntimeError()
except RuntimeError as e:
    print_exc_info(e)
```
That's why we got:
```
2017-05-31 02:51:38 - (sanic)[ERROR]: NoneType: None
```

PR #869, opened 2017-07-26
Fix issue with parsing multipart body from unirest

ir210:
Some libraries e.g. unirest always put Content-Type inside every chunk.
Even though, it is a simple text/plain. This makes Sanic, mistakenly,
treats that chunk as file.

This commit made sure only chunks with "filename" after
Content-Disposition are treated as files.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -265,7 +265,7 @@ def parse_multipart_form(body, boundary):
                 if 'filename' in form_parameters:
                     file_name = form_parameters['filename']
                 field_name = form_parameters.get('name')
-            elif form_header_field == 'content-type':
+            elif form_header_field == 'content-type' and file_name:
                 file_type = form_header_value
 
         post_data = form_part[line_index:-4]


DirkGuijt:
Can we please get this fix merged in? The fix is small but fixes a big problem. Not all form-data with a content-type are files...

r0fls:
Well the tests are failing but sure I’ll take a look...

DirkGuijt:
Great stuff @r0fls ! Thanks :)

DirkGuijt:
Actually, after taking a second look... 

This fix assumes that the Content-Disposition header is handled before the Content-Type header. Now usually this is the case, but the RFC7578 multipart/form-data specification actually does not specify any preferred order of the headers. Maybe @ir210 can take a look at it, or else I can, but I can't commit to this PR directly of course...

DirkGuijt:
Just created a new PR with an improved fix at PR #1109

DirkGuijt:
This can be closed with the merging of #1109

PR #883, opened 2017-08-08
Weboscket subprotocol negotiation

### LINKED ISSUE #874: Websocket subprotocol support
The _websockets_ library that Sanic utilizes supports passing a list of acceptable websocket subprotocols into their server or client implementations, to be used during the opening handshake ("Sec-WebSocket-Protocol").  Sanic does not support this currently and does not do anything w/ subprotocols in the `websocket_handshake` method.  The eventual websocket object that gets passed to the Sanic handler has an "subprotocol" attribute, which would normally be set during the handshake. Currently it is empty.  Is there any plan to support subprotocols in the future?  Thanks for the excellent framework!

miguelgrinberg:
Fixes #874

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -214,9 +214,12 @@ class Sanic:
         return handler
 
     # Decorator
-    def websocket(self, uri, host=None, strict_slashes=False):
+    def websocket(self, uri, host=None, strict_slashes=False,
+                  subprotocols=None):
         """Decorate a function to be registered as a websocket route
         :param uri: path of the URL
+        :param subprotocols: optional list of strings with the supported
+                             subprotocols
         :param host:
         :return: decorated function
         """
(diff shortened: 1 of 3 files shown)


sanic/app.py
```diff
@@ -214,9 +214,12 @@ def add_route(self, handler, uri, methods=frozenset({'GET'}), host=None,
         return handler
 
     # Decorator
-    def websocket(self, uri, host=None, strict_slashes=False):
+    def websocket(self, uri, subprotocols=None, host=None,
```
seemethere: Can we move subprotocols to be the last keyword-arg? I don't want to break anyone's existing functionality by adding a new positional arg in between the already existing ones.

miguelgrinberg: Yeah, very good point.

PR #887, opened 2017-08-11
Add Flask like built-in sessions

AZLisme:
this pull request adds a new attribute `session` to request, it works like dict, and is captured and persists between requests automatically.

## How to use?

```python
@app.route("/")
async def test(request):
    # read
    print(request.session['hello'])
    # write
    request.session['hello'] = 'world'
    # delete
    del request.session['hello']
    return text("That's it, simple right?")
```

There's more examples in `example/set_sessions.py`

## How it works?

This session managenment is basically same as flask's. It encrypts and stores session key-pair in cookies.

## Tested?

Yes, I added unittest to this feature, which can be found in `tests/test_sessions.py`. 

And I tested it manually, embedded in my new project.

## Configurations?

Every configuration has it's default value, but you can customize them as you wish.

Related configurations are listed as blew:

+ SECRET_KEY
+ SESSION_COOKIE_NAME
+ PERMANENT_SESSION_LIFETIME
+ SESSION_COOKIE_SECURE
+ SESSION_COOKIE_HTTPONLY
+ SESSION_COOKIE_DOMAIN
+ SESSION_COOKIE_PATH

## What's next?

I'll add more documents about this feature, and make sure every configuration works as expected.

### DIFF
--- a/sanic/sessions.py
+++ b/sanic/sessions.py
@@ -1,0 +1,312 @@
+# -*- encoding: utf-8 -*-
+
+import hashlib
+
+from itsdangerous import URLSafeTimedSerializer
+
+import ujson as json
+from sanic.exceptions import BadSignature
+
+def total_seconds(td):
+    """Returns the total seconds from a timedelta object.
+    :param timedelta td: the timedelta to be converted in seconds
+    :returns: number of seconds
+    :rtype: int
+    """
+    return td.days * 60 * 60 * 24 + td.seconds
+
+class SessionMixin(object):
+    """Expands a basic dictionary with an accessors that are expected
+    by Flask extensions and users for the session.
+    """
+
+    def _get_permanent(self):
+        return self.get('_permanent', False)
+
+    def _set_permanent(self, value):
+        self['_permanent'] = bool(value)
+
+    #: this reflects the ``'_permanent'`` key in the dict.
+    permanent = property(_get_permanent, _set_permanent)
+    del _get_permanent, _set_permanent
+
+    #: some session backends can tell you if a session is new, but that is
+    #: not necessarily guaranteed.  Use with caution.  The default mixin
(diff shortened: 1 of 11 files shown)


AZLisme:
Ooops, looks like I have a dependency issue.

sanic/sessions.py
```diff
@@ -0,0 +1,303 @@
...
+def total_seconds(td):
+    """Returns the total seconds from a timedelta object.
+    :param timedelta td: the timedelta to be converted in seconds
+    :returns: number of seconds
+    :rtype: int
+    """
+    return td.days * 60 * 60 * 24 + td.seconds
+
+
+class SessionMixin(object):
```
yunstanford: This is essentially copy paste from Flask: https://github.com/pallets/flask/blob/master/flask/sessions.py#L22.
This may not have to be a built-in feature of Sanic. 
https://pythonhosted.org/Flask-Session/, I'd like to say you can import the module or make a Sanic one.

AZLisme: Yes this interface is 90% from flask source, I just made minimum modify to work with sanic. Why bother build a whole new thing if there's already one. 

Extension will definitely work as same, I just thought it would be nice to have it built in.

AZLisme:
OK, I'll close this pr, and add this feature to this [session](https://github.com/subyraman/sanic_session)

sanic/sessions.py
yunstanford: Sorry, I am not suggesting to build a whole thing.  If there has been already one, let's use it. BUT, Make it sharable, like a lib, instead of copying the code. It's not good for maintaining the code base in both sides. Someone should always track changes to check if any bug or upgrade, it's not good pattern, i'd like to see sth has global impact. That's my whole point.

PR #896, opened 2017-08-19
fix #895

### LINKED ISSUE #895: Maybe we can let users choose whether `merge_route` or not?
I use Ruby on Rails before, I like the resourcesful router in Rails, so I try to do the same thing in Sanic.

```python
bp = sanic.Blueprint('posts', url_prefix='posts')

@bp.get('/<post_id>')
async def show(request, post_id):
    pass

@bp.patch('/<post_id>')
async def update(request, post_id):
    pass

@bp.delete('/<post_id>')
async def delete(request, post_id):
    pass
```
But when I use `url_for('posts.update', post_id=post.id)`, It shows `Error: Endpoint with name posts.update was not found`.

Then I found it will `merge_route` into `CompositionView` as it change router name to `posts.show`, it is exactlly not what I want. Maybe we can disable `merge route` by given an options to user?

https://github.com/channelcat/sanic/blob/master/sanic/router.py#L194-L214

r0fls:
There is an issue with `url_for` regarding overlapping methods. Allowing disabling `merge_routes` is a workaround.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -28,7 +28,7 @@ class Sanic:
 
     def __init__(self, name=None, router=None, error_handler=None,
                  load_env=True, request_class=None,
-                 log_config=LOGGING):
+                 log_config=LOGGING, merge_routes=True):
         if log_config:
             logging.config.dictConfig(log_config)
         # Only set up a default log handler if the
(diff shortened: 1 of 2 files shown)


yunstanford:
unit test ?

r0fls:
This is actually not working. Still getting `post.update` not found (see original issue) with the current patch. Closing this since a different approach is needed.

PR #898, opened 2017-08-20
feat(examples): add `add_task_sanic.py`

jiaxiaolei:
In the examples, I do not find a demo with `add_task`. In my opinion, `add_task` is a very interesting
feature in `sanic`.

Hope the demo can be added in `examples`, and `sanic` become more powerful and more popular

Thanks.

### DIFF
--- a/examples/add_task_sanic.py
+++ b/examples/add_task_sanic.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+import asyncio
+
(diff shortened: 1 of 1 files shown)


r0fls:
Can you add a hello world route to this as well, to demonstrate how they both work together? Some route that just says "hello" or something, e.g.:

```
@app.get("/hello")
def hello(request):
    return text("Hello")
```

Issue #897, opened 2017-08-19
default strict_slashes


kianxineki:
I have seen that all strict_slashes are False, is it possible to change the default to True? Maybe it could be an argument from app.config

zhustec:
+1. Same needed.

yunstanford:
Changing the default value may break current Sanic application, you can easily pass `strict_slashes=True` when you define your routes.

r0fls:
I think they just want to be able to _set_ the default value, so to avoid passing it to each route. This seems reasonable to me.

yunstanford:
yeah, i see its value, maybe allow to pass a keyword `strict_slashes` when initializing app or blueprints.
i can push a quick fix.

zhustec:
@yunstanford Maybe set in 'app.config' instead of keywords arguments is better?

zhustec:
@yunstanford Maybe set in 'app.config' instead of keywords arguments is better?

yunstanford:
I am thinking this feature should also apply to `blueprints`, `app.config` doesn't work for that. I can think of a use case, like, different bp could have different default `strict_slashes`  values
```python3
app = Sanic()
bp1 = Blueprint('bp1', strict_slashes=False)
bp2 = Blueprint('bp1', strict_slashes=True)
```

PR #900, opened 2017-08-21
Patch default strict slashes

yunstanford:
Fix https://github.com/channelcat/sanic/issues/897

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -28,7 +28,7 @@ class Sanic:
 
     def __init__(self, name=None, router=None, error_handler=None,
                  load_env=True, request_class=None,
-                 log_config=LOGGING):
+                 log_config=LOGGING, strict_slashes=False):
         if log_config:
             logging.config.dictConfig(log_config)
         # Only set up a default log handler if the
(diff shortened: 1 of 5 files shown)


sanic/app.py
```diff
@@ -111,7 +112,7 @@ def decorator(listener):
 
     # Decorator
     def route(self, uri, methods=frozenset({'GET'}), host=None,
-              strict_slashes=False, stream=False, version=None):
+              strict_slashes=None, stream=False, version=None):
```
r0fls: Instead of `strict_slashes=None` I think it's better to do `strict_slashes=self.strict_slashes` since it's more compact. What do you think?

yunstanford: `self` is not available at that moment i think.

r0fls: I understand your doubt but it seemed to work here: https://github.com/channelcat/sanic/pull/896

I did the same type of thing a couple days ago when I made that, so should work :)

yunstanford: yeah, it works in your case, because that is inside `def route(self, ...)`, `self` is available then.
```python
def route(self, ...):
   # blablabla
   self.router.add(..., merge_routes=self.merge_routes)
```
Correct me if i am looking at wrong place.

r0fls: 🤕  derp, good call

r0fls:
Nice work, once again. Thanks for the help!

r0fls:
@yunstanford final thing -- can you add this to the docs?

PR #901, opened 2017-08-21
add name option for route building

lixxu:


r0fls:
This may also solve https://github.com/channelcat/sanic/issues/895

lixxu:
no, it can not. the later name will be ignored as `sanic` use url as key that caused only first name will be the winner.
```python
app = Sanic('test_dynamic_route')

@app.route('/overload', methods=['GET'], name='route_first')
async def handler1(request):
    return text('OK1')

@app.route('/overload', methods=['POST', 'PUT'], name='name_will_be_ignored')
async def handler2(request):
    return text('OK2')
```

lixxu:
Not sure if this makes sense.
```python
# sanic/router.py

class RouteNameExists(Exception):
    pass


class Router:
    # other codes
    def __init__(self):
        # other codes
        self.routes_names = {}
        
    def _add(self, uri, methods, handler, host=None, name=None):
        # other codes
        if handler_name in self.routes_names:
            raise RouteNameExists(
                "Route name already exists: {}".format(handler_name))
            
        self.routes_names[handler_name] = (uri, route)

    def remove(self, uri, clean_cache=True, host=None):
        if host is not None:
            uri = host + uri
        try:
            route = self.routes_all.pop(uri)
            for handler_name, pairs in self.routes_names.items():
                if pairs[0] == uri:
                    self.routes_names.pop(handler_name)
                    break
                
        except KeyError:
            raise RouteDoesNotExist("Route was not registered: {}".format(uri))

        # other codes

    @lru_cache(maxsize=ROUTER_CACHE_SIZE)
    def find_route_by_view_name(self, view_name):
        # ...
        return self.routes_names.get(view_name, (None, None))
```

lixxu:
I tried, seems work.

r0fls:
@lixxu do you mean that it should be ok to close #895 ?

lixxu:
yes, it's ok to close. but seems there's a bug that when specify the name that same as the handler name for first function (don't know why). like this:
```
# below will not work
@bp.get('/<post_id>', name='show')
async def show(request, post_id):
    pass

# below is ok
@bp.get('/<post_id>', name='show')
async def get(request, post_id):
    pass
```

PR #904, opened 2017-08-21
feat(examples): add `authorized_sanic.py`

jiaxiaolei:
We can check a request if the client is authorized
to access a resource by the decorator `authorized`.

It should be widely used in the projects.

### DIFF
--- a/examples/authorized_sanic.py
+++ b/examples/authorized_sanic.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+
+from sanic import Sanic
+from functools import wraps
+from sanic.response import json
+
+app = Sanic()
+
+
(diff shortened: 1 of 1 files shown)


GoodiesHQ:
The `app.middleware` and `app.listener` seems much more appropriate.

jiaxiaolei:
Because of `app.middleware` and `app.listener `,  It becomes not appropriate to use decorator like `authorized` to do some check for the request and the decorator becomes `chicken ribs`

But, `app.middleware` and `app.listener` work on all the `request` and `response`. The are not flexible like `authorized`.

For example, in all requests(urls), some urls should check user whether login in, and others not.

Above is just my thoughts, maybe it's not right.
Thanks for pointing out my errors.

PR #924, opened 2017-08-30
recovery ujson error 

Zheaoli:
Sanic use ultra-json as its default json parser, but sometimes when ujson dumps some special dict ,it will raise some Exception such as **OverflowError: Invalid Nan value when encoding double** , but the standard json library would not . So I suggest to add a mechanism in Sanic that when the `ujson.dumps` raise Exception , we should use the standard json library and  try it again

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -4,7 +4,7 @@ from os import path
 try:
     from ujson import dumps as json_dumps
 except:
-    from json import dumps as jjson_dumps
+    from json import dumps as json_dumps
 finally:
     from json import dumps as jjson_dumps
 from aiofiles import open as open_async
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -2,10 +2,11 @@ from mimetypes import guess_type
 from os import path
 
 try:
-    from ujson import dumps as json_dumps
+    from ujson import dumps as ujson_dumps
 except:
-    from json import dumps as json_dumps
-
+    from json import dumps as jjson_dumps
+finally:
+    from json import dumps as jjson_dumps
 from aiofiles import open as open_async
 
 from sanic.cookies import CookieJar
@@ -244,8 +245,12 @@ def json(body, status=200, headers=None,
     :param headers: Custom Headers.
     :param kwargs: Remaining arguments that are passed to the json encoder.
     """
-    return HTTPResponse(json_dumps(body, **kwargs), headers=headers,
-                        status=status, content_type=content_type)
+    try:
+        return HTTPResponse(ujson_dumps(body, **kwargs), headers=headers,
+                            status=status, content_type=content_type)
+    except Exception as e:
+        return HTTPResponse(jjson_dumps(body, **kwargs), headers=headers,
+                            status=status, content_type=content_type)
 
 
 def text(body, status=200, headers=None,
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -2,7 +2,7 @@ from mimetypes import guess_type
 from os import path
 
 try:
-    from ujson import dumps as ujson_dumps
+    from ujson import dumps as json_dumps
 except:
     from json import dumps as jjson_dumps
 finally:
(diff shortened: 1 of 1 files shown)


sanic/response.py
```diff
@@ -4,8 +4,9 @@
 try:
     from ujson import dumps as json_dumps
 except:
-    from json import dumps as json_dumps
-
+    from json import dumps as jjson_dumps
```
r0fls: You should leave this as `json_dumps` since otherwise it will raise an exception every time in line 249 when using json instead of ujson.

Zheaoli: OK,I get your means

sanic/response.py
```diff
@@ -244,8 +245,12 @@ def json(body, status=200, headers=None,
     :param headers: Custom Headers.
     :param kwargs: Remaining arguments that are passed to the json encoder.
     """
-    return HTTPResponse(json_dumps(body, **kwargs), headers=headers,
-                        status=status, content_type=content_type)
+    try:
+        return HTTPResponse(json_dumps(body, **kwargs), headers=headers,
```
yunstanford: 1. This line could raise a bunch of exceptions, not only from `json_dump`. We should not retry the same thing if exception is raised by others. You should either extract `json_dump` out or catch more specific exception.
Otherwise, This `try/except` could be problematic.

2. If the application'd trigger a exception per request here, this type of handling is very expensive though. It will slow the application a ton. When someone writes the application, they should expect some sort of results, if it expects some results that `ujson` cannot handle, I'd suggest use `HTTPResponse` instead of this `def json()` function, as such, you could use `json` or `simplejson` directly instead of this `try/except` block. That'd be better.

Zheaoli: With no offensive , here's my opinion

1. **ujson** with `OverflowError: Invalid Nan value when encoding double`had been issued in 2014, but it's not been fixed , you can look this [ujson issue](https://github.com/esnme/ultrajson/issues/146)

2.In Sanic  [guideline](http://sanic.readthedocs.io/en/latest/) , the first sample is using `json`  to decorate the dict that the people want to return . So if there is some people who just start to use Sanic , when they use sample in guideline , there is possibility to raise an Exception because of ujson's bug. 

Because of those reason , I think there may have two way to fix this problem

1. As you say , we can use `HTTPResponse` instead of `json` method to avoid this exception , but we have to warn people for that the `json` method may raise an exception with decorate some special dict because of **ujson**'s bug.
2.I can commit once more to catch an  specific exception and raise other exception

What do you think

Zheaoli: @yunstanford

yunstanford: np, i am just thinking the strategy.
First, i agree, it's `ujson` issue, thanks for bringing up this discussion, as for fixing,
1. if you really wanna fix this way, this may be better
```pthon
try:
    json_resp = json_dumps(body, **kwargs)
except:
    json_resp = jjson_dumps(body, **kwargs)
```
2. or you can just document the solution above.

I am fine with both. Thanks.
cc: @r0fls @seemethere  What do you guys think ?

Zheaoli: I am fine with both too
cc @r0fls @seemethere

seemethere: I feel like you lose the initial error though, like if `ujson` fails it should report it upwards, not just eat the exception and retry with the standard json library.

r0fls: Hmm, that makes sense @seanpar203, but if there is no exception when using `json` instead of `ujson`, then it's probably ok that it swallows the `ujson` exception in that case.

seemethere: @r0fls I think you tagged the wrong person but I'm fine with this the `try except` approach from @yunstanford

Zheaoli: I‘am ok for both of two way. So, who can give me a final decision , and give me five and thumbs up ! Oh, it's a personal issue Hhhhhhh

r0fls: @Zheaoli the gods have spoken, @yunstanford's method is final.

r0fls: @Zheaoli do you have time to switch the implementation to the above and resolve the merge conflicts?

chiuczek:
Is #948 relevant here? It seems like hiding the json processing from the user isn't the best idea — it might make things look easier, but it doesn't feel right to me. I prefer to have a reasonable default (`ujson` in this case) but allow it to be changed if needed (so use standard library `json`).

r0fls:
Ya, I think @Zheaoli should just use the strategy from #948. I'm going to close this soon unless we hear a strong argument for it.

Zheaoli:
@r0fls Sorry about this= =
I closed this PR just now.

PR #929, opened 2017-09-01
If a flag cookie is set to False don't include it

### LINKED ISSUE #920: Cookie secure option not encoded properly
When `Cookies.encode` encounters `response.cookies["<cookie>"]["secure"] = False` then it outputs:

`b'Domain=xad.com; Path=/; Secure=False'`

where it should output:

`b'Domain=xad.com; Path=/;'` when `response.cookies["<cookie>"]["secure"] = False`

and

`b'Domain=xad.com; Path=/; Secure;'` when `response.cookies["<cookie>"]["secure"] = True`

reallistic:
Fixes #920

### DIFF
--- a/sanic/cookies.py
+++ b/sanic/cookies.py
@@ -116,8 +116,9 @@ class Cookie(dict):
                     ))
                 except AttributeError:
                     output.append('%s=%s' % (self._keys[key], value))
-            elif key in self._flags and self[key]:
-                output.append(self._keys[key])
+            elif key in self._flags:
+                if self[key]:
+                    output.append(self._keys[key])
             else:
                 output.append('%s=%s' % (self._keys[key], value))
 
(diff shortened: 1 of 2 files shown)


sanic/cookies.py
```diff
@@ -116,8 +116,9 @@ def encode(self, encoding):
                     ))
                 except AttributeError:
                     output.append('%s=%s' % (self._keys[key], value))
-            elif key in self._flags and self[key]:
-                output.append(self._keys[key])
+            elif key in self._flags:
+                if self[key]:
```
r0fls: This logic should actually be the same as previously. Did you try running the tests without this change, to make sure they actually fail?

r0fls:
Closing in favor of https://github.com/channelcat/sanic/pull/961. There were issues with the original false cookie implementation and its unit test (my bad). They should be fixed now

nappappa:
Just for the sake of completeness: This patch (#929) would have fixed issue #920. Note that the logic is indeed changed @r0fls, since there is an else statement below which must not be called if self[key] == False! Interestingly, the two nested ifs were also part of the initial commit, #707 introduced the problem.

PR #931, opened 2017-09-05
make the prefix for environment variables alterable

Tim-Erwin:
Suppose my app is named "Abc", it is really not obvious, why the configuration prefix for environment variables is "SANIC_". Also, if I happen to have more than one Sanic app in one environment, there would be no way to distinguish those. Hence, I propose to make it possible to use other prefixes.

### DIFF
--- a/docs/sanic/config.md
+++ b/docs/sanic/config.md
@@ -29,9 +29,15 @@ In general the convention is to only have UPPERCASE configuration parameters. Th
 
 There are several ways how to load configuration.
 
-### From environment variables.
+### From Environment Variables
 
-Any variables defined with the `SANIC_` prefix will be applied to the sanic config. For example, setting `SANIC_REQUEST_TIMEOUT` will be loaded by the application automatically. You can pass the `load_env` boolean to the Sanic constructor to override that:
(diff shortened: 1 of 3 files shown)


jrocketfingers:
LGTM, except that it'd be cool if we documented it too.

jrocketfingers:
https://github.com/channelcat/sanic/blob/master/docs/sanic/config.md#loading-configuration

This looks like a fine place to start.

Tim-Erwin:
Great, now I get to be the one to write the docs for something already existing *grumble*

Well, you're right, I'll add some docs :)

Tim-Erwin:
Oh, I just saw, there is already documentation for that, my fault. Of course I'll update it.

Tim-Erwin:
I just realized, that - while this change is sufficient for my use case - many people will probably use the `load_env` option. What do you think about allowing to pass in the prefix via this `load_env` variable? E.g.

```python
app = Sanic(load_env='MYAPP_')
```

Tim-Erwin:
I added the possibility to specify the prefix through the app constructor, added docs and tests.

Also I fixed the existing tests, two of the functions had the same name which caused one of them (the failing one) to be skipped.

PR #939, opened 2017-09-12
Split RequestTimeout, ResponseTimeout, and KeepAliveTimeout into different timeouts

### LINKED ISSUE #902: sanic.exceptions.RequestTimeout: Request Timeout
I glanced at the issues(61), and did not found the problem I encounter.
I find the problem more than one time. There is a example to show the problem.

In the module `method_sanic.py`,  there is just one url. After I run the server by ` python method_sanic.py`, I get the response through type `http://[ip]:8001/get`.

But, few seconds later, the backend throw an exception, `sanic.exceptions.RequestTimeout: Request Timeout`. 

I try it few times, the problem is not occur every time.

```
##### filename: method_sanic.py
# -*- coding: utf-8 -*-

from sanic import Sanic

app = Sanic()

@app.get('/get')
async def get_handler(request):
    return text('GET request - {}'.format(request.args))

if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8001)
```

```
$ python method_sanic.py
2017-08-21 20:28:00 - (sanic)[DEBUG]:
                 ▄▄▄▄▄
        ▀▀▀██████▄▄▄       _______________
      ▄▄▄▄▄  █████████▄  /                 \
     ▀▀▀▀█████▌ ▀▐▄ ▀▐█ |   Gotta go fast!  |
   ▀▀█████▄▄ ▀██████▄██ | _________________/
   ▀▄▄▄▄▄  ▀▀█▄▀█════█▀ |/
        ▀▀▀▄  ▀▀███ ▀       ▄▄
     ▄███▀▀██▄████████▄ ▄▀▀▀▀▀▀█▌
   ██▀▄▄▄██▀▄███▀ ▀▀████      ▄██
▄▀▀▀▄██▄▀▀▌████▒▒▒▒▒▒███     ▌▄▄▀
▌    ▐▀████▐███▒▒▒▒▒▐██▌
▀▄▄▄▄▀   ▀▀████▒▒▒▒▄██▀
          ▀▀█████████▀
        ▄▄██▀██████▀█
      ▄██▀     ▀▀▀  █
     ▄█             ▐▌
 ▄▄▄▄█▌              ▀█▄▄▄▄▀▀▄
▌     ▐                ▀▀▄▄▄▀
 ▀▀▄▄▀

2017-08-21 20:28:00 - (sanic)[INFO]: Goin' Fast @ http://0.0.0.0:8001
2017-08-21 20:28:00 - (sanic)[INFO]: Starting worker [4514]
2017-08-21 20:28:08 - (network)[INFO][172.31.246.188:54142]: GET http://172.28.32.49:8001/get  200 16
2017-08-21 20:29:09 - (sanic)[ERROR]: Traceback (most recent call last):
  File "/root/.virtualenvs/py3.6.1env/lib/python3.6/site-packages/sanic/server.py", line 143, in connection_timeout
    raise RequestTimeout('Request Timeout')
sanic.exceptions.RequestTimeout: Request Timeout

^C2017-08-21 20:29:24 - (sanic)[INFO]: Stopping worker [4514]
2017-08-21 20:29:24 - (sanic)[INFO]: Server Stopped
```

```
2017-08-21 21:24:02 - (sanic)[INFO]: Goin' Fast @ http://0.0.0.0:8001
2017-08-21 21:24:02 - (sanic)[INFO]: Starting worker [4543]



2017-08-21 21:24:09 - (network)[INFO][172.31.246.188:56531]: GET http://172.28.32.49:8001/get  200 16


###### another attempt：
----

2017-08-21 21:24:19 - (network)[INFO][172.31.246.188:56531]: GET http://172.28.32.49:8001/get  200 16
2017-08-21 21:24:21 - (network)[INFO][172.31.246.188:56531]: GET http://172.28.32.49:8001/get  200 16
2017-08-21 21:25:21 - (sanic)[ERROR]: Traceback (most recent call last):
  File "/root/.virtualenvs/py3.6.1env/lib/python3.6/site-packages/sanic/server.py", line 143, in connection_timeout
    raise RequestTimeout('Request Timeout')
sanic.exceptions.RequestTimeout: Request Timeout

2017-08-21 21:25:22 - (sanic)[ERROR]: Traceback (most recent call last):
  File "/root/.virtualenvs/py3.6.1env/lib/python3.6/site-packages/sanic/server.py", line 143, in connection_timeout
    raise RequestTimeout('Request Timeout')
sanic.exceptions.RequestTimeout: Request Timeout

^C2017-08-21 21:25:33 - (sanic)[INFO]: Stopping worker [4543]
2017-08-21 21:25:33 - (sanic)[INFO]: Server Stopped
```

```
##### filename: sanic/server.py
    130     def connection_timeout(self):
    131         # Check if
    132         time_elapsed = current_time - self._last_request_time
    133         if time_elapsed < self.request_timeout:
    134             time_left = self.request_timeout - time_elapsed
    135             self._timeout_handler = (
    136                 self.loop.call_later(time_left, self.connection_timeout))
    137         else:
    138             if self._request_stream_task:
    139                 self._request_stream_task.cancel()
    140             if self._request_handler_task:
    141                 self._request_handler_task.cancel()
    142             try:
    143                 raise RequestTimeout('Request Timeout')
    144             except RequestTimeout as exception:
    145                 self.write_error(exception)
```

Now, the problem is a bug in `sanic`? I use it in a wrong way?

Any response is gratitude.

ashleysommer:
This is a big change to the server backend of Sanic, and my largest Sanic contribution by far. There will need to be some discussion around this.

While investigating what seemed like a relatively simple problem in Issue #902 it was discovered that Sanic uses a single timeout (called RequestTimeout), which is handled by the `connection_timeout()` callback, and controlled by the 'REQUEST_TIMEOUT' config option, to cover three different scenarios:
1. The request took too long to come from the client,
2. The response took too long to come from our application,
3. We are using using a KeepAlive connection, and the KeepAlive timer runs out.

In all three of these scenarios, the server generates a 408 ("RequestTimeout") error and sends a 408 response to the client. While this is correct for case (1), it is incorrect behaviour for scenarios (2) and (3).

Even more confusting is if you look in the test_receive_timeout.py Test code, the test code injects a delay into the response, causing the response to timeout, then throws, catches, and delivers a 408: Request Timeout. This is confusing and wrong.

In the case of (2) when our application takes too long to generate a response to a request, we should use a ResponseTimeout, and send a 5xx error.

In thecase of (3), when the keepalive timeout expires, we should not send an error at all, just log it and move on.

To put it another way,
* A RequestTimeout is a Client error. It is caused by the client taking too long to send a request (after establishing a TCP connection). It gets a 4xx series error response, in this case 408.
* A ResponseTimeout is a Server error. It is our application doing something that is taking too long, so it should send a 5xx series error response. In this case a 503 is the closest match.
* A KeepAlive timeout is *_not_* an error. It is normal and expected that every connection will eventually time out. This is what Issue #902 was about.

This PR has some big changes.
There are now three different timeouts tracked by the sanic backend server, each with their own configurable timeout values, their own default values, their own callback handlers, and their own logical flow.

1) RequestTimeout (error 408), which is a timeout that occurs between the point in time when a TCP connection is created, and the the point in time when the entire HTTP request is received. Default=60 seconds.
2) A ResponseTimeout (error 503), which is a timeout that occurs between the point in time when the server starts to process a request, and the point in time when the server delivers a response. Default=60 seconds.
3) A KeepAliveTimeout (not an error), which is a timeout that occurs between the point in time when the server delivers a response, and the point in time when a new request is received on the same TCP connection. Default = 5 seconds.

Fixes #902

### DIFF
--- a/tests/test_keep_alive_timeout.py
+++ b/tests/test_keep_alive_timeout.py
@@ -1,0 +1,142 @@
+from json import JSONDecodeError
+from sanic import Sanic
+from time import sleep as sync_sleep
+import asyncio
+from sanic.response import text
+from sanic.config import Config
+import aiohttp
+from aiohttp import TCPConnector
+from sanic.testing import SanicTestClient, HOST, PORT
+
+
+class ReuseableTCPConnector(TCPConnector):
+    def __init__(self, *args, **kwargs):
+        super(ReuseableTCPConnector, self).__init__(*args, **kwargs)
+        self.conn = None
+
+    @asyncio.coroutine
+    def connect(self, req):
+        if self.conn:
+            return self.conn
+        conn = yield from super(ReuseableTCPConnector, self).connect(req)
+        self.conn = conn
+        return conn
+
+    def close(self):
+        return super(ReuseableTCPConnector, self).close()
+
+
+class ReuseableSanicTestClient(SanicTestClient):
+    def __init__(self, app):
+        super(ReuseableSanicTestClient, self).__init__(app)
+        self._tcp_connector = None
+        self._session = None
+
+    def _sanic_endpoint_test(
+            self, method='get', uri='/', gather_request=True,
+            debug=False, server_kwargs={},
+            *request_args, **request_kwargs):
+        results = [None, None]
+        exceptions = []
+
+        if gather_request:
+            def _collect_request(request):
+                if results[0] is None:
+                    results[0] = request
+
+            self.app.request_middleware.appendleft(_collect_request)
+
+        @self.app.listener('after_server_start')
+        async def _collect_response(sanic, loop):
+            try:
+                response = await self._local_request(
+                    method, uri, *request_args,
+                    **request_kwargs)
+                results[-1] = response
+            except Exception as e:
+                log.error(
+                    'Exception:\n{}'.format(traceback.format_exc()))
+                exceptions.append(e)
+            self.app.stop()
+
+        server = self.app.create_server(host=HOST, debug=debug, port=PORT, **server_kwargs)
+        self.app.listeners['after_server_start'].pop()
+
+        if exceptions:
+            raise ValueError(
+                "Exception during request: {}".format(exceptions))
+
+        if gather_request:
+            try:
+                request, response = results
+                return request, response
+            except:
+                raise ValueError(
+                    "Request and response object expected, got ({})".format(
+                        results))
+        else:
+            try:
+                return results[-1]
+            except:
+                raise ValueError(
+                    "Request object expected, got ({})".format(results))
+
+    async def _local_request(self, method, uri, cookies=None, *args,
+                             **kwargs):
+        if uri.startswith(('http:', 'https:', 'ftp:', 'ftps://' '//')):
+            url = uri
+        else:
+            url = 'http://{host}:{port}{uri}'.format(
+                host=HOST, port=PORT, uri=uri)
+        if self._session:
+            session = self._session
+        else:
+            if self._tcp_connector:
+                conn = self._tcp_connector
+            else:
+                conn = ReuseableTCPConnector(verify_ssl=False)
+                self._tcp_connector = conn
+            session = aiohttp.ClientSession(cookies=cookies,
+                                            connector=conn)
+            self._session = session
+
+        async with getattr(session, method.lower())(
+                url, *args, **kwargs) as response:
+            try:
+                response.text = await response.text()
+            except UnicodeDecodeError:
+                response.text = None
+
+            try:
+                response.json = await response.json()
+            except (JSONDecodeError,
+                    UnicodeDecodeError,
+                    aiohttp.ClientResponseError):
+                response.json = None
+
+            response.body = await response.read()
+            return response
+
+
+Config.KEEP_ALIVE_TIMEOUT = 30
+Config.KEEP_ALIVE = True
+keep_alive_timeout_app = Sanic('test_request_timeout')
+
+
+@keep_alive_timeout_app.route('/1')
+async def handler(request):
+    return text('OK')
+
+
+def test_keep_alive_timeout():
+    client = ReuseableSanicTestClient(keep_alive_timeout_app)
+    headers = {
+        'Connection': 'keep-alive'
+    }
+    request, response = client.get('/1', headers=headers)
+    assert response.status == 200
+    #sync_sleep(2)
+    request, response = client.get('/1')
+    assert response.status == 200
+
+
--- a/tests/test_keep_alive_timeout.py
+++ b/tests/test_keep_alive_timeout.py
@@ -4,6 +4,7 @@ from time import sleep as sync_sleep
 import asyncio
 from sanic.response import text
 from sanic.config import Config
+from sanic import server
 import aiohttp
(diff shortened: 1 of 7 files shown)


ashleysommer:
In this PR, all three timeouts now have their own group of tests.

There have been no performance benchmarks done yet.

ashleysommer:
Ok, while the tests are passing on my machine, they are _not_ passing on Travis.
Will need to do some more work on the tests.

ashleysommer:
Woo! passing now :)

r0fls:
@ashleysommer sorry still reviewing this. Looks great though.

tests/test_keep_alive_timeout.py
```diff
@@ -0,0 +1,269 @@
...
+        if self.old_proto is not None:
+            if self.old_proto != new_conn._protocol:
+                raise RuntimeError(
+                    "We got a new connection, wanted the same one!")
+        print(new_conn.__dict__)
+        self.old_proto = new_conn._protocol
+        return new_conn
+
+
+class ReuseableSanicTestClient(SanicTestClient):
```
yunstanford: Is there any reason we need this ? can we use `test_client` fixture in `pytest-sanic` ? https://github.com/yunstanford/pytest-sanic, it'd be better to maintain.

ashleysommer: This is here because I needed a custom version of the `SanicTestClient` that could:
1) Honor a keepalive header, and store the connection in the client for reuse.
2) Reuse the existing TCP connection when making a subsequent request, if keepalive was set on the first request.
3) Raise an error if we got a new connection, when we were expecting to reuse the old connection.

These customisations to the `SanicTestClient` are very specific to this particular test (the KeepAliveTimeout test)

I haven't looked, but I don't think the `test_client` fixture in `pytest-sanic` has the ability to be customised in this manner.

yunstanford: No, you can do them all, by simply passing your `ReuseableTCPConnector ` as param to `test_client`.

r0fls:
I'm going to go ahead and merge this now and we can improve it later, if needed. @yunstanford if you have time to make a PR with the pytest-sanic module that'd be nice. Thanks @ashleysommer!

ashleysommer:
@r0fls Ok, thanks for merging.

To be honest, I was expecting more discussion around this before merging, specifically about my choice of default values for the config variables, and the logic around deciding when each timeout is triggered. But I guess those things can be discussed in other PRs or Issues now.

I don't know how many people are using/testing `sanic@master`, but I expect that some people might notice a change in sanic server behavior, particularly if they were relying on the specific nuances of the previous behavior. There may be some new github Issues popping up to around this, as it may appear to some as bugs/regressions.

Another thing I thought of is the documentation. The docs for the `REQUEST_TIMEOUT` config variable will need to be changed, plus docs for `RESPONSE_TIMEOUT` and `KEEP_ALIVE_TIMEOUT` will need to be written. I will open a PR with those docs changes in the next few days, when I get a chance.

r0fls:
@ashleysommer ok thanks. For the record, if a PR is not complete when created, and you think it _requires_  feedback or more work (e.g. docs additions), then you can mark it as WIP to indicate that.

PR #944, opened 2017-09-15
add __repr__ for sanic request

blazehu:
when i use sanic i found that in views print request show {}  . it's not friendly

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -71,10 +71,10 @@ class Request(dict):
 
     def __repr__(self):
         if self.method is None or not self.path:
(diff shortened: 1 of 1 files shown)


sanic/request.py
```diff
@@ -170,8 +178,8 @@ def remote_addr(self):
             remote_addrs = [
                 addr for addr in [
                     addr.strip() for addr in forwarded_for
-                ] if addr
-            ]
+                    ] if addr
+                ]
```
r0fls: Apparently pep8 isn't weighing in here, but I prefer the earlier spacing in this case, because it closes the multi line list at the same indentation that it was created.

Issue #905, opened 2017-08-21
Custom JSON encoder?


olalonde:
I'm getting the following error when I call json() on a dict that contains numpy floats:

```python
    return HTTPResponse(json_dumps(body, **kwargs), headers=headers,
OverflowError: Maximum recursion level reached
```

I'm porting an application from flask to sanic but am stuck there. With flask, I had the following custom JSON encoder:

```python
class CustomJSONEncoder(JSONEncoder):
    '''
    Custom JSON encoder that converts numpy.float64 values to strings.
    '''
    def default(self, val):
        if isinstance(val, numpy.int32):
            return int(val)
        if isinstance(val, numpy.int64):
            return int(val)
        if isinstance(val, numpy.number):
            return str(val)
        if isinstance(val, numpy.ndarray):
            lst = []
            for v in val:
                lst.append(str(v))
            return lst
        return JSONEncoder.default(self, val)
```

I tried `json(data, cls=CustomJSONEncoder)` but it fails since you're using `ujson` which doesn't accept a custom encoder class.

Any suggestion on how to best approach this issue?

olalonde:
Using the following work around:

```python
from sanic.response import json as sanic_json

def is_iterable(x):
    try:
        iter(x)
        return True
    except TypeError:
        return False


def map_anything(x, fn):
    print('map_anything ', x)
    if isinstance(x, str):
        return fn(x)
    if isinstance(x, dict):
        return {k: map_anything(v, fn) for k, v in x.items()}
    if is_iterable(x):
        return [map_anything(ele, fn) for ele in x]
    return fn(x)


def prepare_for_json(val):
    if isinstance(val, numpy.int32):
        return int(val)
    if isinstance(val, numpy.int64):
        return int(val)
    if isinstance(val, numpy.number):
        return str(val)
    if isinstance(val, numpy.ndarray):
        lst = []
        for v in val:
            lst.append(str(v))
        return lst
    return val


def json(data, **kwargs):
    # cls not supported by ujson
    # https://github.com/esnme/ultrajson/issues/124
    return sanic_json(map_anything(data, prepare_for_json), **kwargs)
```

jrocketfingers:
The most elegant solution to this is given in the offending line itself. You could create your own `json_response`, which is essentially a wrapper around `HTTPResponse` (just like in the first line of your exception).

So something along the lines of:
```python
def json(body, status, headers, content_type):
[...]

PR #948, opened 2017-09-19
Use dependency injection to allow alternative json parser or encoder

chiuczek:
Allow a custom/alternative JSON encode/decode function to be specified when sending a response or parsing a request. The `response.json` function gets an extra optional keyword argument `dumps`, with obvious default. When parsing a request, instead of the property `request.json` a method `request.load_json(loads=<new loads>)` can be used.

My use case is returning results from calculations, stored as numpy arrays or pandas dataframes. The pandas library has a nice JSON encode/decode module (derived from ujson) which will sensibly (and quickly) handle numpy arrays.

```{python}
import numpy as np
import pandas as pd
import sanic

def a_response_handler(request)
    request_data = request.load_json(loads=pd.json.loads)

# processing would go here

    response_data = {
        "first": np.array([0.0, 1.0, 4.0]),
        "second": np.random.uniform(0, 1, 10)
    }
    return sanic.response.json(response_data, dumps=pd.json.dumps)
```

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -71,12 +71,7 @@ class Request(dict):
     @property
     def json(self):
         if self.parsed_json is None:
-            try:
-                self.parsed_json = json_loads(self.body)
-            except Exception:
-                if not self.body:
-                    return None
-                raise InvalidUsage("Failed when parsing body as json")
+            self.load_json()
 
         return self.parsed_json
 
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -69,14 +69,24 @@ class Request(dict):
         self.stream = None
 
     @property
-    def json(self, loads=json_loads):
+    def json(self):
         if self.parsed_json is None:
             try:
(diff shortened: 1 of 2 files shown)


chiuczek:
This relates to issue #905, but wasn't inspired by it

chiuczek:
Also, sorry for the username confusion, chiuczek is my personal account and intelliHugh is for work use, but they're both me!

sanic/request.py
```diff
@@ -80,6 +80,16 @@ def json(self):
 
         return self.parsed_json
 
+    def load_json(self, loads=json_loads):
```
r0fls: you should replace https://github.com/chiuczek/sanic/blob/a8f764c1612ab17eb99924ac2363c05110a1911d/sanic/request.py#L74-L81 with a call to this function

chiuczek: of course I should, great suggestion, thanks

PR #962, opened 2017-10-06
Fix logs

ProstoMaxim:
Preferable logging is through percentage sign https://docs.python.org/3/library/logging.html#formatter-objects
It is real issue for integration with Sentry. If message will be trasfered just as formatted string, Sentry will not be able to 'intelligently group messages' (last paragraph).
https://docs.sentry.io/clients/python/integrations/logging/#usage

### DIFF
--- a/sanic/handlers.py
+++ b/sanic/handlers.py
@@ -86,12 +86,13 @@ class ErrorHandler:
             self.log(format_exc())
             if self.debug:
                 url = getattr(request, 'url', 'unknown')
-                response_message = (
-                    'Exception raised in exception handler "{}" '
-                    'for uri: "{}"\n{}').format(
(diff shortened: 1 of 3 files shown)


r0fls:
@ProstoMaxim have you tested this implementation to work with sentry?

r0fls:
This is interesting because I thought the `"{}".format(var)` syntax was preferred in general. Maybe not for logging...

ProstoMaxim:
Yes, this implementation will work with Sentry as it supposed to, otherwise it will generate a lot of similar logging records (which I came across). Format method is indeed preferred in general, but not for logging. As far as I know, logging module have a lot of legacy code and for now this percentage formatting is just something that we have to accept, like Sentry developers did.

Issue #959, opened 2017-10-03
(sanic)[ERROR]: Connection lost before response written


aqeelahmad:
I am facing an issue with some of the requests. It's returning an error when we get quite a few simultaneous requests. The error description is as below:

`web_1  | 2017-10-03 09:24:49 - (network)[INFO][172.17.0.1:55372]: GET http://localhost:8000/api/order_items/123456  200 38
web_1  | 2017-10-03 09:24:50 - (network)[INFO][172.17.0.1:55382]: GET http://localhost:8000/api/order_items/123456 200 38
web_1  | 2017-10-03 09:24:55 - (network)[INFO][172.17.0.1:55392]: GET http://localhost:8000/api/order_items/123456 200 38
web_1  | 2017-10-03 09:24:56 - (sanic)[ERROR]: Connection lost before response 2343 written @ ('172.17.0.1', 55402)
web_1  | 2017-10-03 09:24:56 - (network)[INFO][172.17.0.1:55412]: GET http://localhost:8000/api/order_items/123456 200 38
web_1  | 2017-10-03 09:24:57 - (sanic)[ERROR]: Connection lost before response 2343 written @ ('172.17.0.1', 55424)
web_1  | 2017-10-03 09:24:57 - (network)[INFO][172.17.0.1:55430]: GET http://localhost:8000/api/order_items/123456 200 38`

So as per my understanding, HTTP connection is closing before Sanic can write to it which is fine but I should be able to override the behaviour and hide the error if I wish to which is something I need help with

ashleysommer:
Are you using sanic@master or 0.6.0 release?

I ask, because there was recently some changes to the TCP backend server connection handling logic, and this may be a regression.

Otherwise, if you are on 0.6.0, perhaps you could try sanic@master to see if the behavior is better in that.

samael500:
I have the same issue on sanic `0.6.0` pypi version, and on `sanic@master` git version too

```
2017-10-06 07:31:07 - (sanic)[ERROR]: Connection lost before response written @ ('172.21.0.1', 51732)
2017-10-06 07:31:07 - (sanic)[ERROR]: Connection lost before response written @ ('172.21.0.1', 51730)
2017-10-06 07:31:07 - (sanic)[ERROR]: Connection lost before response written @ ('172.21.0.1', 51728)
2017-10-06 07:31:07 - (sanic)[ERROR]: Connection lost before response written @ ('172.21.0.1', 51726)
[...]

PR #965, opened 2017-10-09
fix issue #959

samael500:
Connection lost occurs when client disconnect before the response will written. So this is not an error.

In AioHTTP was the same issue, and was resolved with just ignore this error on non debug mode https://github.com/aio-libs/aiohttp/commit/fc1a04d9494bfbe55e97c2bf4077dad3237294cb .


aqeelahmad:
Do we know when exactly 0.6.1 will be released?

hansraj1999:
in which version this was fixed?

PR #972, opened 2017-10-13
to fix condition error that used in `log_response`

lanfon72:
`request` class is derived from **`dict`**, so it will never be **`True`**.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -311,7 +311,7 @@ class HttpProtocol(asyncio.Protocol):
             else:
                 extra['byte'] = -1
 
-            if self.request:
+            if self.request is not None:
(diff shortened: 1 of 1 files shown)


r0fls:
>request class is derived from dict, so it will never be True.

Well, that's false:

```python
>>> nothing, something = {}, {'foo': 'bar'}
>>> if nothing:
...    print("you won't see me")
... 
>>> if something:
...    print("hello world")
... 
hello world
```

lanfon72:
I meant something like this:
```python
from sanic.request import Request, RequestParameters

r = Request(b"http://localhost/", 2,3,4, 5)
True if r else False
False if r else True
r, r.items(), r.keys(), r.values(), len(r), bool(r)
```
image

PR #975, opened 2017-10-16
Add documentation for new Timeout values, after #939

ashleysommer:
Add documentation for REQUEST_TIMEOUT, RESPONSE_TIMEOUT and KEEP_ALIVE_TIMEOUT config values.

Fixed some inconsistent default values.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -76,7 +76,7 @@ class HttpProtocol(asyncio.Protocol):
 
     def __init__(self, *, loop, request_handler, error_handler,
                  signal=Signal(), connections=set(), request_timeout=60,
-                 response_timeout=60, keep_alive_timeout=15,
+                 response_timeout=60, keep_alive_timeout=5,
                  request_max_size=None, request_class=None, access_log=True,
                  keep_alive=True, is_request_stream=False, router=None,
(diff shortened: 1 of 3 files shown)


sanic/server.py
```diff
@@ -76,7 +76,7 @@ class HttpProtocol(asyncio.Protocol):
 
     def __init__(self, *, loop, request_handler, error_handler,
                  signal=Signal(), connections=set(), request_timeout=60,
-                 response_timeout=60, keep_alive_timeout=15,
+                 response_timeout=60, keep_alive_timeout=5,
```
r0fls: What's the motivation for changing this?

ashleysommer: That is to keep all of the keep_alive_timeout default values the same.
In #939 I had:
`Config.KEEP_ALIVE_TIMEOUT=5` in config.py
`keep_alive_timeout=60` in server.py `serve()`
`keep_alive_timeout=15` in HttpProtocol.__init__()

By default the `Config.KEEP_ALIVE_TIMEOUT` value was used and passed through to the others, so the 60 and the 15 were never used.

I changed them to 5 to be consistent with the default config value, and the documentation.

Issue #969, opened 2017-10-11
Websockets connections receiving "Response Timeout"


furious-luke:
Hi there, I'm finding that websockets connections are getting cancelled from the server as a result of a 60 second timeout. The error looks like this:

```bash
[2017-10-11 14:43:54 +1100] [11644] [ERROR] Traceback (most recent call last):
  File ".../sanic/sanic/server.py", line 178, in response_timeout_callback
    raise ServiceUnavailable('Response Timeout')
sanic.exceptions.ServiceUnavailable: Response Timeout
```

This issue can be reproduced with a simple server like this:

```python
import time
from sanic import Sanic
from sanic.response import json

app = Sanic()

@app.websocket('/')
async def test(request, ws):
    time.sleep(120)
    ws.send('hello')

if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8000)
```

And a client like this:

```python
import asyncio
import websockets

async def go():
    async with websockets.connect('ws://localhost:8000') as ws:
        await ws.recv()

loop = asyncio.get_event_loop()
loop.run_until_complete(go())
```

Is there something obvious I'm doing wrong?

Thanks for your help!

CharAct3:
`time.sleep(120)` blocks the server, use:
```python
import asyncio


@app.websocket('/')
async def test(request, ws):
    await asyncio.sleep(120)
    ws.send('hello')
```

furious-luke:
Hey @CharAct3, that's a fair point about blocking the server, however I don't think it's related to the websockets getting clobbered. I've altered the server code to use `await asyncio.sleep(120)` and the websockets are still getting killed by a timeout with the same error.

CharAct3:
Hey @furious-luke , you're right.
I was using sanic with version 0.6.0, so I did't meet this problem.
It's caused by `HttpProtocol.response_timeout`.

furious-luke:
It looks like it shouldn't be affecting websockets, at least by the code in `sanic/websocket.py`:

```python
    def connection_timeout(self):
        # timeouts make no sense for websocket routes
        if self.websocket is None:
            super().connection_timeout()
```

Is it possible this method is being ignored for some reason?

CharAct3:
@furious-luke Yes, this method is being ignored.

`WebsocketProtocol` inherits `HttpProtocol`.

And `HttpProtocol.connection_timeout` has been splited into three methods now: 
`request_timeout_callback`
`response_timeout_callback`
[...]

PR #976, opened 2017-10-16
Fix Websocket protocol timeouts after #939

ashleysommer:
Update websocket protocol to accommodate changes in HTTP protocol from https://github.com/channelcat/sanic/pull/939

Fixes https://github.com/channelcat/sanic/issues/969

### DIFF
--- a/sanic/websocket.py
+++ b/sanic/websocket.py
@@ -13,10 +13,18 @@ class WebSocketProtocol(HttpProtocol):
         self.websocket_max_size = websocket_max_size
         self.websocket_max_queue = websocket_max_queue
 
-    def connection_timeout(self):
-        # timeouts make no sense for websocket routes
+    # timeouts make no sense for websocket routes
+    def request_timeout_callback(self):
         if self.websocket is None:
-            super().connection_timeout()
+            super().request_timeout_callback()
+
+    def response_timeout_callback(self):
+        if self.websocket is None:
+            super().response_timeout_callback()
+
+    def keep_alive_timeout_callback(self):
+        if self.websocket is None:
+            super().keep_alive_timeout_callback()
 
     def connection_lost(self, exc):
         if self.websocket is not None:


r0fls:
I still receive the request timeout with this branch. To test you can open the browser with the websocket.py example and wait 60 seconds (or, more intelligibly, decrease the timeout).

In the same vein, the below test also fails on this branch, which I would expect to pass:

```
$ git diff  master tests/test_request_timeout.py
diff --git a/tests/test_request_timeout.py b/tests/test_request_timeout.py
index a1d8a88..d226da3 100644
--- a/tests/test_request_timeout.py
+++ b/tests/test_request_timeout.py
@@ -138,6 +138,15 @@ Config.REQUEST_TIMEOUT = 2
 request_timeout_default_app = Sanic('test_request_timeout_default')
 request_no_timeout_app = Sanic('test_request_no_timeout')
 
+@request_timeout_default_app.websocket('/feed')
+async def feed(request, ws):
+    while True:
+        data = 'hello!'
+        print('Sending: ' + data)
+        await ws.send(data)
+        data = await ws.recv()
+        print('Received: ' + data)
+
 
 @request_timeout_default_app.route('/1')
 async def handler1(request):
@@ -149,6 +158,13 @@ async def handler2(request):
     return text('OK')
 
 
+def test_websocket_request_timeout():
+    client = DelayableSanicTestClient(request_timeout_default_app, None, 3)
+    request, response = client.get('/feed')
+    assert response.status == 200
+    assert response.text == 'hello!'
```

ashleysommer:
> I still receive the request timeout with this branch.

Ok, I see that now too.

I believe this behaviour is actually correct. A request timeout is triggered when the client is taking too long to send the request to the server. In this case, the timeout is occurring _before_ the websocket handshake is completed, so the protocol object does not yet have a `websocket`, so it is treated like an ordinary HTTP request, so the timeout throws an error as expected. If the request timeout occurrs _after_ the websocket handshake, then the protocol object does have a `websocket` object so the timeout is ignored as intended. 

With that said, this PR still fixes #976 because ResponseTimeout and KeepAliveTimeout are always called _after_ a websocket handshake, so those timeouts will always be ignored on a websocket connection.

Please let me know if that makes sense and is valid logic.

Issue #977, opened 2017-10-16
Implement 'endpoint' attribute of request object


olexandr-klymenko:
In my use case I need to have the endpoint attribute in request object which should contain name of coro (1-st argument of endpoint coro) similarly to Flask request.endpoint, i.e.:

```
@app.middleware('response')
def _after_request(request, response):

    def make_request_id():
        return '.'.join([request.endpoint, request.method.lower(), str(uuid4())])

    request_id = make_request_id()
    response.headers[REQUEST_ID_HEADER] = request_id
    return response

```
endpoint can be automatically created with blueprint decorator:

```
@blueprint.get("", strict_slashes=True)
async def get_profile(request):
    return response.json({
        'user_id': '0',
        'profile_id': 0,
        'nickname': 'Nickname'
    }, 200)
```
i.e. in this case _request.endpoint_ is '**_get_profile_**'

ahopkins:
Done.

PR #979, opened 2017-10-17
add handler name to request as endpoint

r0fls:
Fixes: #977

### DIFF
--- a/docs/sanic/request_data.md
+++ b/docs/sanic/request_data.md
@@ -120,3 +120,19 @@ args.get('titles') # => 'Post 1'
 
 args.getlist('titles') # => ['Post 1', 'Post 2']
 ```
+
+## Accessing the handler name with the request.endpoint attribute
+
+The `request.endpoint` attribute holds the handler's name. For instance, the below
+route will return "hello".
+
+```python
+from sanic.response import text
+from sanic import Sanic
+
+app = Sanic()
+
+@app.get("/")
+def hello(request):
+    return text(request.endpoint)
+```
--- a/docs/sanic/request_data.md
+++ b/docs/sanic/request_data.md
@@ -136,3 +136,24 @@ app = Sanic()
 def hello(request):
     return text(request.endpoint)
 ```
+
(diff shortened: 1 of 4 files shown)


docs/sanic/request_data.md
```diff
@@ -120,3 +120,19 @@ args.get('titles') # => 'Post 1'
 
 args.getlist('titles') # => ['Post 1', 'Post 2']
 ```
+
+## Accessing the handler name with the request.endpoint attribute
+
+The `request.endpoint` attribute holds the handler's name. For instance, the below
+route will return "hello".
```
messense: I think in a blueprint, it should also include the blueprint's name? For example: `foo.hello`, `bar.hello`

r0fls: I like that idea

r0fls:
@messense if you see a better way of implementing this with the blueprint name, let me know. I'm not fond of the if statement, as it will be executed on each request.

messense:
Can we make `request.endpoint` a lazy property on `request` class? Ideally it should only be constructed when accessing it.

sanic/app.py
```diff
@@ -247,6 +247,12 @@ def websocket(self, uri, host=None, strict_slashes=None,
         def response(handler):
             async def websocket_handler(request, *args, **kwargs):
                 request.app = self
+                if not getattr(handler, '__blueprintname__', False):
```
profuel: will you please make this (and used in converted_response_type) portion a function.
Also I would drop 'else' statement there, so it looks like:
def _get_endpoint(handler):
      if not getattr(handler, '__blueprintname__', False):
         return handler.__name__
      return getattr(handler,
                                       '__blueprintname__',
                                      '') + handler.__name__

and used as 
request.endpoint = _get_endpoint(handler)

ahopkins:
@r0fls Are we keeping this? Can you either update to resolve conflicts or close it?

ahopkins:
I updated this PR to current master, made a few tweaks, added some tests, but was not allowed to push to this branch. Therefore, I am closing and will submit a new PR.

PR #997, opened 2017-10-31
tests: do not assume that locahost == 127.0.0.1

ignatenkobrain:
Signed-off-by: Igor Gnatenko <ignatenkobrain@fedoraproject.org>


ignatenkobrain:
CI check seems to fail, but it doesn't seem to be related to my changes..

r0fls:
rerunning the job did the trick

PR #1005, opened 2017-11-02
Add strict_slashes to {app, blueprint}.static()

joar:


sanic/static.py
```diff
@@ -18,7 +18,7 @@
 
 def register(app, uri, file_or_directory, pattern,
              use_modified_since, use_content_range,
-             stream_large_files, name='static', host=None):
+             stream_large_files, name='static', host=None, strict_slashes=None):
```
r0fls: Add a newline here before `strict_slashes` to fix the tests: `sanic/static.py:21:80: E501 line too long (80 > 79 characters)`

sanic/static.py
```diff
@@ -103,7 +104,7 @@ def register(app, uri, file_or_directory, pattern,
                     if isinstance(stream_large_files, int):
                         threshold = stream_large_files
                     else:
-                        threshold = 1024*1000
+                        threshold = 1024 * 1000
```
joar: I found this one while acting on https://github.com/channelcat/sanic/pull/1005#discussion_r148696110

sanic/static.py
joar: I had my editor set to 80 character line width, fixed my editor!

When reformatting, i discovered another PEP8 violation, and then a bug in the definition of what i assume is intended to be a default threshold of 1 MiB, I can file a separate PR for the default threshold if requested.

sanic/blueprints.py
```diff
@@ -221,8 +221,12 @@ def static(self, uri, file_or_directory, *args, **kwargs):
         name = kwargs.pop('name', 'static')
         if not name.startswith(self.name + '.'):
             name = '{}.{}'.format(self.name, name)
-
         kwargs.update(name=name)
+
+        strict_slashes = kwargs.pop('strict_slashes', None)
```
r0fls: Can you use `get` instead of `pop` here? No reason to mutate the `kwargs` in this case.

joar: Fixed!

PR #1013, opened 2017-11-10
Add support of custom routers rules

VanyaDNDZ:


sanic/router.py
```diff
@@ -65,17 +65,19 @@ def my_route_with_type(request, my_param: my_type):
     routes_always_check = None
     parameter_pattern = re.compile(r'<(.+?)>')
 
-    def __init__(self):
+    def __init__(self, converters=None):
```
profuel: I think it's better to make converters mandatory and pass whatever is required in each case.

VanyaDNDZ: Not all users should redefine that behavior.

profuel: True, but current implementation breaks DIP design.
And your implementation makes design even harder, since there's some predefined list which you're updating.
Imagine that tomorrow you need another rules or even another order of rules.
*Though this would be a BC break, and should be handled properly during release.*

r0fls: I prefer it as is, since that pattern is common practice in python and throughout Sanic. Though admittedly I had never heard of DIP design before this, so I may be missing the benefit.

VanyaDNDZ: I can suggest solution with default converter which equal to REGEX_TYPES.

r0fls:
@VanyaDNDZ can you add something to the docs about how to use this?

sanic/router.py
```diff
@@ -65,17 +66,17 @@ def my_route_with_type(request, my_param: my_type):
     routes_always_check = None
     parameter_pattern = re.compile(r'<(.+?)>')
 
-    def __init__(self):
+    def __init__(self, converters=REGEX_TYPES):
```
mikekeda: From point of usability, will it be better to define custom_converters?
By custom_converters I mean extending string, int, number, alpha, path converters.

```
def __init__(self, custom_converters =None):
    self._converters = dict(REGEX_TYPES)
    if  custom_converters:
        self._converters.update(custom_converters)
```

sjsadowski:
I'm closing as the PR author failed to follow through on request from reviewers for additional documentation.

vltr:
It's a pitty to close this PR, honestly. This might have been a great addition to Sanic. I'm even working on something similar on a side project. @sjsadowski perhaps I can continue where @VanyaDNDZ left?

sjsadowski:
I agree, but if we're going to add functionality we should have two things: unit tests for coverage and documentation of the new functionality. If we don't have either of those things, I don't think we can responsibly accept a PR.

vltr:
Yes, I know. All I'm saying is that I can work on these missing points to provide a proper PR.

sjsadowski:
Sounds like a good option to me.

PR #1026, opened 2017-11-18
 Add ability to use string path for `sock` instead of forcing user to manually create socket

davidtgq:
Apparently unix domain sockets are faster than localhost loopbacks: https://stackoverflow.com/a/29436429/4698922

Currently, if a string is used as the `sock` in `app.run()`, there's no indication that anything is wrong... it just has no effect and defaults to loopback on localhost. Uvloop has the `loop.create_unix_server` specifically for use with a string, it checks the path and creates the proper socket instance.


davidtgq:
dang I suck at git, how do I remove some of those irrelevant commits under the OP?

davidtgq:
Hmm might not be worth implementing this because uvloop still requires something like a `PurePosixPath` rather than a simple string, might as well just create the socket manually at that point. Closing...

PR #1035, opened 2017-11-26
Adopt new websockets interface

yunstanford:
Similar to this fix https://github.com/channelcat/sanic/commit/bca1e084116335fd939c2ee226070f0428cd5de8

Ref: https://github.com/aaugustin/websockets/blob/4.0/websockets/protocol.py#L696

### DIFF
--- a/sanic/worker.py
+++ b/sanic/worker.py
@@ -115,7 +115,7 @@ class GunicornWorker(base.Worker):
             coros = []
             for conn in self.connections:
                 if hasattr(conn, "websocket") and conn.websocket:
-                    coros.append(conn.websocket.close_connection(force=True))
+                    coros.append(conn.websocket.close_connection(after_handshake=False))
                 else:
                     conn.close()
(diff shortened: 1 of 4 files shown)


sanic/worker.py
```diff
@@ -115,7 +115,9 @@ def run(self):
             coros = []
             for conn in self.connections:
                 if hasattr(conn, "websocket") and conn.websocket:
-                    coros.append(conn.websocket.close_connection(force=True))
+                    coros.append(
+                        conn.websocket.close_connection(after_handshake=False)
```
r0fls: Shouldn't this only be false if the handshake fails?

https://github.com/aaugustin/websockets/blob/4.0/websockets/protocol.py#L704-L705

Or am I reading that wrong?

yunstanford: This line intends to `force` close connection after graceful timeout.

yunstanford: If `after_handshake=True`, it'll still wait for data transfer phase or so. then we can't shutdown server even after timeout. @r0fls

PR #1037, opened 2017-11-27
Boost test speed by pytest-xdist

youknowone:
And no panalty without it.

### DIFF
--- a/tests/test_url_building.py
+++ b/tests/test_url_building.py
@@ -5,7 +5,7 @@ from sanic import Sanic
 from sanic.response import text
 from sanic.views import HTTPMethodView
 from sanic.blueprints import Blueprint
-from sanic.testing import PORT as test_port, HOST as test_host
+from sanic.testing import HOST as test_host
 from sanic.exceptions import URLBuildError
 
 import string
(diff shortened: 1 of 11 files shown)


r0fls:
@youknowone can you investigate why the tests are failing? Are these working locally?

youknowone:
Nope, I can reproduce the problem. I will investigate about it (hopefully soon)

youknowone:
It seems they were different problem or I was something wrong at that time. It is fixed.

tests/test_url_building.py
```diff
@@ -15,11 +15,11 @@
 URL_FOR_ARGS2 = dict(arg1=['v1', 'v2'], _anchor='anchor')
 URL_FOR_VALUE2 = '/myurl?arg1=v1&arg1=v2#anchor'
 URL_FOR_ARGS3 = dict(arg1='v1', _anchor='anchor', _scheme='http',
-                     _server='{}:{}'.format(test_host, test_port), _external=True)
-URL_FOR_VALUE3 = 'http://{}:{}/myurl?arg1=v1#anchor'.format(test_host, test_port)
+                     _server='{}:PORT_PLACEHOLDER'.format(test_host), _external=True)
```
r0fls: Why can't you just use the previous format logic? :thinking:

r0fls:
@youknowone thanks, just have that one question left regarding the `format()` :+1:

tests/test_url_building.py
youknowone: To reach the concurrency goal, I created multiple server instances by each worker.
When you run it with xdist (with -n options in pytest), it will run each server instances with different port. So the port will be decided in runtime.

See tests/conftest.py how I injected the test client port. So the formatting must be delayed after running it.
Not sure but maybe it still can be solved in previous way with more knowledge about sanic & pytest. Any advice?

r0fls: I see, I don't have a better idea right now. You could do `{}:{}.format(HOST, '{}')` but I think it's too contrived.

PR #1047, opened 2017-12-07
Add auto reloading.

Yaser-Amiri:
Hi. I  added auto reloading mechanism to Sanic inspired with Flask and Werkzeug, but It is not compatible with Sanic test client yet.
To enable auto reloading developer must pass "auto_reload=True" to  "run" method of Sanic app.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1,3 +1,4 @@
+import os
 import logging
 import logging.config
 import re
@@ -22,6 +23,7 @@ from sanic.static import register as static_register
 from sanic.testing import SanicTestClient
 from sanic.views import CompositionView
 from sanic.websocket import WebSocketProtocol, ConnectionClosed
+import sanic.reloader_helpers as reloader_helpers
 
 
 class Sanic:
@@ -604,7 +606,7 @@ class Sanic:
     def run(self, host=None, port=None, debug=False, ssl=None,
             sock=None, workers=1, protocol=None,
             backlog=100, stop_event=None, register_sys_signals=True,
-            access_log=True):
+            access_log=True, auto_reload=False):
         """Run the HTTP Server and listen until keyboard interrupt or term
         signal. On termination, drain connections before closing.
 
@@ -638,12 +640,16 @@ class Sanic:
             host=host, port=port, debug=debug, ssl=ssl, sock=sock,
             workers=workers, protocol=protocol, backlog=backlog,
             register_sys_signals=register_sys_signals,
-            access_log=access_log)
+            access_log=access_log, auto_reload=auto_reload)
 
         try:
             self.is_running = True
             if workers == 1:
-                serve(**server_settings)
+                if os.name == 'posix' and auto_reload and \
+                    os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
+                    reloader_helpers.watchdog(2)
+                else:
+                    serve(**server_settings)
             else:
                 serve_multiple(server_settings, workers)
         except BaseException:
@@ -733,7 +739,8 @@ class Sanic:
     def _helper(self, host=None, port=None, debug=False,
                 ssl=None, sock=None, workers=1, loop=None,
                 protocol=HttpProtocol, backlog=100, stop_event=None,
-                register_sys_signals=True, run_async=False, access_log=True):
+                register_sys_signals=True, run_async=False, access_log=True,
+                auto_reload=False):
         """Helper function used by `run` and `create_server`."""
         if isinstance(ssl, dict):
             # try common aliaseses
@@ -799,14 +806,16 @@ class Sanic:
 
         if self.configure_logging and debug:
             logger.setLevel(logging.DEBUG)
-        if self.config.LOGO is not None:
+
+        if self.config.LOGO is not None and \
+            os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
             logger.debug(self.config.LOGO)
 
         if run_async:
             server_settings['run_async'] = True
 
         # Serve
-        if host and port:
+        if host and port and os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
             proto = "http"
             if ssl is not None:
                 proto = "https"
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -646,7 +646,7 @@ class Sanic:
             self.is_running = True
             if workers == 1:
                 if os.name == 'posix' and auto_reload and \
-                    os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
+                        os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
                     reloader_helpers.watchdog(2)
                 else:
                     serve(**server_settings)
@@ -808,7 +808,7 @@ class Sanic:
             logger.setLevel(logging.DEBUG)
 
         if self.config.LOGO is not None and \
-            os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
+                os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
             logger.debug(self.config.LOGO)
 
         if run_async:
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -645,8 +645,13 @@ class Sanic:
         try:
             self.is_running = True
             if workers == 1:
-                if os.name == 'posix' and auto_reload and \
-                        os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
+                if auto_reload and os.name != 'posix':
+                    # This condition must be removed after implementing
+                    # auto reloader for other operating systems.
+                    raise NotImplementedError
+
+                if auto_reload and \
+                        os.environ.get('SANIC_SERVER_RUNNING') != 'true':
                     reloader_helpers.watchdog(2)
                 else:
                     serve(**server_settings)
@@ -808,14 +813,14 @@ class Sanic:
             logger.setLevel(logging.DEBUG)
 
         if self.config.LOGO is not None and \
-                os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
+                os.environ.get('SANIC_SERVER_RUNNING') != 'true':
             logger.debug(self.config.LOGO)
 
         if run_async:
             server_settings['run_async'] = True
 
         # Serve
-        if host and port and os.environ.get('MAIN_PROCESS_RUNNED') != 'true':
+        if host and port and os.environ.get('SANIC_SERVER_RUNNING') != 'true':
             proto = "http"
             if ssl is not None:
                 proto = "https"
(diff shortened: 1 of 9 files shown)


marijngiesen:
I don't think this is something that needs to be in the Sanic core.

Why not just add a `dev.py` to your project and run that when developing?

```python
import time
from multiprocessing import Process

from os import path
from watchdog.events import FileSystemEventHandler
from watchdog.observers import Observer

HOST = "127.0.0.1"
PORT = 8000
PATH = "."


class Reloader:
    class Handler(FileSystemEventHandler):
        def __init__(self, reloader):
            self.reloader = reloader

        def on_any_event(self, event):
            self.reloader.reload()

    def __init__(self, directory, callback):
        self.observer = None
        self.process = None
        self.handler = Reloader.Handler(self)
        self.directory = path.abspath(directory)
        self.callback = callback

    def watch(self, first=True):
        if first:
            self.reload()
        else:
            self.observer = Observer()
            self.observer.schedule(self.handler, self.directory, recursive=True)
            self.observer.start()

    def reload(self):
        if self.observer:
            self.observer.stop()
            self.process.terminate()
            self.process.join()
        try:
            self.process = Process(target=self.callback)
            self.process.start()
        finally:
            self.watch(first=False)


def open_browser():
    import webbrowser
    time.sleep(2)
    webbrowser.open("{}:{}".format(HOST, PORT), new=2)


def run_server():
    from website import run
    run()


Reloader(PATH, run_server).watch()
open_browser()
while True:
    try:
        time.sleep(1)
    except KeyboardInterrupt:
        break
```

Yaser-Amiri:
I think every thing that used by every programmer should be in the core of framework, and auto reloading has this conditions, writing/adding a simple and similar code in every project is unattractive.
See Django, Flask, ... !

marijngiesen:
That might be so, but my opinion is that Sanic is a tiny, high performance framework. I think it's better to add this functionality to it by creating an extension.

Just my 2 cents..

ashleysommer:
@Yaser-Amiri 
Also probably not a good idea to introduce `requests` as a new dependency for running tests.

Yaser-Amiri:
@ashleysommer 
Yes, you're right! I will fix it!

sanic/reloader_helpers.py
```diff
@@ -0,0 +1,120 @@
...
+    :param proc: worker process (process ID)
+    :return: Nothing
+    """
+    kill_process_children_unix(proc.pid)
+    proc.terminate()
+    os._exit(0)
+
+
+def watchdog(sleep_interval):
+    """Whatch project files, restart worker process if a change happened.
```
seemethere: Typo here, `Whatch` -> `Watch`

sanic/reloader_helpers.py
```diff
@@ -0,0 +1,120 @@
...
+
+    for child_pid in children_list_pid:
+        children_proc_path = "/proc/%s/task/%s/children" % \
+            (child_pid, child_pid)
+        if not os.path.isfile(children_proc_path):
+            continue
+        with open(children_proc_path) as children_list_file_2:
+            children_list_pid_2 = children_list_file_2.read().split()
+        for _pid in children_list_pid_2:
+            os.kill(int(_pid), signal.SIGTERM)
```
seemethere: Is this compatible with windows?

sanic/reloader_helpers.py
```diff
@@ -0,0 +1,120 @@
...
+    return rv
+
+
+def restart_with_reloader():
+    """Create a new process and a subprocess in it
+    with the same arguments as this one.
+    """
+    args = _get_args_for_reloading()
+    new_environ = os.environ.copy()
+    new_environ['MAIN_PROCESS_RUNNED'] = 'true'
```
seemethere: Can we change this to something `sanic` specific like `SANIC_SERVER_RUNNING`?

Also if we make it a constant then there's less of a chance of us mis-typing it somewhere else.

sanic/reloader_helpers.py
```diff
@@ -0,0 +1,120 @@
...
+    return worker_process
+
+
+def kill_process_children_unix(pid):
+    """Find and kill child process of a process (maximum two level).
+
+    :param pid: PID of process (process ID)
+    :return: Nothing
+    """
+    root_process_path = "/proc/%s/task/%s/children" % (pid, pid)
```
seemethere: can we replace with either f strings or `.format`?

Same goes for the rest of the formatting done like this.

tests/test_auto_reload.py
```diff
@@ -0,0 +1,85 @@
...
+            _file.write(sanic_project_content_two)
+        sleep(2) # wait for completing server start process
+        response = requests.get("http://127.0.0.1:8000/").json()
+        assert response == {"test": 2}
+
+        thread.join(1)
+        os.remove(sanic_app_file_path)
+
+    def teardown_method(self, method):
+            if process_id:
```
seemethere: Seems like the spacing is off here.

Yaser-Amiri: @seemethere 
Thanks for your review :)
I will fix this problems as soon as I can, but there is one point:
I'm using Linux and actually I can't write programs for windows. Now this mechanism works on Linux  and I added an condition to check platform.
If somebody add windows support to it, it will be great :)

seemethere: No problem! Thank you for your contribution!, if you could add a quick failure, raising a `NotImplementedError` should do the trick, for those on windows until we can have someone on windows verify that this method works.

Yaser-Amiri: I will work on it in this weekend!

tests/test_middleware.py
```diff
@@ -146,4 +146,4 @@ def test_middleware_order():
     request, response = app.test_client.get('/')
 
     assert response.status == 200
-    assert order == [1,2,3,4,5,6]
```
seemethere: Are all these changes to other tests necessary for this PR?

Yaser-Amiri: Actually no :/
My editor flake checker noticed me some problems and I fixed them, If you think that they should remain virgin, I'll respect that.

seemethere: If we could leave those out of this PR and put them in as a separate PR that would be awesome!

Yaser-Amiri: I fixed that, I will send a PR to solve flake problems.
If there are other this kind of problems in my code, warn me to don't repeat it in next pull requests!

r0fls:
@seemethere have all of your comments been resolved? I'd like to merge this if they have.

r0fls:
>That might be so, but my opinion is that Sanic is a tiny, high performance framework. I think it's better to add this functionality to it by creating an extension.

@marijngiesen while we stick to that philosophy in general, this wouldn't effect performance when not enabled, and is arguably a core feature. We've seen at least 3 total attempted PRs now, so it seems the community thinks it's worth adding to the core.

sanic/reloader_helpers.py
arnulfojr: I don’t think this is compatible with Windows, but I can recommend you to use a Docker container :)

Yaser-Amiri:
Please reject or accept this PR. I did the wrong thing and added auto reloading to my master branch and now sometimes I forget this PR and  I'm pushing on my master and it causes changes in this PR.

sanic/reloader_helpers.py
```diff
@@ -0,0 +1,121 @@
...
+    return rv
+
+
+def restart_with_reloader():
+    """Create a new process and a subprocess in it with the same arguments as
+    this one.
+    """
+    args = _get_args_for_reloading()
+    new_environ = os.environ.copy()
+    new_environ['SANIC_SERVER_RUNNING'] = 'true'
```
arnulfojr: I think that the name of this env variable is a bit misleading (in my point fo view). for me it sounds like if this variable is set in the server module (or by the server) instead of being set by the reloader, I would suggest renaming it to something as `SANIC_APP_REALOADING` (and changing the logic accordingly).

Hope it helps :)

r0fls:
>Please reject or accept this PR. I did a mistake and added auto reloading to my master branch and now sometimes I forget this PR and I'm pushing on my master and it causes changes in this PR.

@Yaser-Amiri please address the existing comments before pressuring for further review or acceptance

r0fls:
>I did a mistake and added auto reloading to my master branch 

You can clean that up pretty easily if you need to. For example, something like `git reset --hard origin/master`

sanic/reloader_helpers.py
Yaser-Amiri: Thanks for your review.
I think this logic is more understandable, Werkzeug uses this way too and I inspired by Werkzeug!

sanic/reloader_helpers.py
Yaser-Amiri: Done!

sanic/reloader_helpers.py
Yaser-Amiri: Done!

sanic/reloader_helpers.py
Yaser-Amiri: Done!

tests/test_auto_reload.py
Yaser-Amiri: Done!

Yaser-Amiri:
@r0fls Sorry but I'm not pressuring for acceptance! I just said it's better to close this PR(Reject or Accept) because I did a mistake is using Git! Forget it!
Thanks.

r0fls:
@seemethere you ok with merging this?

sanic/reloader_helpers.py
arnulfojr: I understand and I'm also a big fan of Werkzeug but please evaluate the name one more time :)
And if you can please give me another reason other than werkzeug using it, would be helpful for me then to understand more throughly  :)

kolapapa:
When will it be merged into master?

r0fls:
This looks good enough for me. Let's go ahead and merge it.

sanic/app.py
```diff
@@ -604,7 +606,7 @@ def test_client(self):
     def run(self, host=None, port=None, debug=False, ssl=None,
             sock=None, workers=1, protocol=None,
             backlog=100, stop_event=None, register_sys_signals=True,
-            access_log=True):
+            access_log=True, auto_reload=False):
```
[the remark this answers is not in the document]
Yaser-Amiri: I wanted to do this but it made some problems in running tests.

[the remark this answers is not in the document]
r0fls: No worries, we can always add it later

sanic/reloader_helpers.py
abuckenheimer: there are libraries that attempt to be cross platform compatible here `psutil` comes to mind but it also requires a c tool chain which is a pain

PR #1049, opened 2017-12-09
Fix root logger config

MarSoft:
According to the [doc](https://docs.python.org/3/library/logging.config.html#dictionary-schema-details), root logger configuration should be given as a separate key, not within `loggers` array.
Previous configuration would create a new (regular) logger named `root` instead of configuring root logger which in turn remained with its default `WARNING` loglevel.

### DIFF
--- a/sanic/log.py
+++ b/sanic/log.py
@@ -6,11 +6,11 @@ LOGGING_CONFIG_DEFAULTS = dict(
     version=1,
     disable_existing_loggers=False,
 
+    root={
+        "level": "INFO",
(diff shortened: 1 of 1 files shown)


ashleysommer:
You probably already noticed, but this cannot be merged because it breaks the test_logging test.
You may need to refactor that test so that it passes with the new configuration, or perhaps actually the test failing points to something wrong in the new configuration.

harshanarayana:
PR #1397 addresses these changes.

PR #1053, opened 2017-12-11
Fix ip and socket data format on V6 

danpalmer:
Currently Sanic throws errors when serving IPv6 requests. This fixes the error.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -1,5 +1,6 @@
 import sys
 import json
+import socket
 from cgi import parse_header
 from collections import namedtuple
 from http.cookies import SimpleCookie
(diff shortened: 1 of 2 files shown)


sanic/request.py
```diff
@@ -185,9 +185,18 @@ def socket(self):
         return self._socket
 
     def _get_address(self):
-        self._socket = (self.transport.get_extra_info('peername') or
-                        (None, None))
-        self._ip, self._port = self._socket
+        sock = self.transport.get_extra_info('socket')
+
+        if sock.family == socket.AF_INET:
```
r0fls: it looks like you forgot to import the socket module: `import socket`

danpalmer:
Whoops apologies! Fixed.

PR #1055, opened 2017-12-14
Cancel request tasks when response timeout is triggered

youknowone:
Before: Even after raising ResponseTimeout, server still processes
remaining tasks until it is done
After: Before raising ResponseTimeout, server stops working task.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -174,6 +174,10 @@ class HttpProtocol(asyncio.Protocol):
                                      self.response_timeout_callback)
             )
         else:
+            if self._request_stream_task:
+                self._request_stream_task.cancel()
+            if self._request_handler_task:
+                self._request_handler_task.cancel()
             try:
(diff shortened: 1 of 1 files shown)


r0fls:
@youknowone could you add a test for this?

r0fls:
Or, @ashleysommer would you be able to help review this?

ashleysommer:
@r0fls  @youknowone 
This looks good to me.
Actually, I believe this was an oversight or bug when I wrote the new timeout handlers.

Before this PR, the `request_timeout_callback` handler has the request task `cancel()` calls, but the ` response_timeout_callback` does not have them. This should have been the other way around. My modification should have had the `cancel()` calls in `response_timeout_callback` but not in `request_timeout_callback`.

However it is probably a good idea to have the `cancel()` calls in both timeouts as this PR provides.

PR #1063, opened 2017-12-22
try to inject the app in add_task method

### LINKED ISSUE #1062: Accessing app in a background task
The example in the docs shows a nice way of scheduling a long-running task:

```python
async def notify_server_started_after_five_seconds():
    await asyncio.sleep(5)
    print('Server successfully started!')

app.add_task(notify_server_started_after_five_seconds())
```

If I've stashed my `notify_server_started_after_five_seconds()` function in `tasks.py`, What's the best way to access the `app` instance from it?

I've looked at dirty solutions like making `app` global... but I'm surprised that `app` and `loop` don't get injected like they do on the decorated methods.

Thanks in advance!

r0fls:
fixes #1062

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -88,14 +88,20 @@ class Sanic:
         """
         try:
             if callable(task):
-                self.loop.create_task(task())
+                try:
+                    self.loop.create_task(task(self))
+                except TypeError:
+                    self.loop.create_task(task())
             else:
                 self.loop.create_task(task)
         except SanicException:
             @self.listener('before_server_start')
             def run(app, loop):
                 if callable(task):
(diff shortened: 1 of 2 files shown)


dvf:
This is great! 

If `loop.app` is always available, should we even accept `loop` as a secondary arg in the decorated tasks?

r0fls:
That is a good question. Seems the primary reason is that the `app.loop` property waits to check if the loop is running. Not sure why that's there. Shouldn't be necessary.

dvf:
Should we update the docs?

r0fls:
https://github.com/channelcat/sanic/pull/1071

PR #1073, opened 2017-12-29
remove uvloop for windows setup

Kurlov:
Hello

This PR matters for windows users only. I understand that windows user might provide `SANIC_NO_UVLOOP=True` var, but I guess it would be convenient to remove `uvloop` from `setup.py` for windows by default. Just to make it a little easy for tinkering/studying.

Currently, `pip install sanic` leads to an error on windows since uvloop hasn't implemented yet for this platform.


fourjr:
Could we do a release for this?

PR #1081, opened 2018-01-07
Fix: the Chinese URI

howie6879:
Hello, 
I have a question about routing, for example:
``` python
from sanic import Sanic
from sanic.response import text, redirect


app = Sanic()


@app.route("/你好")
async def test(request):
    return text('你好')

if __name__ == "__main__":

# Visit:http://0.0.0.0:8000/你好
# Output: Error: Requested URL /%E4%BD%A0%E5%A5%BD not found
```
Maybe we could use the unquote() to solve this problem

``` python
# Visit:http://0.0.0.0:8000/你好
# Output: 你好
```


sanic/router.py
```diff
@@ -373,6 +374,7 @@ def _get(self, url, method, host):
         :param method: request method
         :return: handler, arguments, keyword arguments
         """
+        url = unquote(url)
         url = host + url
```
r0fls: can you switch this to `url = unquote(host + url)`?

r0fls:
@howie6879 can you also please add the below unit test to `tests/test_routes.py`:

```python
def test_unicode_routes():                                                                                                  
    app = Sanic('test_unicode_routes')                                                                                      
                                                                                                                            
    @app.get('/你好')                                                                                                       
    def handler(request):                                                                                                   
        return text('OK')                                                                                                   
                                                                                                                            
    request, response = app.test_client.get('/你好')                                                                        
    assert response.text == 'OK'   
```

howie6879:
@r0fls Done😊

PR #1087, opened 2018-01-13
Fix typo

Stranger6667:
There is no `_get_socket` method, that sets `_socket` attribute, but there is `_get_address` method, that does the tsing (and sets `_ip` and `_port` as well)


r0fls:
Thanks!

PR #1094, opened 2018-01-19
update class_based_views

caitinggui:
Decorating single method is necessary sometimes, for example:

```
GET /users    # get user info  and need to login
POST /users   # regist user, and don't need to login
```

If you use login_required decorator in HTTPMethodView, both methods need to check if logined, but we just want to check if logined in "GET".

### DIFF
--- a/docs/sanic/class_based_views.md
+++ b/docs/sanic/class_based_views.md
@@ -98,7 +98,7 @@ class ViewWithDecorator(HTTPMethodView):
 app.add_route(ViewWithDecorator.as_view(), '/url')
 ```
 
-But if you just want to decorator some functions and not all functions, you can use as followed.
+But if you just want to decorate some functions and not all functions, you can do as follows:
 
 ```python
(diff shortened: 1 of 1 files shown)


docs/sanic/class_based_views.md
```diff
@@ -92,10 +92,27 @@ class ViewWithDecorator(HTTPMethodView):
...
     return text('Hello I have a decorator')
 
+  def post(self, request, name):
+    return text("Hello I also have a decorator")
+
 app.add_route(ViewWithDecorator.as_view(), '/url')
 ```
 
-#### URL Building
+But if you just want to decorator some functions and not all functions, you can use as followed.
```
r0fls: Can you please make a small change to make this say `But if you just want to decorate some functions and not all functions, you can do as follows:` instead?

caitinggui: Of course.

PR #1097, opened 2018-01-22
Add parameter check

howie6879:
Hi there:

There is a basic route looks like the following:

``` python
from sanic import Sanic

app = Sanic()


@app.route("/")
async def test():
    return text('Hello World!')


if __name__ == "__main__":
```
It can be seen that despite the lack of the request parameter service can still run.

When you write a route, it is likely to forget to add a parameter called `request` in the function.

Maybe we can add a parameter check mechanism:

``` python
python  app.py

ValueError: Required parameter `request` missing in the test()
```

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -154,7 +154,9 @@ class Sanic:
                                 version=version, name=name)
                 return handler
             else:
-                raise ValueError('Required parameter `request` missing in the {0}()'.format(handler.__name__))
+                raise ValueError(
+                    'Required parameter `request` missing in the {0}()'.format(
+                        handler.__name__))
 
         return response
 
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -5,7 +5,7 @@ import warnings
 from asyncio import get_event_loop, ensure_future, CancelledError
 from collections import deque, defaultdict
 from functools import partial
-from inspect import isawaitable, stack, getmodulename
+from inspect import getmodulename, isawaitable, signature, stack
 from traceback import format_exc
 from urllib.parse import urlencode, urlunparse
 from ssl import create_default_context, Purpose
@@ -25,7 +25,6 @@ from sanic.websocket import WebSocketProtocol, ConnectionClosed
 
 
 class Sanic:
-
     def __init__(self, name=None, router=None, error_handler=None,
                  load_env=True, request_class=None,
                  strict_slashes=False, log_config=None,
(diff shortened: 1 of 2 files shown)


bosskwei:
Well, I think this check is never needed, some ide will warn an unused warning if we do not use the request param next. So let the user decide whether to keep this param.

howie6879:
Hi @bosskwei : 

Maybe you don't understand what I mean, as this example, although the service can be successfully started, but when I visit `http://0.0.0.0:8000/`, The program will throw a error:

``` python
TypeError: test() takes 0 positional arguments but 1 was given
```

So this parameter is necessary

bosskwei:
@howie6879 My wrong, it's necessary..😅

yunstanford:
It's pretty normal and conventional to have `request` as param. It's not very valuable to add extra overhead cost for checking this. And it might also break some libs built on top of that by breaking the convention.

r0fls:
>It's not very valuable to add extra overhead cost for checking this.

Well, it's only happening during the route registration, so it wouldn't effect the actual response time. Might be worth it to add something like this to make the error more intuitive.

yunstanford:
Ignore that part, :). Didn't look carefully. But still feel like it's not very good to break the convention. And I don't see a lot benefits. It's just my opinion. :). If you feel like it's useful, go ahead and get it merged

howie6879:
@yunstanford @r0fls : 
As the routing increases,  it prevents the program from discovering errors while accepting the request

yunstanford:
Sorry, I didn't get your point

yunstanford:
Ah, sorry.. I think I misunderstood the PR and what you described. 

I was totally misunderstood what you described.. 

Sorry, the PR looks good.

howie6879:
@yunstanford 

I apologize for my ambiguous description.

I mean it is better to check this parameter earlier......

And i need to learn English well... :100:

sanic/app.py
```diff
@@ -143,12 +144,19 @@ def route(self, uri, methods=frozenset({'GET'}), host=None,
...
+                if stream:
+                    handler.is_stream = stream
+
+                self.router.add(uri=uri, methods=methods, handler=handler,
+                                host=host, strict_slashes=strict_slashes,
+                                version=version, name=name)
+                return handler
+            else:
+                raise ValueError(
+                    'Required parameter `request` missing in the {0}()'.format(
```
r0fls: Can you make this say ```Required parameter `request` missing in the {0}() route```? Otherwise LGTM

howie6879: Done

PR #1101, opened 2018-01-27
Provide information about sanic-oauth extension

SirEdvin:
Old Sanic-OAuth repository not updated and don't release real version, so I create another sanic-oauth extension, could you please add it to extension list?


abuckenheimer:
I'm not really sure how this library counts as a sanic extension? It doesn't really interact with sanic in any meaningful way

SirEdvin:
@abuckenheimer Thanks, that you notice this! I just add base integration with sanic, that will help you to use this library.

Issue #1093, opened 2018-01-18
KeepAlive Timeout: what is it?


smlbiobot:
Since upgrading from 0.5.0 to 0.7.0 I am seeing a lot of these in my logs:

```
[2018-01-18 17:37:16 +0800] [5425] [INFO] KeepAlive Timeout. Closing connection.
[2018-01-18 17:37:16 +0800] [5425] [INFO] KeepAlive Timeout. Closing connection.
[2018-01-18 17:37:16 +0800] [5425] [INFO] KeepAlive Timeout. Closing connection.
[2018-01-18 17:37:16 +0800] [5425] [INFO] KeepAlive Timeout. Closing connection.
[2018-01-18 17:37:16 +0800] [5425] [INFO] KeepAlive Timeout. Closing connection.
[2018-01-18 17:37:16 +0800] [5425] [INFO] KeepAlive Timeout. Closing connection.
```

What are these and are they normal or is there something in my code that I need to fix?

arthurlenoir:
Same here but everything's going well anyway.  ¯\_(ツ)_/¯

frnkvieira:
They are perfectly normal (in most cases). In HTTP/1.1 connections are keep-alive by default which means your browser keeps the connection open after it made all the requests. After some time if you don't request a new page this connection will timeout. Not sure why Sanic devs log such thing in INFO mode. IMHO this would be a DEBUG level since most users don't even understand what's going on.

smlbiobot:
Also, I kept on getting these, which also didn’t use to happen:

```
WARNING:asyncio:Executing <Handle <TaskWakeupMethWrapper object at 0x108fbf2e8> created at /usr/local/Cellar/python3/3.6.4_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/tasks.py:639> took 0.828 seconds
```

What am I being warned about exactly? The site runs fine and I don’t quite understand what I am supposed to fix.

arnulfojr:
As @frnkvieira said, this are perfectly normal messages, you can take a look into the [docs](http://sanic.readthedocs.io/en/latest/sanic/config.html?highlight=keep%20alive#what-is-keep-alive-and-what-does-the-keep-alive-timeout-value-do) for more information.
I would only disagree in the logging level that frnkviera says, as this is configurable (can be switched off) makes sense to have it in INFO level, as well as it is a connection-related information.

yunstanford:
I'm also thinking this could be DEBUG level

[...]

PR #1104, opened 2018-01-30
KeepAlive Timeout log level change to debug

arnulfojr:
As discussed in [this thread](https://github.com/channelcat/sanic/issues/1093), I just made the logging change from `INFO` to `DEBUG` level when the Keep Alive Times out.

FYI @ashleysommer

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -195,7 +195,7 @@ class HttpProtocol(asyncio.Protocol):
                                      self.keep_alive_timeout_callback)
             )
         else:
-            logger.info('KeepAlive Timeout. Closing connection.')
+            logger.debug('KeepAlive Timeout. Closing connection.')
             self.transport.close()
             self.transport = None
 


ashleysommer:
LGTM 👍 
But I would still like to see input from others about this too, some may argue that is should stay at INFO level.

r0fls:
I think `DEBUG` level makes sense, epsecially if it's confusing people.

arnulfojr:
For my point of view, it’s an `INFO` log level, but i saw that a lot of people are strung confused so maybe is best if we set this as debug and we avoid getting issues that ask about that. Specially now that Asyncio is just taking off and people are getting on board and there’s going to be a lot of new devs into sanic.
In this sort of thinking, I give more weight on being clear+transparent to the community :)

r0fls:
Yep, 👍  I will merge this soon unless someone provides an argument against it.

PR #1109, opened 2018-02-01
fixed bug in multipart/form-data parser

DirkGuijt:
Sanic automatically assumes that a form field is a file if it has a content-type header, even though the header is text/plain or application/json. This is a fix for it, I took into account the RFC7578 specification regarding the defaults.

This is an improvement of the incorrect fix in PR #869

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -18,7 +18,7 @@ except ImportError:
         json_loads = json.loads
 
 from sanic.exceptions import InvalidUsage
-from sanic.log import error_logger
+from sanic.log import error_logger, logger
 
 DEFAULT_HTTP_CONTENT_TYPE = "application/octet-stream"
 
@@ -309,18 +309,21 @@ def parse_multipart_form(body, boundary):
                 content_type = form_header_value
                 content_charset = form_parameters.get('charset', 'utf-8')
 
-        post_data = form_part[line_index:-4]
-        if file_name:
-            file = File(type=content_type, name=file_name, body=post_data)
-            if field_name in files:
-                files[field_name].append(file)
+        if field_name:
+            post_data = form_part[line_index:-4]
+            if file_name:
+                file = File(type=content_type, name=file_name, body=post_data)
+                if field_name in files:
+                    files[field_name].append(file)
+                else:
+                    files[field_name] = [file]
             else:
-                files[field_name] = [file]
+                value = post_data.decode(content_charset)
+                if field_name in fields:
+                    fields[field_name].append(value)
+                else:
+                    fields[field_name] = [value]
         else:
-            value = post_data.decode(content_charset)
-            if field_name in fields:
-                fields[field_name].append(value)
-            else:
-                fields[field_name] = [value]
+            logger.debug('Form-data field does not have a name parameter in the Content-Disposition header')
 
     return fields, files
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -312,8 +312,9 @@ def parse_multipart_form(body, boundary):
         if field_name:
             post_data = form_part[line_index:-4]
             if file_name:
-                form_file = \
-                    File(type=content_type, name=file_name, body=post_data)
+                form_file = File(type=content_type,
+                                 name=file_name,
+                                 body=post_data)
                 if field_name in files:
                     files[field_name].append(form_file)
                 else:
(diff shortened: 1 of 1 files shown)


sanic/request.py
```diff
@@ -304,19 +305,21 @@ def parse_multipart_form(body, boundary):
             if form_header_field == 'content-disposition':
                 if 'filename' in form_parameters:
                     file_name = form_parameters['filename']
-                field_name = form_parameters.get('name')
+                field_name = form_parameters['name']
```
r0fls: please leave this as `field_name = form_parameters.get('name')` to avoid throwing an error here and causing the sanic process to die

sanic/request.py
```diff
@@ -304,19 +305,21 @@ def parse_multipart_form(body, boundary):
...
             elif form_header_field == 'content-type':
-                file_type = form_header_value
+                content_type = form_header_value
+                if 'charset' in form_parameters:
+                    content_charset = form_parameters['charset']
```
r0fls: let's just do `content_charset = form_parameters.get('charset')` and remove the `if 'charset' in form_parameters:`

r0fls: then it will default to `utf-8` if not specified

sanic/request.py
DirkGuijt: Okay, good. The "name" parameter is required in the Content-Disposition header according to the specification, but it is indeed nice to not cause a crash when it is missing of course.

DirkGuijt: It is now weird when it is missing though... There will be a 'None' key in the files/form dictionary.

sanic/request.py
DirkGuijt: I'll change it :)

DirkGuijt:
Also, #PR869 can be closed if this will be merged.

sanic/request.py
r0fls: hmm is there a different default that makes sense? Just don't want the server to crash even if it's a rare edge case

DirkGuijt: I think we should not allow the field/file to end up in the request at all if there is no 'name' parameter in the Content-Disposition header. Like I said, it is required by the spec.

We can't define a default value because you should be able to POST any field name in the form-data. For example, if we say the default value is the string value 'None', then the behavior will be weird when somebody wants to use 'None' as a field name and has a form-field without a 'name'. It's a big edge case of course, but still...

DirkGuijt: I made a commit to this PR to omit the form field from the request when it does not have a name parameter. When it is omitted, there is a log sent to the logger.DEBUG. I wasn't sure what the logging policy is though, should it go to error_logger instead?

sanic/request.py
```diff
@@ -302,24 +303,28 @@ def parse_multipart_form(body, boundary):
...
-            file = File(type=file_type, name=file_name, body=post_data)
-            if field_name in files:
-                files[field_name].append(file)
+                content_type = form_header_value
+                content_charset = form_parameters.get('charset', 'utf-8')
+
+        if field_name:
+            post_data = form_part[line_index:-4]
+            if file_name:
+                file = File(type=content_type, name=file_name, body=post_data)
```
r0fls: `file` is a keyword in python. Can you replace this with a different variable name, to avoid overshadowing that? e.g. response_file

DirkGuijt: erh, yeah sure 👍

DirkGuijt:
Does this need more changes/reviews?

sanic/request.py
```diff
@@ -302,24 +303,29 @@ def parse_multipart_form(body, boundary):
...
-            file = File(type=file_type, name=file_name, body=post_data)
-            if field_name in files:
-                files[field_name].append(file)
+                content_type = form_header_value
+                content_charset = form_parameters.get('charset', 'utf-8')
+
+        if field_name:
+            post_data = form_part[line_index:-4]
+            if file_name:
+                form_file = \
```
r0fls: Please change this to use newlines (see my comment outside of here, because it won't let me put newlines in this comment)

r0fls:
Please use the below spacing instead:

```python
form_file = File(type=content_type,
                 name=file_name,
                 body=post_data)
```

DirkGuijt:
Done.

Issue #1102, opened 2018-01-28
How to response 204 using response.json() 


alaminopu:
Usually the `DELETE` request should return a [204](https://httpstatuses.com/204) response. How can I send 204 response using sanic. I was trying something like below. 

```python
return response.json(
        None,
        status=204
    )
```

But it doesn't work, it response with null.

ahopkins:
That looks right to me. Can you post the raw response. Are you sure you are not getting the right status code? I do something similar, albeit with `response.text`

    text('', status=204)

Works just fine.

alaminopu:
If I use `response.text()` it's working fine. Why can't get the same response using `response .json()`? 

The following response with 204 but with `""` string in the response body.  For 204 response, content-length should be 0. 

```python
return response.json(
       '',
        status=204
    )
```

ahopkins:
Try this:

    return response.json(None, status=204)

alaminopu:
As I mentioned in my first comment. the following doesn't work. 

```python
return response.json(None, status=204)
```

It return a response with 204 but with `null` in the response body

r0fls:
@alaminopu are you using a `delete` route? With the below example, I do get a 0 length response:

```python
from sanic import Sanic
from sanic.response import json

app = Sanic()

@app.delete("/")
async def test(request):
    return json(None, status=204)

if __name__ == "__main__":
    app.run(host="localhost", port=8010)
```

```
$ curl -s -XDELETE localhost:8010 | wc -l
0
```

However, the `Content-Length` header is incorrect.

```
 curl -v -XDELETE localhost:8010
* Rebuilt URL to: localhost:8010/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8010 (#0)
> DELETE / HTTP/1.1
> Host: localhost:8010
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 204 No Content
< Connection: keep-alive
< Keep-Alive: 5
< Content-Length: 4
< Content-Type: application/json
< 
* Excess found in a non pipelined read: excess = 4 url = / (zero-length body)
* Connection #0 to host localhost left intact
```

arnulfojr:
[...]

PR #1110, opened 2018-02-01
Issue: 1102 HTTP 204 status code

arnulfojr:
According to [Issue 1102](https://github.com/channelcat/sanic/issues/1102) this is a proposal as for how the HTTPResponse should behave when returning an HTTP status code 204 (No-Content)

Previous behavior:
- When passing `None` in the json response method with status 204:
    + Returns a Response with content length header length `4`  (from `len('null')`)

Current behavior:
- On HTTP status code 204, no body will be sent and content length header will be set to 0, regardless of the body passed.

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -239,7 +239,8 @@ def json(body, status=200, headers=None,
     :param headers: Custom Headers.
     :param kwargs: Remaining arguments that are passed to the json encoder.
     """
-    return HTTPResponse(dumps(body, **kwargs), headers=headers,
+    _body = dumps(body, **kwargs) if body else None
+    return HTTPResponse(_body, headers=headers,
                         status=status, content_type=content_type)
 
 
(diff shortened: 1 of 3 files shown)


sanic/response.py
```diff
@@ -218,7 +222,7 @@ def output(
                    b'keep-alive' if keep_alive else b'close',
                    timeout_header,
                    headers,
-                   self.body
+                   self.body if self.status is not 204 else b''
```
r0fls: Instead of this can you do the same thing you did with content-length? So, move a variable `body = ''` up there to line ~200 and in `if self.status is not 204:` set `body=self.body`. Finally use `body` here instead of `self.body`

r0fls: That way we do only one check of the status code instead of 2

h-wei: @r0fls It's not bug, just fix content-length = 0 for status code is 204

arnulfojr: @r0fls sure, will do! thanks for the feedback :)

arnulfojr:
@r0fls @h-wei I'll close this PR as also the status code 304 is behaving in the same way and in this I only consider 204.

Issue #728, opened 2017-05-15
Trying to send 304 response results in unwanted headers


mike-hart:
An HTTP 304 (Not Modified) response should be zero length and should contain headers appropriate to the 200 that would have resulted otherwise.

There does not seem to be a way to send a response without either a Content-Length or Content-Type header (neither of which are appropriate for a zero-length response) and either of which could upset the cache content if the cache were to update its headers from this response.

cf. https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.5

r0fls:
>The 304 response MUST NOT contain a message-body

arnulfojr:
I can see that this relates to the same [issue 1102](https://github.com/channelcat/sanic/issues/728) @r0fls shall I include this status code in there as well?

r0fls:
closing per #1113

mike-hart:
Hi, it looks like #1113 still emits a Content-Length header for a 304, but the RFC states at the very least that it SHOULD NOT, and in particular cases that it MUST NOT:

> If the conditional GET used a strong cache validator (see section 13.3.3), the response SHOULD NOT include other entity-headers. Otherwise (i.e., the conditional GET used a weak validator), the response MUST NOT include other entity-headers; this prevents inconsistencies between cached entity-bodies and updated headers.

The headers it permits are Date, ETag, Vary, Content-Location, Expires, and Cache-Control.

Have I misread the RFC?

r0fls:
Thanks @mike-hart

r0fls:
Sounds like it should remove these other ones as well for 304 responses, `Connection`, `Keep-Alive` and `Content-Type`

```
< HTTP/1.1 304 Not Modified
< Connection: keep-alive
< Keep-Alive: 5
< Content-Type: application/json
```

mike-hart:
I'm not sure there should necessarily be an outright removal of things, but perhaps the option of having a user-added-only set of headers for a response would be good. The dev can break as many rules as they please, but the framework won't make that any worse for them :)

mike-hart:
Status code 206 for example may be similarly afflicted.

arnulfojr:
Hmmm... well I guess that removing/adding certain headers per http response is sort of a bigger PR than the one I did, as the one PR I did was basically only covering the content-length mismatch.

[...]

PR #1113, opened 2018-02-02
Content Length header on 204 and 304 responses

arnulfojr:
## Issues
According to [Issue 1102](https://github.com/channelcat/sanic/issues/1102) this is a proposal as for how the HTTPResponse should behave when returning an HTTP status code 204 (No-Content)
According to [Issue 728](https://github.com/channelcat/sanic/issues/728) this is a proposal as for how the HTTPResponse should behave when returning an HTTP status code 304 (Not-Modified)

## Reference
[Reference for 1102](https://tools.ietf.org/html/rfc2616#section-10.2.5)
[Reference for 728](https://tools.ietf.org/html/rfc2616#section-10.3.5)

## Behavior
Previous behavior:
- When passing `None` in the json response method with status 204:
    + Returns a Response with content length header length `4`  (from `len('null')`)
- When returning a Response status code 304 with or without body:
    + Returns content length header of length `> 0`

Current behavior:
- On HTTP status code 204, no body will be sent and content length header will be set to 0, regardless of the body passed.
- On HTTP status code 304, no body will be sent and content length header will be set to 0, regardless of the body passed.

### Note
This PR substitutes [PR 1110](https://github.com/channelcat/sanic/pull/1110)

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -239,7 +239,8 @@ def json(body, status=200, headers=None,
     :param headers: Custom Headers.
     :param kwargs: Remaining arguments that are passed to the json encoder.
     """
-    return HTTPResponse(dumps(body, **kwargs), headers=headers,
+    _body = dumps(body, **kwargs) if body else None
+    return HTTPResponse(_body, headers=headers,
                         status=status, content_type=content_type)
 
 
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -196,11 +196,13 @@ class HTTPResponse(BaseHTTPResponse):
         if keep_alive and keep_alive_timeout is not None:
             timeout_header = b'Keep-Alive: %d\r\n' % keep_alive_timeout
 
+        body = b''
         content_length = 0
         if self.status is not 204:
+            body = self.body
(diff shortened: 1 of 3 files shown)


sanic/response.py
```diff
@@ -195,8 +195,14 @@ def output(
         timeout_header = b''
         if keep_alive and keep_alive_timeout is not None:
             timeout_header = b'Keep-Alive: %d\r\n' % keep_alive_timeout
-        self.headers['Content-Length'] = self.headers.get(
-            'Content-Length', len(self.body))
+
+        body = b''
+        content_length = 0
+        if self.status is not 204 and self.status != 304:
```
r0fls: Whoops, I approved this already but a minor nitpick here: `if self.status is not 204 and self.status != 304` can we use the same comparison operator? Or, another idea is to put these into a constant list at the top: `EMPTY_RESPONSE_CODES = [204, 304]` and then do `if self.status not in EMPTY_RESPONSE_CODES`. This way we can add more response codes easily if needed.

arnulfojr: Sure I will fix them, I’ll do the empty list thing :)

PR #1114, opened 2018-02-02
do not include content-length header in 304

r0fls:
fix: #728


r0fls:
looks like more thought will have to be put into this, as there are other headers that should be excluded and need to be removed in server.py rather than respone.py.

Issue #916, opened 2017-08-24
Delete `ujson` and `uvloop` from requirements.txt


zhustec:
`ujson` and `uvloop` are not nessesary for sanic, so maybe we can delete them from requirements.txt, and give a message to user to manually install them for faster running?

https://github.com/channelcat/sanic/blob/7b66a56cad9820b3172850fde9a61f0cb865dcf5/sanic/response.py#L4-L7

https://github.com/channelcat/sanic/blob/eb06e6ba51b493f871d71dbddd9c7b4271e799b4/sanic/server.py#L22-L25

zhustec:
#855

yunstanford:
httptools may still not work on Windows. But I don't have a windows machine, so can't test it out.

zhustec:
True, currently it sill not work on Windows, but why not try to make it work on Windows step by step.

abuckenheimer:
I'm a fan of the batteries included but swap-able design. People who come to sanic for the purported speed may be dissuaded when it doesn't work as advertised out of the box.

rdehouss:
I understand your point, but for example, `ujson` is not a drop-in replacement for `json` and it might lead to issues, so, at least, when it's not 100% compatible, it shouldn't be forced.
So far, I'm obliged to `pip install sanic && sanic -y uninstall ujson` which is kind of stupid :)

abuckenheimer:
my intuition is that something like ujson and uvloop in general should be [extras](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies). That way you could ```pip install sanic``` or ```pip install sanic[uvloop]``` or ```pip install sanic[all]``` which has the opt in your looking for. But the project has a stated "[main goal](http://sanic.readthedocs.io/en/latest/sanic/contributing.html#warning)" of speed so I think its probably more in line with the project to be an opt-out feature a la ```SANIC_NO_UJSON=true pip install sanic```.

rdehouss:
The sad thing is that SANIC_NO_UJSON=true is not taken into account because in requirements.txt
And there are no solution in pip right now to be able to avoid that.

abuckenheimer:
yea that's annoying :disappointed:, in some senses this is a pip ergonomics problem, it's possible this could be alleviated with [conda features](https://conda.io/docs/user-guide/tasks/build-packages/features.html). I think someone started the process of making a conda package here #848

messense:
[...]

PR #1117, opened 2018-02-06
only install ujson and uvloop with cpython on non windows machines

abuckenheimer:
suggested by @messense in #916

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,6 +7,6 @@ httptools
 flake8
 pytest==3.3.2
 tox
-ujson; sys_platform != "windows" and implementation_name == "cpython"
-uvloop; sys_platform != "windows" and implementation_name == "cpython"
+ujson; sys_platform != "win32" and implementation_name == "cpython"
+uvloop; sys_platform != "win32" and implementation_name == "cpython"
 gunicorn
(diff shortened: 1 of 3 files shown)


r0fls:
We've seen a lot of PRs claiming to fix this. I would like to see multiple Windows users verify this fix works for them before merging this.

abuckenheimer:
Obviously this doesn't help with httptools on windows, still need to go through the visual c++ rigamarole there but I've gotten this installed on windows 10/7 no problem with python 3.6.3.

r0fls:
That makes sense. By "seen a lot of" I actually mean we've merged a lot of potential Windows fixes around removing unsupported dependencies 😂  which is why I would like to see another Windows user chime in here and verify this helps in some way, particularly since I thought we were doing this exact thing (or something very similar) at one point ¯\\_(ツ)_/¯

There is always docker for windows by the way ;)

abuckenheimer:
😆 certainly one way to test something, if you can find any of those old merges shoot me one, I'd be curious to see what didn't work. Also unfortunately no docker on Windows 7 yet haha

r0fls:
I guess just these: https://github.com/channelcat/sanic/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+windows

alexandertsema:
I've tested on my Windows 10 x64, works like a charm. Vote for merge.

alexandertsema:
when will it be released with a new version? @seemethere

Issue #1120, opened 2018-02-09
Passing the list into route decorator's host argument


mexicarne:
Hello,

according to the [docs](http://sanic.readthedocs.io/en/latest/sanic/routing.html#http-request-types) :
> There is also an optional host argument (which can be a list or a string)

But if I try to pass a list into host argument, I got 
```python
File "/path/lib/python3.6/site-packages/sanic/router.py", line 216, in merge_route
    uri, ','.join(list(duplicated))))
sanic.router.RouteExists: Route already registered: test.com/post/ [POST]
```

Here is the test sample:
```python
from sanic import Sanic
from sanic.response import text

app = Sanic()

@app.route('/post', methods=['POST'], host=['test.com', 'test2.com'])
async def post_handler(request):
    return text('POST request for test.com - {}'.format(request.json))

@app.route('/post', methods=['POST'])
async def post_handler(request):
    return text('POST request for default - {}'.format(request.json))

if __name__ == '__main__':
    app.run(host='127.0.0.1', port=8000)
```

I get the same exception if I'm passing the list with single value in it (`host=['test.com']`). Everything works as expected if host argument is a string though. Am I doing something wrong or it's a bug in documentation/sanic?

Sanic is installed from pip: sanic==0.7.0

tandalf:
Hello @mexicarne it's actually a bug in sanic. I was able to reproduce your issue.
A quick hack for this would be to set strict_slashes to True when calling @app.route. I'v pushed a PR for this. Pending it's approval, you could use this quick fix.

mexicarne:
@tandalf , that was fast! I can't reproduce the problem after applying the patch from your PR. Now both string and list working flawlessly. Thank you!

tandalf:
My pleasure. I'm glad it helped.

PR #1121, opened 2018-02-09
Fixed bug when passing a list into route decorator's host argument #1120

tandalf:
Fixing #1120 
There happens to be a bug in a recursive call to `Router._add` that tries to apply the `strip slashes`  logic wrongly. Normally when the host param passed to `Router._add` is a string, the method runs smoothly (the easy, good case). However, setting host to a list like 

```python
@app.route('/post', methods=['POST'], host=['test.com', 'test2.com'])
def handler():
    return text('OK')
```

triggers a recursive call to Router._add. More details are stated in the commit message.

### DIFF
--- a/sanic/router.py
+++ b/sanic/router.py
@@ -128,6 +128,13 @@ class Router:
         if strict_slashes:
             return
 
+        if not isinstance(host, str) and host is not None:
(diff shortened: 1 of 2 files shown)


r0fls:
Thanks for the fix!

PR #1131, opened 2018-02-18
Fix cookie parser (change to Django way)

### LINKED ISSUE #840: Sanic fails to read cookies if one of them is broken
example broken cookie `inbound_referral_site: Direct Traffic`
Than cookie contains non escaped space and http.cookie.SimpleCookie fails to parse all of them and returns empty dictionary

Following code fails (sanic uses the same aproach) 
```
c = SimpleCookie()
s1 = 'inbound_referral_site=Direct Traffic; good_cookie=123'
c.load(s1)
print(c.items())

Output: dict_items([])
```

It should return at least good cookies

Yaser-Amiri:
This code resolves #840.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -171,7 +171,6 @@ class Request(dict):
 
         return self._cookies
 
-
     @property
     def ip(self):
         if not hasattr(self, '_socket'):
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -3,7 +3,7 @@ import json
 import socket
 from cgi import parse_header
 from collections import namedtuple
-from http.cookies import SimpleCookie
+from http import cookies
 from httptools import parse_url
 from urllib.parse import parse_qs, urlunparse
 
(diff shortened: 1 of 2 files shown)


Yaser-Amiri:
And CI system has problems :/

sanic/request.py
```diff
@@ -158,15 +158,20 @@ def raw_args(self):
...
+                        key, val = chunk.split('=', 1)
+                    else:
+                        key, val = '', chunk
+                    key, val = key.strip(), val.strip()
+                    if key or val:
+                        self._cookies[key] = cookies._unquote(val)
+
         return self._cookies
 
+
```
r0fls: Remove this newline

Yaser-Amiri: oops! sorry! Done!

Yaser-Amiri: Done!

sanic/request.py
```diff
@@ -158,13 +158,17 @@ def raw_args(self):
...
             cookie = self.headers.get('Cookie')
+            self._cookies = {}
             if cookie is not None:
-                cookies = SimpleCookie()
-                cookies.load(cookie)
-                self._cookies = {name: cookie.value
-                                 for name, cookie in cookies.items()}
-            else:
-                self._cookies = {}
+                for chunk in cookie.split(';'):
```
r0fls: I'm worried this logic is going to be significantly slower than whatever was happening before.

Yaser-Amiri:
I wrote a simple example for this!
[Link](https://github.com/channelcat/sanic/issues/840)
I don't know why _http.cookies_ can't parse it correctly :/ It's a question for me too!

sanic/request.py
Yaser-Amiri: I see. It's depended on your goals and viewpoint. Do you see it as a bug? :/ or you are accepting it to increase performance?! :/
(I didn't run benchmark on it and I don't know how much this code increases or decreases performance)

yunstanford: yep, we might just follow the builtin lib behavior for now and keep an eye on that.

Yaser-Amiri: If you want to follow _http.cookie_ Please mention it and it's problems in your documentation :/

sjsadowski:
I'm going to close this; the initial reviewers opted to not upcheck this and remain with the current behavior. We could open an issue for improving documentation though.

ahopkins:
If there is still interest I would encourage discussion over whether to change the behavior or not to move to the community forums: https://community.sanicframework.org/c/moving-forward/feature-requests

PR #1149, opened 2018-03-02
use travis_retry on tox

r0fls:
this will rerun the tox command when it fails up to three times to prevent one-off failures, which we've seen before

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
     - env: TOX_ENV=check
       python: 3.6
 install: pip install -U tox
-script: tox -e $TOX_ENV
+script: travis_retry tox -e $TOX_ENV
 deploy:
   provider: pypi
   user: channelcat


.travis.yml
```diff
@@ -19,7 +19,7 @@ matrix:
     - env: TOX_ENV=check
       python: 3.6
 install: pip install -U tox
-script: tox -e $TOX_ENV
+script: travis_retry tox -e $TOX_ENV
```
yunstanford: We may also wanna upgrade `dist: precise` as it has been deprecated. but i recall that we have a unit test failed on `dist: trusty` ?

https://blog.travis-ci.com/2017-08-31-trusty-as-default-status

r0fls: Ya, it's still failing with trusty. Looks like it just stops responding: https://travis-ci.org/channelcat/sanic/jobs/349989523

PR #1155, opened 2018-03-11
Change deprecated ssl option in aiohttp

nirgn975:
Change `verify_ssl` to `ssl` in `aiohttp.TCPConnector`, because `verify_ssl` is now deprecated.


hatarist:
@r0fls but `aiohttp>=2.3.0` at the moment installs the `3.1.3`; maybe it's worth locking the version at `==2.3.0` or the latest 2.x one, `==2.3.10`?

r0fls:
Ah thanks @hatarist, we could also pin the later version since 3.x should work fine.

r0fls:
Or we could just merge this, but if the version is actually 2.x it will fail so it seems we should bump the requirements to me

aviv-ebates:
mmm.... Since the `aiohttp` is only mentioned in `requierements-dev.txt`, user code isn't bound by it; That means that my code is getting the deprecation error when I run my tests.
Do you know if there's a reason not to just bump to aiohttp 3.0?

ahopkins:
@nirgn975 Can you update your branch?

nirgn975:
> @nirgn975 Can you update your branch?

@ahopkins of course, did it and the travis is passed.

ahopkins:
Thanks @nirgn975!

uda:
When will this be published to pypi?

Issue #1151, opened 2018-03-05
`request` in exception handler is None


Luftzig:
Hello,

I have the following request handler:
```
        @self._app.exception(Exception)
        def to_json(request, exception):
            if request:
                error_logger.exception(log_message(request, exception))
                return json({"error": str(exception), "request": request.body},
                            status=exception.status_code if hasattr(exception, 'status_code') else 500)
            else:
                error_logger.exception("Error occurred outside request handling %s" % exception)
                return json({"error": str(exception)},
                            status=exception.status_code if hasattr(exception, 'status_code') else 500)
```
For some reason I see errors handled by the `else` branch, specifically this:

    aiohttp.client_exceptions.ClientResponseError: 404, message='Not Found'

I need the context of the incoming request because I include a header with a request ID propagated to all services involved.

Expect Behaviour:

request should not be None when handling errors originated from aiohttp.

r0fls:
Do all 404s hit the `else` branch? 🤔

go2sh:
You hit a very tricky case in python. Your if check does two things:
1) Check for not `None`
2) Check for `__len__()` > 0 (since request is a subclass of dict)
Nearly all of your objects will hit false in the 2. case and you fall through in the else clause. It's better to check for `request is not None`.

go2sh:
Btw see: https://docs.python.org/3.6/library/stdtypes.html#truth-value-testing

r0fls:
Maybe we should override the length method? Or is that a silly idea?

sunny94:
As pointed out by @go2sh in [comment](https://github.com/channelcat/sanic/pull/1156#issuecomment-372136646) we should override __bool__ method instead of __len__.

sjsadowski:
@r0fls should we work on a PR for this one?

PR #1156, opened 2018-03-11
Request.py: Updated base class from dict to object in Request wrapper…

sunny94:
Fixed issue #1151 which occurred because Request wrapper used in Sanic was inheriting from dict base class and in Request constructor,  base class dict constructor was not getting called and also Request was not overriding __len__ method of base class dict which was giving 0 as output everytime it is being called in "if request" in exception handler method. Thus updated base class from dict to object as no particular functionality of dict class is being used in Request class.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -42,7 +42,7 @@ class RequestParameters(dict):
         return super().get(name, default)
 
 
-class Request(dict):
+class Request(object):
     """Properties of an HTTP request such as URL, headers, etc."""
     __slots__ = (
         'app', 'headers', 'version', 'method', '_cookies', 'transport',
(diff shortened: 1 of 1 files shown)


go2sh:
Your change breaks stuff. The dict capability is needed in sanic. It is better to override `def __bool__(self):`.

go2sh:
See #1151

sunny94:
Yes, found that in failed build. I think we are using dict inheritance just for saving random key-value pairs in request, but we can do that with request.args also.

sunny94:
As pointed out by @go2sh in [comment](https://github.com/channelcat/sanic/pull/1156#issuecomment-372136646) we should override __bool__ method instead of __len__.

ahopkins:
I am closing this now because I think this is stale and not needed. If I am wrong, please let me know and we can reignite the conversation.

Issue #1158, opened 2018-03-13
[Suggestion]: auto reload


AberSheeran:
When the code be changed with DEBUG=True，Server will reload like flask.

Why not do that？

hustshawn:
You can use sanic-admin

AberSheeran:
I've already written a tool to manage it. Thank your help. 
But I think it would be better if sanic built this feature.

Berkodev:
You can use the library [AoikLiveReload ](https://github.com/AoiKuiyuyou/AoikLiveReload) to do that.

Take a look at the demo here:
https://github.com/AoiKuiyuyou/AoikLiveReload/blob/master/src/aoiklivereload/demo/sanic_demo.py

Yaser-Amiri:
To enable auto reloading developer must pass "auto_reload=True" to "run" method of Sanic app.

AberSheeran:
Really? I don't know that but I will try soon after class. I'm a student😳

AberSheeran:
@Yaser-Amiri 
Hey, I try it, but get a TypeError
```python
Traceback (most recent call last):
  File "main.py", line 55, in <module>
    main.run("0.0.0.0", 8000, debug=True, auto_reload=True)
TypeError: run() got an unexpected keyword argument 'auto_reload'
```

Is there anything I've overlooked?

yunstanford:
it's `app.run` https://github.com/channelcat/sanic/blob/master/sanic/app.py#L652.

AberSheeran:
Thank your help. Don't have this parameter in the version I installed with PIP. I will update Sanic from github. @yunstanford

yunstanford:
@Yaser-Amiri can you double check the unit test here ? https://github.com/channelcat/sanic/blob/master/tests/test_auto_reload.py
it's actually failing on my side, and might also try refactor a little bit, it's not very pytest style.

PR #1159, opened 2018-03-14
default to auto_reload in debug mode

r0fls:
fixes: https://github.com/channelcat/sanic/issues/1158

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -649,7 +649,7 @@ class Sanic:
     def run(self, host=None, port=None, debug=False, ssl=None,
             sock=None, workers=1, protocol=None,
             backlog=100, stop_event=None, register_sys_signals=True,
-            access_log=True, auto_reload=False):
+            access_log=True, **kwargs):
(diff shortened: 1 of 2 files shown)


arnulfojr:
@r0fls now that the `auto_reload` option is set to True in debug mode, maybe we can add it to the documentation? I have created a PR for it if you have any ideas or feedback it will be welcomed :)

r0fls:
hmm i rebased but the build is still hanging

r0fls:
@yunstanford any ideas?

Yaser-Amiri:
oops! @r0fls we can't do this now! We talked [here](https://github.com/channelcat/sanic/pull/1047#discussion_r159031631) before. Enabling auto-reloading causes failing some tests those need DEBUG=True so I separated DEBUG mode from auto-reloading.

r0fls:
@Yaser-Amiri thanks for the heads up, I just disabled `auto_reload` in the test client for now: https://github.com/channelcat/sanic/pull/1159/files#diff-8ff5a7907a8a3e558aa2868b65263a87R48

PR #1162, opened 2018-03-16
Fix hang build and failed builds

yunstanford:
The builds for all other PRs are failing now. @r0fls 

1. quick fix for unblocking other PRs.
     a. remove pytest-xdist integration for now, some race condition for sharing same port
     b.`test_autoload.py` might cause some zombie process, we should come up a more robust test, 
     let's remove it for now.
2. fix hang build, fix another race condition. (https://github.com/channelcat/sanic/blob/master/tests/test_multiprocessing.py#L22)
3. migrate to `dist: trusty`


.travis.yml
```diff
@@ -1,5 +1,4 @@
 sudo: false
-dist: precise
```
yunstanford: `precise` has been deprecated, let's move to default `trusty`.

tests/test_multiprocessing.py
```diff
@@ -19,7 +19,7 @@ def stop_on_alarm(*args):
             process.terminate()
 
     signal.signal(signal.SIGALRM, stop_on_alarm)
-    signal.alarm(1)
-    app.run(HOST, app.test_port, workers=num_workers)
```
yunstanford: fix a potential race condition, which could cause build hanging there.

PR #1163, opened 2018-03-16
Add __weakref__ to Request slots

vopankov:


r0fls:
Out of curiosity, what is the use-case here?

vopankov:
Hello!

In my case I tried to attach DataLoader object to the request. DataLoader is a batch loader and cache (e.g. for for GraphQL and other needs). It should have the same lifetime as the request has and in my case it should hold the reference to the request to be able to access request context (app, headers, ...). So I need to use weak reference here to avoid cyclic references.

```python
import weakref

from aiodataloader import DataLoader


class RequestDataLoader(DataLoader):
    def __init__(self, request, options=None, **kwargs):
        super().__init__(**kwargs)

        # Produces TypeError: cannot create weak reference to 'Request' object
        self.request = weakref.proxy(request)
        self.options = options if options is not None else {}

    @classmethod
    def for_request(cls, request, options=None, **kwargs):
        data_loader_key = ('data-loader', '.'.join([cls.__module__, cls.__name__]))
        if data_loader_key in request:
            return request[data_loader_key]
        else:
            instance = cls(request, options, **kwargs)
            request[data_loader_key] = instance
            return instance
``` 

Unfortunately the Request object is not weak-referenceable directly due to the lack of ```__weakref__```. I hope my case is not the exclusive case. So I decided to make pull request.

PR #1175, opened 2018-03-24
Fix try_everything example.

PyManiacGR:
Simple change log w/ logger to work.


r0fls:
we've gotta do something about these failing tests :(

PyManiacGR:
Yep. It looks ugly 😺

r0fls:
oh well, this is definitely correct anyway so let's merge it and deal with the consequences later :)

Issue #1176, opened 2018-03-26
Architectural problems of Sanic framework


sindzicat:
[Article on Reddit](https://www.reddit.com/r/Python/comments/876msl/sanic_python_web_server_thats_written_to_die_fast/)

**A copy of content of this article:**

Performance of Sanic-based web servers is pretty good, sure.

The only problem is that people want not only fast but stable servers.

From this perspective the Sanic is awful.

Malicious software can crash any Sanic server easy by out-of-memory error.

Let me analyze several different attack vectors:

<ol>
<li>
Send a POST to any (even not existing) PATH.

Push "Content-Length" HTTP header with the maximum available value without sending a body.

You should discover what is a maximum allowed size for POST HTTP body. Sanic by default limiting it down to insane 100 megabytes but Reverse Proxy Server like NGINX may reduce it to more reasonable 1 MB for example. Even 1 MB is enough.

Now we have 2 options:

a. Without closing the connection open as many concurrent connections as you can and push all of them into the state when HTTP headers are sent but HTTP body transition is postponed. Pretty classic attack for locking all TCP ports on a server by opened idle connections.

Sanic will drop the connection after 60 seconds by default but a minute may be enough for pushing the server into Deny-Of-Service state. The problem is not specific to Sanic, consequences are relative innocent.

b. The more interesting case is sending almost whole BODY but without a couple bytes at the end.

In fact, Sanic performs a routing and request handling only after fetching the whole BODY.

It means that BODY is COLLECTED IN MEMORY before starting of PATH/HEADERS analyzing and processing. You can declare 100 MB request's BODY, send 99 MB of garbage random data and stop sending after that.

Open another concurrent request and do the same. Repeat multiple times. Most likely the server will run out of memory before getting run out of free ports.

Why out-of-memory error is more dramatic than free-ports problem? Because of it, not only web process(es) stops processing incoming requests but the whole server goes to unresponsible state: physical memory is out, everything becomes swapped to disk, CPU is overheated by kernel swapper worker, as the result even connecting to a problematic server by SSH becomes deadly slow.

The problem can be reduced by adding monitoring tools for looking at Sanic processes memory and killing problematic web workers. But default configuration has no such monitors.

I bet that most web servers in the world are not configured properly (at least very many of them). Also killing a web process is a painful procedure, sometimes is not easy to distinguish normal occasional high memory consumption from malicious attack. As result, normal processing of user data will be killed.
</li><li>
Use Sanic streaming.

[...]

PR #1179, opened 2018-03-29
Pausable response streams

ashleysommer:
This commit adds handlers for the asyncio/uvloop protocol callbacks for pause_writing and resume_writing.

These are needed for the correct functioning of built-in tcp flow-control provided by uvloop and asyncio. See #1176 
This is somewhat of a breaking change, because the `write` function in user streaming callbacks now must be `await`ed.
This is necessary because it is possible now that the http protocol may be paused, and any calls to write may need to wait on an async event to be called to become unpaused.

Updated examples and tests to reflect this change.

This change does not apply to websocket connections. A change to websocket connections may be required to match this change.

### DIFF
--- a/tests/test_request_stream.py
+++ b/tests/test_request_stream.py
@@ -83,7 +83,7 @@ def test_request_stream_app():
                 body = await request.stream.get()
                 if body is None:
                     break
-                response.write(body.decode('utf-8'))
+                await response.write(body.decode('utf-8'))
         return stream(streaming)
 
     @app.put('/_put')
@@ -100,7 +100,7 @@ def test_request_stream_app():
                 body = await request.stream.get()
                 if body is None:
                     break
-                response.write(body.decode('utf-8'))
+                await response.write(body.decode('utf-8'))
         return stream(streaming)
 
     @app.patch('/_patch')
@@ -117,7 +117,7 @@ def test_request_stream_app():
                 body = await request.stream.get()
                 if body is None:
                     break
-                response.write(body.decode('utf-8'))
+                await response.write(body.decode('utf-8'))
         return stream(streaming)
 
     assert app.is_request_stream is True
@@ -177,7 +177,7 @@ def test_request_stream_handle_exception():
                 body = await request.stream.get()
                 if body is None:
                     break
-                response.write(body.decode('utf-8'))
+                await response.write(body.decode('utf-8'))
         return stream(streaming)
 
     # 404
@@ -231,7 +231,7 @@ def test_request_stream_blueprint():
                 body = await request.stream.get()
                 if body is None:
                     break
-                response.write(body.decode('utf-8'))
+                await response.write(body.decode('utf-8'))
         return stream(streaming)
 
     @bp.put('/_put')
@@ -248,7 +248,7 @@ def test_request_stream_blueprint():
                 body = await request.stream.get()
                 if body is None:
                     break
-                response.write(body.decode('utf-8'))
+                await response.write(body.decode('utf-8'))
         return stream(streaming)
 
     @bp.patch('/_patch')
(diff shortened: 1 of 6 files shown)


ashleysommer:
Tests are failing because new aiohttp brings in compatibility problems.
Requires this PR: https://github.com/channelcat/sanic/pull/1180

asvetlov:
Much better but as you've mentioned the change is not backward compatible.
Moreover the breakage is awful: old existing code stops writing streamed data without any exception (old code has no `await` before `write(data)` call).

I don't know is it ok or not (Sanic is in Beta stage, this assumes some place for breaking changes if needed).

ashleysommer:
Damn, that hanging bug in the tests from last week is back.

ashleysommer:
Is there a way to force the travis test to run again, without pushing another change?

seemethere:
There should be a `Restart Build` button, but that may not be there for people without write access 🤔

ashleysommer:
No, I don't see the "restart build" button because I don't have write access.
Doesn't matter, I checked the output of the travis error log more closely and noticed there was actually a failure on the flake8 check, so I needed to push a correction for that anyway.

ashleysommer:
Ok, well since it's likely a major version bump, are there any other breaking changes we've been wanting to make but holding off?

DirkGuijt:
Any news on this? I would be very happy with a new release :)

yunstanford:
cc: @seemethere

ashleysommer:
@seemethere @yunstanford Any update on getting this merged?
There's significant push by enthusiastic and passionate users at the moment to get a new version of Sanic published. I believe important PRs such as this one, and #1220 (which is a similar idea, except for requests) should be merged asap in order to be tested by upstream `master` users before a new release goes out.

ashleysommer:
I've just rebased this PR to current master HEAD. There wasn't any conflicts, but I didn't want it to drift too far away from current.

r0fls:
@ashleysommer looks like there are some tox/pep8 errors:

```
sanic/server.py:80:5: E303 too many blank lines (2)
sanic/server.py:399:9: E303 too many blank lines (2)
```

ashleysommer:
@rofls, those must've been caused by the rebase. I've just fixed the formatting up, should pass now.

tests/test_request_stream.py
```diff
@@ -83,7 +83,7 @@ def test_request_stream_app():
                 body = await request.stream.get()
                 if body is None:
                     break
-                response.write(body.decode('utf-8'))
+                await response.write(body.decode('utf-8'))
```
r0fls: PR looks good. Hoever, won't this require an update to the docs to use the `await` syntax for streaming?

ashleysommer: Yes. You're right.
I changed the response-streaming example, but I forgot to change the response-streaming docs.

ashleysommer:
@r0fls 
Updated docs with await syntax

tests/test_response.py
```diff
@@ -168,20 +170,39 @@ def test_stream_response_includes_chunked_header():
...
+        response.protocol.transport.write(data)
+
+    response.protocol.push_data = mock_push_data
+    response.protocol.drain = mock_drain
 
     @streaming_app.listener('after_server_start')
     async def run_stream(app, loop):
         await response.stream()
-        assert response.transport.write.call_args_list[1][0][0] == (
+        # assert response.protocol.push_data.call_args_list[1][0][0] == (
```
r0fls: these comments seem accidental -- did you mean to leave them here? They seem to be the exact same as the tests themselves

ashleysommer: TBH, I don't remember.
They look accidental. I will remove them.

PR #1180, opened 2018-03-29
Fix failing tests when aiohttp>=3.1.0

ashleysommer:
Some of the tests in Sanic (test_request_timout, test_response_timeout, test_keep_alive_timeout) use a custom SanicClient with modified methods. This relies on overriding internal aiohttp Client class methods.

In aiohttp 3.1.0 there were some breaking changes that caused the custom methods to be no longer compatible with latest upstream aiohttp Client class.
See: https://github.com/aio-libs/aiohttp/commit/903073283fa85bb3e8bb7568d3832a8fd9c06435
and: https://github.com/aio-libs/aiohttp/commit/b42e0ced4653ff62fa1829941e4f680d712b830a

This commit adds aiohttp version checks to adapt to these changes.

### DIFF
--- a/tests/test_request_timeout.py
+++ b/tests/test_request_timeout.py
@@ -58,18 +58,36 @@ class DelayableTCPConnector(TCPConnector):
             t = req.loop.time()
             print("sending at {}".format(t), flush=True)
             conn = next(iter(args)) # first arg is connection
-            try:
-                delayed_resp = self.orig_send(*args, **kwargs)
-            except Exception as e:
-                return aiohttp.ClientResponse(req.method, req.url)
+            if aiohttp.__version__ >= "3.1.0":
+                try:
+                    delayed_resp = await self.orig_send(*args, **kwargs)
+                except Exception as e:
+                    return aiohttp.ClientResponse(req.method, req.url,
+                        writer=None, continue100=None, timer=None,
+                        request_info=None, auto_decompress=None, traces=[],
+                        loop=req.loop, session=None)
+            else:
+                try:
+                    delayed_resp = self.orig_send(*args, **kwargs)
+                except Exception as e:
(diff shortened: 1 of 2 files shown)


tests/test_request_timeout.py
```diff
@@ -58,18 +58,36 @@ def close(self):
             t = req.loop.time()
             print("sending at {}".format(t), flush=True)
             conn = next(iter(args)) # first arg is connection
-            try:
-                delayed_resp = self.orig_send(*args, **kwargs)
-            except Exception as e:
-                return aiohttp.ClientResponse(req.method, req.url)
+            if aiohttp.__version__ >= "3.1.0":
```
r0fls: Why would this be greater than 3.1.0 given https://github.com/channelcat/sanic/pull/1180/files#diff-f4c265073ea4274a80eca85440875e56R2

🤔 

Whatever the case since your tests are passing, this is a good sign. They've all been failing lately 😢

ashleysommer: good question.
It is because tox.ini chooses `aiohttp>=2.3`
https://github.com/channelcat/sanic/blob/8a07463a67c61dd6303f3bf222600f0c5656565d/tox.ini#L15
So that is why tests were passing on my local machine with `aiohttp==1.3.5` as per requirements-dev.txt, but failing on Travis in tox virtual environment.
I changed the requirements-dev.txt to match what tox uses, and then I started seeing these problems with latest aiohttp at 3.1.0.

The other option would be simply pin requirements-dev.txt and tox.ini to `aiohttp==2.3.0`, so we don't have to deal with changes in new versions of aiohttp.

r0fls: That makes sense, thanks for the explanation. I think it's ok to leave the `>=`, at least for now. I like living on the cutting edge 😄

PR #1188, opened 2018-04-05
Modified code to allow blueprint's to have handlers which are class/i…

mashaalmemon:
With respect to #659, modified code to allow blueprint's to have handlers which are class/instance methods, in addition to already working plain vanilla functions.

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -65,9 +65,9 @@ class Blueprint:
 
         # Routes
         for future in self.routes:
-            # attach the blueprint name to the handler so that it can be
-            # prefixed properly in the router
-            future.handler.__blueprintname__ = self.name
+            # attach the blueprint name to the handler or method's class
+            # so that it can be prefixed properly in the router
+            BlueprintNameHelper.set_blueprintname_for_handler(future.handler, self.name)
             # Prepend the blueprint URI prefix if available
             uri = url_prefix + future.uri if url_prefix else future.uri
 
@@ -83,9 +83,9 @@ class Blueprint:
                       )(future.handler)
 
         for future in self.websocket_routes:
-            # attach the blueprint name to the handler so that it can be
-            # prefixed properly in the router
-            future.handler.__blueprintname__ = self.name
+            # attach the blueprint name to the handler or method's class
+            # so that it can be prefixed properly in the router
+            BlueprintNameHelper.set_blueprintname_for_handler(future.handler, self.name)
             # Prepend the blueprint URI prefix if available
             uri = url_prefix + future.uri if url_prefix else future.uri
             app.websocket(uri=uri,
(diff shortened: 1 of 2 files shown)


vltr:
That would be a nice feature :+1:

mashaalmemon:
It looks like my commit has some problems with style conventions as well as another small error. I will change and resubmit.

mashaalmemon:
I have some longer method names for clarity - I like my code to be easy to read months and years later when I forgot what I was doing! :) . I believe with Flake, line length is to be limited to 78chars. Is that a hard rule - that is should I go ahead and shorten those lines/method names up appropriately?

Any other feedback would be welcome so I can go ahead and finalize this pull request.

vltr:
@mashaalmemon have you tried [`black`](https://github.com/ambv/black) for formatting and fixing some of the issues from the [build log](https://travis-ci.org/channelcat/sanic/jobs/367865913)? Like:

```
...
sanic/blueprints.py:298:1: E302 expected 2 blank lines, found 1
sanic/blueprints.py:304:9: E722 do not use bare except'
sanic/blueprints.py:315:9: E722 do not use bare except'
sanic/blueprints.py:322:51: E502 the backslash is redundant between brackets
sanic/blueprints.py:323:44: E502 the backslash is redundant between brackets
sanic/blueprints.py:324:21: E128 continuation line under-indented for visual indent
sanic/blueprints.py:324:41: E502 the backslash is redundant between brackets
sanic/blueprints.py:325:25: E128 continuation line under-indented for visual indent
sanic/blueprints.py:333:17: E131 continuation line unaligned for hanging indent
sanic/blueprints.py:342:17: E131 continuation line unaligned for hanging indent
sanic/blueprints.py:349:1: W391 blank line at end of file
sanic/router.py:266:70: E502 the backslash is redundant between brackets
sanic/router.py:267:21: E131 continuation line unaligned for hanging indent
...
```

You can test running `tox` and `flake8` (now `pycodestyle` (I _think_)) in your code to see if everything is fine before commiting, too :wink:

mashaalmemon:
@vltr,

Planning to get all tests passing. Have been pulled away to another project but will get to this soon. Stay tuned.

pyx:
I don't see the `_for_handler` part is needed if all you do is `get_blueprint_name(handler)`.
One-liner helper functions not general enough are usually unnecessary in python.
Also those methods are brittle, for example, what if `__get_bpname_from_class_method` returns `None`... 

The problem of line length is superficial, **the real problem is `class` was being (ab)used solely as a namespace here**.  It seems to me Java code in python syntax.

Rule of thumb, stop writing classes if
- all you have are classmethods and staticmethods, you are using class as namespace, please use modules and packages to organize your code
- the class contains only a few public facing methods, especially no meaningful use of `self` in those methods (that is, those method do not share any environment), again you are using class as namespace, using a set of functions working on the same datatype is usually a better solution.
- plain 'o function will do, with shorter and cleaner code

This PR is a little bit of all three.

> The problem with object-oriented languages is they’ve got all this implicit environment that they carry around with them. You wanted a banana but what you got was a gorilla holding the banana and the entire jungle.
> 
> Joe Armstrong

ahopkins:
Is this still relevant? I am closing for now because it seems stale. feel free to carry on the conversation and we can see if we want this.

PR #1207, opened 2018-04-25
Apache common/combined Capable logs

arnulfojr:
I was actually not going to open this PR but well I would like some comments in the idea at least.

## Background
I have been playing with sanic for quite a long time for microservices and what thing that bothers me quite often is the access logs format.

## Goal
Replace current format for Apache Common (and possibility to have combined format)

## Benefits

* Standard format out of the box
* make my life easier and not base my self in a fork
* show real ip if found one from the `remote_addr` field of the request

## Alternative

I went ahead and developed a Plugin so maybe we can have a picture of the alternative:

https://github.com/arnulfojr/sanic_apache_accesslogs/

## Questions

* Maybe is best to have a plugin for this?


ahopkins:
This seems like a good idea. But, I agree that it is more suited to a plugin. Making that change now would have a big backwards compatibility issue. I am closing this mow, but perhaps this is worth bringing up for conversation on the community forums? community.sanicframework.com

Issue #603, opened 2017-03-28
HTTP 206 Partial Content


nszceta:
Static files partial content support is lacking. HTML5 video cannot be seeked without proper 206 support.

My current, potentially flawed workaround:
```python
@app.get('/video/<path>')
async def video_get(request, path):
    _, ext = os.path.splitext(path)
    headers = {'Content-Type': f'video/{ext}', 'Accept-Ranges': 'Accept-Ranges: bytes'}
    try:
        with open('static/' + os.path.basename(path), 'rb') as f:
            range_start = None
            range_end = None
            f.seek(0, 2)
            file_length = f.tell()
            f.seek(0)
            try:
                range_ = request.headers['Range'].split('=')[1]
                range_split = range_.split('-')
                range_start = int(range_split[0])
                f.seek(range_start)
                range_end = int(range_split[1])
            except ValueError:
                pass
            if range_start:
                if not range_end:
                    range_end = file_length
                read_length = range_end - range_start
                status_code = 206
            else:
                range_start = 0
                read_length = file_length
                status_code = 200
            f.seek(range_start)
            headers['Content-Length'] = read_length
            headers['Content-Range'] = f'bytes {range_start}-{range_end}/{read_length}'
            return response.HTTPResponse(
                body_bytes=f.read(read_length), status=status_code, headers=headers)
    except FileNotFoundError:
        return response.HTTPResponse(status=404)
```

seemethere:
Have you tried something similar to: https://github.com/channelcat/sanic/pull/378/files#diff-e5ad9a50378f9f4f2b96eeb462c4209aR56

I know there's not a full fledged example for it but content range is supported.

nszceta:
Yep, I've tried serving static with the use_content_range=True flag. It's a no-go unless the logic is implemented pretty much exactly as I have in my last comment above.

PR #1219, opened 2018-05-10
Return 206 when answering `Range` request

logannc:
Per RFC 7233, Sanic should return 206 when answering `Range` requests.


logannc:
_May_ fix #603. Not sure since I don't know what is actually wrong in his use case.

arnulfojr:
Can you please be more specific on what you aiming to solve with this PR?
As this only affects the response’s file_stream and stream functionality and not responding 206 to all requests that include the Range header

logannc:
That's fair. I had a bit of tunnel vision because I was working on a file server so the only relevant response types for me were `file` and `file_stream`. 

Also, I am not really aware of a use for the Range header outside of file serving, nor would I know how to properly handle that generically, for all requests. `file` and `file_stream` are already 'Range aware' which makes it easy. For an arbitrary route, I could throw away stuff from the beginning and ends of the response to enforce Range and override the route's status code, but would that usually be what the developer would want? Certainly for a `stream` response or something, most implementations would be able to seek in a much more efficient way than reading and dropping from the front of the stream. Likewise, they can simply stop producing bytes instead of reading and dropping from the end of the stream. And overriding their status code, while RFC compliant, may not be what they want.

So, to be specific, I want to return 206 when serving up partial files due to a `Range` header, per the RFC. While the RFC may prescribe returning 206 in other scenarios (I can imagine partial content of data routes would be useful in some scenarios), there is not a clearly useful way to generically handle those, so my PR is only partially compliant and leaves the rest to the application developer.

sjsadowski:
@logannc can you update your fork and resolve the conflicts there so that we can easily verify the merge with master?

logannc:
Sorry, I'd meant to do this the next day but it slipped my mind.

> And overriding their status code, while RFC compliant, may not be what they want.

This is now relevant to the PR. Previously, the methods I changed didn't give the option to choose a status code so my overriding it was okay. I've updated the function signature so that the default is added later iff they don't specify a status code. It is less pretty, but it works.

logannc:
It looks like these changes were incorporated in a different PR.

¯\\\_(ツ)_/¯

PR #1232, opened 2018-05-29
Fixed #1231 - release resource no matter what

### LINKED ISSUE #1231: Possible memory leak in websocket_handler function
Hey! It seems that I found a possible memory leak in `websocket_handler` function inside `Sanic.websocket` https://github.com/channelcat/sanic/blob/master/sanic/app.py#L301

If arbitrary exception occurred in websocket handler, it won't be catched down there and `fut` object will stay in `self.websocket_tasks` list. Little by little this list will become bigger and will consume more memory.

Probably it makes sense to catch all exceptions in `try: except:` block, not only `(CancelledError, ConnectionClosed)`?

pyx:
**Disclaimer:
I did not investigate the code, took the words of OP(#1231) as the fact, and provide a solution based on the description and from my experience.**

In case the issue really happened, we have it covered and one merge away. :smile:

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -303,7 +303,8 @@ class Sanic:
                     await fut
                 except (CancelledError, ConnectionClosed):
                     pass
-                self.websocket_tasks.remove(fut)
+                finally:
+                    self.websocket_tasks.remove(fut)
                 await ws.close()
 
             self.router.add(uri=uri, handler=websocket_handler,


r0fls:
Hmm seems to have broken the build 😢 why didn't the build run on this?

yunstanford:
@r0fls might be related to version bump of aiohttp.

pyx:
According to [the error message](https://travis-ci.org/channelcat/sanic/jobs/388987711#L572) 

> `ValueError: Exception during request: [TypeError("connect() got an unexpected keyword argument 'timeout'",)]`

`connect()` somehow being called like so, `connect(..., timeout=<whatever>)`, it's weird as I did a search, 
https://github.com/channelcat/sanic/search?q=timeout&unscoped_q=timeout, could not even find the call site nor an exact match of `timeout`.

Spent like 10 minutes reading the code, and I don't even have a clue about the control flow :rofl: , might take a closer look if I got more time later.

**Off-topic, no flame war intended**. That's why, IMHO, for mere mortal like me, abstractions by "referential transparency" always beats "Liskov substitution principle".

yunstanford:
the error is raised by aiohttp

PR #1249, opened 2018-06-18
Make auto reloader work on Mac

garyo:
The existing version relies on the `/proc` filesystem so it doesn't work on Mac. This adds a version that uses `pkill`.


r0fls:
Nice, thank you!

PR #1254, opened 2018-06-24
fix #1253

### LINKED ISSUE #1253: request.form() lost many bytes of binary data
Hello,
I found a bug of request.form(), it call parse_qs() with default encoding="utf8", when there are binary data (such as image data) in request.body, it would lost some bytes, because some binary data sequence is not in UTF-8 order.

------------------ Reproducing the bug: ------------------------------
1. server code:
#!/usr/bin/env python3
from sanic import Sanic
from sanic import response

app = Sanic(__name__)
 
@app.route('/', methods=['POST'])
async def api_img(request):
    img_data = request.form.get('img_data', '') 
    print('img_data:', len(img_data))                                                                                                                                                                       
    img_data_binary = img_data.encode('utf8')
    print('get img_data : {}, img_data_binary: {}'.format(len(img_data), len(img_data_binary)))
 
    result = { 
        'img_data': len(img_data),
        'img_data_binary': len(img_data_binary),
    }   
    return response.json(result)
 
if __name__ == '__main__':
    app.run(host='127.0.0.1', port=7000)

2. Client code:
#!/usr/bin/env python

import requests
img_data = open('dog.jpg', 'rb').read()
post_data = {
    'img_data': img_data,
    'img_type': 'jpg',
}
url = 'http://127.0.0.1:7000/'
r = requests.post(url, data=post_data)
print('client send img_data:', len(img_data))
print('server got: ', r.text)

3. run server code and client code, client code will print the msg:
('client send img_data:', 150156)
('server got: ', u'{"img_data":142079,"img_data_binary":269713}')
that means, client sent a image of 150156 bytes, but server request.form() only got 142079 bytes, something lost!

--------------- How to fix it? ------------------
Just in function request.form() call parse_qs() with encoding="latin1"
change file sanic/request.py, line 130. change to:
`parse_qs(self.body.decode('utf-8'),` encoding='latin1'))

veelion:
This fix refer to tornado (another python web framework), which solve the same problem in function parse_qs_bytes() in file tornado/escape.py:

    def parse_qs_bytes(qs, keep_blank_values=False, strict_parsing=False):
        """Parses a query string like urlparse.parse_qs, but returns the
        values as byte strings.

        Keys still become type str (interpreted as latin1 in python3!)
        because it's too painful to keep them as byte strings in
        python3 and in practice they're nearly always ascii anyway.
        """
        **# This is gross, but python3 doesn't give us another way.
        # Latin1 is the universal donor of character encodings.
        result = _parse_qs(qs, keep_blank_values, strict_parsing,
                           encoding='latin1', errors='strict')**
        encoded = {}
        for k, v in result.items():
            encoded[k] = [i.encode('latin1') for i in v]
        return encoded

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -127,7 +127,7 @@ class Request(dict):
             try:
                 if content_type == 'application/x-www-form-urlencoded':
                     self.parsed_form = RequestParameters(
-                        parse_qs(self.body.decode('utf-8'), encoding='latin1'))
+                        parse_qs(self.body.decode('latin1'), encoding='latin1'))
                 elif content_type == 'multipart/form-data':
                     # TODO: Stream this instead of reading to/from memory
                     boundary = parameters['boundary'].encode('utf-8')
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -127,7 +127,7 @@ class Request(dict):
             try:
                 if content_type == 'application/x-www-form-urlencoded':
                     self.parsed_form = RequestParameters(
-                        parse_qs(self.body.decode('utf-8')))
+                        parse_qs(self.body.decode('utf-8'), encoding='latin1'))
                 elif content_type == 'multipart/form-data':
                     # TODO: Stream this instead of reading to/from memory
                     boundary = parameters['boundary'].encode('utf-8')


r0fls:
@veelion can you post an example where this helps, or add a unit test?

ashleysommer:
@veelion are you sure this fixes the problem?

I thought that if it was going to lose bytes, it would lose them when decoding the body bytes with `self.body.decode('utf-8')`.

Would be be better to do:
`parse_qs(self.body.decode('latin1'), encoding='latin1')`?

sjsadowski:
Looks like @ashleysommer provided a reasonable solution.

PR #1258, opened 2018-07-05
Add url_bytes to Request

johndoe46:
For some application, we need to have access to the raw unparsed URL from the Request object.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -48,10 +48,11 @@ class Request(dict):
         'app', 'headers', 'version', 'method', '_cookies', 'transport',
         'body', 'parsed_json', 'parsed_args', 'parsed_form', 'parsed_files',
         '_ip', '_parsed_url', 'uri_template', 'stream', '_remote_addr',
-        '_socket', '_port', '__weakref__'
+        '_socket', '_port', '__weakref__', 'raw_url'
     )
 
     def __init__(self, url_bytes, headers, version, method, transport):
+        self.raw_url = url_bytes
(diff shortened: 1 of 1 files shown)


r0fls:
Why not just convert it back to the raw url afterwards?

johndoe46:
That would be no "raw". We have no way to convert the URL back since some information is lost in the parsing done by parse_url.

Here are some urls we need to work with:

http://domain.tld/ark:/1234/567?
http://domain.tld/ark:/1234/567??

Note the question marks (yes, as crazy it seems, they're significant in the spec we have to implement). parse_url does this:

```
In [1]: from httptools import parse_url

In [2]: parse_url(b"http://domain.tld/ark:/1234/567")
Out[2]: <URL schema: b'http', host: b'domain.tld', port: None, path: b'/ark:/1234/567', query: None, fragment: None, userinfo: None>

In [3]: parse_url(b"http://domain.tld/ark:/1234/567?")
Out[3]: <URL schema: b'http', host: b'domain.tld', port: None, path: b'/ark:/1234/567', query: None, fragment: None, userinfo: None>

In [4]: parse_url(b"http://domain.tld/ark:/1234/567??")
Out[4]: <URL schema: b'http', host: b'domain.tld', port: None, path: b'/ark:/1234/567', query: b'?', fragment: None, userinfo: None>

```

johndoe46:
Please Say Yes

r0fls:
okcat

johndoe46:
Awesome

PR #1261, opened 2018-07-09
Add subprotocols param to add_websocket_route

ciscorn:
The `websocket_route` decorator can take `subprotocols` parameter.

However the corresponding `add_websocket_route` helper method can't.


r0fls:
Nice, thanks @ciscorn!

Issue #761, opened 2017-05-31
Run under an ASGI interface.


tomchristie:
I've taken a first pass at implementing a Gunicorn worker that uses `uvloop` and `httptools` and interfaces with the application framework using an ASGI consumer interface... https://github.com/tomchristie/uvicorn (Using `aiohttp` and `sanic` as guidelines)

The intent here is to take the same set of tooling that sanic uses, but have the [ASGI consumer interface](http://channels.readthedocs.io/en/stable/reference.html#consumers) as a proper layer of separation between the server and application, in the same way that we currently have for synchronous frameworks with WSGI.

I'm interested in hearing from you folks if Sanic might consider adopting the ASGI consumer interface?

* Have shared server implementations capable of interfacing with any Python framework that exposes itself as an ASGI consumer callable.
* Remove the requirement of maintaining worker and server in sanic directly.
* Add the ability to run Sanic behind a channel layer (eg. to support websocket broadcast across a cluster of servers)
* Far easier to support new protocol types (ASGI is a general purpose messaging interface)

seemethere:
It's definitely something we have considered in the past.

I think the general consensus that we had was that we would wait until ASGI was more mature in terms of having concrete guidelines. I'd love to revisit the idea though.

tomchristie:
Yes, I think there may still some thrashing out to do wrt. the consumer callable interface. However I'm not in any doubt that having a clear server/framework interface separation is absolutely something that we ought to be pushing towards. I'll plan to keep you folks up to date with progress on the server implementation there (which now has a name, uvicorn)

seemethere:
Definitely digging the name @tomchristie! 😄

tomchristie:
🦄  Had to do it, really.

tomchristie:
I've done enough work on the `uvicorn` package now that the server/application interface I'm aiming for should be clear now. Any feedback most welcome.

(Other stuff still on the way: websocket support, broadcast support)

tomchristie:
So uvicorn, and the ASGI-based messaging interface is definitely now in a state where it's a viable server for asyncio frameworks. I've been having conversations about what & why in a few different places...

* django-dev - https://groups.google.com/forum/#!topic/django-developers/_314PGl3Ao0
* aio-libs - https://groups.google.com/forum/#!topic/aio-libs/7rJ8Pb1y7aA and https://github.com/aio-libs/aiohttp/pull/2035

[...]

PR #1265, opened 2018-07-18
WIP: ASGI Integration

tomchristie:
Quick proof-of-concept for adding ASGI support to Sanic.
Refs. #761

Makes the following possible:

```python
from sanic import Sanic

app = Sanic()

@app.route('/')
async def test(request):
    return json({'hello': 'world'})
```

Then:

```shell
uvicorn example:app  # Or `daphne` or `hypercorn`.
```

I've just added this in as dumb-as-possible way for now, as a proof-of-concept.

* Doesn't currently handle streaming requests, always just reads body to memory.
* Doesn't currently handle streaming responses.
* Doesn't handle websockets.
* Doesn't currently interact properly with Sanic's exception handling.
* Naively drops the gunicorn callable, in favor of an ASGI callable.

Note that Ctrl-C to quit may not work properly here, I think due to not having plugged Sanic's exception handling in properly yet.

A more committed approach to moving to ASGI might instead introduce changes to the request/response classes, rather than just adding a dumb shim over the top of Sanic's existing implementation. Additionally it might also adapt the server implementation slightly so that
it actually calls into Sanic applications using ASGI.

Some benefits from moving to ASGI would be:

* HTTP 2 support. (Currently available in Hypercorn, Daphne. Planned for Uvicorn)
* PyPy support. (Currently available in Uvicorn and Daphne)
* Windows support. (Currently available in Daphne. Very nearly available in Uvicorn.)
* Implementations that fix write-buffer flow control, eg #1176. (In this PR always send back the response using an asynchronous API, allowing server implementations to pause while the write buffer is flushed, if needed, which I know Uvicorn and Daphne both do, and I expect Quart also does.)
* Enable a shared ecosystem of ASGI middleware. (Across APIStar, Sanic, Quart, Starlette for starters.)
* Interoperable server implementations. (Allow folks to develop more robust or faster server implementations, and have that work shared across all ASGI frameworks.)

This PR is more to start a conversation than anything else, and to demonstrate that Sanic's close enough already that moving to ASGI might not *necessarily* require a massive set of changes.

Obviously I *personally* think there'd be a big long-term benefit to Sanic moving to an ASGI interface, both for Sanic itself, and for the community as a whole. Tho this PR is more of a proof-of-concept at this stage, than a serious attempt.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -7,6 +7,7 @@ from asyncio import get_event_loop, ensure_future, CancelledError
 from collections import deque, defaultdict
 from functools import partial
 from inspect import getmodulename, isawaitable, signature, stack
+from multidict import CIMultiDict
 from traceback import format_exc
 from urllib.parse import urlencode, urlunparse
 from ssl import create_default_context, Purpose
@@ -16,6 +17,7 @@ from sanic.constants import HTTP_METHODS
 from sanic.exceptions import ServerError, URLBuildError, SanicException
 from sanic.handlers import ErrorHandler
 from sanic.log import logger, error_logger, LOGGING_CONFIG_DEFAULTS
+from sanic.request import Request
 from sanic.response import HTTPResponse, StreamingHTTPResponse
 from sanic.router import Router
 from sanic.server import serve, serve_multiple, HttpProtocol, Signal
@@ -630,10 +632,17 @@ class Sanic:
                 )
 
         # pass the response to the correct callback
-        if isinstance(response, StreamingHTTPResponse):
+        if write_callback is None or isinstance(response, StreamingHTTPResponse):
             await stream_callback(response)
         else:
             write_callback(response)
+
+    # -------------------------------------------------------------------- #
+    # ASGI
+    # -------------------------------------------------------------------- #
+
+    def __call__(self, scope):
+        return ASGIApp(self, scope)
 
     # -------------------------------------------------------------------- #
     # Testing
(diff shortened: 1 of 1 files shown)


skewty:
This seems stalled again.  Is this still a priority?  

This along with Gino makes **sanic** quite appealing for a project we just got approval for at work.  Without working web sockets, it is a non starter for us (unless I am understanding things incorrectly).

tomchristie:
Sanic has working websocket support, so I'm not sure if this issue is relevant to you or not.
Personally I'm up for helping progressing this issue myself, but I don't know what the sanic team's stance on it is.

ahopkins:
I think this is something that is definitely on the radar. It makes sense that if `asgi` is going to have a broader applicability and become adopted outside `django-channels` (for example), then we should look into applying it here. The point about a benefit for both Sanic and the Python community is a motivating factor.

As said above, websockets are a part of sanic now. So, if that is what you need to move forward, [you should be able to](https://sanic.readthedocs.io/en/latest/sanic/routing.html#websocket-routes).

Right now the Sanic team is doing a little reorganizing to shift to being a [community run organization](https://github.com/huge-success). Part of the discussions have included placing a greater emphasis on websockets. 

That, plus this being a "proof of concept" PR puts a little pause on this before we would merge this asgi support PR. @tomchristie graciously put some effort into making this work inside Sanic, and I do believe this (or something that achieves this) will be adopted.

tomchristie:
All sounds super positive! Up for being involved/helping out as needed. Cheers!

sjsadowski:
@tomchristie can you update your fork so we can get the tests to run again against current master?

sjsadowski:
Tagged this for inclusion for 19.03

tomchristie:
Discussion: https://community.sanicframework.org/t/considering-asgi-support/70

ahopkins:
We can leave this one closed in favor of #1475. I am going to start working on the test client issue.

PR #1267, opened 2018-07-19
Add content_type flag to Sanic.static

### LINKED ISSUE #1266: Define Static File Content Type
While trying to serve a utf-8 encoded HTML document via `Sanic.static`, I noticed that `content_type` cannot be set using `Sanic.static`.

`Sanic` header content_type [defaults to `text/plain`](https://github.com/channelcat/sanic/blob/master/sanic/static.py#L101)
if the type cannot be guessed by `mime_types.guess_type`, which seems to only analyze the filename.

Currently, the following is not possible (same goes for blueprints):
```python
app.static('/', get_data('index.html'), content_type='text/html; charset=utf-8')
```

Prior to this PR, you need to do the following to serve a document with a custom content-type:
```python
@app.route('/')
async def handle_index(request):
    return await response.file(
        'index.html',
        headers={'Content-Type': 'text/html; charset=utf-8'}
    )
```

I'll be submitting a PR shortly with a proposed fix.

MyNameIsCosmo:
Fixes #1266 

With this PR, we can do:
```python
import tempfile
from sanic.config import Config

app = Sanic()

indexhtml = "<html><body><pre>" + Config().LOGO + "</pre></body></html>"

temp = tempfile.NamedTemporaryFile(delete=False)
temp.write(str.encode(indexhtml))
temp.close()

app.static('/', temp.name, content_type='text/html; charset=utf-8')

app.run()
```

Comments and feedback are appreciated.


sanic/static.py
```diff
@@ -95,10 +96,12 @@ def register(app, uri, file_or_directory, pattern,
                         del headers['Content-Length']
                         for key, value in _range.headers.items():
                             headers[key] = value
+            headers['Content-Type'] = content_type \
+                or guess_type(file_path)[0] or 'text/plain'
             if request.method == 'HEAD':
                 return HTTPResponse(
                     headers=headers,
-                    content_type=guess_type(file_path)[0] or 'text/plain')
+                    content_type=content_type)
```
r0fls: what happens here if `content_type` is still `None`?

MyNameIsCosmo: Ah, good catch!
Since `headers['Content-Type']` is defined from lines 99-100, `response.HTTPResponse` will default to `headers['Content-Type']` instead of `content_type` regardless if `content_type` is defined:
https://github.com/channelcat/sanic/blob/master/sanic/response.py#L146

Because of the above case, should this line be removed,
or
should this become `content_type=headers['Content-Type']` (feels redundant)?

MyNameIsCosmo:
Removed `content_type=content_type` as it is redundant:
https://github.com/channelcat/sanic/pull/1267#discussion_r204190913

PR #1278, opened 2018-08-06
Gracefully handle when the request_handler_task is cancelled.

ashleysommer:
This PR contains modifications to the `handle_request` function to detect and gracefully handle the case that the request_handler Task is canceled by the sanic server while it is handling the request. One common occurrence of this is when the server encounters a ResponseTimeout error, it cancels the response_handler_task.

In this change, the CancelledError exception handler purposely sets `response` to `None` to drop references to the handler coroutine, in an attempt to preemptively release resources.

This commit also fixes a possible reference-before-assignment of the `response` variable in the `handle_request` function.

Finally, another byproduct of this change is that ResponseMiddleware will no longer run if the `response` is `None`.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -571,6 +571,10 @@ class Sanic:
 
         :return: Nothing
         """
+        # Define `response` var here to remove warnings about
+        # allocation before assignment below.
+        response = None
+        cancelled = False
         try:
             # -------------------------------------------- #
             # Request Middleware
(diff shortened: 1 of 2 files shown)


ashleysommer:
Turns out it wasn't that hard to test for this. I've pushed a new commit that adds the test.

abuckenheimer:
this looks great, open question to you, do you think server.write_error should call self.cleanup() like I do here https://github.com/channelcat/sanic/compare/master...abuckenheimer:fix_response_timeout-1273#diff-bd84a6e523017a7f3d6bd72ee689dce2R422 ?

PR #1280, opened 2018-08-06
disabled auto_reload by default in windows

abuckenheimer:
this half undoes the work in #1159, it doesn't really make sense to enable auto_reload by default (as a debug feature) on a windows if the feature doesn't work (raises NotImplementedError).

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -671,8 +671,8 @@ class Sanic:
         """
         # Default auto_reload to false
         auto_reload = False
-        # If debug is set, default it to true
-        if debug:
+        # If debug is set, default it to true (unless on windows)
+        if debug and os.name == 'posix':
             auto_reload = True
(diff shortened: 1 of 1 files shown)


sanic/app.py
```diff
@@ -671,8 +671,8 @@ def run(self, host=None, port=None, debug=False, ssl=None,
         """
         # Default auto_reload to false
         auto_reload = False
-        # If debug is set, default it to true
-        if debug:
+        # If debug is set, default it to true (unless on windows)
+        if debug and os.name != 'posix':
```
r0fls: shouldn't this be `os.name == 'posix'`?

abuckenheimer: :rofl: wow really thought I could handle a 2 line pr

r0fls: Happens to the best of us 😄 that's why review is good!

PR #1283, opened 2018-08-13
A kind of `middleware` for certain handler

fduxiao:
Now we have nested blueprint support, which is desired by who wants to reuse. In my situation, I build a restful CRUD api for some database operations (a blueprint), but in my application different urls have different intended users i.e. they have different privileges. I want to only add the privilege check for a certain url instead of always building a new blueprint, so there has to be some mechanics to add a `filter` to an existing route, which can return new response or call the old handler and modify the result.
As you commented [here](https://github.com/channelcat/sanic/blob/212da1029e9a0ed786b5a1bfaf10183bfef9a790/sanic/router.py#L243), it's too complex for me to find a suitable way to replace the old handler, so I just write an immature implementation for blueprint. Do you have any better idea?


ashleysommer:
@fduxiao I like the idea behind this, but I think it is something that should be implemented by a Plugin, rather than by Sanic itself.

fduxiao:
@ashleysommer alright, let me write a Plugin then.

ahopkins:
@fduxiao Did you ever make a plugin for this? I am going to close this PR since I agree that this is not something we want in sanic core right now.

Issue #1285, opened 2018-08-16
recent auto_reload code change broke auto_reload on Linux


ashleysommer:
See this commit: https://github.com/channelcat/sanic/commit/01257f65a6f262df953f63fafb45993ddf4b7df8
Specifically this line: https://github.com/channelcat/sanic/commit/01257f65a6f262df953f63fafb45993ddf4b7df8#diff-9e9169b19156123d155a011bada481d4R97
In the python `sys` module, `sys.platform` can never == 'posix'.
See this doc for more details:
https://docs.python.org/3.5/library/sys.html
The only possible values are "linux", "win32", "cygwin", and "darwin".

So as a result, `kill_process_children()` no longer works on Linux.

I have written a fix, I will file a PR.

PR #1286, opened 2018-08-16
Fix auto_reload in Linux

ashleysommer:
Fix two problems with the auto reloader in Linux.
1) Change 'posix' to 'linux' in sys.plaform check, because 'posix' is an invalid value and 'linux' is the correct value to use here.
2) In kill_process_children, don't just kill the 2nd level procs, also kill the 1st level procs.

Also in kill_process_children, catch and ignore errors in the case that the child proc is already killed.
Fixes https://github.com/channelcat/sanic/issues/1285


ashleysommer:
Woops, looks like a flake8 error. Let me fix it.

ashleysommer:
Fixed. Merge-able now.

ahopkins:
awesome..... I've been using https://github.com/AoiKuiyuyou/AoikLiveReload

PR #1305, opened 2018-09-02
Reuse app fixture in tests

Stranger6667:
Hello!

This PR replaces the manual creation of `Sanic` app with a pytest fixture (similar to the one in the docs), which reduces the code duplication and makes the test suite more in line with the examples in the documentation.

Cheers


sjsadowski:
I think this is a positive change and would like to see it in 18.12 LTS. @ashleysommer and @yunstanford could you eyeball this and verify that I wouldn't be crazy for merging. I'm comfortable with the pass that @r0fls gave it, but I want to just be triple-sure before we do anything that is a hefty change.

ashleysommer:
@sjsadowski Yep, as long as all tests are still passing then I see this as a good change to have in the test suite.

PR #1310, opened 2018-09-10
Created methods to append and finish body content on Request

vltr:
Hi! I just created two methods to append and finish the body content "feed" on request.py so the underlying body instance can have certain flexibility (example: one could use `BytesIO` instead of a `list` for storing the body data if a custom Request-like class is provided to the `Sanic` constructor).

I also modified server.py to reflect the changes on the `Request` class.

If any documentation is required regarding this, please let me know. Also, I'm not sure if this change is the best approach to the `Request.body` attribute so I'm open to suggestions :wink:

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -84,6 +84,12 @@ class Request(dict):
             return True
         return False
 
+    def body_append(self, data):
+        self.body.append(data)
+
+    def body_finish(self):
+        self.body = b''.join(self.body)
+
     @property
     def json(self):
         if self.parsed_json is None:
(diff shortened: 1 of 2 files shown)


seemethere:
@vltr would you mind rebasing your PR?

vltr:
@seemethere yeah, no problem. In fact, one problem: `unknown repository`. How did I managed to do that? :scream:

If I don't figure out if there's a way to fix this, I'll just make another PR replacing this one.

vltr:
I was able to get the code from the PR, rebase and etc; but `unknown repository` PRs are readonly, so it is not possible to append directly :unamused:

So, I made a new PR #1379 to replace this one. Sorry for the mess.

PR #1314, opened 2018-09-14
Change content type to file type on File.type

### LINKED ISSUE #1311: Request type is used instead of file type on file.type
Heya!

I've noticed that `file.type` over [here](https://sanic.readthedocs.io/en/latest/sanic/request_data.html#request-data) (see `files`) is not a mimetype for the file but for the request, grabbed from [`cgi.parse_header`](https://docs.python.org/3.5/library/cgi.html#cgi.parse_header).

This feels like an incorrect behavior as we (@elixire) previously assumed that this was for file mimetype, and it caused a bug that affecting many users.

I propose doing one (or more) of these:
- Updating documentation to state that it's not file type but request type
- Swapping it out for file's mimetype
- Removing `.type`
- Adding a `.mime` field to file with mimetype

We can get the proper file mimetype from either the file's magic header with [`python-magic`](https://github.com/ahupp/python-magic) or with the name through [mimetypes](https://docs.python.org/3.5/library/mimetypes.html) library. While `python-magic` would give more accurate results, it requires installing libmagic through distro's package manager which might not be favorable for some.

I'd just send a PR but wasn't sure which approach was the most favorable, so I'll send one once a decision is made regarding which approach will be taken.

aveao:
As I detailed on #1311, the current behavior seems incorrect or at the very least, confusing. Please read that first.

This change moves File.type from content type (determined from request's mimetype, which seems to always be `application/octet-stream`) to file's actual type (determined with `mimetypes` library based on file's extension, same as how type determination is done on rest of the code).

---

~~With this change the `content_type` variable on `parse_multipart_form` ends up being unused, I can remove or comment that if needed.~~

I also removed `content_type` variable on `parse_multipart_form` as travis detected it as an unused variable and caused CI to fail.

---

Closes #1311

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -291,7 +291,6 @@ def parse_multipart_form(body, boundary):
     form_parts = body.split(boundary)
     for form_part in form_parts[1:-1]:
         file_name = None
-        content_type = 'text/plain'
         content_charset = 'utf-8'
         field_name = None
         line_index = 2
@@ -313,7 +312,6 @@ def parse_multipart_form(body, boundary):
                 file_name = form_parameters.get('filename')
                 field_name = form_parameters.get('name')
             elif form_header_field == 'content-type':
-                content_type = form_header_value
                 content_charset = form_parameters.get('charset', 'utf-8')
 
         if field_name:
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -5,6 +5,7 @@ from cgi import parse_header
 from collections import namedtuple
 from http.cookies import SimpleCookie
 from httptools import parse_url
+from mimetypes import guess_type
 from urllib.parse import parse_qs, urlunparse
 
 try:
@@ -318,7 +319,8 @@ def parse_multipart_form(body, boundary):
         if field_name:
             post_data = form_part[line_index:-4]
             if file_name:
-                form_file = File(type=content_type,
+                file_type = guess_type(file_name)[0] or 'text/plain'
+                form_file = File(type=file_type,
                                  name=file_name,
                                  body=post_data)
                 if field_name in files:


sanic/request.py
```diff
@@ -312,13 +312,13 @@ def parse_multipart_form(body, boundary):
...
                 field_name = form_parameters.get('name')
             elif form_header_field == 'content-type':
-                content_type = form_header_value
                 content_charset = form_parameters.get('charset', 'utf-8')
 
         if field_name:
             post_data = form_part[line_index:-4]
             if file_name:
-                form_file = File(type=content_type,
+                file_type = guess_type(file_name)[0] or 'text/plain'
```
r0fls: I think this should still use `form_header_value` when present, e.g.: `file_type = form_header_value  or guess_type(file_name)[0] or 'text/plain'`

aveao: How would that work exactly?

`Content-Disposition` is the only header that we generate a `File` on. `Content-Type` is a whole other header.

Are you proposing bringing back the `content_type` variable and storing in that?

---

What about the cases where `Content-Type` comes after `Content-Disposition`? It'll not use that Content-Type in such cases. Old code would also cause issues with that, sure, but should we really reintroduce a bug?

---

Additionally, bringing back `Content-Type` brings 2 issues:

- Users can fake the type of files, which might be undesirable (currently they need to change the extension to fake it, which is relatively more secure in cases where a whitelist is introduced (as people can't just upload their exes and send a `Content-Type` of say, `image/png`, and in the current state users would need to rename the file to run them as executables))

- `Content-Type` doesn't necessarily show the type of file. Originally when we had this issue the code wasn't returning `text/plain` (which is default) but `application/octet-stream`, as that's what the request had as `Content-Type`

vltr: I'm a little confused here. `Content-Disposition` is a browser related header, it only provides information regarding how the browser should handle with the content it receives in the body. From the [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition):

> In a regular HTTP response, the Content-Disposition response header is a header indicating if the content is expected to be displayed inline in the browser, that is, as a Web page or as part of a Web page, or as an attachment, that is downloaded and saved locally.

AFAIK, `Content-Type` is almost a must have to differentiate what's the kind of data is being sent over the body. `application/octet-stream` is the _de facto_ value for "all I know is that this is binary data", but this can be further refined to meet some end. See [this answer](https://stackoverflow.com/a/20509354) on SO regarding a similar topic.

Perhaps we should get the RFC and strict to the expected behavior and, if one needs, just override it, for request or response.

aveao: I mentioned `Content-Disposition` as we only create a `File` object when we're processing that (so checking `form_header_value` before creating it would have `Content-Disposition` info).

Also it's more than that, it also provides info about filename and such, which is why it's used by sanic.

---

`application/octet-stream` is all we get, no matter type of file. That's not helpful.

---

Which RFC would be relevant to this? I'm willing to check it out.

vltr: I know that `Content-Disposition` may provide a fine grained control over received files, but it is not something we'll always have at our disposal (and we really should take that under consideration). If that's the case, what do you suggest could be the default behavior?

You can check RFC 1867 (obsoleted), 2388 (proposed, obsoleted), 2854 (informational), 7578 (proposed).

aveao: I don't know what `Content-Disposition` could be replaced with. I suppose in rare cases where the header is not present, the file name and type could be left as an empty str or just as null. Though I'm not quite sure how a file's presence would be checked in the first place.

---

To be clear, when I say `application/octet-stream is all we get, no matter type of file. That's not helpful.`, I mean we as in @elixire, not sanic (though we do use sanic in prod and had this behavior).

The behavior might be different under different environments and different clients, however this was the case for us.

---

I didn't get a chance to read the RFCs but I'll try to read them tomorrow and might add a second comment.

seemethere:
@aveao would you mind rebasing your PR?

ahopkins:
@aveao Is this something you plan to update? If not we will close it.

PR #1327, opened 2018-09-25
Rename the `http` module to `helpers`

### LINKED ISSUE #1323: Trouble with import conflict in request.py
```
  File "/env/lib/python3.6/site-packages/sanic/__main__.py", line 4, in <module>
    from sanic.log import logger
  File "/env/lib/python3.6/site-packages/sanic/__init__.py", line 1, in <module>
    from sanic.app import Sanic
  File "/env/lib/python3.6/site-packages/sanic/app.py", line 21, in <module>
    from sanic.server import serve, serve_multiple, HttpProtocol, Signal
  File "/env/lib/python3.6/site-packages/sanic/server.py", line 31, in <module>
  File "/env/lib/python3.6/site-packages/sanic/request.py", line 6, in <module>
    from http.cookies import SimpleCookie
ModuleNotFoundError: No module named 'http.cookies'; 'http' is not a package
```

It seems it tries to use sanic's `http.py` instead of the builtin Python library. The `http.py` itself is used only twice, in the `response.py` & `exceptions.py`.
Am I missing something and/or have something misconfigured?

hatarist:
to prevent conflicts with the built-in Python http library (fixes #1323)


vltr:
Yes, this will probably break backward compatibility and it's not something we _desire_, but eventually we'll have to deal with these kind of changes to fix something as simple as a conflict with Python's own stdlib.

sjsadowski:
Could we target this change for the december release and be very clear about the breaking change in the announcement?

PR #1333, opened 2018-09-29
Sanic fix #1139 the status of response to request with range should be 206 and  shuold contain range's last byte

### LINKED ISSUE #1139: Static files handling makes a mistake when deal with "Range: bytes=0-" headers.
I use the code below to deal with static files with  request with "Range: 0-" headers.
```python
@app.route("/media/<filename>")
async def static(request, filename):
    filename = os.path.join('media',  filename)
    stats = await stat(filename)
    _range = None
    try:
        _range = ContentRangeHandler(request, stats)
    except HeaderNotFound:
        pass
    else:
        # _range.end -= 1
    resp = await sanic.response.file_stream(filename, _range=_range)
    # if _range: resp.status=206
    return resp
```
But unfortunately, it doesn't work as expected.
Assume the file size is 1024 bytes.  
Firstly, the `_range`  returned by `ContentRangeHandler` in the code produces a header 'Range: bytes 0-1024/1024', which exactly should be 'Range: bytes 0-1023/1024'. I should add `_range.end -= 1` to fix it.When dealing with video files, this mistake will make video on browsers un-seekable.
Secondly, If `_range`  is not None, the `file_stream` returns a `StreamingHTTPResponse` with a status 200. I should change is to 206 by myself , which is annoying.

ryanrain2016:
fix these below:
1. Http request with a range header return status 200, which should be 206.
2. Range header `Range: x-y` means content contains last byte content[y],  but sanic handler it like slice.


ashleysommer:
Looks great! Thanks @ryanrain2016

sjsadowski:
@ashleysommer can you upcheck this again when you get a chance (after the branch update is complete)

sanic/response.py
```diff
@@ -316,6 +317,7 @@ def html(body, status=200, headers=None):
     if _range:
         headers['Content-Range'] = 'bytes %s-%s/%s' % (
             _range.start, _range.end, _range.total)
+        status = 206
```
abuckenheimer: this may be an overblown worry but what if someone sets a specific status for their request, this will silently change it on them if its a range request. should we set the default to None so we can tell if the caller provided a status?

sjsadowski: @abuckenheimer Or just check to see if status is already set, and if it isn't, then set the status?

vltr: That does makes sense, but the default `status` argument value given to the function is `200` and the `_range` keyword needs to be given to the function as well to enter in this condition, which **should** send a `status` value of `206`, so I think that hard-coding it might not be an issue :wink:

vltr: [update/continue] unless an error occur or something, will the status still be `206` in this case? Just if some kind of read error occurs, which will be prior to the `status` change ... Is there any test in case it seeks for a start point that doesn't exists or try to reads more bytes than it is requested?

ahopkins:
I ran `git rebase -i e89342b36c24afe876bf925ae67b1511d050969b` to try and squash the commits for @ryanrain2016, ... but it just added them as a new commit at the end.

@ryanrain2016, Can you try and undo my changes on your branch and then submit the file changes as a single commit?

sjsadowski:
Closing in favor of #1402

Issue #1325, opened 2018-09-25
Should modify Blueprint's default url_prefix to empty string?


chenjr0719:
The current default value of `url_prefix` is `None`(check [here](https://github.com/huge-success/sanic/blob/30e6a310f132752669a74927530e8bc52a51e98e/sanic/blueprints.py#L18)). This will raise an `TypeError: must be str, not NoneType` if use this default value with `Blueprint.group()` because of [line 57](https://github.com/huge-success/sanic/blob/30e6a310f132752669a74927530e8bc52a51e98e/sanic/blueprints.py#L57) in `blueprints.py`.

Considering to modify the default value from `None` to `''` or add an if statement before line 57.

ahopkins:
Adding an `if` statement would likely to have less of an impact on compatibility issues.

chenjr0719:
I send a PR #1334 to fix this issue and use `os.path.normpath` to avoid `url_prefix` become invalid format like `/api//v1`.
You can also check the test which I added in PR #1334,  it explains why I encounter this issue.
Should I just close this issue now?

vltr:
@chenjr0719 using `os.path` wouldn't break this under Windows?

@sjsadowski @yunstanford I think we should add AppVeyor to our CI list if we are going to support Windows. What's your thoughts?

sjsadowski:
@vltr I think that supporting windows requires a larger group discussion. Previously the maintainers have included many patches for windows compat - like excepting uvloop and ujson from windows installs - so it seems like it's a passive goal. 

I'm fine with making it an active goal, and appveyor does provide free hosting for OSS projects. I'm not sure I want to do it for the 18.12 release though.

chenjr0719:
@vltr well, it really breaks on windows. How about use `pathlib.Path('/api//v1').as_posix()` to deal the compatibility problem? This can also work nicely on Linux and Windows.

vltr:
> @vltr I think that supporting windows requires a larger group discussion.

@sjsadowski that is completely understandable. I _thought_ that Windows was a goal for Sanic (blame me), mostly for some comments regarding being able to install it under Windows with the 0.8.x version, along with [some of the code](https://github.com/huge-success/sanic/blob/master/setup.py#L54) inside `setup.py` (or perhaps lack of it, such as the `Operating System` classifier). I think this could be set for a later discussion as well.

[...]

PR #1334, opened 2018-09-29
Fix TypeError when use Blueprint.group() to group blueprint with defa…

chenjr0719:
…ult url_prefix, Use os.path.normpath to avoid invalid url_prefix like api//v1


sanic/blueprints.py
```diff
@@ -54,7 +55,10 @@ def chain(nested):
                     yield i
         bps = []
         for bp in chain(blueprints):
+            if bp.url_prefix is None:
+                bp.url_prefix = ''
             bp.url_prefix = url_prefix + bp.url_prefix
+            bp.url_prefix = os.path.normpath(bp.url_prefix)
```
yunstanford: I don't think it should call normpath here.

vltr: @yunstanford there's a discussion regarding this on #1325, @chenjr0719 just needs to update the PR afaik.

chenjr0719: @vltr @yunstanford I just remove `normpath` to make this PR simple and focus on the original issue about default value of `url_prefix`.

chenjr0719:
@ahopkins Oops, my bad. The css file is created by `pytest-html` which I used to trace the output and log of unit test. I forgot to clean it up.

PR #1335, opened 2018-09-29
unittests passing on windows again

abuckenheimer:
I think this is mostly straightforward, the two big action points this brought up for me were

1. in order to re-enable the multiprocessing / server events tests on windows we have to come up with an alternative to `signal.alarm`, this has been done [before](https://stackoverflow.com/questions/366682/how-to-limit-execution-time-of-a-function-call-in-python/601168#601168) so I don't think it should be too hard
2. I noticed in test_response.py that the windows version of the test client returns 1 extra byte (a carriage return character) which surprisingly only breaks one test. I'm not exactly sure if this is expected or not, if it is not then I was thinking we may want to come up with some optional mocking functionality so tests are truly cross platform but in the meantime I changed the test to expect different things in different places.

Open to feedback on this but hopefully having passing unit tests is a pretty obvious win 😄

### DIFF
--- a/sanic/testing.py
+++ b/sanic/testing.py
@@ -1,30 +1,12 @@
-import sys
 import traceback
 from json import JSONDecodeError
 from sanic.log import logger
 from sanic.exceptions import MethodNotSupported
 from sanic.response import text
 
-try:
-    try:
-        # direct use
-        import packaging as _packaging
-        version = _packaging.version
-    except (ImportError, AttributeError):
-        # setuptools v39.0 and above.
-        try:
-            from setuptools.extern import packaging as _packaging
-        except ImportError:
-            # Before setuptools v39.0
-            from pkg_resources.extern import packaging as _packaging
-    version = _packaging.version
-except ImportError:
-    raise RuntimeError("The 'packaging' library is missing.")
-
 
 HOST = '127.0.0.1'
 PORT = 42101
-is_windows = sys.platform in ['win32', 'cygwin']
 
 
 class SanicTestClient:
--- a/sanic/testing.py
+++ b/sanic/testing.py
@@ -1,12 +1,30 @@
+import sys
 import traceback
 from json import JSONDecodeError
 from sanic.log import logger
 from sanic.exceptions import MethodNotSupported
 from sanic.response import text
 
+try:
+    try:
+        # direct use
+        import packaging as _packaging
+        version = _packaging.version
+    except (ImportError, AttributeError):
+        # setuptools v39.0 and above.
+        try:
+            from setuptools.extern import packaging as _packaging
+        except ImportError:
+            # Before setuptools v39.0
+            from pkg_resources.extern import packaging as _packaging
+    version = _packaging.version
+except ImportError:
+    raise RuntimeError("The 'packaging' library is missing.")
+
 
 HOST = '127.0.0.1'
 PORT = 42101
+is_windows = sys.platform in ['win32', 'cygwin']
 
 
 class SanicTestClient:
--- a/sanic/worker.py
+++ b/sanic/worker.py
@@ -12,207 +12,199 @@ except ImportError:
 
 try:
     import uvloop
-    import gunicorn.workers.base as base
     asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 except ImportError:
-    base = None
     pass
+import gunicorn.workers.base as base
 
 from sanic.server import trigger_events, serve, HttpProtocol, Signal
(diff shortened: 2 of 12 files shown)


sjsadowski:
@abuckenheimer I think this looks okay, flake8 failed for some reason. Probably need to dig in to that.

abuckenheimer:
sorry didn't realize I missed some linting in worker, will post that tonight

sanic/testing.py
```diff
@@ -1,12 +1,30 @@
...
+        # setuptools v39.0 and above.
+        try:
+            from setuptools.extern import packaging as _packaging
+        except ImportError:
+            # Before setuptools v39.0
+            from pkg_resources.extern import packaging as _packaging
+    version = _packaging.version
+except ImportError:
+    raise RuntimeError("The 'packaging' library is missing.")
+
```
yunstanford: We should not move this piece of code inside sanic module, it should stay in the unit tests file at least.

I'd like to refactor and get rid of this piece of code actually, i think there is one unit test relies on this because of aiohttp versions.

tests/test_worker.py
```diff
@@ -4,16 +4,27 @@
...
 import pytest
+from sanic.app import Sanic
+try:
+    from sanic.worker import GunicornWorker
+except ImportError:
+    pytestmark = pytest.mark.skip(
+        reason="GunicornWorker Not supported on this platform"
+    )
+    # this has to be defined or pytest will err on import
+    GunicornWorker = object
```
yunstanford: We can use `@pytest.mark.skipif` instead.

sanic/worker.py
```diff
@@ -12,199 +12,207 @@
...
-                       self.loop)
-        self._server_settings['before_start'] = ()
-
-        self._runner = asyncio.ensure_future(self._run(), loop=self.loop)
-        try:
-            self.loop.run_until_complete(self._runner)
-            self.app.callable.is_running = True
-            trigger_events(self._server_settings.get('after_start', []),
+# gunicorn is not available on windows
+if base is not None:
```
yunstanford: This is actually unnecessary.

Gunicorn doesn't support windows, and this module requires Gunicorn. I don't think users on windows want to import this worker module.

yunstanford: Also note, gunicorn is not a direct dependency of Sanic.

sanic/testing.py
abuckenheimer: this gets used in both test_keep_alive_timeout and test_request_timeout, testing seemed like the natural place to consolidated this logic. Its complicated enough that I feel like we shouldn't write it twice

tests/test_worker.py
abuckenheimer: does skipif work on class definitions? The issue I had here was that GunicornTestWorker would fail to build if we didn't define GunicornWorker

sanic/worker.py
abuckenheimer: workers seemed like a generic enough idea that I left it but you may be right, should I just let the import error raise when workers tries to import gunicorn then?

sanic/testing.py
yunstanford: 1. `test_keep_alive_timeout` and `test_request_timeout` are 2 unit tests that lives in `tests/` folder, which will not be packaged, uploaded and published. They're used for Sanic's Unit testing and Development. But when you move this code under `Sanic/`, it'll be packaged and published to pypi. I don't think this piece of code is useful to any user whoever `pip install Sanic`.
2. Although `packaging` and `setuptools` are universal tools, i don't think we should import `packaging` and `setuptools` inside Sanic modules, and they're not dependencies of Sanic. Indeed, python dependency management tools like `pip`, `pipenv` rely on `setuptools/packaging`, but tools like `poetry` don't rely on it.
3. If you think it could be reused, `tests/utils.py` probably is good place. But personally, I'd prefer to get rid of this chunk of code and refactor `test_keep_alive_timeout` and `test_request_timeout` a little bit. This piece of code is used for checking aiohttp version and handle them differently, we probably should specify a little bit more strict version constraint for aiohttp.

sanic/worker.py
yunstanford: yes, since the worker module is useless on windows.

tests/test_worker.py
yunstanford: ah i see your point. Then you can use `collect_ignore.append("test_worker.py")` in conftest.py when detecting the platform is windows.

abuckenheimer:
@yunstanford how do these changes look?

yunstanford:
@abuckenheimer Great! Thanks a lot!

PR #1336, opened 2018-09-29
Try not to stringify exception in logging messages

### LINKED ISSUE #1332: Don't use `logger.error(format_exc())`
Whenever an error happens, sanic stringifies the exception and logs it. This interacts really poorly with [Sentry.io](http://sentry.io/) (disclaimer: I am affiliated) and probably every similar tool out there because an error log of the stringified exception will be seen as a distinct error from the error already caught via errorhandlers.

There are a few possible fixes that we can make in our sanic integration without having to touch sanic:

1. Ignore the `"root"` logger. Sanic [logs to a logger just called root](https://github.com/huge-success/sanic/blob/076cf51fb295e7624968db766fd613aa3948289d/sanic/handlers.py#L86). This is undesirable because that logger might actually contain useful errors as well.
2. Error deduplication based on comparing stringified tracebacks. Let's just not go there.

But rather I'd like to change how Sanic logs those errors. Instead of `error(format_exc())`, what about this?

`error("Failed to run handler", exc_info=sys.exc_info())`

Or this?

`logger.exception("Failed to run handler")`

untitaker:
This just fixes the worst offenders that trip up error reporting tools
like Sentry.io

fix #1332

### DIFF
--- a/sanic/handlers.py
+++ b/sanic/handlers.py
@@ -79,7 +79,7 @@ class ErrorHandler:
         response = None
         try:
             if handler:
-                response = handler(req, exception)
+                response = handler(request, exception)
             if response is None:
                 response = self.default(request, exception)
         except Exception:
--- a/sanic/handlers.py
+++ b/sanic/handlers.py
@@ -1,5 +1,5 @@
 import sys
-from traceback import extract_tb
+from traceback import extract_tb, format_exc
 
 from sanic.exceptions import (
     ContentRangeError,
--- a/sanic/handlers.py
+++ b/sanic/handlers.py
@@ -79,10 +79,11 @@ class ErrorHandler:
         response = None
         try:
             if handler:
-                response = handler(request, exception)
+                response = handler(req, exception)
             if response is None:
                 response = self.default(request, exception)
         except Exception:
+            self.log(format_exc())
             try:
                 url = repr(request.url)
             except AttributeError:
@@ -99,12 +100,11 @@ class ErrorHandler:
 
     def log(self, message, level='error'):
         """
-        Override this method in an ErrorHandler subclass to prevent
-        logging exceptions.
+        Deprecated, do not use.
         """
-        getattr(logger, level)(message)
 
     def default(self, request, exception):
+        self.log(format_exc())
         try:
             url = repr(request.url)
         except AttributeError:
--- a/sanic/handlers.py
+++ b/sanic/handlers.py
@@ -1,5 +1,5 @@
 import sys
-from traceback import format_exc, extract_tb
+from traceback import extract_tb
 
 from sanic.exceptions import (
     ContentRangeError,
@@ -83,16 +83,16 @@ class ErrorHandler:
             if response is None:
                 response = self.default(request, exception)
         except Exception:
-            self.log(format_exc())
-            if self.debug:
-                url = getattr(request, 'url', 'unknown')
-                response_message = ('Exception raised in exception handler '
-                                    '"%s" for uri: "%s"\n%s')
-                logger.error(response_message,
-                             handler.__name__, url, format_exc())
+            try:
+                url = repr(request.url)
+            except AttributeError:
+                url = "unknown"
+            response_message = ('Exception raised in exception handler '
+                                '"%s" for uri: %s')
+            logger.exception(response_message, handler.__name__, url)
 
-                return text(response_message % (
-                    handler.__name__, url, format_exc()), 500)
+            if self.debug:
+                return text(response_message % (handler.__name__, url), 500)
             else:
                 return text('An error occurred while handling an error', 500)
         return response
@@ -105,7 +105,14 @@ class ErrorHandler:
         getattr(logger, level)(message)
 
     def default(self, request, exception):
-        self.log(format_exc())
+        try:
+            url = repr(request.url)
+        except AttributeError:
+            url = "unknown"
+
+        response_message = ('Exception occurred while handling uri: %s')
+        logger.exception(response_message, url)
+
         if issubclass(type(exception), SanicException):
             return text(
                 'Error: {}'.format(exception),
@@ -115,9 +122,6 @@ class ErrorHandler:
         elif self.debug:
             html_output = self._render_traceback_html(exception, request)
 
-            response_message = ('Exception occurred while handling uri: '
-                                '"%s"\n%s')
-            logger.error(response_message, request.url, format_exc())
             return html(html_output, status=500)
         else:
             return html(INTERNAL_SERVER_ERROR_HTML, status=500)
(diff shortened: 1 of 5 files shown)


untitaker:
cc @sjsadowski

sanic/handlers.py
```diff
@@ -83,16 +83,16 @@ def response(self, request, exception):
...
         except Exception:
-            self.log(format_exc())
+            try:
+                url = repr(request.url)
+            except AttributeError:
+                url = "unknown"
+            response_message = ('Exception raised in exception handler '
+                                '"%s" for uri: %s')
+            logger.exception(response_message, handler.__name__, url)
+
```
yunstanford: curious why move this piece out of if block ?

sjsadowski:
I'm fine with merging this and closing #1332 once it passes review.

sanic/server.py
```diff
@@ -441,7 +441,7 @@ def bail_out(self, message, from_error=False):
             logger.error("Transport closed @ %s and exception "
                          "experienced during error handling",
                          self.transport.get_extra_info('peername'))
-            logger.debug('Exception:\n%s', traceback.format_exc())
+            logger.debug('Exception:', exc_info=True)
```
abuckenheimer: `bail_out` this is kinda a wonky function signature in general now that I'm thinking about it, if the caller supplies `from_error=True` then the function ignores the required message argument. I'm not exactly sure why we specify `from_error` or not here but I think we could simplify this by having message be an optional argument and this branch triggering if its not supplied

abuckenheimer: not sure if this needs to be addressed in this PR either, seems like a logical local change in the context of whats already here

sanic/handlers.py
```diff
@@ -105,7 +105,14 @@ def log(self, message, level='error'):
         getattr(logger, level)(message)
 
     def default(self, request, exception):
-        self.log(format_exc())
+        try:
+            url = repr(request.url)
+        except AttributeError:
+            url = "unknown"
```
abuckenheimer: this may be a personal preference so I'll leave it to the committer but I think `repr(getattr(request, 'url', "unknown"))` is a bit more concise when there's a logical default. EFAP seems technically more "pythonic" but I always found it kinda verbose🤷‍♂️

sanic/handlers.py
```diff
@@ -83,16 +83,16 @@ def response(self, request, exception):
             if response is None:
                 response = self.default(request, exception)
         except Exception:
-            self.log(format_exc())
```
abuckenheimer: took me a second to grok this but I think part of the point of have the `self.log` function on the `ErrorHandler` class was that the class should funnel logs through it so the method could be overridden in a user subclass. I think removing the call here would be backwards incompatible with user code so instead of doing `logger.exception(...)` below (line 92) we need to do `self.log(..., 'exception')`. the problem with that is the function only takes 1 argument where as `logger.exception` takes multiple which makes using `self.log` how you want also backwards incompatible 😞 . I think changing the signature of log to `def log(message, *args, level='error')` is probably less backwards incompatible but would like to hear some other opinions on this

ashleysommer: I agree, we still need to use `self.log()` here rather than `logger.exception()`, however we need to be able to pass in the extra arguments.
I think changing the log signature to `log(message, *args, level='error')` is a great solution. I'd like to see that changed as part of this PR.

ashleysommer:
@untitaker just curious, does this remove _all_ occurrences of `format_exc()` in the codebase? 
If not, are there are places it could be removed?

sanic/handlers.py
ashleysommer: I actually prefer `repr(getattr(request, 'url', "unknown"))` too, and in this case might be slightly faster than `try/except`.

yunstanford: I'm fine with both, but `try/except` should be faster than `repr(getattr(request, 'url', "unknown"))`. And in case of "url" exists, `try/except` is significantly faster than the other one.

untitaker: Your suggestion is not logically equivalent to this code. In my code it sets `url = "unknown"` whereas in yours it would set `url = repr("unknown")`

sanic/handlers.py
untitaker: How is this more backwards compatible? Anybody who previously overwrote `log` will not be able to deal with this invocation. If anything this makes it worse because it will crash while crashing vs just not using their `log`

sanic/handlers.py
abuckenheimer: good point! actually not entirely sure why you are using repr in the pr, `request.url` is a plain string I thought?

abuckenheimer: @yunstanford did some baseline benchmarking 

```
In [7]: a.url
Out[7]: 'http://127.0.0.1:42101/'

In [8]: %%timeit
   ...: try:
   ...:     url = a.url
   ...: except AttributeError:
   ...:     url = "unknown"
   ...:
5.51 µs ± 250 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

untitaker: Because the last thing I want to have at this point is to crash because of some stupid encoding problem with the URL.

vltr:
@sjsadowski sometimes Python 3.5 gives timeout on Travis for reasons unknown (_yey_ ...). All you have to do is hit "rebuild" (I hope this option is available for you).

untitaker:
@ashleysommer 

>@untitaker just curious, does this remove all occurrences of format_exc() in the codebase?
If not, are there are places it could be removed?

It's still used in response bodies but not for logging

untitaker:
i replied to all review comments, do you have strong opinions on `self.log`?

abuckenheimer:
this is tough, I think maybe the best thing to do here is subclass `ErrorHandler` and override the signature of `log` in that subclass, then the user can opt into the new style handling by passing the new handler to `app.Sanic`, we can then go through a deprecation process with the old style handling and eventually default to the new style

vltr:
@abuckenheimer I think there's a good point in there, either to slowly adopt a new logging approach or even provide a "choose your logging approach".

untitaker:
I would just release 0.9.0. The other breaking change in here is that the `root` logger got renamed.

vltr:
@untitaker we have recently changed the way releases are going to be held with Sanic. Take a look in the announcement: https://sanicframework.org/sanic-is-now-an-organization

untitaker:
How about this? We deprecate `log` and make it a noop by default. We call it exactly like before.

sanic/server.py
untitaker: I'm going to ignore this for this PR. To be honest my only objective here is to have exc_info available on messages.

abuckenheimer:
I think your right, a migration plan seems like more work then strictly necessary here. I'm fine with making `.log` a noop by default

PR #1338, opened 2018-09-30
Check error message and fix some lint error in test config.

hramezani:


tests/test_config.py
```diff
@@ -82,5 +90,6 @@ class Config:
 
 
 def test_missing_config(app):
-    with pytest.raises(AttributeError):
+    with pytest.raises(AttributeError) as e:
         app.config.NON_EXISTENT
+    assert str(e.value) == ("Config has no 'NON_EXISTENT'")
```
vltr: Shouldn't this be inside the `with` scope, since `e` should (at least) be valid there?

hramezani: @vltr Thanks, I moved them inside the `with` scope.

PR #1342, opened 2018-10-03
Handle syntax error in load config file.

hramezani:


tests/test_config.py
```diff
@@ -72,6 +72,17 @@ def test_load_from_missing_envvar(app):
...
 
+def test_load_from_envvar_invalid_syntax(app):
+    config = b"VALUE = some value"
+    with NamedTemporaryFile() as config_file:
+        config_file.write(config)
+        config_file.seek(0)
+        environ['APP_CONFIG'] = config_file.name
+
+        with pytest.raises(SyntaxError):
+            app.config.from_envvar('APP_CONFIG')
```
abuckenheimer: why do from_envvar here? your change is more directly relevant to from_pyfile

sanic/config.py
```diff
@@ -83,6 +83,8 @@ def from_pyfile(self, filename):
         except IOError as e:
             e.strerror = 'Unable to load configuration file (%s)' % e.strerror
             raise
+        except Exception:
+            raise SyntaxError('Invalid syntax in config file.')
```
abuckenheimer: I'm a bit confused here because when I `compile('VALUE = some value', 'name.py', 'exec')` I actually get a `SyntaxError`, which means the original implementation should raise a `Syntaxerror`, what were you seeing?

tests/test_config.py
hramezani: Sorry I missed it. I replace it with `from_pyfile`

sanic/config.py
hramezani: If you put `VALUE = 1/0` in the config file you get a `ZeroDivisionError`. I put `except Exception` because I want to catch all type of exceptions.

yunstanford: I don't think we should wrap all exceptions up, and raise `Syntaxerror` instead. It also hides the true error. If it raises `ZeroDivisionError` because of `VALUE = 1/0`, it should just raise `ZeroDivisionError`.

hramezani: @yunstanford I changed it and added an extra test

sanic/config.py
```diff
@@ -83,6 +83,8 @@ def from_pyfile(self, filename):
         except IOError as e:
             e.strerror = 'Unable to load configuration file (%s)' % e.strerror
             raise
+        except Exception as e:
+            raise
```
vltr: Is this just to raise the "unknown" `Exception` instance `e`? In that case, wouldn't be better to `raise e` or `raise Exception from e` ?

hramezani:
@vltr thanks, I added `happened in config file` to error. maybe it helps the user to find the location of the error.
If you feel it doesn't help I can undo the changes in `sanic/config.py` and just keep the tests. if not we can close the PR :)

abuckenheimer:
I still like the spirit of this I just think we should use error chaining instead of manipulating the arguments, if we created a custom exception we could do something like this
```
class PyFileError(Exception):
    def __init__(self, file):
        super().__init__('could not execute config file %s', file)

... # back in from_pyfile
except Exception as e:
    raise PyFileError(filename) from e
```

vltr:
@abuckenheimer that would be awesome.
@hramezani sorry for stepping on your toes on this :grimacing:

sjsadowski:
Had to revert this. Please see the build failure reports: https://travis-ci.org/huge-success/sanic/builds/440156553

vltr:
There's something missing in there ...

```
NameError: name 'NamedTemporaryFile' is not defined
```

:confused:

sjsadowski:
@vltr reverted it. That's not the only problem. Care to approve #1357 ?

hramezani:
the problem is that `NamedTemporaryFile` is removed. I rebased my local branch with master and fix it in my local machine. how can I deliver it?

PR #1349, opened 2018-10-06
Fix logo in config.py

hramezani:


vltr:
A broader question: will we still use this logo?

hramezani:
@vltr I don't know. I just compare it with the logo in `README.rst` and saw an extra character on it.

vltr:
@hramezani oh yes, that I noticed, sorry. I was just thinking out loud, let's see what other @huge-success/sanic-core-devs thinks of this (about having a logo in the `config.py` file).

sjsadowski:
I'm not a fan of the logo in the config. First thing I had to figure out how to do was disable it so it wouldn't clutter my logs.

abuckenheimer:
shoot I didn't mean to approve these, I think we actually need that extra backslash because it prevents a newline escape

yunstanford:
gonna close this, and i agree we probably don't need this in config.py

PR #1361, opened 2018-10-12
Cancel request when connection lost

yunstanford:
Fix: https://github.com/huge-success/sanic/issues/1360


ahopkins:
This looks good, my only question is if this should be configurable some how? Dis it make sense when sometimes you would NOT want this to happen and the process to continue? I think probably yes.

yunstanford:
yeah, I also thought about this, but the only use case i can think of is triggering/scheduling tasks ? But in that case, the api could be just schedule a callback or async task and return immediately.

vltr:
But isn't there already a good number of tasks inside the protocols to take care of timeouts, the current server time (this being a useless random bug factory), etc? This task is "just another one" to the packet ...

yunstanford:
This is not timeout. If connection is lost, server should not waste resource for continuing serving the request.

vltr:
I know, I just mentioned there are other tasks already for handling a bunch of stuff, like request / response timeouts and another task is really not a big deal.

PR #1365, opened 2018-10-14
Codestyle black

yunstanford:
Start a conversation around adopting `black` for code style. This PR includes,

1. Run black against `sanic` module
2. Added black check in tox, `black --check --verbose sanic`


ahopkins:
I am in favor of "uncompromising" consistency.

vltr:
@yunstanford can't we add some more rules regarding `isort` and `flake8`, as mentioned by the [black documentation](https://github.com/ambv/black#the-black-code-style) itself? Also, I started a forum discussion [regarding this](https://community.sanicframework.org/t/python-and-documentation-code-rules/38/3) some days ago, it might worth looking as well.

yunstanford:
@vltr will take a look, thanks!

vltr:
Thanks, @yunstanford !

yunstanford:
@vltr i'm fine with `isort`, does it provide any check command like `black` ? or just always run isort against the repo ?

And also why do we want to ignore `C901` ?

vltr:
Oh, `C901` I think is residual from my old days with Twisted, just ignore it (in the forums I put just a sample, not a solution for Sanic).

As for `isort`, it plays well with some editors, as `black` respects its rules. It can also be used as a command line check, exactly like `black`. Here's the command line I use inside my `tox.ini` files:

```
    isort --verbose --check-only --diff --recursive src tests setup.py
```

PR #1368, opened 2018-10-16
Add '%' to quote_plus's `safe` parameter in response.redirect

yunstanford:
Fix https://github.com/huge-success/sanic/issues/1356


mohd-akram:
Could we get a release with this? Redirects are broken without it.

PR #1371, opened 2018-10-18
codestyle: Integrate isort

yunstanford:
suggested by @vltr 

This PR includes:
1. ran `isort` against sanic folder
2. add isort check in `tox -e lint`


sanic/app.py
```diff
@@ -1,29 +1,31 @@
...
+from asyncio import CancelledError, ensure_future, get_event_loop
+from collections import defaultdict, deque
 from functools import partial
 from inspect import getmodulename, isawaitable, signature, stack
+from ssl import Purpose, create_default_context
 from traceback import format_exc
 from urllib.parse import urlencode, urlunparse
-from ssl import create_default_context, Purpose
+
+import sanic.reloader_helpers as reloader_helpers
```
abuckenheimer: should we change this to `from sanic import reloader_helpers` so it gets put in the right bucket?

sanic/reloader_helpers.py
```diff
@@ -1,9 +1,10 @@
 import os
-import sys
 import signal
 import subprocess
-from time import sleep
+import sys
+
 from multiprocessing import Process
+from time import sleep
```
abuckenheimer: weird, maybe we add these to `known_standard_library`

vltr: Perhaps a "leaner" `isort` config,  so it won't add a blank line between `import ...` and `from ... import ...` ?

yunstanford: I'm fine with one blank between different import types. no blank is fine too for me.

PR #1376, opened 2018-10-19
Allow sanic test client to bind to a random port

relud:
I like to run tests with detox, but using `sanic.Sanic.test_client` causes issues because of multiple servers trying to simultaneously bind to `sanic.testing.PORT`.

This solves that problem by having `SanicTestClient` use `socket` to bind to a random port when `port=None`.

### DIFF
--- a/docs/sanic/testing.md
+++ b/docs/sanic/testing.md
@@ -59,6 +59,23 @@ the available arguments to aiohttp can be found
 [in the documentation for ClientSession](https://aiohttp.readthedocs.io/en/stable/client_reference.html#client-session).
 
 
+## Using a random port
+
+If you need to test using a free unpriveleged port chosen by the kernel
+instead of the default with `SanicTestClient`, you can do so by specifying
+`port=None`. On most systems the port will be in the range 1024 to 65535.
+
+```python
+# Import the Sanic app, usually created with Sanic(__name__)
+from external_server import app
+from sanic.testing import SanicTestClient
+
+def test_index_returns_200():
+    request, response = SanicTestClient(app, port=None).get('/')
+    assert response.status == 200
+```
+
+
 ## pytest-sanic
 
 [pytest-sanic](https://github.com/yunstanford/pytest-sanic) is a pytest plugin, it helps you to test your code asynchronously.
--- a/sanic/testing.py
+++ b/sanic/testing.py
@@ -3,6 +3,7 @@ from json import JSONDecodeError
 from sanic.exceptions import MethodNotSupported
 from sanic.log import logger
 from sanic.response import text
+from socket import socket
 
 
 HOST = "127.0.0.1"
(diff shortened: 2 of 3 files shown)


relud:
Whoops, missed that. Should be all fixed up now.

docs/sanic/testing.md
```diff
@@ -59,6 +59,22 @@ the available arguments to aiohttp can be found
 [in the documentation for ClientSession](https://aiohttp.readthedocs.io/en/stable/client_reference.html#client-session).
 
 
+## Using a random port
+
+If you need to test using a random port instead of the default with 
+`SanicTestClient`, you can do so by specifying `port=None`
```
harshanarayana: From a general usability standpoint, wouldn't it be good to reverse the behavior? 
i.e. the `port` param defaults to `None` and picks a random port, however, you can specify a port and use for running test cases. IMHO, this would be more readable and easy to understand. 

As an additional option, we can even add an ENV variable `TEST_CLIENT_PORT` and use that to provide a much dynamic port assignment.

sanic/testing.py
```diff
@@ -10,17 +11,21 @@
 
 class SanicTestClient:
     def __init__(self, app, port=PORT):
+        """Use port=None to bind to a random port"""
         self.app = app
         self.port = port
 
-    async def _local_request(self, method, uri, cookies=None, *args, **kwargs):
+    async def _local_request(
+        self, method, uri, sock=None, cookies=None, *args, **kwargs
```
abuckenheimer: My main issue here is that this function signature now has duplicate information, it seems like if we have all the information we need in the caller to infer the url then it should be the callers responsibility for doing so.The easiest way to do that would be `self.port = sock.getsockname()[1]` after 92, lazily setting the port _should_ be fine but it may be worth thinking about. Another option though is we have a short circuit on 23 for uri's that look like urls, we could similarly do `uri = "http://{sock[0]}:{sock[1]}{uri}".format(sock=sock.getaddrname(), uri=uri)` at after 92 as well.

docs/sanic/testing.md
abuckenheimer: I like this idea but unless other people pipe up on it I think adding this functionality and changing the default can be addressed separately

ahopkins:
@relud  Are you still pursuing this one? 

As for the question of changing the functionality, or whether there should be something separate (`TEST_CLIENT_PORT` for example), I would lean towards making it an opt in feature and not changing the default.

sanic/testing.py
relud: I like the idea, but setting `self.port` causes multiple requests to bind the same port, which may fail if the OS assigned the random port to something else between requests. I think it would be better to set `self.sock` to `None` in `__init__` and set and unset it from `_collect_request()` below.

relud:
@ahopkins yes, I'm still pursuing this, I just haven't had time to @abuckenheimer's request because of a very busy december. I'm planning to add the requested test and documentation next week after Tuesday.

ahopkins:
🤘 Happy New Year 🥳

docs/sanic/testing.md
seemethere: Yeah I think changing the default functionality should come with a different PR

Nelapa:
Vote for PR, feature crashes our CI pipelines too.

sjsadowski:
I'm moving to 19.6, hopefully we can get the changes requested implemented in the PR by then.

@relud if you do get this done before 10-Mar ping me directly so I can get people to review and I'll merge it if the reviewers are satisifed.

sanic/testing.py
relud: > we could similarly do `uri = "http://{sock[0]}:{sock[1]}{uri}".format(sock=sock.getaddrname(), uri=uri)` at after 92 as well.

I ended up going with that, more or less

relud:
@sjsadowski I believe this has now been updated to address reviews.

PR #1378, opened 2018-10-22
Fix some test files lint errors.

hramezani:


yunstanford:
Just fyi, we don't run lint against `tests/` folder. Personally i don't care codestyle in unit tests, but i'm not oppose to it and if we wanna fix those, can we bring up a PR for addressing them all once ? (since i saw a similar PR before)

hramezani:
@yunstanford I fixed all tests lint error.
@ahopkins, first I fix them with `autopep8` and then check it and change them by hand

Issue #1374, opened 2018-10-19
readthedocs - tag not available


sjsadowski:
**Describe the bug**
When using the tag to view specific version documentation, the docs are unavailable.

**Code snippet**
Visit https://sanic.readthedocs.io/en/latest and select the tag for '0.8.3' or visit https://sanic.readthedocs.io/en/0.8.3/ directly

**Expected behavior**
Documentation is available

This is more irksome than a real problem at the moment, however when we release 18.12 and for further versions we should be making sure that 'latest' and 'stable' are appropriately linked to the version they represent, as well as having documentation for the release version (especially for LTS) directly available.

yunstanford:
And the readthedoc build is failing for long time.

yunstanford:
Actually there is another error that we need figure out, you can take a look at above link

yunstanford:
The link in that PR

ahopkins:
I just looked at the comment about the build failing....

Looks to me like an rst issue. Take a look at that message. There is a location that needs and extra line break after the code block definition. I'm on my phone so I cannot make the change right now. Can e should reopen the PR, fix that and then run sphinx to make sure it is solved.

sjsadowski:
@huge-success/sanic-core-devs someone double-check me but it looks like it's failing because of a gunicorn dependency not being met?

http://readthedocs.org/projects/sanic/builds/7989537/

ahopkins:
@sjsadowski This is what I was referring to. It looks like a rst compile error in that there is a missing li e break after the code block line. I meant to check when I was in the office earlier and forgot.

sjsadowski:
Ah, I see now. Thanks for the clarification @ahopkins

ahopkins:
See PR #1383, there were 5 files that were causing an issue.

PR #1383, opened 2018-10-26
Fix documentation build errors

ahopkins:
There were a number of build issues relating #1374 for the documentation. See also: https://readthedocs.org/projects/sanic/builds/7989537/

Merging this should restart the documentation build process to pass on RTD.

### DIFF
--- a/docs/sanic/middleware.md
+++ b/docs/sanic/middleware.md
@@ -17,7 +17,7 @@ string representing its type: `'request'` or `'response'`.
 
 The simplest middleware doesn't modify the request or response at all:
 
-```python
+```
 @app.middleware('request')
 async def print_on_request(request):
 	print("I print when a request is received by the server")
@@ -33,7 +33,7 @@ Middleware can modify the request or response parameter it is given, *as long
 as it does not return it*. The following example shows a practical use-case for
 this.
 
-```python
+```
 app = Sanic(__name__)
 
 @app.middleware('response')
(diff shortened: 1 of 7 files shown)


ahopkins:
I made no changes to `master` other than the docs....

[Weird failure](https://travis-ci.org/huge-success/sanic/jobs/446574436)

docs/sanic/middleware.md
```diff
@@ -60,7 +60,7 @@ and the response will be returned. If this occurs to a request before the
 relevant user route handler is reached, the handler will never be called.
 Returning a response will also prevent any further middleware from running.
 
-```python
+```
```
yunstanford: Why remove python ?

yunstanford: Also weird that we're mixing md and rst

ahopkins: Because it doesn't recognize the syntax as python because of the async/await language. It was creating a build error. Really these MD files need to be switched to RST. I was thinking of doing some documentation overhaul work after the LTS.

sjsadowski: Switching all the docs to rst is pretty high on my list too.

yunstanford: Did you install the sphinx asyncio plugin when testing locally ? I think async/await should be supported. But we can take look into this later.

ahopkins: I did not.

ahopkins: I figured there must be a way to get it to understand async await... I just was punting it to another day so we don't need a dependency for it now.

PR #1387, opened 2018-10-30
Resolve build of latex documentation relating to markdown lists

ahopkins:
I jumped the gun on #1383. It only resolved the HTML docs (which are now updated on RTD). There was still a failure for latex relating to enumerated lists in MD.

Again, I opted for the easier route on this one just to get it built since we will be redoing the docs in the near future to move everything to RST.


ahopkins:
Passing doc build: https://readthedocs.org/projects/sanic/builds/8027279/ 💪

Issue #1390, opened 2018-10-31
Remove kwargs from app.run and/or raise Exception on loop as param


ahopkins:
The [`app.run`](https://github.com/huge-success/sanic/blob/master/sanic/app.py#L823) method accepts `**kwargs`. I do not think it needs to, as there is only one call to it in the method that could easily be removed.

The issue is that it falsely allows developers to do:

```
loop = asyncio.get_event_loop()
app.run(..., loop=loop)
```

See #1388 for example.

To get this working as intended, the solution is to use `app.create_server` as [mentioned in the docs](https://sanic.readthedocs.io/en/latest/sanic/deploying.html#asynchronous-support).

---

I think we can do one better either by removing `**kwargs` or raising an Exception if `loop` is passed to the method. Doing this would be helpful for the developer to get them to be able to diagnose the issue on their own.

AndresSan6:
Hello! I can help with issue, I will work on it today. 

I can try by removing **kwargs and making an exception to let the user know when they pass an unexpected argument.

ahopkins:
Awesome! Thanks.

If we remove kwargs, then python will provide the exception. I'm thinking the best for the developer might be too leave kwargs, and raise an exception if loop is passed. I've seen this a number of times. If we rely upon the default  python exception it us less helpful than one that specifically tells them how to accomplish the goal.

AndresSan6:
I made a PR #1391 for this issue, please let me know if this changes helps you.

PR #1391, opened 2018-10-31
Handle "loop" exception in app.py

AndresSan6:
This PR solves issue #1390

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -835,6 +835,11 @@ class Sanic:
         access_log=True,
         **kwargs
     ):
+        try:
+            assert "loop" not in kwargs
+        except AssertionError:
+            raise TypeError("loop is not a valid argument. For asynchronous support, check https://sanic.readthedocs.io/en/latest/sanic/deploying.html#asynchronous-support")
+
         """Run the HTTP Server and listen until keyboard interrupt or term
         signal. On termination, drain connections before closing.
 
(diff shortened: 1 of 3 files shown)


sanic/app.py
```diff
@@ -835,6 +835,11 @@ def run(
         access_log=True,
         **kwargs
     ):
+        try:
+            assert "loop" not in kwargs
```
ahopkins: Rather than do it as an `assert` statement (which can be turned off), I think it would be better just to do:

```
if "loop" in kwargs:
```

sanic/app.py
```diff
@@ -835,6 +835,11 @@ def run(
...
+        except AssertionError:
+            raise TypeError("loop is not a valid argument. For asynchronous support, check https://sanic.readthedocs.io/en/latest/sanic/deploying.html#asynchronous-support")
```
ahopkins: How about:

"loop is not a valid argument. To use an existing loop, change to create_server().\nSee more: https://sanic.readthedocs.io/en/latest/sanic/deploying.html#asynchronous-support"

sjsadowski:
@AndresSan6 - linting is failing. Can you clean up your commit? https://travis-ci.org/huge-success/sanic/jobs/449073363

ahopkins:
Not your fault. Some linting issues made their way into `master`. Checkout this PR for the linting fixes: https://github.com/huge-success/sanic/pull/1373

AndresSan6:
Now is fixed. Sorry for all the last commits, I was not understanding properly which was the reason for failing the test. Thanks for your comments!

AndresSan6:
Hello @ahopkins , any updates?

ahopkins:
👍

PR #1393, opened 2018-11-04
Fix pickling blueprints Fixes #1392

### LINKED ISSUE #1392: Blueprint with multiple workers not usable on Windows due to pickling error
**Describe the bug**
When using a blueprint with multiple workers on windows, sanic fails on startup due to a failure to pickle the route:

> Exception has occurred: _pickle.PicklingError
Can't pickle <class 'sanic.blueprints.Route'>: attribute lookup Route on sanic.blueprints failed

**Code snippet**
```python
from sanic import Sanic
from sanic import Blueprint

blueprint = Blueprint("API_blueprint")

@blueprint.route("/")
async def test(request):

app = Sanic()

app.blueprint(blueprint)

def main():
    app.run(workers=2)
```

**Expected behavior**
It should be possible to run with multiple workers using blueprints in the same way as it is using the sanic app object directly

 - OS: Windows
 - Version 7

**Additional context**
A similar issue with pickling was [recently fixed](https://github.com/ashleysommer/sanic-cors/issues/14#issuecomment-434968189) in sanic-cors. No idea if the resolution to that one will be helpful

ashleysommer:
Change the string passed in the "name" section of the namedtuples in Blueprint to match the name of the Blueprint module attribute name.
This allows blueprints to be pickled and unpickled, without errors, which is a requirment of running Sanic in multiprocessing mode in Windows.
Added a test for pickling and unpickling blueprints
Added a test for pickling and unpickling sanic itself
Added a test for enabling multiprocessing on an app with a blueprint (only useful to catch this bug if the tests are run on Windows).
Fixes #1392

### DIFF
--- a/tests/test_multiprocessing.py
+++ b/tests/test_multiprocessing.py
@@ -69,7 +69,7 @@ def test_pickle_app(app, protocol):
 
 
 @pytest.mark.parametrize('protocol', [3, 4])
-def test_pikcle_app_with_bp(app, protocol):
+def test_pickle_app_with_bp(app, protocol):
     from sanic import Blueprint
     bp = Blueprint('test_text')
     bp.route('/')(handler)


ashleysommer:
I'll squash all these commits when I get the linter tests to pass

ashleysommer:
Ok, commits squashed and lints passing now.

sanic/blueprints.py
```diff
@@ -17,11 +17,15 @@
         "name",
     ],
 )
-FutureListener = namedtuple("Listener", ["handler", "uri", "methods", "host"])
```
yunstanford: not related, I like `attrs` over `namedtuple`.
And `namedtuple` is also slow when initializing, but doesn't matter in this case.

Issue #1386, opened 2018-10-29
Middleware for Blueprint.group()


chenjr0719:
**Is your feature request related to a problem? Please describe.**
Not related to any problem but about better usage of `Blueprint.group()`. I've to say that middleware of `Blueprint()` is a really useful feature. And I hope that `Blueprint.group()` can also support this useful feature.

**Describe the solution you'd like**
For example, I have lots of blueprints and all of them are grouped by `Blueprint.group()`. In this case, if I want to apply a function(authorization for example) to all blueprints, I would like to just add middleware to `Blueprint.group()`.

**Additional context**
Here is an example:
```python
from sanic import Blueprint

# Assume there are already lots of blueprints are created.
bp_1 = Blueprint('bp_1', url_prefix='/bp_1')
.
.
.
bp_10 = Blueprint('bp_10', url_prefix='/bp_10')

# Use Blueprint.group() to group all blueprint.
bp_api_group = Blueprint.group(bp1, ..., bp_10, url_prefix='/api)

# Apply an authorization function to Blueprint.group()
@bp_api_group.middleware('request')
def authorization(request):
    pass
```

Let me know your idea.

harshanarayana:
@chenjr0719 @ahopkins Let me take a stab at this?

chenjr0719:
@harshanarayana Yes please go get a try. Actually, I have no idea how to implement this feature.:sweat_smile:

harshanarayana:
@chenjr0719 @ahopkins @sjsadowski I just created a new PR #1399 for the same.

ahopkins:
@harshanarayana Thanks! I will take a look. @vltr has been working on [`sanic-boom`](https://github.com/vltr/sanic-boom) that does add this ability now. 

So, if you want to use it now, maybe give that a try.

I am marking this as a Future Release as I do not see this as a feature we want to add for 18.12. We are trying to keep the upcoming December (and **first**) LTS release with a limited amount of "new" features.

I will let @sjsadowski and @yunstanford decide if they want to add that to 19.03 or not.

vltr:
@harshanarayana as @ahopkins mentioned, I had my time working on a solution for this. Since I wanted a [more performant router](https://github.com/vltr/very-stupid-simple-routers-bench) for my Sanic applications, I chose to add "route specific" middlewares inside the routing tree (I created a project called [xrtr](https://xrtr.readthedocs.io/en/latest/) for that, based on a Radix Tree).

[...]

PR #1399, opened 2018-11-06
Enable Middleware Support for Blueprint Groups

harshanarayana:
This commit will enable the users to implement a middleware at the blueprint group level whereby enforcing the middleware automatically to each of the available Blueprints that are part of the group.

This will enable a simple way in which a certain set of common features and criteria can be enforced on a Blueprint group. i.e. authentication and authorization

This commit will address the feature request raised as part of Issue #1386


harshanarayana:
@sjsadowski I am working on adding the documents. I will update the PR with the changes shortly.

harshanarayana:
@vltr Sure. Consider it done.

harshanarayana:
@sjsadowski I added the basic documentation on usage and tested a local `sphinx` build to ensure the changes are working as expected. 

@vltr  I've added additional unit test cased with few more HTTP methods as well as `url_prefix` and some parameterized `route`

sjsadowski:
@harshanarayana thanks. This won't go in to 18.12 but it will go in after that. Appreciate your help.

harshanarayana:
@sjsadowski I'm OK with this getting merged into any version you see fit. No issues from my side. Just give me a heads-up if anything else needs to be addressed in this PR.

sanic/blueprint_group.py
```diff
@@ -0,0 +1,93 @@
...
+    class as a list/tuple inside the existing implementation.
+    """
+
+    def __init__(self, url_prefix=None):
+        """
+        Create a new Blueprint Group
+
+        :param url_prefix: URL: to be prefixed before all the Blueprint Prefix
+        """
+        self._blueprints = list()
```
yunstanford: `self._blueprints = []`

sanic/blueprint_group.py
```diff
@@ -0,0 +1,93 @@
...
+        :param blueprint: Instance of Blueprint
+        :return: None
+        """
+        self._blueprints.append(blueprint)
+
+    def __iter__(self):
+        """Tun the class Blueprint Group into an Iterable item"""
+        return self
+
+    def __next__(self):
```
yunstanford: It's basically iterating self._blueprints list here...I'd prefer
```
def __iter__(self):
    return iter(self. _blueprints)

# No need define __next__
```
And in current implementation, `_iter_position` always increments ? it can only iterate once ?

sanic/blueprint_group.py
```diff
@@ -0,0 +1,93 @@
+class BlueprintGroup(object):
```
yunstanford: it's default to new-style classes in python3, no need `object` here.

sanic/blueprints.py
```diff
@@ -284,6 +287,11 @@ def register_middleware(_middleware):
             middleware = args[0]
             args = []
             return register_middleware(middleware)
+        elif len(args) == 2 and len(kwargs) == 0 and callable(args[0]):
```
yunstanford: this check condition is a little bit error-prone...

sanic/blueprints.py
```diff
@@ -71,15 +72,17 @@ def chain(nested):
             for i in nested:
                 if isinstance(i, (list, tuple)):
                     yield from chain(i)
+                elif isinstance(i, (BlueprintGroup)):
```
yunstanford: don't need `()`

sanic/blueprint_group.py
harshanarayana: I usually prefer `list()` and `dict()` because of readability.  But if `[]` is the standard, then I can change it.

sanic/blueprint_group.py
harshanarayana: @yunstanford Makes sense. The lesser code is better.

sanic/blueprint_group.py
harshanarayana: Done

sanic/blueprints.py
harshanarayana: @yunstanford No need to check for `kwargs` ? Or you don't like the idea of `len(args)` check? A bit more context would help me make the necessary changes.

sanic/blueprint_group.py
harshanarayana: @yunstanford But i think it's still a good idea to keep `__next__` because, it can help if you are using `next(x)` constructs to handle the iteratable item in different stages.

yunstanford: As for `next()`, you could also utilize sth. like
```
bp_group_iter = iter(bp_group)
bp = next(bp_group_iter)
```

yunstanford: I don't wanna make the class itself a iterator, which doesn't make sense.

But the class could provide methods like `.iterator()` or use `iter()`, that returns iterator.

yunstanford: It'll have problems if you make the class itself iterator.

e.g. in this implementation, you maintain the iterator position `_iter_position`, and there is no way for two coroutines to iterate it at the same time, since they'll always have the same position.

also note, `_iter_position` always increments ? it can only iterate once ?

harshanarayana: @yunstanford As per the specs in PEP, isn't it expected that you get a `StopIteration` exception on exhaustion of the items in case of using `next()`?

> Iterator objects returned by either form of iter() have a next()
    method.  This method either returns the next value in the
    iteration, or raises StopIteration (or a derived exception class)
    to signal the end of the iteration.

The `__iter__` method has been modified to return `iter(self._blueprints)` so you can iterate over it multiple times in multiple places like using them in list comprehension etc. Am I missing something ?

sanic/blueprints.py
harshanarayana: @yunstanford Would you mind adding a bit more context to this item so that I can address it the right way. Thanks again

sanic/blueprint_group.py
yunstanford: It'll raise `StopIteration` when exhausted all items. `iter([])` basically returns an iterator.

harshanarayana: @yunstanford That's what is being done now in the code right? The `__iter__` method supports the re-useable mechanism to iterate over the class object directly while `__next__` provides another mode while raising a `StopIteration`

yunstanford: I don't see any benefit around adding current `__next__`, and it's quite confusing since it only supports to iterate once and after one full iteration, `_iter_position` is totally useless...

harshanarayana: @yunstanford No worries. Let me get rid of it.

sanic/blueprints.py
yunstanford: Yeah, none of these conditions is specific to register blueprint group. I'm a little worry about that this'd be hard to maintain in the future when adding more things.

harshanarayana: @yunstanford Understandable. Logically speaking, we should completely get rid of this entire check section. It should be a generic implementation where we pass everything we get to the method and let method decide how it wants to handle. 

Let me see what I can do. I don't want to change the existing one too much

harshanarayana: @yunstanford I tweaked the way the blueprint group middleware was being identified to make it more definitive instead of the length check as you suggested. Please let me know if this looks okay.

ahopkins:
@yunstanford Are you good with this one?

harshanarayana:
@huge-success/sanic-core-devs Can you please take a look at this and let me know if there are any other changes required to be done on this PR?

ashleysommer:
@harshanarayana
Overall this looks pretty good.
When this gets merged, tools like SanicPluginsFramework, and other plugins may no longer work as expected, because they tend to assume the only classes that can have middlewares on them are Sanic App and Blueprints. They may need to be modified to be BlueprintGroup aware, but that is inevitable fallout.

The only code issue I see is the use of `group.blueprint = bp` where you override the property setter to append `bp` to a list. That is a counter-intuitive pattern to me, as usually the python assignment operator `=` is used to replace the target, not append to it.

The way I've usually implemented this kind of functionality is to create a method called `.append()` on the class, and call `.append(bp)` as if the class itself is a list. This also matches the other list-like functionality on BlueprintGroup like the iteration behaviour.

harshanarayana:
@ashleysommer 

>tools like SanicPluginsFramework, and other plugins may no longer work as expected, because they tend to assume the only classes that can have middlewares on them are Sanic App and Blueprints

The idea was to keep almost all the existing behaviors of `Blueprint.group` intact so that the other plugins and frameworks don't run into any major breaking issues. I will take a look around the existing plugins and see if any of them will actually have a problem because of this change. 

> The way I've usually implemented this kind of functionality is to create a method called `.append()` on the class, and call `.append(bp)` as if the class itself is a list.

That sounds like a really good thing to do. Let me get that change sorted.

ashleysommer:
> The idea was to keep almost all the existing behaviors of Blueprint.group intact so that the other plugins and frameworks don't run into any major breaking issues.

Yeah, it's a good implementation. I didn't mean to suggest it might cause breaking changes. I don't think that. I just mean that some plugins will need to be BlueprintGroup aware, specifically if it is a middleware-based plugin and it wants to apply its plugin/middleware functionality to blueprint groups.

harshanarayana:
> I didn't mean to suggest it might cause breaking changes

Sorry if my message came across that way. I wasn't meaning that. After you mentioned, I realized that there might be some breaking changes wrt these plugins if they are doing an `isinstance` like check for something.

ashleysommer:
@sjsadowski
Why was this closed?
I thought it should be merged.

sjsadowski:
Ugh. It has to do with how projects work. I closed all the PRs that looked like they were merged from the project screen. Gotta go fix now...

ahopkins:
I agree. I think this is a positive breaking change.

PR #1400, opened 2018-11-06
Add test for sanic.root logger and update the docs of logging

chenjr0719:
This PR aims to add a test case to prevent non-functional default logger and also update the example in the document of logging.

BTW, I find out that `reset_logging()` in previous tests will break default logger(because of `reload(logging)` I guess), so I disable it and all tests still pass, this quite confusing me.:confused:


yunstanford:
Hey, can you resolve conflicts ?

chenjr0719:
@yunstanford Sorry, didn't notice that conflicts. I will update my PR.

docs/sanic/logging.md
```diff
@@ -9,17 +9,53 @@ A simple example using default settings would be like this:
...
 async def test(request):
-    return response.text('Hello World!')
+    logger.info('Here is your log')
+    return text('Hello World!')
 
 if __name__ == "__main__":
   app.run(debug=True, access_log=True)
 ```
 
+After the server is running, you can see some messages looks like:
```
ahopkins: Can you take this logo part out of the documentation?

chenjr0719: @ahopkins Sure, no problem. But is that mean the `debug` option in `app.run()` also needed to modify? Or there may some user will be confused why there is a logo show up in logs.

ahopkins: I mean there is no reason to add it into the docs right now because we likely will be removing it in the future. I think the likelihood of confusion from it being in someone's logs but not in the documentation is low.

Issue #1398, opened 2018-11-06
Random Failures during tox run on Mac


harshanarayana:
**Describe the bug**
When running Unit Test in local machine (mac), few test cases randomly fail and re-cover during re-run.

**Error Output**
```bash
# Test session starts (platform: darwin, Python 3.5.6, pytest 3.3.2, pytest-sugar 0.9.1)
    def test_default_server_error_request_timeout():
        client = DelayableSanicTestClient(request_timeout_default_app, None, 3)
        request, response = client.get('/1')
>       assert response.status == 408
E       assert 200 == 408
E        +  where 200 = <test_request_timeout.DelayableTCPConnector.RequestContextManager object at 0x10723ccf8>.status
tests/test_request_timeout.py:194: AssertionError
```

```bash
# Test session starts (platform: darwin, Python 3.7.0, pytest 3.3.2, pytest-sugar 0.9.1)
    def test_default_server_error_request_timeout():
        client = DelayableSanicTestClient(request_timeout_default_app, None, 3)
        request, response = client.get('/1')
>       assert response.status == 408
E       assert 200 == 408
E        +  where 200 = <test_request_timeout.DelayableTCPConnector.RequestContextManager object at 0x112d87470>.status
tests/test_request_timeout.py:194: AssertionError
```

```bash
# py37-no-ext develop-inst-nodeps: <path>
# py37-no-ext installed: aiofiles==0.4.0,aiohttp==3.2.1,async-generator==1.10,async-timeout==3.0.1,attrs==18.2.0,beautifulsoup4==4.6.3,chardet==2.3.0,coverage==4.5.1,gunicorn==19.9.0,httptools==0.0.11,idna==2.7,multidict==4.4.2,pluggy==0.6.0,py==1.7.0,pytest==3.3.2,pytest-cov==2.6.0,pytest-sanic==0.1.13,pytest-sugar==0.9.1,-e git+git@github.com:harshanarayana/sanic.git@e3a27c2cc485d57aa1ff87d9f69098e4ab12727e#egg=sanic,six==1.11.0,termcolor==1.1.0,websockets==6.0,yarl==1.2.6
# py37-no-ext run-test-pre: PYTHONHASHSEED='27345397'
# py37-no-ext runtests: commands[0] | pytest tests --cov sanic --cov-report=
[...]

PR #1403, opened 2018-11-07
Fix Content-Length Mismatch while using json and ujson

harshanarayana:
The current implementation of `sanic` attempts to make use of `ujson` if it's available in the system and if not, it will default to the inbuilt `json` module provided by python.

The current implementation of `ujson` does not provide a mechanism to provide a custom `separators` parameter as part of the `dumps` method invocation and the default behavior of the module is to strip all the spaces around separators such as `:` and `,`. This leads to an inconsistency in the response length when the response is generated using the `ujson` and inbuilt `json` module provided by python.

To maintain the consistency, this commit overrides the default behavior of the `dumps` method provided by the `json` module to add a `separators` argument that will strip the white spaces around these character like the default behavior of `ujson`

This addresses the issue referenced in #1398

PS: Fixed some basic linter reported issue in the test files as well.

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -1,3 +1,4 @@
+from functools import partial
 from mimetypes import guess_type
 from os import path
 from urllib.parse import quote_plus
@@ -12,7 +13,11 @@ from sanic.helpers import STATUS_CODES, has_message_body, remove_entity_headers
 try:
     from ujson import dumps as json_dumps
 except BaseException:
-    from json import dumps as json_dumps
+    from json import dumps
+
+    # This is done in order to ensure that the JSON response is
+    # kept consistent across both ujson and inbuilt json usage.
+    json_dumps = partial(dumps, separators=(",", ":"))
 
 
 class BaseHTTPResponse:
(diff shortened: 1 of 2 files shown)


ahopkins:
Awesome catch! Taking a look now.

harshanarayana:
@ahopkins Adding a few more unit test with responses containing different data types. Please wait before themerge.

harshanarayana:
@ahopkins Done.

ahopkins:
No worries. I will not merge. 

As a general practice, we've adopted a strategy to have two release managers for a year cycle. Their job is to do all merges to master and steer the release cycle. Right now it is @sjsadowski and @yunstanford. But the implementation looks sound. And more tests can never hurt.

harshanarayana:
Looks like the CI jobs is stuck somewhere. It ran the [Unit Tests for the second commit as well](https://travis-ci.org/huge-success/sanic/builds/451976354) , but the results aren't being reported. Strange.

yunstanford:
I've restarted travis ci build

Issue #1380, opened 2018-10-25
Port not provided in URLs produced with `url_for`


eric-spitler:
**Describe the bug**
The default port for a Sanic app is 8000, which by HTTP standards is a custom port. This port can also be set during calls to `app.run()`.

When using `app.url_for('handler.name', _external=True)`, the resultant URL does **not** contain the port without manual intervention.

This line is the final step before the application returns the URL. The URL is build using `urlunparse` which takes a `tuple` of values as would be producted by [urlparse](https://docs.python.org/3.6/library/urllib.parse.html#urllib.parse.urlparse). The snippet below shows that only the first 6 arguments are provided, which does not include the port (unless included as part of `netloc`).
https://github.com/huge-success/sanic/blob/905c51bef03818c629173938a1183b38dd0ebc65/sanic/app.py#L695

In order to get port, the value of `app.confg.SERVER_NAME` **must** be set to the hostname and port
```python
app.config.SERVER_NAME = '{}:{}'.format(host, port)
app.run(host=host, port=port)
```
Thus, the default behavior for creating the URL is incorrect as it requires manual attribute setting for every single application.

**Expected behavior**
Calls to `app.url_for()` with the `_external` flag should build the correct URL, including the port number, such as `http://hostname:8000` without forcing every application to specify the `SERVER_NAME` config attribute.

**Environment (please complete the following information):**
 - OS: CentOS 7.5
 - Version 0.7.0

spitzaf:
Thinking on this more last night, I can see the use case for things like Cloud Foundry deploys where the internal port being served would be obfuscated by the cloud instance, so the application should report `:80` or `:443`, even if its running at `:8000`.

Perhaps a flag then to enable port inclusion?

sjsadowski:
Port 80 is the default port for HTTP and 443 is default for https, if the external URL is serving something like 'http://www.example.com/blah' then there's no real reason for port inclusion - however in every other case where the port is non-standard, I agree with your assessment that it should be included in the url. Possibly just inclusion of the port for the external uri should be enough. I'm sure one of the better devs will weigh in on it.

eric-spitler:
Sorry, when I said "should report `:80`" I meant that the application would need to generate URLs as if it were on that port (despite running on 8000), meaning the behavior as it stands right now. But this is also just for the case where there is an obfuscation layer or load balancer between callers and the application.

ahopkins:
[...]

PR #1406, opened 2018-11-08
add _port to url_for, update routing doc

lixxu:


sanic/app.py
```diff
@@ -690,6 +698,16 @@ def url_for(self, view_name: str, **kwargs):
 
         # parse the remainder of the keyword arguments into a querystring
         query_string = urlencode(kwargs, doseq=True) if kwargs else ""
+
+        if external and port:
+            port = int(port)
```
vltr: I know this may sound silly, but shouldn't we wrap this call with a `try/except` in case `int` raises a `ValueError` or `TypeError`? Just in case someone tries to pass some random string or some other weird object to the `_port` argument or `SERVER_PORT` configuration key? At least we will prevent any issues in the future with this ...

lixxu: yes, sounds reasonable.

vltr: Great, thanks @lixxu !

sanic/app.py
```diff
@@ -700,13 +700,17 @@ def url_for(self, view_name: str, **kwargs):
...
-            if (
-                (scheme == "http" and port != 80)
-                or (scheme == "https" and port != 443)
-                or (scheme in ("ftp", "ftps") and port != 21)
-            ):
-                netloc = "{}:{}".format(netloc, port)
+            try:
+                port = int(port)
+            except (ValueError, TypeError):
+                pass  # just ignore it
```
vltr: Yeah, I think we can just ignore this (for now). If someone complains in the future, we can perhaps just add a `logger.warn` call here and that should suffice.

lixxu: Don't know why lint check failed.

vltr: @lixxu try to run `black` to format the code (in the root of the repo):
```
$ black ./sanic
```

lixxu: oops, 69 files reformatted, 19 files left unchanged. Should I commit these changes? a lot files, may conflict with other PRs.

vltr: Have you rebased with the last master?

lixxu: sorry, I ran the cmd outside of sanic folder, now only the 2 files updated. :)

vltr: Ohh, that makes sense :smile:

lixxu:
22 is used for `ssh`, and I think (maybe wrong) it may have no chance to generate urls for `ssh`. Or you means for `sftp`?

ahopkins:
Yes, `sftp`, which ... as far as I know is just a cover for `ssh` on port 22.

Regardless, is that even something that we want to be supporting? Why have any ftp at all?

vltr:
I agree with @ahopkins . We should stick to `http` and `https` since Sanic is a web framework [...]

lixxu:
ftp and ftps are from testing.py, they are there maybe in case that someone use sanic to write ftp/ftps server?

vltr:
That would have the need of a completely different protocol implementation ... As well as a lot of FTP specific features like mapping all the commands, perhaps build a proper FS integration and etc, etc ... I think it'll be too much work to make it run on top of Sanic - unless someone already had the trouble to do something like that (and I would be really impressed).

If I needed (today) to implement my own custom FTP server, I would go with [Twisted](https://twistedmatrix.com/documents/current/_downloads/ftpserver.py) or [aioftp](https://aioftp.readthedocs.io/). No need to reinvent the wheel ...

lixxu:
so, should remove ftp stuff from `url_for`?

vltr:
IMO it makes more sense (inside Sanic) to remove `ftp` from the "port normalization" part of the function.

sanic/app.py
```diff
@@ -637,14 +637,22 @@ def url_for(self, view_name: str, **kwargs):
...
+            if netloc and "://" in netloc[:8]:
                 netloc = netloc.split("://", 1)[-1]
 
+            if not netloc:
+                scheme = ""
+                external = False
+
+        port = kwargs.pop("_port", None)
+        if port is None:
+            port = self.config.get("SERVER_PORT", "")
```
harshanarayana: Wouldn't it be much simpler to merge all of these into a single line and let python handle it rather than the conditional branching?

```python
port = kwargs.pop('_port', self.config.get('SERVER_PORT', ''))
```

lixxu: single line format will not get the value from configuration if someone passes `_port=None` as parameter.

harshanarayana: @lixxu 
The following should give you the same output.

```python
port = kwargs.pop('_port', self.config.get('SERVER_PORT', '')) or ''
```

harshanarayana: Just FYI, 

```bash
>>> timeit.timeit(stmt=lookup, number=1000000)
0.472626347996993
>>> timeit.timeit(stmt=cond, number=1000000)
0.5024505830078851
```

lixxu:
sorry, I made the commits ugly. Is there anyway to fix it? :(

sanic/app.py
lixxu: ok, updated for it. but I made the commits history ugly. :( wonder how to squash it again.

lixxu:
tks @harshanarayana, seems worked (with force pushed :( ).

harshanarayana:
@lixxu That's expected. Since you `rebase`d it, now the commit IDs in github and the one you are trying to push are all wibly wobly. So you have to use `-f` more or less always when you use `git rebase` commands.

sjsadowski:
I'm tagging this for 19.03 since it is a change, not a fix.

lixxu:
the issue #1380 was tagged for 18.12 lts.

sjsadowski:
That is because #1380 is pointing out a bug. This fixes a bug, yes, but it also adds functionality.

sanic/app.py
```diff
@@ -637,14 +637,19 @@ def url_for(self, view_name: str, **kwargs):
...
                 else:
                     scheme = "http"
 
-            if "://" in netloc[:8]:
+            if netloc and "://" in netloc[:8]:
                 netloc = netloc.split("://", 1)[-1]
 
+            if not netloc:
+                scheme = ""
+                external = False
```
abuckenheimer: the branching here is getting a bit complicated, I'd maybe suggest something like
```
scheme = re.match(
    r"(?P<scheme>https?)?(?(scheme)://)(?P<netloc>\w+)",
    kwargs.pop("_scheme", "")
).groupdict()
if scheme:
    scheme, netloc = scheme['scheme'], scheme['netloc']
else:
    scheme, netloc = ""
```
but we can always address this seperately

sanic/app.py
abuckenheimer: I actually like this more as it was, makes it impossible to use it wrong. the try/except logic lets you pass a nonsense port and doesn't even complain

yunstanford:
This PR looks like out of date, could you update it ?

sanic/app.py
MuhistheHolvi: TLDR; Fail fast and fail loud.
I saw that @lixxu had already implemented it first without try:except then they were asked to put it in a try:except block. In this case, I believe failing loudly would have better point that failing silently here. A user might wrongly pass some object that can't be converted into integer and this function will fail silently.

lixxu:
Sorry, how to update it?

MuhistheHolvi:
@lixxu You can either merge master to your local branch and then merge or you can rebase your branch onto master branch. You need to resolve the conflicts as shown in github.

sjsadowski:
@lixxu can you update your branch?

lixxu:
> @lixxu can you update your branch?

updated, please review.

sjsadowski:
@lixxu can you back that out, do a squash and merge or rebase instead? This is gonna be a huge amount of clutter in the commits to merge back in.

lixxu:
oops, I lost the code lines (sorry, not familiar with squash, rebase), how to find them back?

PR #1412, opened 2018-11-15
router parameter pattern supports negative integer

tairitian:
such as:
/my/url/<int_param:int>

now int_param can be a negative integer.

### DIFF
--- a/sanic/router.py
+++ b/sanic/router.py
@@ -13,7 +13,7 @@ Parameter = namedtuple('Parameter', ['name', 'cast'])
 
 REGEX_TYPES = {
     'string': (str, r'[^/]+'),
-    'int': (int, r'\d+'),
+    'int': (int, r'-?\d+'),
     'number': (float, r'[0-9\\.]+'),
     'alpha': (str, r'[A-Za-z]+'),
     'path': (str, r'[^/].*?'),


ahopkins:
@tairitian Can you give me an example of why we would want to support this? I mean I suppose it sort of makes sense, and it seems harmless enough.

I am somewhat hesitant to add this now because we are trying to limit the changes that get added to 18.12. And, there may be some larger changes relating to the router coming down the road.

You could always achieve this now with [custom regex](https://sanic.readthedocs.io/en/latest/sanic/routing.html#request-parameters).

andreymal:
@ahopkins https://stackoverflow.com/users/-1

vltr:
> https://stackoverflow.com/users/-1

Isn't this an edge case?

Either way, you can ovewrite the current `int` implementation by monkey patching it:

```python
from sanic import Sanic
from sanic import router

router.REGEX_TYPES["int"] = (int, r'-?\d+')

app = Sanic()
# ...
```

You can see [more examples](https://github.com/huge-success/sanic/issues/1282#issuecomment-421376741), even creating a [more complex type](https://community.sanicframework.org/t/is-this-a-bug-or-not-hex-and-octal-int/41/3?u=vltr) if that's the case ... IMHO I see no point on changing the Router (for now).

sjsadowski:
As there are already ways to implement this, I'm going to close the PR.

PR #1414, opened 2018-11-17
Add "extra" parameter in FutureRoute

sevaho:
The problem that I face is the lack of route discovery information.

I would like to have an endpoint where I can see all my routes with some information like:

```json
                {
                    "scoped": ["admin"],
                    "method": "POST",
                    "description": "TODO",
                    "version": "v1",
                    "path": "/v1/api/users",
                    "body": {},
                    "responses": {201: {"users": "users"}, 401: {"TODO"}},
                },
```

But at the moment Future route has only this paramters:

```
        "handler",
        "uri",
        "methods",
        "host",
        "strict_slashes",
        "stream",
        "version",
        "name",
```

And to obtain information about a route I have an endpoint like this atm:

```python
API = Blueprint("apiv, url_prefix="/api")

@API.get("/all", version="v1", name="routes")
async def get_routes(request):

   resp = {}
    for routesin API.routes:
        resp[route.uri] = route.uri
        ....

   return response.json(resp)
```

But what is missing in my opinion is a way to retrieve information about the payload needed for a route and the response payload. Like in the json above a way of telling that the specific route needs an empty body and returns with a list of users woud be really handy for route discovery.

Maybe there is a better way to do this? But what I propose is to add a parameter in the FutureRoute called "extra" where you could be free to add information about the route, for example extra={"body": {"id": "str"}} which would mean that the route requires a json formatted payload containing a string of key "id".

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -152,6 +152,7 @@ class Sanic:
         stream=False,
         version=None,
         name=None,
+        extra=None,
     ):
         """Decorate a function to be registered as a route
 
@@ -161,6 +162,7 @@ class Sanic:
         :param strict_slashes:
         :param stream:
         :param version:
+        :param extra:
         :param name: user defined route name for url_for
         :return: decorated function
         """
@@ -190,6 +192,7 @@ class Sanic:
                     strict_slashes=strict_slashes,
                     version=version,
                     name=name,
+                    extra=extra,
                 )
                 return handler
             else:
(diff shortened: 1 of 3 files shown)


harshanarayana:
@sevaho did you try out the `sanic-openapi` that can provide you both swagger documentation for api as well as a generated Open api Spec document? That might be more suited for your need than enabling it via the `sanic` core.

sevaho:
I did not, will look into that.

PR #1415, opened 2018-11-17
Fix `remove_entity_headers` helper function

hramezani:
Based on the comment on `remove_entity_headers` helper function:

>  Removes all the entity headers present in the headers given.
    According to RFC 2616 Section 10.3.5,
    Content-Location and Expires are allowed as for the
    "strong cache validator".

`Content-Location` and `Expires` should be allowed.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -34,3 +34,41 @@ def test_is_hop_by_hop_header():
     )
     for header, expected in tests:
         assert helpers.is_hop_by_hop_header(header) is expected
+
+
+def test_remove_entity_headers():
+    tests = (
+        (
+            {},
+            {}
+        ),
+        (
+            {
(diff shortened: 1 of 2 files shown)


tests/test_helpers.py
```diff
@@ -34,3 +34,26 @@ def test_is_hop_by_hop_header():
...
+                "Content-Type": "application/json",
+                "Expires": "Wed, 21 Oct 2015 07:28:00 GMT",
+                "Foo": "Bar"
+            },
+            {
+                "Expires": "Wed, 21 Oct 2015 07:28:00 GMT",
+                "Foo": "Bar"
+            },
+        ),
+    )
```
abuckenheimer: can you add a test that includes a "content-location" header?

hramezani: @abuckenheimer test added.

PR #1416, opened 2018-11-20
Add tests for static and update document

chenjr0719:
Notice there is a [call to arms](https://community.sanicframework.org/t/call-to-arms-testing-coverage/85) on community discussion, so here I am.

This PR mainly add tests for `sanic/static.py` in `Sanic`. The current coverage of `sanic/static.py` is **67.19%**(from [here](https://codecov.io/gh/huge-success/sanic/src/master/sanic/static.py)), after adding some tests, it should up to a better number now.

And, this PR also add two example about use `app.static()` with `host` and `stream_large_files` arguments.

### DIFF
--- a/sanic/static.py
+++ b/sanic/static.py
@@ -114,7 +114,7 @@ def register(
                 return HTTPResponse(headers=headers)
             else:
                 if stream_large_files:
-                    if isinstance(stream_large_files, int):
+                    if type(stream_large_files) == int:
                         threshold = stream_large_files
                     else:
                         threshold = 1024 * 1024
--- a/docs/sanic/static_files.md
+++ b/docs/sanic/static_files.md
@@ -1,7 +1,7 @@
 # Static Files
 
 Static files and directories, such as an image file, are served by Sanic when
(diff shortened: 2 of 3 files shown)


ahopkins:
@chenjr0719 you rock!

sanic/static.py
```diff
@@ -114,7 +114,7 @@ def register(
                 return HTTPResponse(headers=headers)
             else:
                 if stream_large_files:
-                    if isinstance(stream_large_files, int):
+                    if type(stream_large_files) == int:
```
ahopkins: @chenjr0719 I am looking at this again ... what is the reason to change from `isinstance` to `type`?

chenjr0719: @ahopkins When use `instance()` between `True` and `int`, it will always `True`(here is [reference](https://stackoverflow.com/a/37888668)). So, I use `type()` to fix this unreachable case.

ahopkins: Ah yes. Booleans are subclasses of integers in python.

PR #1422, opened 2018-12-03
Add in some server.py __slots__ attribute names that are missing.

ashleysommer:
While doing some digging in the server.py file this morning, I found that there are a bunch of attributes used in the HTTPProtocol class which are not declared in the `__slots__` tuple at the top of the class.

Normally not having the attribute name in `__slots__` means the class cannot write to that attribute at runtime (because it doesn't exist) but in this case it never mattered because HTTPProtocol is a subclass of asyncio.Protocol, which itself is a Normal class (no `__slots__`). So having `__slots__` on the HTTPProtocol subclass to restrict its possible attributes is kind-of useless in this case because it doesn't restrict anything.

But anyway, for correctness sake this PR adds the missing attribute names to the `__slots__` of Sanic server.py HTTPProtocol.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -62,6 +62,7 @@ class HttpProtocol(asyncio.Protocol):
         "request_class",
         "is_request_stream",
         "router",
+        "error_handler",
         # enable or disable access log purpose
         "access_log",
         # connection management
@@ -73,6 +74,12 @@ class HttpProtocol(asyncio.Protocol):
         "_last_response_time",
         "_is_stream_handler",
         "_not_paused",
+        "_request_handler_task",
+        "_request_stream_task",
+        "_keep_alive",
+        "_header_fragment",
+        "state",
+        "_debug",
     )
 
     def __init__(


vltr:
@ashleysommer sorry for being the "devil's advocate" in here, but ... Since you already "compiled" why it's useless to have `__slots__` in the `HttpProcol` class, my question is: do we need `__slots__` there in the first place?

I mean, it's useless. It won't "restrict" the attributes set in a instance ... And [with a simple benchmark](https://gist.github.com/vltr/f081e44502af02ae97bf24caeb2e6611), generally a subclass with `__slots__` tends to be marginally slower to instantiate than subclasses without it - although this can be futile, so to speak ...

ashleysommer:
@vltr
You're right. We probably should remove the __slots__ altogether here. There's really no reason to have it.

The reason I added the missing entries in this PR is because it was a smaller change to make than removing the whole slots declaration.

seemethere:
`__slots__` are supposed to allow for faster attribute access, since `asyncio.Protocol` doesn't actually implement anything I think we'd be pretty safe keeping slots in just for faster attribute access.

See https://github.com/python/cpython/blob/master/Lib/asyncio/protocols.py#L64-L102

ashleysommer:
@vltr 
> @ashleysommer sorry, I don't want to sound stingy or something, it's just that the `__slots__` are dead code in this case

I know. I'm not disagreeing with you.

@seemethere  @vltr 
This page is a good overview of what we're actually looking at in this case:
https://stackoverflow.com/questions/1816483/python-how-does-inheritance-of-slots-in-subclasses-actually-work

Even though `asyncio.Protocol` has no attributes of its own, it doesn't define `__slots__` in its definition therefore it gets an internal `__dict__` allocated. When we subclass this in `HTTPProtocol`, we add our own `__slots__` but it still inherits the `__dict__` from the parent, and as seen by the fact this PR exists, that `__slots__` wasn't doing anything because all `HTTPProtocol` attributes were allocated on the `__dict__` anyway.

vltr:
Well, it does works - but it is kind of tricky:

```python
>>> class Foo: pass
... 
>>> class Bar(Foo):
...     __slots__ = ("test",)
... 
>>> x = Bar()
>>> x
<__main__.Bar object at 0x7f7436445848>
>>> dir(x)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__slots__', '__str__', '__subclasshook__', '__weakref__', 'test']
>>> x.test = "hello"
>>> x.not_in_slots = "world"
>>> print(x.test)
hello
>>> print(x.__dict__)
{'not_in_slots': 'world'}
>>> import timeit
>>> def get_set_delete_slots():
...     x.test = "foo"
...     x.test
...     del x.test
...
>>> def get_set_delete_no_slots():
...     x.foo = "foo"
...     x.foo
...     del x.foo
...
>>> min(timeit.repeat(get_set_delete_slots))
0.18530158700014
>>> min(timeit.repeat(get_set_delete_no_slots))
0.22946998499537585
>>> z = Bar()
>>> dir(z)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__slots__', '__str__', '__subclasshook__', '__weakref__', 'test']
```

This means that we can have attributes in the instance that are not listed in the `__slots__`  - that will end up in the `__dict__` - we'll just end up having to take extra care to have all attributes into `__slots__`.

ashleysommer:
I did some experiments.

1) Yes, the `HTTPProtocol` does get allocated an internal `__dict__` structure because its parent (`asyncio.Protocol`) doesn't define `__slots__`
```assert hasattr(self, '__dict__')```

2) I was wrong, when a class has both `__dict__` _and_ `__slots__` the attributes are assigned to the `__slots__` if the attribute name is in there, else it adds the attribute to the `__dict__`.

Adding a debugging print `self.__dict__` in server.py at runtime:
With some attributes missing from the `__slots__`:
```
<class 'dict'>: {'_header_fragment': b'', 'state': {'requests_count': 1}, '_debug': False}
```
When ensuring all attributes are added to the `__slots__`:
```
<class 'dict'>: {}
```

**Comment by vltr** [comment]
@ashleysommer same experiments :smile:

**Comment by ashleysommer** [comment]
@vltr another interesting observation.
Adding:
```
__dict__ = None
```
To the top of the `HTTPProtocol` class shows `self.__dict__` is `None` at runtime.
```
self.__dict__
<NoneType>
```

But the code can still write non-slot attributes to the class.
```
self.cat = "Hello"
```
Turns out in this case it adds the attribute to the _Parent Class's_ `__dict__`:
```
self.__dict__
<NoneType>

super(HttpProtocol, self).__dict__
<class 'dict'>: {'cat': "Hello"}
```
Not sure the implications of that. But anyway, that observation doesn't affect Sanic.

vltr:
> Or, are you saying that there is some benefit if the attribute is in `__slots__` and then it just uses `__dict__` as a fall back?

Exactly ... Go figure.

ashleysommer:
@ahopkins 
> If adding __slots__ doesn't add performance increase on a subclass, then why bother?

From these experiments, it looks like it still adds a tiny bit of performance, but you don't get the memory saving (from not allocating the `__dict__`) that you would normally get when using `__slots__`.

harshanarayana:
> Turns out in this case it adds the attribute to the _Parent Class's_ `__dict__`:

This happened due to how python handles it's object model. Self, if not possible, then attempt to perform the operation on the parent. So any `object.attribue = value` is set using the `object.__class__.__mro__` based lookup in case of inherited items. 

So, if the current class doesn't provide a way to set `__dict__`, it will attempt to do the same under the base class.

vltr:
@harshanarayana thanks for pointing out the MRO, it makes total sense this behavior now :+1:

abuckenheimer:
good sleuthing folks, so it seems to me like the right answer for now is to approve this PR? seems to bring modest attribute access speedup, reduce memory usage and while it may be a slightly more complicated implementation than a no `__slots__` world it is still a less complicated change.

vltr:
@abuckenheimer yeah, for me it is good to go.

vltr:
If only we could access the underlying protocol through the `request` instance ... This test would be as simple as `assert len(request.protocol.__dict__) == 0`.

PR #1423, opened 2018-12-05
basic request streaming support with flow control

yunstanford:
basic request streaming support, could be useful in cases like large file upload.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -294,7 +294,9 @@ class HttpProtocol(asyncio.Protocol):
                 self.request
             )
             if self._is_stream_handler:
-                self.request.stream = StreamBuffer(self.request_body_buffer_queue_size)
+                self.request.stream = StreamBuffer(
+                    self.request_body_buffer_queue_size
+                )
                 self.execute_request_handler()
 
     def on_body(self, body):
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -304,8 +304,8 @@ class HttpProtocol(asyncio.Protocol):
             self._request_stream_task = self.loop.create_task(
                 self.body_append(body)
             )
-            return
-        self.request.body_push(body)
+        else:
+            self.request.body_push(body)
 
     async def body_append(self, body):
         if self.request.stream.is_full():
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -22,7 +22,7 @@ from sanic.exceptions import (
     ServiceUnavailable,
 )
 from sanic.log import access_logger, logger
-from sanic.request import Request
+from sanic.request import Request, StreamBuffer
 from sanic.response import HTTPResponse
 
 
@@ -59,6 +59,7 @@ class HttpProtocol(asyncio.Protocol):
         "response_timeout",
         "keep_alive_timeout",
         "request_max_size",
+        "request_body_buffer_queue_size",
         "request_class",
         "is_request_stream",
         "router",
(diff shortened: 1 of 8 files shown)


sanic/server.py
```diff
@@ -291,17 +294,27 @@ def on_headers_complete(self):
...
     def on_body(self, body):
         if self.is_request_stream and self._is_stream_handler:
             self._request_stream_task = self.loop.create_task(
-                self.request.stream.put(body)
+                self.body_append(body)
             )
             return
         self.request.body_push(body)
 
+    async def body_append(self, body):
```
ahopkins: Do you think we need to figure out a way to test this (the pausing and resuming of the `transport`?

sanic/server.py
```diff
@@ -291,17 +294,27 @@ def on_headers_complete(self):
...
+                )
                 self.execute_request_handler()
 
     def on_body(self, body):
         if self.is_request_stream and self._is_stream_handler:
             self._request_stream_task = self.loop.create_task(
-                self.request.stream.put(body)
+                self.body_append(body)
             )
             return
```
ahopkins: Why do we have this `return` here since the caller is not expecting a return? Seems a simpler `if/else` construct would be more self-explanatory.

yunstanford: yeah, agree

sanic/server.py
yunstanford: yeah, but a real integration test (or maybe an example) might be more helpful in this situation. and we probably should assume `pause_reading` and `resume_reading ` working as it is, https://docs.python.org/3/library/asyncio-protocol.html#asyncio.ReadTransport.pause_reading.

yunstanford:
Anyone take a look ?

abuckenheimer:
@yunstanford can review tomorrow

yunstanford:
@abuckenheimer agree in general, but httptools doesn't provide the transport that we can control how many bytes to read from the socket. We can build another layer in `StreamBuffer` for `read_n(bytes)`, but the data has been feed in memory already at that time.

This is a simple fix for https://github.com/huge-success/sanic/issues/546, providing basic data flow control.

We can look into `httptools` later and see if we can do more around that, and vender it if needed.

abuckenheimer:
Gotcha, I still think REQUEST_BUFFER_QUEUE_SIZE is kind of meaningless though, we need to restrict based on amounts of bytes in the queue somehow. Making an actual async buffer is kinda complicated, I'm fine with moving forward on this as long as there is some sort of "alpha/beta" marker on streaming incoming requests because I have to imagine the API will change.

yunstanford:
yeah, the API doesn't need change as we can set default value for `.read(bytes=100)`, but we need to document that in the future.

abuckenheimer:
Each item could potentially be any size in bytes, how do you plan on rechunking the bytes across items in the queue?

Also ideally the default would be return all bytes like a real buffer. I realize this is kind of an anti pattern because your really no longer streaming the request but, I'd expect `body = stream.read()` to return the whole body. I think `n=100` would be easier to misuse than the alternative.

yunstanford:
We'll need introduce another layer in StreamBuffer for rechunking, but didn't think deep into this. and introduce another API like `.read_any(bytes=some)` or some similar if you think it's less confusing. 

Also in reality, the common use case, like uploading large file, usually is in a loop, users might be not very interested in the exact bytes in each single read. 

We can definitely look into improving this in the future as mentioned before (not small amount of work). and we can provide the functionality/support first.

pypycoder:
@yunstanford @ahopkins Can we get this feature in next release ?

ahopkins:
@pypycoder Unfortunately not. We put a freeze on features a few weeks back for the 18.12 release (which is ready to ship). It should be added to 19.03

andreymal:
@yunstanford @ahopkins this PR broke the graceful shutdown of streaming responses in 19.03.1

(Perhaps I'll create an issue or PR later)

PR #1427, opened 2018-12-12
fixed setup.py url, added vscode to .gitignore

LBWright:
Updated repo url inside of the `setup.py` file to point to the proper repo.

It looks like VSCode creates a local folder whenever the user manually selects an interpreter (and virtual environment so I added the folder to the `.gitignore`. Can remove if unnecessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ docs/_build/
 docs/_api/
 build/*
 .DS_Store
+.vscode/
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ with open_local(['README.rst']) as rm:
 setup_kwargs = {
     'name': 'sanic',
     'version': version,
-    'url': 'http://github.com/channelcat/sanic/',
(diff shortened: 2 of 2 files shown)


LBWright:
All tests passing. 

screen shot 2018-12-11 at 8 43 59 pm

vltr:
Well, I use VSCode myself and have some configurations here and there for specific projects (Sanic being one of them), so I really don't mind having it on `.gitignore` (furthermore, [I have a strong opinion](https://github.com/github/gitignore/pull/2838#issuecomment-430224927) regarding `.gitignore` to basically put **everything** that's not code in there. It's a simple control file, anyway!).

As for the `setup.py` ... We would eventually get it fixed (prior to the next release), but every contribution is welcome and counts, so thank you for paying an extra attention to these details. Sometimes they might pass unnoticed! :metal:

LBWright:
Thanks! I'm trying to pick up what I can. This is a great framework that I use often and would love to get involved. :)

PR #1436, opened 2018-12-26
Config from object string

jotagesales:
Implemented support in config_from_object method to receive a path by sting equal similar to Flask.


vltr:
Also, take a look at #1424 (as reference) to provide documentation to the `import_string` function and for the official docs :wink:

docs/sanic/config.md
```diff
@@ -53,6 +53,13 @@ import myapp.default_settings
 app = Sanic('myapp')
 app.config.from_object(myapp.default_settings)
 ```
+or also by path to config:
+
+```
```
harshanarayana: Would it be possible to add a sample Configuration class as well in this section so that it can be handy when one does Ctrl+C > Ctrl + V

setup.py
```diff
@@ -63,6 +63,7 @@ def open_local(paths, mode='r', encoding='utf8'):
     'aiofiles>=0.3.0',
     'websockets>=6.0,<7.0',
     'multidict>=4.0,<5.0',
+    'import-string>=0.1.0'
```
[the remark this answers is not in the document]
jotagesales: @ahopkins was removed dependence and implemented a helper function using the module import_lib.

sanic/helpers.py
```diff
@@ -131,3 +135,21 @@ def remove_entity_headers(headers, allowed=("content-location", "expires")):
...
+    """
+    import a module or class by string path.
+
+    :module_name: str with path of module or path to import and
+    instanciate a class
+    :returns: a module object or one instance from class if
+    module_name is a valid path to class
+
+    """
+    module, klass = module_name.rsplit(".", 1)
```
harshanarayana: I might be overthinking this, but would it be a good idea to support relative string paths?

docs/sanic/config.md
```diff
@@ -53,6 +53,13 @@ import myapp.default_settings
...
+app = Sanic('myapp')
+app.config.from_object('config.path.config.Class')
```
jotagesales: @harshanarayana  here the doc in config.md

sanic/helpers.py
jotagesales: well, i use this function in Flask for configure app using config_from_object and is very good.
Because is common in application has config classes.
example:
```
class Config:
      DEBUG = False

class Develop(Config):
    DEBUG = True

class Production(Config):
      pass
```

harshanarayana: > In this case the user instanciate Sanic class and configure according from environment pass the string path of the class get from environment variable for example.

Oh I am totally with you on this feature. I was just curious if it might be possible to support relative path as well. 

i.e. Instead of `module1.module2.Class`, if you are already inside `module1`, then `module2.Class1` should suffice.

harshanarayana: @vltr It works, but the method invocation might need to be tweaked a bit to provide another helper argument. `import_module(name, package=None)`

sjsadowski: @jotagesales looks like this is still outstanding - can you make changes so we can get this in soon?

tests/test_config.py
```diff
@@ -17,17 +17,30 @@ def temp_path():
         yield Path(td, "file")
 
 
-def test_load_from_object(app):
-    class Config:
-        not_for_config = "should not be used"
-        CONFIG_VALUE = "should be used"
+class Config:
```
harshanarayana: Can you please rename this class from `Config` to something else? This line is overriding the import done by `from sanic.config import Config`

sanic/helpers.py
jotagesales: @sjsadowski ready ;)

tests/test_config.py
jotagesales: @harshanarayana i renamed the class, thank.

ahopkins: Renaming this class seems to be the only thing holding it up.

PR #1439, opened 2018-12-27
WIP: PR Benchmarking - Prototype

harshanarayana:
Hi @huge-success/sanic-core-devs , this is a very early edition of the PR benchmarking feature that can be plugged into `sanic`'s Pull request lifecycle to ensure that new changes do not impact the performance in any way. 

The idea is to add a cut-off boundary that is configurable for each parameter and we keep a standard sample server code which runs and each PR gets benchmarked against the standard values. This can possibly be extended with `ab`/`abi` / `bombadier` to perform a benchmarking comparison across frameworks as well. 

This PR is only to get an early feedback.

### DIFF
--- a/benchmarking/__init__.py
+++ b/benchmarking/__init__.py
@@ -1,0 +1,149 @@
+from abc import ABCMeta, abstractmethod
+from subprocess import Popen, PIPE
+from os import path
+from sys import executable
+from shutil import which
+from json import load
+from os import kill
+from signal import SIGTERM
+
+try:
+    from loguru import logger
+except:
+    import logging
+
+    logger = logging.getLogger("benchmark")
+    logger.setLevel(logging.DEBUG)
+    logger.addHandler(logging.Handler(level=logging.DEBUG))
+
+
+class AbstractBenchmarkRunner(metaclass=ABCMeta):
+    def __init__(self, server_name):
+        self._server_name = server_name
+        self._command = []
+        self._config = {}
+
+    @property
+    def config(self):
+        return self._config
+
+    @config.setter
+    def config(self, config):
+        self._config = config
+
+    @property
+    def command(self):
+        return self._command
+
+    @command.setter
+    def command(self, command):
+        self._command = command
+
+    @property
+    def server_name(self):
+        return self._server_name
+
+    @server_name.setter
+    def server_name(self, server_name):
+        self._server_name = server_name
+
+    def __enter__(self):
+        return self._enter()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._exit(exc_type, exc_val, exc_tb)
+
+    @abstractmethod
+    def _load_configuration(self):
+        raise NotImplemented("Implementation Required")
+
+    @abstractmethod
+    def _enter(self):
+        raise NotImplemented("Implementation Required")
+
+    @abstractmethod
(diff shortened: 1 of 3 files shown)


harshanarayana:
@chenjr0719  I absolutely agree about taking the benchmarking server code from available resources. There is no reason why we should we writing them again on our own. We can even use the available items as a base reference to set our cut-off limit for performance. 

> I also notice that you are using `boom` as the benchmark runner. But at the [README](https://github.com/tarekziade/boom) of `boom`, it says `boom` is deprecated now. No sure would it still suitable for use as the benchmark runner.

As for `boom` it was just a prototype I picked and not the final one by any standard. (I am more of a `Locust` guy) `boom` had a simple to use API that could be called from python with no additional changes required. That is the reason I used it here.

ahopkins:
One thing I want to make sure we do not lose sight of is that we also want to benchmark individual components of Sanic, and not just the whole.

For example, we should have a set of tests to fill the router with different sizes and complexities of routes (10, 50, 100+, etc) and then run speed tests on how fast it resolves to the proper view.

vltr:
I agree with @ahopkins - take [this routing benchmark](https://github.com/vltr/very-stupid-simple-routers-bench) I made as example. The Sanic router is clearly slower when not hitting the `Router.lrucache(get)` and/or with more "complicated" routes ... So, we might need to cover some more areas when "benchmarking" ... I mean, it's easy to check if a route returns always a "static" `hello world`, put that into an AST and create a "wow" effect, right? :wink:

vltr:
@harshanarayana I like the idea. I'm just feeling this could be an entire new project, with its own repository - instead of being inside Sanic. That way, more people can contribute to this, making tweaks and etc without touching the Sanic repo itself. Thoughts, anyone?

harshanarayana:
@vltr 
> I'm just feeling this could be an entirely new project, with its own repository - instead of being inside Sanic.

Agreed. I will write up the basic things and get them to work the way it is supposed to. Once we are okay with the full prototype, then we can move it a common location. For now, [I have a new repo in my personal space](https://github.com/harshanarayana/hfb) where I will be making changes. This PR was created just to get quick feedback on the idea and nothing more. 

Also,  the benchmarking related things are not necessary to be part of this repository anyway. It would be a lot of unwanted code that doesn't have anything directly to do with the framework itself from an end-user standpoint.

vltr:
@harshanarayana nice! I'll watch the repo and I agree with you that benchmarking code is not necessary for us (in here), but certainly will be for other reasons, such as testing the impact on performance of a new feature and etc, in a overall manner.

harshanarayana:
@vltr  
> I'll watch the repo
Suggestions and feedbacks are welcome too. 😃

ahopkins:
Wow, this is a really interesting idea. Much deeper than I was originally thinking. It certainly is ambitious, but has the potential to fill a void in the python world. I like the idea of branching it into it's own repo under the same umbrella. And, given the emphasis on performance, it certainly fits. I will try and clone the repo and give some substantive thoughts. If you want me to open something under the community organization let me know.

ahopkins:
I opened a new conversation here: 
https://community.sanicframework.org/t/benchmarking-utility/163

harshanarayana:
Closing this PR in favor of a much simpler and cleaner approach from #1499

Issue #1432, opened 2018-12-20
Disabling the LOGO


rbw:
**Describe the bug**
```
[D 2018-12-20 02:20:18.032 8537] False
[I 2018-12-20 02:20:18.032 8537] Goin' Fast @ http://0.0.0.0:8080 
```

Kept getting this in my Sanic log.

Started debugging and realized setting LOGO to a false value other than None is treated as a logo.

**Code snippet**
`app.config.LOGO=False`

**Expected behavior**
No logging of course.

**Additional context**
Let's get real - no one in their right mind would set LOGO to False, empty string, 0 or such and expect Sanic to treat it as an actual logo.

I think it's safe to assume that changing this:
https://github.com/huge-success/sanic/blob/e3a27c2cc485d57aa1ff87d9f69098e4ab12727e/sanic/app.py#L1107

to:
`not self.config.LOGO`

...wouldn't cause any breakage or upset anyone

rbw:
Oh, yeah- and thank you for a great piece of software. Love it :+1:

abuckenheimer:
Fine by me if you want to put in a PR for this

ahopkins:
Agreed. Should be a simple PR to get approved.

ahopkins:
See #1443 with the new logo and some changes to the LOGO.

PR #1441, opened 2018-12-29
set config.LOGO to False, disable logo output

emctoo:
related issue [1432](https://github.com/huge-success/sanic/issues/1432)
fixing as described in the issue.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1256,7 +1256,7 @@ class Sanic:
             logger.setLevel(logging.DEBUG)
 
         if (
-            self.config.LOGO is not None
+            not self.config.LOGO
             and os.environ.get("SANIC_SERVER_RUNNING") != "true"
         ):
             logger.debug(self.config.LOGO)


emctoo:
Sorry, something is not right, close now, maybe PR later.

sanic/app.py
```diff
@@ -1256,7 +1256,7 @@ def _helper(
             logger.setLevel(logging.DEBUG)
 
         if (
-            self.config.LOGO is not None
```
MyNameIsCosmo: You could possibly do
```python
self.config.LOGO not in [None, False, ""]
```

This should cover the cases in #1432

MyNameIsCosmo: The caveat is setting `self.config.LOGO` to `True` will literally print `True` (or similar).
Perhaps this could be handled by changing line 1262 to
```python
logger.debug(self.config.LOGO if self.config.LOGO is True else Config.LOGO)
```

ahopkins:
We just [adopted a new logo](https://community.sanicframework.org/t/new-project-logo/22/60?u=ahopkins) for the project. I am pushing a PR to add it and will address this issue at the same time.

ahopkins:
See #1443 with the new logo and some changes to the LOGO. I took the strategy from both @emctoo and @MyNameIsCosmo . Thank you both.

PR #1442, opened 2018-12-29
add an option to change access_log using gunicorn

Amanit:
There is still no easy way to disable an access logs running the server via Gunicorn.
And that logs are affecting the performance dramatically

```
[fedoruk@localhost wrk]$ ./wrk -d10s -t8 -c400 http://127.0.0.1:8100
Running 10s test @ http://127.0.0.1:8100
  8 threads and 400 connections
    Latency    10.12ms    9.08ms 137.15ms   91.71%
    Req/Sec     5.69k     1.01k   12.03k    70.75%
  453230 requests in 10.03s, 54.89MB read
Requests/sec:  45188.92
Transfer/sec:      5.47MB

[fedoruk@localhost wrk]$ ./wrk -d10s -t8 -c400 http://127.0.0.1:8100
Running 10s test @ http://127.0.0.1:8100
  8 threads and 400 connections
    Latency    23.55ms   13.83ms 120.72ms   72.80%
    Req/Sec     2.20k   322.17     4.55k    65.12%
  175537 requests in 10.03s, 21.26MB read
Requests/sec:  17503.14
Transfer/sec:      2.12MB
```

First run is with access_log=False and second is with access_log=True

I suggest to set the access_log config variable on base of log-level passed to gunicorn -
access logs are level INFO, if log-level is higher than INFO - disable access_log.

I also added some more information about access_log in readme.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -5,12 +5,12 @@ import re
 import typing
 import warnings
 
-from asyncio import CancelledError, ensure_future, get_event_loop, Protocol
+from asyncio import CancelledError, Protocol, ensure_future, get_event_loop
 from collections import defaultdict, deque
 from functools import partial
 from inspect import getmodulename, isawaitable, signature, stack
 from socket import socket
-from ssl import Purpose, create_default_context, SSLContext
+from ssl import Purpose, SSLContext, create_default_context
 from traceback import format_exc
 from urllib.parse import urlencode, urlunparse
 
@@ -969,17 +969,17 @@ class Sanic:
 
     def run(
         self,
-        host: typing.Optional[str]=None,
-        port: typing.Optional[int]=None,
-        debug: bool=False,
-        ssl: typing.Union[dict, SSLContext, None]=None,
-        sock: typing.Optional[socket]=None,
-        workers: int=1,
-        protocol: typing.Type[Protocol]=None,
-        backlog: int=100,
-        stop_event: typing.Any=None,
-        register_sys_signals: bool=True,
-        access_log: bool=None,
+        host: typing.Optional[str] = None,
+        port: typing.Optional[int] = None,
+        debug: bool = False,
+        ssl: typing.Union[dict, SSLContext, None] = None,
+        sock: typing.Optional[socket] = None,
+        workers: int = 1,
+        protocol: typing.Type[Protocol] = None,
+        backlog: int = 100,
+        stop_event: typing.Any = None,
+        register_sys_signals: bool = True,
+        access_log: typing.Optional[bool] = None,
         **kwargs: typing.Any
     ) -> None:
         """Run the HTTP Server and listen until keyboard interrupt or term
@@ -1003,7 +1003,8 @@ class Sanic:
         :param backlog: a number of unaccepted connections that the system
             will allow before refusing new connections
         :type backlog: int
-        :param stop_event: event to be triggered before stopping the app - deprecated
+        :param stop_event: event to be triggered
+            before stopping the app - deprecated
         :type stop_event: None
         :param register_sys_signals: Register SIG* events
         :type register_sys_signals: bool
@@ -1043,12 +1044,7 @@ class Sanic:
             )
         # if access_log is passed explicitly change config.ACCESS_LOG
         if access_log is not None:
-            if isinstance(access_log, bool):
-                self.config.ACCESS_LOG = access_log
-            else:
-                raise ServerError(
-                    ("'access_log' passed in 'run' should be boolean")
-                )
+            self.config.ACCESS_LOG = access_log
 
         server_settings = self._helper(
             host=host,
@@ -1107,7 +1103,7 @@ class Sanic:
         protocol: typing.Type[Protocol] = None,
         backlog: int = 100,
         stop_event: typing.Any = None,
-        access_log: bool = None,
+        access_log: typing.Optional[bool] = None,
     ) -> None:
         """
         Asynchronous version of :func:`run`.
@@ -1136,7 +1132,8 @@ class Sanic:
         :param backlog: a number of unaccepted connections that the system
             will allow before refusing new connections
         :type backlog: int
-        :param stop_event: event to be triggered before stopping the app - deprecated
+        :param stop_event: event to be triggered
+            before stopping the app - deprecated
         :type stop_event: None
         :param access_log: Enables writing access logs (slows server)
         :type access_log: bool
@@ -1159,15 +1156,7 @@ class Sanic:
             )
         # if access_log is passed explicitly change config.ACCESS_LOG
         if access_log is not None:
-            if isinstance(access_log, bool):
-                self.config.ACCESS_LOG = access_log
-            else:
-                raise ServerError(
-                    (
-                        "'access_log' passed in 'create_server' "
-                        "should be boolean"
-                    )
-                )
+            self.config.ACCESS_LOG = access_log
 
         server_settings = self._helper(
             host=host,
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -2,7 +2,6 @@ import logging
 import logging.config
 import os
 import re
-import typing
 import warnings
 
 from asyncio import CancelledError, Protocol, ensure_future, get_event_loop
@@ -12,6 +11,7 @@ from inspect import getmodulename, isawaitable, signature, stack
 from socket import socket
 from ssl import Purpose, SSLContext, create_default_context
 from traceback import format_exc
+from typing import Any, Optional, Type, Union
 from urllib.parse import urlencode, urlunparse
 
 from sanic import reloader_helpers
@@ -969,18 +969,18 @@ class Sanic:
 
     def run(
         self,
-        host: typing.Optional[str] = None,
-        port: typing.Optional[int] = None,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
         debug: bool = False,
-        ssl: typing.Union[dict, SSLContext, None] = None,
-        sock: typing.Optional[socket] = None,
+        ssl: Union[dict, SSLContext, None] = None,
+        sock: Optional[socket] = None,
         workers: int = 1,
-        protocol: typing.Type[Protocol] = None,
+        protocol: Type[Protocol] = None,
         backlog: int = 100,
-        stop_event: typing.Any = None,
+        stop_event: Any = None,
         register_sys_signals: bool = True,
-        access_log: typing.Optional[bool] = None,
-        **kwargs: typing.Any
+        access_log: Optional[bool] = None,
+        **kwargs: Any
     ) -> None:
         """Run the HTTP Server and listen until keyboard interrupt or term
         signal. On termination, drain connections before closing.
@@ -1095,15 +1095,15 @@ class Sanic:
 
     async def create_server(
         self,
-        host: typing.Optional[str] = None,
-        port: typing.Optional[int] = None,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
         debug: bool = False,
-        ssl: typing.Union[dict, SSLContext, None] = None,
-        sock: typing.Optional[socket] = None,
-        protocol: typing.Type[Protocol] = None,
+        ssl: Union[dict, SSLContext, None] = None,
+        sock: Optional[socket] = None,
+        protocol: Type[Protocol] = None,
         backlog: int = 100,
-        stop_event: typing.Any = None,
-        access_log: typing.Optional[bool] = None,
+        stop_event: Any = None,
+        access_log: Optional[bool] = None,
     ) -> None:
         """
         Asynchronous version of :func:`run`.
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -2,13 +2,15 @@ import logging
 import logging.config
 import os
 import re
+import typing
 import warnings
 
-from asyncio import CancelledError, ensure_future, get_event_loop
+from asyncio import CancelledError, ensure_future, get_event_loop, Protocol
 from collections import defaultdict, deque
 from functools import partial
 from inspect import getmodulename, isawaitable, signature, stack
-from ssl import Purpose, create_default_context
+from socket import socket
+from ssl import Purpose, create_default_context, SSLContext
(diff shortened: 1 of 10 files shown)


sanic/worker.py
```diff
@@ -57,6 +57,10 @@ def run(self):
             if self.app.callable.websocket_enabled
             else self.http_protocol
         )
+
+        # set ACCESS_LOG on base of logging level
+        self.app.callable.config.ACCESS_LOG = self.log.loglevel <= logging.INFO
```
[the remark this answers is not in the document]
Amanit: I have tried - 
`
app = Sanic()
app.config.ACCESS_LOG = False
...
and

my_settings = {
    'ACCESS_LOG': False
}

app.config.update(my_settings)
`
Nothing has changed..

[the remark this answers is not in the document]
Amanit: `my_settings = {
    'ACCESS_LOG': False
}

app.config.from_object(my_settings)`

also don't work

Amanit:
There is an easy way to change that option using app.run() and no easy way to do it using gunicorn.
Anyway the documentation about that option and how to change it should be improved - the difference in performance is almost 3x.

sanic/worker.py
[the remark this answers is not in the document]
yunstanford: i think the issue is here: https://github.com/huge-success/sanic/blob/master/sanic/app.py#L1089,L1118

It has been always override to `true` when using gunicorn.

yunstanford:
Totally agree, it's just how should we tackle and fix it.

sanic/worker.py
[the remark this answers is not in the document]
chenjr0719: It seems that `access_log` in `app.run()` and `app.create_server()` will always override `app.config.ACCESS_LOG`.
https://github.com/huge-success/sanic/blob/72f2e18a84f748f028cb4795c31e7db8b05663d2/sanic/app.py#L1031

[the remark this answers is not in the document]
ahopkins: I am not opposed to something like this. **HOWEVER**, if `ACCESS_LOG` is going to be overridden based upon whether the `loglevel` is or is not above `logging`, then I would suggest throwing a `WARNING` to stdout to clear up potential confusion.

Amanit:
Ok, I removed that logic from worker. And I have fixed changing ACCESS_LOG in the config.
Also I have fixed a bug in Boolean Environment Variables used in config.
And I had to remove some global Config changes in the tests, replacing them with app.config - that Config affected every test.

sanic/config.py
```diff
@@ -6,10 +6,28 @@
...
+    "WEBSOCKET_READ_LIMIT": 2 ** 16,
+    "WEBSOCKET_WRITE_LIMIT": 2 ** 16,
+    "GRACEFUL_SHUTDOWN_TIMEOUT": 15.0,  # 15 sec
+    "ACCESS_LOG": True,
+}
+
 
 class Config(dict):
-    def __init__(self, defaults=None, load_env=True, keep_alive=True):
+    def __init__(self, defaults=None, load_env=True, keep_alive=None):
```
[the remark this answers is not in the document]
Amanit: Default value for keep_alive was moved in DEFAULT_CONFIG. If keep_alive is passed explicitly - we will use it. Otherwise even if you pass KEEP_ALIVE in defaults and don't pass it as kwarg it will be rewrited by default value. e.g.
`
config = Config(defaults={"KEEP_ALIVE": False})
config.KEEP_ALIVE == True

config = Config(defaults={"KEEP_ALIVE": False}, keep_alive=False)
config.KEEP_ALIVE == False
`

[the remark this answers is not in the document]
ahopkins: I see it now. That's what I get for doing a code review on my phone.

sanic/config.py
```diff
@@ -130,4 +136,7 @@ def load_environment_vars(self, prefix=SANIC_PREFIX):
                     try:
                         self[config_key] = float(v)
                     except ValueError:
-                        self[config_key] = v
+                        if v in ["True", "False"]:
```
ahopkins: Should we also add other truthy strings? 

`"true", "yes", "Yes", "Y", etc`

sanic/config.py
```diff
@@ -6,10 +6,28 @@
...
+    "GRACEFUL_SHUTDOWN_TIMEOUT": 15.0,  # 15 sec
+    "ACCESS_LOG": True,
+}
+
 
 class Config(dict):
-    def __init__(self, defaults=None, load_env=True, keep_alive=True):
+    def __init__(self, defaults=None, load_env=True, keep_alive=None):
+        super().__init__(DEFAULT_CONFIG)
         super().__init__(defaults or {})
```
ahopkins: Why run `super().__init__` twice? And not:

```initial = dict()
initial.update(DEFAULT_CONFIG)
initial.update(defaults or {})
super().__init__(initial)
```

Unless one is more performant than the other, I think it is a little more explicit what is going on in this version. I feel like it is not so common to see `super().__init__` called more than once in `__init__`.

sanic/app.py
```diff
@@ -1027,8 +1027,10 @@ def run(
                 "stop_event will be removed from future versions.",
                 DeprecationWarning,
             )
-        # compatibility old access_log params
-        self.config.ACCESS_LOG = access_log
+        # if access_log is passed explicitly change config.ACCESS_LOG
+        if access_log is not None:
```
ahopkins: Should we make sure that `access_log` is a boolean?

sanic/config.py
Amanit: how about:
```python
def __init__(self, defaults=None, load_env=True, keep_alive=None):
    defaults = defaults or {}
    super().__init__({**DEFAULT_CONFIG, **defaults})
```
?

sanic/app.py
Amanit: I didn't saw any type checks in that method. I think that if somebody will pass a string or something in there he will see an error quite fast. I also like type hinting but it is almost not used in Sanic.

sanic/config.py
ahopkins: Much cleaner :+1:

sanic/app.py
ahopkins: Yes ... but something we need to fix. There is no time like the present :sunglasses: But I was suggesting one step further than just type hinting.

sanic/config.py
ahopkins: The `setup.py` uses `distutils.util.strtobool` to [process some environment variables ](https://sanic.readthedocs.io/en/latest/sanic/getting_started.html#install-sanic). One thing that has bothered me in the past is that we are not consistent in how environment variables are handled throughout.

If setting `SANIC_NO_UVLOOP=true` yields a `True` value, I would expect `SANIC_KEEP_ALIVE=true` to also.

sanic/app.py
Amanit: Do you want to raise an Exception or just ignore the value if it is not a bool?

ahopkins: Being that it is something that happens on startup, and should control an important feature, I think Exception is warranted.

(PS - Thanks for the work on this :beers: )

sanic/app.py
```diff
@@ -1027,8 +1027,15 @@ def run(
...
+            if access_log in [True, False]:
```
jotagesales: @Amanit  `isinstance(True, bool)` ??

:)

Amanit: Yeah, it was a late night, thanks :)

sanic/config.py
Amanit: There is something strange with tests *'-no-ext' when I add 
```
from distutils.util import strtobool
```
in `config.py`

```
E       assert ('asyncio', 4...f2244870da0>') == ('sanic.root',....0.1:42101/'")
E         At index 0 diff: 'asyncio' != 'sanic.root'
E         Use -v to get the full diff

tests/test_middleware.py:106: AssertionError
```
No matter where I put it - in the head or in the function.
Any idea?

ahopkins: I added it to `config.py` on a clean version of the `master` branch and was able to run both `tox -e py36-no-ext` and `tox -e py37-no-ext`.

Now, looking at `test_middleware.py` I see exactly what the issue is because I just dealt with this issue on `test_logo.py`. Why it is doing this for you right now ONLY with that import ... that makes no sense, unless somehow logging is being changed somewhere to `DEBUG`.

Basically what is happening is there is an extra debug that is logged, so `caplog` is one row off of stdout from where it should be. At least that is what I was dealing with (see `test_logo.py` to see how I dealt with it). What is that full output?

Amanit:
```
E       assert ('asyncio', 4...f609420dd68>') == ('sanic.root',....0.1:42101/'")
E         At index 0 diff: 'asyncio' != 'sanic.root'
E         Full diff:
E         - ('asyncio',
E         + ('sanic.root',
E         40,
E         +  "Exception occurred while handling uri: 'http://127.0.0.1:42101/'")
E         -  'Unclosed client session\n'
E         -  'client_session: <aiohttp.client.ClientSession object at 0x7f609420dd68>')

tests/test_middleware.py:106: AssertionError
```
And it is only fail when I run all the tests. If I run only `test_middleware.py` everything is fine.

sanic/app.py
```diff
@@ -1027,8 +1027,15 @@ def run(
                 "stop_event will be removed from future versions.",
                 DeprecationWarning,
             )
-        # compatibility old access_log params
-        self.config.ACCESS_LOG = access_log
+        # if access_log is passed explicitly change config.ACCESS_LOG
+        if access_log is not None:
+            if isinstance(access_log, bool):
```
yunstanford: I don't think this is a good pattern for type checking. If we assume users can input any types, then other params will have the same issue, and we do the same thing for all params ?

IMO, we should just put all param types in docstring and shows them in documentation.

sanic/app.py
```diff
@@ -1118,7 +1123,15 @@ def __call__(self):
             )
         # if access_log is passed explicitly change config.ACCESS_LOG
         if access_log is not None:
-            self.config.ACCESS_LOG = access_log
+            if isinstance(access_log, bool):
```
yunstanford: same here

sanic/app.py
```diff
@@ -967,34 +969,47 @@ def test_client(self):
...
+        debug: bool = False,
+        ssl: typing.Union[dict, SSLContext, None] = None,
+        sock: typing.Optional[socket] = None,
+        workers: int = 1,
+        protocol: typing.Type[Protocol] = None,
+        backlog: int = 100,
+        stop_event: typing.Any = None,
+        register_sys_signals: bool = True,
+        access_log: typing.Optional[bool] = None,
+        **kwargs: typing.Any
```
yunstanford: Thanks for adding typings, can we refactor a little bit ? like `from typing import Any, Union, ...` and use them directly here.

ahopkins:
@yunstanford Fair enough, but startup params probably should be verified. If not there could certainly be unintended consequences. And if not providing that we probably should work on upping testing coverage.

ahopkins:
:laughing: No worries. I am fine pulling it out here for consistency purposes, but I do think that separately we should be validating on startup because type checking is not enough, and I do not think it is a good practice to rely upon other tooling to alert the developer that they made a mistake.

Perhaps we can use `warnings` instead of an `Exception`.

sanic/app.py
```diff
@@ -4,12 +4,14 @@
...
-from asyncio import CancelledError, ensure_future, get_event_loop
+from asyncio import CancelledError, Protocol, ensure_future, get_event_loop
 from collections import defaultdict, deque
 from functools import partial
 from inspect import getmodulename, isawaitable, signature, stack
-from ssl import Purpose, create_default_context
+from socket import socket
+from ssl import Purpose, SSLContext, create_default_context
 from traceback import format_exc
+from typing import Any, Optional, Type, Union
```
vltr: To "shorten" code (just) a little **and** to "encourage" developers to use more `typing` hints, I always do the following:

```python
import typing as t
...
def foo(
    bar: t.Type[object],
    baz: t.Optional[int],
):
    pass
```

Does that makes sense?

sanic/worker.py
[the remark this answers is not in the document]
vltr: I agree with @ahopkins in this: the `loglevel` must be in tune for this to be "manageable", or else a **huge** warning (or even an `Exception`, IMO) should appear.

Also, @chenjr0719 made a good point on `app.run()` and `app.create_server()`, although this is a choice of the developer on how to launch Sanic; you won't use both, will you? As for overriding `app.config.ACCESS_LOG`, this does not feels right :confused:

vltr:
Just for clarification over this subject.

@Amanit 
> And that logs are affecting the performance dramatically

Yes, and this is expected to happen, specially if your access log is being thrown to `stdout`. The reason is I/O bound and simple: every `stdout` goes to your terminal (assuming you're calling your Sanic app from a terminal), that requires an underlying fb (framebuffer) and, if virtual, will need transformation to end up in your `vte` and so on. This does have a **huge** impact on performance. My advice to testing (benchmarking) on your local machine is to sending the application `stdout` to a file (if a production environment) or simply `/dev/null` to minimize this overhead:

```
$ gunicorn ... > /dev/null 2>&1
```

Let me know if this changes your benchmark results (even if this is just a test, not the "ideal" production tip).

Amanit:
> ```
> $ gunicorn ... > /dev/null 2>&1
> ```
> Let me know if this changes your benchmark results (even if this is just a test, not the "ideal" production tip).

```
[fedoruk@localhost opt]$ ./wrk -d10s -t8 -c400 http://127.0.0.1:8100
Running 10s test @ http://127.0.0.1:8100
  8 threads and 400 connections
    Latency    14.65ms    5.47ms  65.88ms   72.74%
    Req/Sec     3.45k   310.07     4.37k    72.75%
  274891 requests in 10.04s, 33.29MB read
Requests/sec:  27392.82
```

It is better but still 2x slower.

vltr:
> It is better but still 2x slower.

I know :wink: This was just to pinpoint the impact on having all `stdout` to a terminal (so to speak), not a complete solution for the problem.

Amanit:
@ahopkins @yunstanford is there anything else I should do to get that PR approved?

PR #1447, opened 2019-01-01
request locals

halaei:
Based on what I get from documentation, if I want to add some data to the request object, e.g. the user that sends the request or any other info generated by a middleware, I can add it to request.headers property. However, I think this is a little bit wired because headers property is designed for another purpose. So I think request needs a dictionary to contain anything that a developer desire. The idea is based on [express.js res.locals](https://expressjs.com/en/4x/api.html#res.locals) but I suppose other async frameworks have a similar property in their request or response objects.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -89,6 +89,7 @@ class Request(dict):
         "_port",
         "__weakref__",
         "raw_url",
+        "_locals",
     )
 
     def __init__(self, url_bytes, headers, version, method, transport):
@@ -318,6 +319,12 @@ class Request(dict):
         return urlunparse(
             (self.scheme, self.host, self.path, None, self.query_string, None)
         )
+
+    @property
+    def locals(self):
+        if not hasattr(self, "_locals"):
+            self._locals = {}
+        return self._locals
 
 
 File = namedtuple("File", ["type", "body", "name"])


ahopkins:
First of all, thank you very much for your interest and taking a step forward with this PR. Here are a few points I would like to make.

1. You did not implement a `setter` so there is no way to change `request.locals` without changing the underlying `request._locals`. I am not sure what this layer of privacy adds, except I suppose it has a savings of some memory by not creating the `dict` until it is needed. With this design pattern though, I would think there would also be a `@locals.setter`.
1. I am not sure this is a feature we are going to implement right now since it can be achieved without this.

```python
@app.middleware('request')
async def get_user(request):
    request['user'] = "me"

@app.route("/")
async def test(request):
    return json({"user": request.get('user')})
```

Perhaps this points the finger at a different problem: **documentation**. That you read the documents and took away that the only way to modify the request object is by modifying the headers is a problem (for Sanic, not yours). We need to do a better job of documenting that the request object is afterall a `dict` and therefore would be subject to adding and removing properties just as any other `dict`.

So, for now I am closing this. Perhaps there is a case to be made for actually storing this stuff on `request.locals` rather than a `request.get()` pattern. But it is probably up for a discussion first.

halaei:
Oh, you are totally right. For me it was easy to know that request is a `dict`, but I messed up object properties and dict keys and __slots__ concepts because I am python newbie :)
Thanks for your help.

Issue #1448, opened 2019-01-01
Documentation on amending request object 


ahopkins:
The documentation on how to amend the request object (with Middleware for example) could use some more explanation that it works like a regular dict. See #1447.

chenjr0719:
@ahopkins I can handle this.

PR #1449, opened 2019-01-02
Add example of amending request object

chenjr0719:
Related to #1447, #1448. Add an example to add key, pop key, and check key exists in request object.

### DIFF
--- a/docs/sanic/middleware.md
+++ b/docs/sanic/middleware.md
@@ -36,20 +36,31 @@ this.
 ```
 app = Sanic(__name__)
 
+
+@app.middleware('request')
+async def add_key(request):
+    # Add a key to request object like dict object
+    request['foo'] = 'bar'
+
+
 @app.middleware('response')
 async def custom_banner(request, response):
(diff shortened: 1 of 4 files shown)


docs/sanic/middleware.md
```diff
@@ -36,20 +36,31 @@ this.
...
 	response.headers["x-xss-protection"] = "1; mode=block"
 
+
 app.run(host="0.0.0.0", port=8000)
 ```
 
-The above code will apply the two middleware in order. First, the middleware
+The above code will apply the three middleware in order. The first middleware
+**add_key** will add a new key `foo` into `request` object. This worked due to
+`request` object can be manipulated like `dict` object. Then, the second middleware
```
ahopkins: "This worked _because the_ `request` object can be..."

ahopkins:
`./examples/*.py` are not excluded from test coverage?

docs/sanic/middleware.md
chenjr0719: Oops, That's a little embarrassing.:flushed:

PR #1457, opened 2019-01-03
enforce integer for max-age cookie

r0fls:
Closes: https://github.com/huge-success/sanic/issues/1452

PR #1463, opened 2019-01-08
Fix test dependencies

chenjr0719:
Notice Travis CI did not pass at #1457, and the reason is current version of `pytest-cov`(2.6.1) use `pytest>=3.6` in install _requires(see [here](https://github.com/pytest-dev/pytest-cov/blob/v2.6.1/setup.py#L124)). This is conflict with `Sanic`'s test_requires(it use `pytest==3.3.2`). Set `pytest-cov==2.6.0` in test_requires and tox.ini can fix this issue.

BTW, remove `pytest==3.3.2` is also a possible solution. But I don't know why use this version in test_requires, any idea?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,7 @@ tests_require = [
     "pytest==3.3.2",
     "multidict>=4.0,<5.0",
     "gunicorn",
-    "pytest-cov",
+    "pytest-cov==2.6.0",
     "aiohttp>=2.3.0,<=3.2.1",
     "beautifulsoup4",
     uvloop,
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ setenv =
 deps =
     coverage
     pytest==3.3.2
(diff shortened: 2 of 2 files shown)


harshanarayana:
@chenjr0719 Nice catch.

yunstanford:
Thanks for the fix, I've fixed this locally, but didn't push it out. And I'm thinking we should push forward, instead of backpining.

harshanarayana:
@chenjr0719 Logically speaking, we only need `tox` in our `tests_require` in `setup.py` if you are running tests inside the tox environment. However, if you choose to run the tests outside the `tox` context, you need the bare minimum requirements in `setup.py`.

chenjr0719:
@yunstanford Agreed, let me see is there any error after remove pinned version of `pytest`.

@harshanarayana I know, I just make them consistent to make sure this issue will not happen either using `tox` or outside the `tox`.

PR #1465, opened 2019-01-08
Respect `X-Forward-*` headers and generate correct URLs in `url_for`

BananaWanted:
Resolving these issues:
- https://github.com/huge-success/sanic/issues/1380
- https://github.com/huge-success/sanic/issues/958


sanic/request.py
```diff
@@ -253,6 +258,33 @@ def _get_address(self):
         self._ip = self._socket[0]
         self._port = self._socket[1]
 
+    @property
+    def server_name(self):
+        """
+        server name.
+        The difference between `Request.server_name` and `Request.host`
```
harshanarayana: Can you please ensure you use the right docstring formatting so that the API documentation is rendered properly?

sanic/request.py
```diff
@@ -274,6 +306,12 @@ def remote_addr(self):
 
     @property
     def scheme(self):
+        forwarded_proto = self.headers.get(
+            "x-forwarded-proto"
+        ) or self.headers.get("x-scheme")
```
harshanarayana: Isn't this header `X-Url-Scheme` and not `X-Scheme` ?

tests/test_requests.py
```diff
@@ -698,3 +777,35 @@ def test_request_form_invalid_content_type(app):
...
+
+    @app.get("/another_view/")
+    def view_name(request):
+        return text("OK")
+
+    request, response = app.test_client.get("/", headers={
+        'X-Forwarded-Proto': 'https',
+    })
+    assert app.url_for('view_name') == '/another_view'
+    assert app.url_for('view_name', _external=True) == 'http:///another_view'
```
harshanarayana: That result is fishy. `http:///` ?

sanic/request.py
vltr: I'll add a full comment regarding this.

BananaWanted:
Hi, @vltr @harshanarayana thanks for the comments. Regarding the header naming problem, `X-Fordered-*` are documented in MDN and widely used. `X-Scheme` is used by HAProxy, which is also a widely used name for years. 

There're only a limited number of headers used like de-facto standard, and we don't need to support fully customized headers like `X-My-Special-Scheme`.

In this PR I'm mainly trying to make Sanic **works with Nginx / HAProxy out of the box**. The flexibility requirements could be addressed as another issue.

Besides, there's also a standard `Forwarded` header not implemented in this PR because I personaly don't have much experience handle multi field headers in practice. I would be really appreciating if someone can help.

Also, as meintioned by @andreymal in https://github.com/huge-success/sanic/issues/801 

> looks unconfigurable and unsafe

It's surely a security risk using these headers, though in a good production setup it's not a issue since these headers would be filtered out or replaced by a proxy layer, but it still nice to have some configuration options to control this behavior.

tests/test_requests.py
BananaWanted: That's the missing UT case for `Sanic.url_for`. But there's no good way to fix it since getting a valid hostname needs `Request` context as in this PR.

vltr:
> Hi, @vltr @harshanarayana thanks for the comments. Regarding the header naming problem, `X-Fordered-*` are documented in MDN and widely used. `X-Scheme` is used by HAProxy, which is also a widely used name for years.
> 
> There're only a limited number of headers used like de-facto standard, and we don't need to support fully customized headers like `X-My-Special-Scheme`.
> 
> In this PR I'm mainly trying to make Sanic **works with Nginx / HAProxy out of the box**. The flexibility requirements could be addressed as another issue.

I understand you completely. There's a lot of other tools out there (such as AWS (ELB), Cloudflare, etc) that might have other headers as well and that's why I thought on making them somewhat configurable: we'll be swiping two problems at once. @huge-success/sanic-core-devs , your input would be mostly welcome on this.

tests/test_requests.py
vltr: You can provide some `kwargs` to `url_for`, such as `_server` that can have your valid hostname embed into, like `_server=request.headers.get("Host")` or `_scheme`, like `_scheme=request.headers.get("X-Forwarded-Proto")`. You can check most of them [here](https://github.com/huge-success/sanic/blob/2af229eb1ad89e4d064196d98b3f31630c0423d3/sanic/app.py#L777-L785). I think we should properly document those `kwargs` from the `url_for` method.

sjsadowski:
I think we do need to implement X-Forwarded-* - they're included both by common load balancers and for application routers like Traefik. I'm fine with making them configurable in the future, but I don't know that future features should be a blocker for this PR.

sanic/request.py
BananaWanted: updated

sjsadowski:
@vltr @harshanarayana could you review again when you get a chance?

andreymal:
@BananaWanted can you fix/rebase/rework this PR? I don't know what the Sanic maintainers think about it now, but I think this is an important change, but it seems that it conflicts with my PR https://github.com/huge-success/sanic/pull/1539

ahopkins:
@andreymal Speaking for myself, I agree that it is important.

ahopkins:
@yunstanford or @sjsadowski Can you run this as a "squash and merge"?

Issue #1466, opened 2019-01-09
Server cold start issue


denismakogon:
**Describe the bug**

We’ve (me and my folks) been trying to solve the cold start issue of our pet project (https://github.com/fnproject/fdk-python). At this moment we use aiohttp as the base unix http server.
However, when we started to do a solid performance testing, we’ve figured out that aiohttp takes almost 9 seconds to start within 128Mb RAM (this is our criteria: a server must start within 3 seconds, not longer).
Then I decided to look around to see what else we can use and then i came across sanic.
I was able to reduce the cold start time up to 6.5 seconds, however, this is still x2 of the required cold start time.

I figured out that sanic doesn’t have any importing rules and when i attempt to do things like "from sanic. import "tons of unnecessary imports being executed (for instance, if i don’t use blueprints - i don’t want the import to be loaded; if I don’t use websocket - i don’t want it to affect the cold start, etc.)

```
import time: self [us] | cumulative | imported package
import time:     75054 |      75054 | zipimport
import time:       971 |        971 | _frozen_importlib_external
import time:       166 |        166 |     _codecs
import time:     91712 |      91877 |   codecs
import time:      4969 |       4969 |   encodings.aliases
import time:     11474 |     108320 | encodings
import time:       810 |        810 | encodings.utf_8
import time:       501 |        501 | _signal
import time:      4961 |       4961 | encodings.latin_1
import time:       170 |        170 |     _abc
import time:      1861 |       2031 |   abc
import time:      5073 |       7104 | io
import time:     73518 |      73518 |       _stat
import time:      2963 |      76480 |     stat
import time:      1404 |       1404 |       genericpath
import time:    100572 |     101976 |     posixpath
import time:     99712 |      99712 |     _collections_abc
import time:      8394 |     286560 |   os
import time:      4962 |       4962 |   _sitebuiltins
import time:       797 |        797 |   sitecustomize
import time:       465 |        465 |   usercustomize
import time:      5821 |     298603 | site
import time:      2182 |       2182 |     types
import time:     87238 |      87238 |     _collections
import time:      8005 |      97424 |   enum
import time:       154 |        154 |     _sre
import time:      1343 |       1343 |       sre_constants
import time:      7231 |       8573 |     sre_parse
import time:      7200 |      15925 |   sre_compile
import time:       120 |        120 |     _functools
import time:       178 |        178 |         _operator
import time:     93176 |      93353 |       operator
import time:      1744 |       1744 |       keyword
[...]

PR #1468, opened 2019-01-12
Cold start issue

denismakogon:
# Intro
 
In Serverless Fn Project we use UNIX sockets to ship requests to functions,
 at function side, we have an FDK (function development kit),
 for an HTTP trigger, we use an HTTP server over UNIX socket.

# Problem area

 The problem is a cold start thing: we need an HTTP server to start within 3 seconds on 128Mb RAM.
 AIOHTTP takes almost 10 seconds to start, but Sanic can do that faster - 6 seconds.

# Improvement

 With some importing tweaks, I was able to go low up to 1.5-2 seconds with no harm to the rest of API.

 # Side change

 uvloop is quite outdated and doesn't support Python 3.7, specifically "start_serving=True",
 we need this feature in order to do some manipulations with Unix socket between the following stages:
 - server created, but not started
 - server can be started using "*.start_serving()"
 - a server can be started in an infinite loop with "*.serve_forever()"

 unfortunately, Sanic's app method "create_server" doesn't actually create a server,
 but make it start serving right away.

 so, this side change introduces a new method "sanic.server.create_server" that will create an instance of asyncio server.
 this new method is compatible with both uvloop and asyncio event loop.

 "sanic.app.Sanic.create_server" is also changed in order to work with "sanic.server.create_server"

Closes: #1466

### DIFF
--- a/sanic/__init__.py
+++ b/sanic/__init__.py
@@ -1,7 +1,12 @@
-from sanic.app import Sanic
-from sanic.blueprints import Blueprint
-
-
 __version__ = "18.12.0"
 
 __all__ = ["Sanic", "Blueprint"]
+
+
+def __getattr__(name):
+    if name == "Sanic":
+        from sanic.app import Sanic
+        return Sanic
+    if name == "Blueprint":
+        from sanic.blueprints import Blueprint
+        return Blueprint
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -22,11 +22,11 @@ from sanic.handlers import ErrorHandler
 from sanic.log import LOGGING_CONFIG_DEFAULTS, error_logger, logger
 from sanic.response import HTTPResponse, StreamingHTTPResponse
 from sanic.router import Router
-from sanic.server import HttpProtocol, Signal, serve, serve_multiple
-from sanic.static import register as static_register
-from sanic.testing import SanicTestClient
+from sanic.protocol import HttpProtocol
+from sanic.server import Signal, serve, create_server, serve_multiple
+
 from sanic.views import CompositionView
-from sanic.websocket import ConnectionClosed, WebSocketProtocol
+from sanic import websocket_exceptions
 
 
 class Sanic:
@@ -478,7 +478,8 @@ class Sanic:
                 self.websocket_tasks.add(fut)
                 try:
                     await fut
-                except (CancelledError, ConnectionClosed):
+                except (CancelledError,
+                        getattr(websocket_exceptions, "ConnectionClosed")):
                     pass
                 finally:
                     self.websocket_tasks.remove(fut)
@@ -660,6 +661,7 @@ class Sanic:
         :param content_type: user defined content type for header
         :return: None
         """
+        from sanic.static import register as static_register
         static_register(
             self,
             uri,
@@ -978,11 +980,20 @@ class Sanic:
 
     @property
     def test_client(self):
+        from sanic.testing import SanicTestClient
         return SanicTestClient(self)
 
     # -------------------------------------------------------------------- #
     # Execution
     # -------------------------------------------------------------------- #
+
+    def ensure_protocol(self, protocol):
+        if protocol is None:
+            if self.websocket_enabled:
+                from sanic.websocket import WebSocketProtocol
+                return WebSocketProtocol
+            return HttpProtocol
+        return protocol
 
     def run(
         self,
@@ -1048,10 +1059,7 @@ class Sanic:
         if sock is None:
             host, port = host or "127.0.0.1", port or 8000
 
-        if protocol is None:
-            protocol = (
-                WebSocketProtocol if self.websocket_enabled else HttpProtocol
-            )
+        protocol = self.ensure_protocol(protocol)
         if stop_event is not None:
             if debug:
                 warnings.simplefilter("default")
@@ -1121,6 +1129,7 @@ class Sanic:
         backlog: int = 100,
         stop_event: Any = None,
         access_log: Optional[bool] = None,
+        server_kwargs: Optional[dict] = None,
     ) -> None:
         """
         Asynchronous version of :func:`run`.
@@ -1160,10 +1169,7 @@ class Sanic:
         if sock is None:
             host, port = host or "127.0.0.1", port or 8000
 
-        if protocol is None:
-            protocol = (
-                WebSocketProtocol if self.websocket_enabled else HttpProtocol
-            )
+        protocol = self.ensure_protocol(protocol)
         if stop_event is not None:
             if debug:
                 warnings.simplefilter("default")
@@ -1192,8 +1198,12 @@ class Sanic:
             server_settings.get("before_start", []),
             server_settings.get("loop"),
         )
-
-        return await serve(**server_settings)
+        server_kwargs = server_kwargs if server_kwargs else {}
+        if server_kwargs.get("start_serving", True):
+            return await serve(**server_settings)
+        else:
+            return create_server(
+                server_kwargs=server_kwargs, **server_kwargs)
 
     async def trigger_events(self, events, loop):
         """Trigger events (functions or async)
--- a/sanic/protocol.py
+++ b/sanic/protocol.py
@@ -1,0 +1,586 @@
+import asyncio
+import traceback
+
+from httptools import HttpRequestParser
+from httptools.parser.errors import HttpParserError
+from multidict import CIMultiDict
+
+from sanic.exceptions import (
+    InvalidUsage,
+    PayloadTooLarge,
+    RequestTimeout,
+    ServerError,
+    ServiceUnavailable,
+)
+from sanic.log import access_logger, logger
+from sanic.request import Request, StreamBuffer
+from sanic.response import HTTPResponse
+
+
+try:
+    import uvloop
+
+    asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
+except ImportError:
+    pass
+
+
+current_time = None
+
+
+class Signal:
+    stopped = False
+
+
+class HttpProtocol(asyncio.Protocol):
+    """
+    This class provides a basic HTTP implementation of the sanic framework.
+    """
+
+    __slots__ = (
+        # event loop, connection
+        "loop",
+        "transport",
+        "connections",
+        "signal",
+        # request params
+        "parser",
+        "request",
+        "url",
+        "headers",
+        # request config
+        "request_handler",
+        "request_timeout",
+        "response_timeout",
+        "keep_alive_timeout",
+        "request_max_size",
+        "request_buffer_queue_size",
+        "request_class",
+        "is_request_stream",
+        "router",
+        "error_handler",
+        # enable or disable access log purpose
+        "access_log",
+        # connection management
+        "_total_request_size",
+        "_request_timeout_handler",
+        "_response_timeout_handler",
+        "_keep_alive_timeout_handler",
+        "_last_request_time",
+        "_last_response_time",
+        "_is_stream_handler",
+        "_not_paused",
+        "_request_handler_task",
+        "_request_stream_task",
+        "_keep_alive",
+        "_header_fragment",
+        "state",
+        "_debug",
+    )
+
+    def __init__(
+        self,
+        *,
+        loop,
+        request_handler,
+        error_handler,
+        signal=Signal(),
+        connections=None,
+        request_timeout=60,
+        response_timeout=60,
+        keep_alive_timeout=5,
+        request_max_size=None,
+        request_buffer_queue_size=100,
+        request_class=None,
+        access_log=True,
+        keep_alive=True,
+        is_request_stream=False,
+        router=None,
+        state=None,
+        debug=False,
+        **kwargs
+    ):
+        self.loop = loop
+        self.transport = None
+        self.request = None
+        self.parser = None
+        self.url = None
+        self.headers = None
+        self.router = router
+        self.signal = signal
+        self.access_log = access_log
+        self.connections = connections or set()
+        self.request_handler = request_handler
+        self.error_handler = error_handler
+        self.request_timeout = request_timeout
+        self.request_buffer_queue_size = request_buffer_queue_size
+        self.response_timeout = response_timeout
+        self.keep_alive_timeout = keep_alive_timeout
+        self.request_max_size = request_max_size
+        self.request_class = request_class or Request
+        self.is_request_stream = is_request_stream
+        self._is_stream_handler = False
+        self._not_paused = asyncio.Event(loop=loop)
+        self._total_request_size = 0
+        self._request_timeout_handler = None
+        self._response_timeout_handler = None
+        self._keep_alive_timeout_handler = None
+        self._last_request_time = None
+        self._last_response_time = None
+        self._request_handler_task = None
+        self._request_stream_task = None
+        self._keep_alive = keep_alive
+        self._header_fragment = b""
+        self.state = state if state else {}
+        if "requests_count" not in self.state:
+            self.state["requests_count"] = 0
+        self._debug = debug
+        self._not_paused.set()
+
+    @property
+    def keep_alive(self):
+        """
+        Check if the connection needs to be kept alive based on the params
+        attached to the `_keep_alive` attribute, :attr:`Signal.stopped`
+        and :func:`HttpProtocol.parser.should_keep_alive`
+
+        :return: ``True`` if connection is to be kept alive ``False`` else
+        """
+        return (
+            self._keep_alive
+            and not self.signal.stopped
+            and self.parser.should_keep_alive()
+        )
+
+    # -------------------------------------------- #
+    # Connection
+    # -------------------------------------------- #
+
+    def connection_made(self, transport):
+        self.connections.add(self)
+        self._request_timeout_handler = self.loop.call_later(
+            self.request_timeout, self.request_timeout_callback
+        )
+        self.transport = transport
+        self._last_request_time = current_time
+
+    def connection_lost(self, exc):
+        self.connections.discard(self)
+        if self._request_handler_task:
+            self._request_handler_task.cancel()
+        if self._request_stream_task:
+            self._request_stream_task.cancel()
+        if self._request_timeout_handler:
+            self._request_timeout_handler.cancel()
+        if self._response_timeout_handler:
+            self._response_timeout_handler.cancel()
+        if self._keep_alive_timeout_handler:
+            self._keep_alive_timeout_handler.cancel()
+
+    def pause_writing(self):
+        self._not_paused.clear()
+
+    def resume_writing(self):
+        self._not_paused.set()
+
+    def request_timeout_callback(self):
+        # See the docstring in the RequestTimeout exception, to see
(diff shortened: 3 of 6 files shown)


yunstanford:
I don't think this is a right direction to deal with cold restart, and introduce a lot code nasty-ness.

The issue is more because of Python Interpreter startup time, doesn't make sense to work around fix in application project layer.

Also, this is known issue in Python Community, and will be addressed in future release.

denismakogon:
I kind of diageee with a term you add mere: “code nasty-ness”, zero argument provided.

this PR introduces lazy-like imports, the reason for this change is that we can’t use importlib right away because it limits python version to 3.7

None of the future releases have cold start issue fix on their roadmap and I’m sorry to say but, sanic’s cold start is not python’s problem rather than a consequence of a very controversial decisions of developers who decided to put redundant code to init files along with doing tough imports when they are not required at all.

yunstanford:
1. Sure, we can totally get rid of importlib and also Thanks for refactoring work.
2. A lot inline imports inside functions is not good coding style and also hard to maintain in the future.
3. As for those lazy-imports, I don't how much benefits we can gain from that.  Generally, cold/slow start up in Python is more concerned in  Command line tools.
4. I've been seen a few times, Core developers have discussed Python interpreter Slow start up and will address in the future release. From benchmarks, 3.7 seems worse than 3.6 and 2.7 regarding start up time.

yunstanford:
Also,  in serverless arch like aws lambda, container warmup time is more concerned than Python interpreter start time

denismakogon:
> As for those lazy-imports, I don't how much benefits we can gain from that. Generally, cold/slow start up in Python is more concerned in Command line tools.

It’s not. Current problem of sanic for me is heavy imports, but not the CLI tools. Sanic spends 6 seconds to imports websockets, to import testing client, event loop, etc., and only less than 1 second to its own imports.


> A lot inline imports inside functions is not good coding style and also hard to maintain in the future.

That’s fun thing, because sanic already doing this, see testing client.
Inline import is a normal thing. With good scoping/placement there’s no problem in maintaining that type of code.

denismakogon:
> Also, in serverless arch like aws lambda, container warmup time is more concerned than Python interpreter start time

FYI AWS Lambda doesn’t use containers. This is a separate concern, for now a docker container can start within 300ms and sanic takes 7 seconds to start. So, container is not a weak point, but sanic is.

for now the problem is in sanic, aiohttp (and the rest of frameworks that never considered cold start as a real problem) specifically, a lot of redundant imports, code in init files, etc.

yunstanford:
1. Can you share more context on that ? Curious how it could take several seconds for import ? Didn't observe that.
2. Yeah, it might exist, for sure, refactor work needs to be done for

denismakogon:
AWS Lambda works on top VMs.

yunstanford:
6-10s seems pretty long. Which kind of file system/device are you using ? Will try if can reproduce. The main bottlenecks of Python imports are (i) Disk IO (ii) allocating a lot small pieces of memories. 

Pex and shiv might help.

denismakogon:
Docker-4-mac, debian slim strech (base python image python:3.7-slim-stretch).

For now, the problem is in imports time, like i said, if i want to use an HTTP server why should i ever import websockets? I want my http server stay as tiny as possible.

ahopkins:
@denismakogon That seems much longer than I experience. Whether running barebones, in a Docker container, or in a container managed by K8S.

That aside, I do not disagree that there need not be extra imports that are unnecessary. One of my priorities that I have been working on is #1265 (ASGI Integration). One of the side impacts of this is also to overhaul how `websockets` are handled in Sanic. Do a search for `websockets` on the Issues board, and you will see [a number of issues](https://github.com/huge-success/sanic/issues?utf8=%E2%9C%93&q=is%3Aopen+websockets) needed to tackle how they are integrated currently. Expect this in 2019. I hope for the `19.3` release, just not sure.

On the other hand, startup time (while important) is not itself a main goal for Sanic. If we can improve it, great. But I would be hesitant to make changes that will make the maintainability of the project more difficult or confusing in the future.

There needs to be a balance between the importance of the goal towards the success of the project, the feature/gain from the change, and the impact the change has on the the developers in the future.

denismakogon:
> That aside, I do not disagree that there need not be extra imports that are unnecessary. One of my priorities that I have been working on is #1265 (ASGI Integration). One of the side impacts of this is also to overhaul how websockets are handled in Sanic. Do a search for websockets on the Issues board, and you will see a number of issues needed to tackle how they are integrated currently. Expect this in 2019. I hope for the 19.3 release, just not sure.

I don’t have any problem with websockets. But I don’t understand a need in importing code that is related to websockets when my server doesn’t need them, it’s a huge overhead.


> On the other hand, startup time (while important) is not itself a main goal for Sanic.

I do understand this. But let’s be honest, according to the way sanic was implemented, you never considered cold start issue as real issue.

> But I would be hesitant to make changes that will make the maintainability of the project more difficult or confusing in the future.

Better code placement will bring zero impact on maintainability of the project overall.


> There needs to be a balance between the importance of the goal towards the success of the project, the feature/gain from the change, and the impact the change has on the the developers in the future.

That’s not the way to build a great community. You basically say - live with fork, we might not probably accept your contribution, because we might not believe in issues you’ve pointed out.

ahopkins:
@denismakogon I think you misunderstood me. That is not what I am saying at all. My point was that we need to take the entirety of the project into account when deciding on a change. Sometimes it is an easy no brainer, sometimes there needs to be conversation. All I was saying is that I think we all need to be cognizant of the fact that we need a balanced approach for the health of the project and community.

To be 100% clear, I am not taking a stance as to whether this is a good idea or not. I have not looked at your PR yet. I just read the comments, and was concerned when I saw @yunstanford:

> A lot inline imports inside functions is not good coding style and also hard to maintain in the future.

If there is some easy choices we can make that have a positive impact on the startup time, and "zero impact on maintainability of the project overall", then yes, that sounds excellent! But, if there is some impact, then I was merely putting more context into this comment.

We have said this also in regards to a new feature that may have a negative impact on performance. If (for example) a PR slows down the performance, we would need to think long and hard to make a decision that the penalty is acceptable given what the new feature adds.

Speaking for myself, I for one love to see Sanic being used in a way that it was not originally intended. And if there are things we can do to improve that, I appreciate the time and effort you have put into this.

abuckenheimer:
@denismakogon thanks for looking into this! FWIW I find the PR fairly grok-able and would love to have a more optimized cold start. However I see how without the context of this PR these changes will look counter intuitive, if I had opened up the sanic source code for the first time and saw the whole logic of web sockets inlined in a function I would be super confused and possibly turned off from contributing a change to that logic. I think we could get over that with some serious documentation but I think maybe we should start with some testing. We're talking in [this thread](https://community.sanicframework.org/t/benchmarking-utility/163/5) about benchmarking server performance, I think we could concurrently try to mock something up to test cold start performance. If we had a repeatable way to produce metrics on cold start performance we could make informed decisions about what changes are worth the complexity add.

ahopkins:
@yunstanford, I have a folder with about a dozen different sample uses of sanic. Plus I fired up a production code base in both bare metal and docker. They all booted up so quick that I didn't bother trying to bench mark. Bad on me.

imqqq:
@ahopkins +1, we have several micro services running in prod within docker containers, they all booted up fast, didn't notice 6-10s delay.

denismakogon:
@imqqq you can build a docker image and then run it with those constraints. That’s the requirements - a server must start within 3 seconds within those constraints.

vltr:
Hello, @denismakogon !  Thanks a lot for your interest on Sanic. Indeed, the question you brought us is very interesting to discuss and might be a good idea, as @abuckenheimer mentioned, to add into our benchmarking tool. In our community, everything is taken into account - it really doesn't matter what, as long as we have a consensus - that's why we are a community, so, I understand if you're trying to find a perfect fit for your project (and perhaps your deadline), but a good discussion is worth prior to any change in the code. First things first.

- We are aware that some things _might_ be bloated and we'll work hard to make things right. We already had a lot of discussions, like: put websockets into their own package; removing the test client from the Sanic package (on our plans to do so); add ASGI interface (thus removing some of the required logic) and so on - and now with your issue, cold start will probably be another one.
- Some of these questions take time: testing, community support, code standards (and readability), our own time (we really, really do all these things related to Sanic in our spare time);
- Since we turned Sanic into an organization (late September), our main focus was to find people willing to help the project, provide better documentation, increase testing coverage and release a more stable version, taking over _a lot_ of stalled issues and PRs here on Github, so it was a huge community effort and we are very happy with the results :smiley:

I've seen a lot of frameworks taking some of the measures you provided in your PR to optimize cold start, like Twisted Matrix (that I used for a long time), but this made their code very hard to maintain - thus just a handful of people works constantly in the project and it's hard for newcomers to dive into the code and understand the workflow. We want Sanic to be as developer friendly as possible, thus trying to avoid inline imports or ["weird black magic"](https://github.com/twisted/constantly/blob/39887b6e131a72b04a338919519e972de668c586/constantly/_constants.py#L198) into the codebase - hence the discussions we have to remove the test client to another package and etc, as an example.

But this does not mean we won't accept the PR. It does not mean we don't care about. _Au contraire_, we do care about and we love all contributions!

I made a simple test with docker in my machine with the image you mentioned, `python:3.7-slim-stretch`. Here's my `Dockerfile`:

```
FROM python:3.7-slim-stretch

RUN set -ex \
    \
    && apt-get update \
    && apt-get install -y \
        build-essential \
        make \
        gcc \
        locales \
    && pip install --no-cache-dir \
        ujson \
        uvloop \
        sanic \
    # && (test -d /root/.cache/pip && rm -rf /root/.cache/pip) \
    && apt-get remove -y --purge make gcc build-essential \
    && apt-get autoremove -y \
    && rm -rf /var/lib/apt/lists/*

RUN set -ex \
    \
    && dpkg-reconfigure locales \
    && locale-gen C.UTF-8 \
    && /usr/sbin/update-locale LANG=C.UTF-8

ENV LC_ALL C.UTF-8
# ENV PYTHONPROFILEIMPORTTIME=1

EXPOSE 8000

COPY ./sanicapp.py /sanicapp.py

CMD ["python", "/sanicapp.py"]
```

I commented out the `PYTHONPROFILEIMPORTTIME` because of the code I wrote:

```python
from datetime import datetime


def msg(m):
    print("[{}] - {}".format(datetime.utcnow().isoformat(), m))


msg("starting imports")

from sanic import Sanic

msg("imports done. creating sanic instance")

app = Sanic()

msg("sanic instance done. appending route")


@app.route("/")
async def test(request):
    return text("hello, world")


msg("route done. appending server listener")


async def notify_server_started(app, loop):
    msg("server started")


msg("server listener done. starting sanic ...")


def main():
    app.run(host="0.0.0.0", workers=1, port=8000)


if __name__ == "__main__":
    main()
```

And after calling the built image, here's what I've got:
```
$ docker run -ti --cpu-quota=10000 --cpu-period=100000 --cpu-shares=128 --memory=128m --kernel-memory=128m --memory-swap=128m --name sanic-simple-cold-start -p 8000:8000 sanic-simple
[2019-01-14T13:12:29.852041] - starting imports
[2019-01-14T13:12:33.162196] - imports done. creating sanic instance
[2019-01-14T13:12:33.557510] - sanic instance done. appending route
[2019-01-14T13:12:33.558038] - route done. appending server listener
[2019-01-14T13:12:33.558100] - server listener done. starting sanic ...
[2019-01-14 13:12:33 +0000] [1] [INFO] Goin' Fast @ http://0.0.0.0:8000
[2019-01-14T13:12:33.751751] - server started
[2019-01-14 13:12:33 +0000] [1] [INFO] Starting worker [1]
[2019-01-14 13:12:35 +0000] - (sanic.access)[INFO][172.17.0.1:45758]: GET http://127.0.0.1:8000/  200 12
```

My docker `check-config.sh` output (for curiosity):

```
$ bash ./check-config.sh
info: reading kernel config from /proc/config.gz ...

Generally Necessary:
- cgroup hierarchy: properly mounted [/sys/fs/cgroup]
- CONFIG_NAMESPACES: enabled
- CONFIG_NET_NS: enabled
- CONFIG_PID_NS: enabled
- CONFIG_IPC_NS: enabled
- CONFIG_UTS_NS: enabled
- CONFIG_CGROUPS: enabled
- CONFIG_CGROUP_CPUACCT: enabled
- CONFIG_CGROUP_DEVICE: enabled
- CONFIG_CGROUP_FREEZER: enabled
- CONFIG_CGROUP_SCHED: enabled
- CONFIG_CPUSETS: enabled
- CONFIG_MEMCG: enabled
- CONFIG_KEYS: enabled
- CONFIG_VETH: enabled (as module)
- CONFIG_BRIDGE: enabled (as module)
- CONFIG_BRIDGE_NETFILTER: enabled (as module)
- CONFIG_NF_NAT_IPV4: enabled (as module)
- CONFIG_IP_NF_FILTER: enabled (as module)
- CONFIG_IP_NF_TARGET_MASQUERADE: enabled (as module)
- CONFIG_NETFILTER_XT_MATCH_ADDRTYPE: enabled (as module)
- CONFIG_NETFILTER_XT_MATCH_CONNTRACK: enabled (as module)
- CONFIG_NETFILTER_XT_MATCH_IPVS: enabled (as module)
- CONFIG_IP_NF_NAT: enabled (as module)
- CONFIG_NF_NAT: enabled (as module)
- CONFIG_NF_NAT_NEEDED: enabled
- CONFIG_POSIX_MQUEUE: enabled

Optional Features:
- CONFIG_USER_NS: enabled
- CONFIG_SECCOMP: enabled
- CONFIG_CGROUP_PIDS: enabled
- CONFIG_MEMCG_SWAP: enabled
- CONFIG_MEMCG_SWAP_ENABLED: enabled
    (cgroup swap accounting is currently enabled)
- CONFIG_LEGACY_VSYSCALL_NONE: enabled
    (containers using eglibc <= 2.13 will not work. Switch to
     "CONFIG_VSYSCALL_[NATIVE|EMULATE]" or use "vsyscall=[native|emulate]"
     on kernel command line. Note that this will disable ASLR for the,
     VDSO which may assist in exploiting security vulnerabilities.)
- CONFIG_BLK_CGROUP: enabled
- CONFIG_BLK_DEV_THROTTLING: enabled
- CONFIG_IOSCHED_CFQ: enabled
- CONFIG_CFQ_GROUP_IOSCHED: enabled
- CONFIG_CGROUP_PERF: enabled
- CONFIG_CGROUP_HUGETLB: enabled
- CONFIG_NET_CLS_CGROUP: enabled (as module)
- CONFIG_CGROUP_NET_PRIO: enabled
- CONFIG_CFS_BANDWIDTH: enabled
- CONFIG_FAIR_GROUP_SCHED: enabled
- CONFIG_RT_GROUP_SCHED: missing
- CONFIG_IP_NF_TARGET_REDIRECT: enabled (as module)
- CONFIG_IP_VS: enabled (as module)
- CONFIG_IP_VS_NFCT: enabled
- CONFIG_IP_VS_PROTO_TCP: enabled
- CONFIG_IP_VS_PROTO_UDP: enabled
- CONFIG_IP_VS_RR: enabled (as module)
- CONFIG_EXT4_FS: enabled (as module)
- CONFIG_EXT4_FS_POSIX_ACL: enabled
- CONFIG_EXT4_FS_SECURITY: enabled
- Network Drivers:
  - "overlay":
    - CONFIG_VXLAN: enabled (as module)
      Optional (for encrypted networks):
      - CONFIG_CRYPTO: enabled
      - CONFIG_CRYPTO_AEAD: enabled
      - CONFIG_CRYPTO_GCM: enabled
      - CONFIG_CRYPTO_SEQIV: enabled
      - CONFIG_CRYPTO_GHASH: enabled
      - CONFIG_XFRM: enabled
      - CONFIG_XFRM_USER: enabled (as module)
      - CONFIG_XFRM_ALGO: enabled (as module)
      - CONFIG_INET_ESP: enabled (as module)
      - CONFIG_INET_XFRM_MODE_TRANSPORT: enabled (as module)
  - "ipvlan":
    - CONFIG_IPVLAN: enabled (as module)
  - "macvlan":
    - CONFIG_MACVLAN: enabled (as module)
    - CONFIG_DUMMY: enabled (as module)
  - "ftp,tftp client in container":
    - CONFIG_NF_NAT_FTP: enabled (as module)
    - CONFIG_NF_CONNTRACK_FTP: enabled (as module)
    - CONFIG_NF_NAT_TFTP: enabled (as module)
    - CONFIG_NF_CONNTRACK_TFTP: enabled (as module)
- Storage Drivers:
  - "aufs":
    - CONFIG_AUFS_FS: missing
  - "btrfs":
    - CONFIG_BTRFS_FS: enabled (as module)
    - CONFIG_BTRFS_FS_POSIX_ACL: enabled
  - "devicemapper":
    - CONFIG_BLK_DEV_DM: enabled (as module)
    - CONFIG_DM_THIN_PROVISIONING: enabled (as module)
  - "overlay":
    - CONFIG_OVERLAY_FS: enabled (as module)
  - "zfs":
    - /dev/zfs: missing
    - zfs command: missing
    - zpool command: missing

Limits:
- /proc/sys/kernel/keys/root_maxkeys: 1000000
```

So, in _my machine_, it looks like it took almost 3 seconds from importing Sanic and the server to reporting to be running, being the last couple of lines a simple `curl` call I made from another terminal. But, that's relative to my machine (that's quite fast, even though it's two years old already). In other machines? I don't know. Cold start really is relative. If I put this on a 1st gen raspberry pi with a very, _very slow_ sd / micro-sd card, wouldn't you agree that this might be even slower? As well as being very, very fast if using a state of the art server running an array of nvme drives for storage? There's something we might have to think of as well.

So, there's a lot of variables on the table for cold starting Sanic: from design choices to the metal that's running it. We will need a plan for that :wink: And if you are keen to help, well, welcome onboard! I think a thread in our [community forums](https://community.sanicframework.org/) would make the perfect place for this, since in there we are also having a lot of other discussions (like the ones I mentioned). Please, join! :smiley:

denismakogon:
Hi, I actually posted cold start issue at forum in first place but nobody responded to that for 5 days in a row: https://community.sanicframework.org/t/cold-start-issue/182

Regarding relativity of cold start issue. Indeed, on a different devices it can be a lot slower, however, the technology we’d use for serving http requests over Unix socket must start within less than 3 seconds on a docker, we don’t really care about overall performance because we’re not going over 1K requests per second, and on 128mb if people see a delay they’d get a retriable error code, so, we’re kinda fine with not being able to serve a lot requests, but the framework must start in less than 3 seconds at maximum.

For instance, in ASGI, hypercorn takes almost 6 seconds for its imports, quart takes 10+ seconds, etc.

vltr:
Sorry for missing that on the community forums - I'm just answering things _ad-hoc_ for the last couple of weeks (if not more). I understand that performance-wise you're not concerned, only the startup time.

Have you tried `uvicorn` with the **most simple server possible** using Python to see the startup time?

```python
class App:
    def __init__(self, scope):
        self.scope = scope

    async def __call__(self, receive, send):
        await send({
            'type': 'http.response.start',
            'status': 200,
            'headers': [
                [b'content-type', b'text/html; charset=utf-8'],
            ],
        })
        await send({
            'type': 'http.response.body',
            'body': b'hello, world!',
        })
```

To call:

```
$ uvicorn thefile:App
```

Issue #1451, opened 2019-01-02
Remove unwanted None check for __repr__ in `Request` class


harshanarayana:
# Description
The current implementation of the `Request` class provided by `sanic` has a `__repr__` method that performs an unnecessary `None` check on `method` and `path` and there it is not possible that they can be `None`.  We can remove these condition checks and have a much simpler `__repr__` implementation. 

# Reference
1. `__repr__` [implementation](https://github.com/huge-success/sanic/blob/096c44b910443e74c821dbf21adf2f86ccda8611/sanic/request.py#L97-L102)
2. Community Discussion [Thread](https://community.sanicframework.org/t/housekeeping-items/91/3)

chenjr0719:
Just a reminder, don't forget the test.
https://github.com/huge-success/sanic/blob/cea1547e08230b6ad49eb7777fd8db5335382b7a/tests/test_requests.py#L558-L567

szepnapot:
Hey, can I take this one? :)

PR #1472, opened 2019-01-16
Remove unwanted None check for __repr__ in `Request` class

xxNB:
about #1451
i just delete and add "[]"😅


xxNB:
@ahopkins Hi ahopkins. I am done.

PR #1476, opened 2019-01-18
add Request.not_grouped_args, deprecation warning Request.raw_args

Amanit:
`Request.raw_args` has a huge bug - it takes only the first value of the parameter.
In case of the URLs with `?key1=value1&key2=value2&key1=value3` querystring, `key1` will get `["value1", "value3"]` in `Request.args` and only `value1` in `Request.raw_args`. 

And you can't call it even raw - it is parsed and then reduced.

As it is no opportunity to return multiple equal keys using `dict` for the backward compatibility I created another property `Request.not_grouped_args` that returns list of tuples from `urllib.parse.parse_qsl` method.
I think it is better avoid using `raw_args` and cut it from the future versions.

### DIFF
--- a/docs/sanic/request_data.md
+++ b/docs/sanic/request_data.md
@@ -19,6 +19,8 @@ The following variables are accessible as properties on `Request` objects:
   URL that resembles `?key1=value1&key2=value2`. If that URL were to be parsed,
   the `args` dictionary would look like `{'key1': ['value1'], 'key2': ['value2']}`.
   The request's `query_string` variable holds the unparsed string value.
+  Property is providing the default parsing strategy. If you would like to change it look to the section below
+  (`Changing the default parsing rules of the queryset`).
 
   ```python
   from sanic.response import json
@@ -28,11 +30,54 @@ The following variables are accessible as properties on `Request` objects:
       return json({ "parsed": True, "args": request.args, "url": request.url, "query_string": request.query_string })
   ```
 
-- `not_grouped_args` (list) - On many cases you would need to access the url arguments in
-  a less packed form. For same previous URL `?key1=value1&key2=value2`, the
-  `not_grouped_args` list would look like `[('key1', 'value1'), ('key2', 'value2')]`.
+- `query_args` (list) - On many cases you would need to access the url arguments in
+  a less packed form. `query_args` is the list of `(key, value)` tuples.
+  Property is providing the default parsing strategy. If you would like to change it look to the section below
+  (`Changing the default parsing rules of the queryset`).
+  For the same previous URL queryset `?key1=value1&key2=value2`, the
+  `query_args` list would look like `[('key1', 'value1'), ('key2', 'value2')]`.
   And in case of the multiple params with the same key like `?key1=value1&key2=value2&key1=value3`
-  the `not_grouped_args` list would look like `[('key1', 'value1'), ('key2', 'value2'), ('key1', 'value3')]`.
+  the `query_args` list would look like `[('key1', 'value1'), ('key2', 'value2'), ('key1', 'value3')]`.
+
+  The difference between Request.args and Request.query_args
+  for the queryset `?key1=value1&key2=value2&key1=value3`
+
+  ```python
+  from sanic import Sanic
+  from sanic.response import json
+
+  app = Sanic(__name__)
+
+
+  @app.route("/test_request_args")
+  async def test_request_args(request):
+      return json({
+          "parsed": True,
+          "url": request.url,
+          "query_string": request.query_string,
+          "args": request.args,
+          "raw_args": request.raw_args,
+          "query_args": request.query_args,
+      })
+
+  if __name__ == '__main__':
+      app.run(host="0.0.0.0", port=8000)
+  ```
+
+  Output
+
+  ```
+  {
+    "parsed":true,
+    "url":"http:\/\/0.0.0.0:8000\/test_request_args?key1=value1&key2=value2&key1=value3",
+    "query_string":"key1=value1&key2=value2&key1=value3",
+    "args":{"key1":["value1","value3"],"key2":["value2"]},
+    "raw_args":{"key1":"value1","key2":"value2"},
+    "query_args":[["key1","value1"],["key2","value2"],["key1","value3"]]
+  }
+  ```
+
+  `raw_args` contains only the first entry of `key1`. Will be deprecated in the future versions.
 
 - `files` (dictionary of `File` objects) - List of files that have a name, body, and type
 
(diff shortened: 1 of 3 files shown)


Amanit:
If I understand well the coverage test is not passed because of the
```
if self.app.debug:
    warnings.simplefilter("default")
```
Should I really test it?

ahopkins:
I will look over the PR more closet later. As for the warning question, that looks fine to add a noqa

Amanit:
@ahopkins noqa did not affected the coverage, I have set # pragma: no cover.
But it is still - 0.01% of coverage :)

yunstanford:
thanks for the PR! i'm thinking about the name, it's not very intuitive right now..

sanic/request.py
```diff
@@ -212,9 +215,24 @@ def args(self):
...
+        warnings.warn(
+            "Use of raw_args will be deprecated in "
+            "the future versions. Please use args or not_grouped_args "
+            "properties instead",
+            DeprecationWarning,
+        )
         return {k: v[0] for k, v in self.args.items()}
 
+    @property
+    def not_grouped_args(self) -> list:
```
harshanarayana: I might be nitpicking but, I think it might be a good idea to rename this property to something more simpler like `query`/`args_as_tuple`/`arg_tuple` or something similar,

tests/test_requests.py
```diff
@@ -626,6 +626,34 @@ def handler(request):
...
 
+def test_request_not_grouped_args(app):
+    # test multiple params with the same key
+    params = [('test', 'value1'), ('test', 'value2')]
+
+    @app.get("/")
+    def handler(request):
+        return text("pass")
+
+    request, response = app.test_client.get("/", params=params)
```
harshanarayana: Mind adding a test case to simulate the actual call with `?test=value1&test=value2` directly in the URL to your test case?

sanic/request.py
```diff
@@ -212,9 +215,24 @@ def args(self):
...
+        if self.parsed_not_grouped_args is None:
+            if self.query_string:
+                self.parsed_not_grouped_args = parse_qsl(self.query_string)
```
harshanarayana: Can we make use of the  `keep_blank_values` so that the no value query string keys are retained ?

sanic/request.py
ahopkins: The current properties are: 

- `args`
- `parsed_args`
- `raw_args`

My suggestion would be to follow this pattern and use either: `query_args` or `tuple_args`.

docs/sanic/request_data.md
```diff
@@ -28,9 +28,11 @@ The following variables are accessible as properties on `Request` objects:
...
   ```
 
-- `raw_args` (dict) - On many cases you would need to access the url arguments in
-  a less packed dictionary. For same previous URL `?key1=value1&key2=value2`, the
-  `raw_args` dictionary would look like `{'key1': 'value1', 'key2': 'value2'}`.
+- `not_grouped_args` (list) - On many cases you would need to access the url arguments in
+  a less packed form. For same previous URL `?key1=value1&key2=value2`, the
+  `not_grouped_args` list would look like `[('key1', 'value1'), ('key2', 'value2')]`.
+  And in case of the multiple params with the same key like `?key1=value1&key2=value2&key1=value3`
+  the `not_grouped_args` list would look like `[('key1', 'value1'), ('key2', 'value2'), ('key1', 'value3')]`.
```
ahopkins: I think this is a confusing way to show this information in the documentation.

This could be solved with:

```python
return json({
    "parsed": True,
    "url": request.url,
    "query_string": request.query_string,
    "args": request.args,
    "parsed_args": request.parsed_args,
    "raw_args": request.raw_args,
    "query_args": request.query_args,  # or whatever the new one will be
})
```

And then showing the actual JSON output to the reader.

tests/test_requests.py
ahopkins: :+1:

sanic/request.py
harshanarayana: +1 for `query_args`

sanic/request.py
Amanit: Or another way to handle it without changing default behavior:
```python
    def get_args(self, keep_blank_values=False, strict_parsing=False, encoding='utf-8', errors='replace'):
        if self.parsed_args is None:
            if self.query_string:
                self.parsed_args = RequestParameters(
                    parse_qs(self.query_string, keep_blank_values, strict_parsing, encoding, errors)
                )
            else:
                self.parsed_args = RequestParameters()
        return self.parsed_args

    args = property(get_args)
```
The same for `query_args`.

Amanit: If someone need to change some default parameters - just call `get_args` method directly.

sanic/request.py
```diff
@@ -200,21 +203,73 @@ def files(self):
 
         return self.parsed_files
 
-    @property
-    def args(self):
-        if self.parsed_args is None:
+    def get_args(
```
harshanarayana: This is a major breaking change for someone who is already using this feature. Instead of removing it directly, should go via the deprecate -> remove phase. It can be a major pain otherwise for anyone trying to upgrade to a latest version

sanic/request.py
```diff
@@ -200,21 +203,73 @@ def files(self):
...
+            "Use of raw_args will be deprecated in "
+            "the future versions. Please use args or query_args "
+            "properties instead",
+            DeprecationWarning,
+        )
         return {k: v[0] for k, v in self.args.items()}
 
+    def get_query_args(
+        self,
+        keep_blank_values: bool = False,
```
harshanarayana: The Original suggestion was the add the `keep_blank_values` to the `parse_qsl` directly and not impact any of the existing behaviors. This would exist only in the new method/property being added and not modify any of the existing behaviors.

sanic/request.py
Amanit: Nothing has changed for who is using `args` property. Look below -
```python
args = property(get_args)
```

sanic/request.py
Amanit: It is not normal that 1 parser `args` will return values with `keep_blank_values=False` and new one will use `keep_blank_values=True`, so I should have changed both.
I think that it is even worst to just change the default behavior.
And now for those who already use `args` - nothing has changed. 
If they need `keep_blank_values=True` - they will just use `get_args(keep_blank_values=True)`.
If somebody will need to change some other parameters of `parse_qs` / `parse_qsl` now it is very easy.

Amanit:
@ahopkins for now `raw_args` is not working properly, but there is no possibility to fix it without changing the output type. I think that for now it is just enough to not mention it in the documentation and show a Warning on use.

sanic/request.py
harshanarayana: @Amanit My bad. Sorry for the confusion. I totally missed that line.

sanic/request.py
harshanarayana: @Amanit  Works for me 👍  Thanks for the clarification.   Can you please make sure that the API method is documented too with proper documentation? I know there is a valid example, having APIs documented helps too.

Amanit: Added docstrings for these methods.

Amanit:
@ahopkins @yunstanford Any suggestions on this PR?

Amanit:
> Are we going to deprecate raw_args? Let's address this and make a decision. If we are, the next thing we need to talk about is how to communicate that.

@ahopkins who we need to call to make this decision? :)

sjsadowski:
@ahopkins @harshanarayana @huge-success/sanic-core-devs are we comfortable accepting this PR at this point?

Amanit:
@sjsadowski seems that `unquote` import has been lost during the merge. I have pushed the fix.

sjsadowski:
@Amanit Thanks - just waiting on some folks to weigh in.

Issue #1354, opened 2018-10-08
`environment.yml` dependency


vltr:
https://github.com/huge-success/sanic/blob/5cc12fd945bf4c5e0d03c0003ad1178567c70e8c/environment.yml#L21

We need to check if this dependency is still needed or if `docutils` already works nicely with `conda` _et al_.

sjsadowski:
This should be fixed with #1478

PR #1478, opened 2019-01-21
Upgrade setuptools version and use native docutils in doc build

chenjr0719:
The doc builds are always failed about one week ago. The reason is `recommonmark` has released a new version a week ago and install this new version with `setuptools==20.3`(`Sanic` use this version as doc build environment, check [this](https://github.com/huge-success/sanic/blob/2758a3ade6d05a5456c3b8522f092a8226d9c52e/environment.yml#L7)) cause an error. Use a newer version of `setuptools` can fix this issue. I also use native `docutils` in the doc build environment and it works properly(see #1354).

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,21 +1,18 @@
-name: py35
+name: py36
 dependencies:
-- openssl=1.0.2g=0
-- pip=8.1.1=py35_0
-- python=3.5.1=0
-- readline=6.2=2
-- setuptools=40.4.3=py35_0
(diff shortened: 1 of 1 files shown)


yunstanford:
pip version also seems old, can we also update that ? maybe also use py36

chenjr0719:
@yunstanford Sure, use python-3.6.5 with pip 18.1 now. I also remove some unnecessary dependencies and make sure all dependencies from docs extras require included in this `environment.yml`.

PR #1482, opened 2019-01-30
Add stream support for bp.add_route()

chenjr0719:
This PR adds `stream` argument for `bp.add_route()` to provide another solution if someone wants to using blueprint with request stream.


chenjr0719:
Just a record, `Travis` is failed at the lint check now. I think the reason is new releases of `flake8`(3.7.0 and 3.7.1). Try to fix it now.

PR #1483, opened 2019-01-31
sanic-zipkin

kevinqqnj:
add zipkin plugin.

Features:
- adding "Request span" by default
- if Request is from another micro-service endpoint, span will be attached (Inject/Extract) to that endpoint
- use "logger" decorator to create span for "methods" calls
- use "sz_rpc" method to create sub-span for RPC calls, attaching to parent span


sjsadowski:
I'm trying to identify what's going on with linting. We'll get this merged once that gets cleaned up.

PR #1489, opened 2019-02-14
Added "databases"

tomchristie:
Adds https://github.com/encode/databases to the "Database Integration" section.

### DIFF
--- a/docs/sanic/extensions.md
+++ b/docs/sanic/extensions.md
@@ -33,6 +33,7 @@ A list of Sanic extensions created by the community.
 - [Sanic CRUD](https://github.com/Typhon66/sanic_crud): CRUD REST API generation with peewee models.
 - [sanic-graphql](https://github.com/graphql-python/sanic-graphql): GraphQL integration with Sanic
 - [GINO](https://github.com/fantix/gino): An asyncio ORM on top of SQLAlchemy core, delivered with a Sanic extension. ([Documentation](https://python-gino.readthedocs.io/))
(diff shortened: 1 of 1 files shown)


vltr:
@tomchristie `databases` looks awesome! I could see it's in alpha yet, but out of curiosity: are you using it already to prototype some future service? I'll give it a shot tomorrow probably :wink:

tomchristie:
I have something that I'll start using it with, yes.

Tho I'm pretty focused on building out more OSS tooling, rather than service-building right now. It's feasible that next up will be "and here's a fully-fledged ORM that sits on top of `databases`", but let's not get too far ahead of ourselves. 😝

vltr:
@tomchristie hahaha yeah, that makes sense. SQLAlchemy is there now for more than 10 years and is rock solid enough to trust on it with eyes blindfolded. I think that if the necessity really grows, SQLAlchemy might end up with an async implementation by itself - we already have [GINO](https://github.com/fantix/gino) and if my memory does not betrays me, @zzzeek once [wrote in his blog](https://techspot.zzzeek.org/2015/02/15/asynchronous-python-and-databases/) (regarding the arrival of asyncio and etc) that the required changes in SQLAlchemy are actually "doable" to achieve that (section "SQLAlchemy? Asyncio? Yes?") - but he's the author, I think that thought must be deep buried in his mind ... But I'm eager to see more async stuff coming out of (or for) SQLAlchemy :sweat_smile:

PR #1497, opened 2019-02-22
Recognizes non-ASCII filenames in RFC 2231, and suport filename length is zero for multipart/form-data.

PWZER:
- Suport filename length is zero for multipart/form-data.
    ```
    ------sanic
    Content-Disposition: form-data; filename=""; name="test"

    ok

    ------sanic--
    ```
- Recognizes non-ASCII filenames in [RFC 2231](https://www.rfc-editor.org/rfc/rfc2331.txt) and [RFC 5987](https://www.rfc-editor.org/rfc/rfc5987.txt), "filename*" format.
    ```
    ------sanic
    Content-Disposition: form-data; filename*="utf-8''filename_%C2%A0_test"; name="test"

    ok

    ------sanic--
    ```
- Add some test cases in tests/test_requests.py::test_request_multipart_files, like those above.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -364,7 +364,7 @@ def parse_multipart_form(body, boundary):
 
         if field_name:
             post_data = form_part[line_index:-4]
-            if file_name:
+            if file_name is not None:
                 form_file = File(
                     type=content_type, name=file_name, body=post_data
                 )
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -2,10 +2,11 @@ import asyncio
 import json
 import sys
 
+import email.utils
 from cgi import parse_header
 from collections import namedtuple
 from http.cookies import SimpleCookie
-from urllib.parse import parse_qs, urlunparse
+from urllib.parse import parse_qs, urlunparse, unquote
 
 from httptools import parse_url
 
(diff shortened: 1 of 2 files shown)


sanic/request.py
```diff
@@ -356,28 +357,35 @@ def parse_multipart_form(body, boundary):
             )
 
             if form_header_field == "content-disposition":
-                file_name = form_parameters.get("filename")
                 field_name = form_parameters.get("name")
+                file_name = form_parameters.get("filename")
+
+                # non-ASCII filenames in RFC2231, "filename*" format
+                if file_name is None and form_parameters.get("filename*"):
```
harshanarayana: Since the `form_parameters` lookup is being done twice, maybe we can move the lookup to a single line and use a variable instead? (this is just a good to have change)

```python
filename_rfc5987 = form_parameters.get("filename*")

if not file_name and filename_rfc5987:
    encoding, _, value = email.utils.decode_rfc5987(filename_rfc2231)
    file_name = unquote(value, encoding=encoding)
```
Also the format you are using is `RFC 5987` standard and not `RFC2231` if I remember correctly. (Sorry, for the nitpicking)

sanic/request.py
```diff
@@ -356,28 +357,35 @@ def parse_multipart_form(body, boundary):
...
+                    )
+                    file_name = unquote(value, encoding=encoding)
             elif form_header_field == "content-type":
                 content_type = form_header_value
                 content_charset = form_parameters.get("charset", "utf-8")
 
         if field_name:
             post_data = form_part[line_index:-4]
-            if file_name:
+            if file_name is None:
```
harshanarayana: Maybe this can be changed to `if not file_name`? Also, is there a reason to switch and invert the existing check here?

PR #1501, opened 2019-03-01
Fix the auto_reloader to work when the executable was launched with a module

### LINKED ISSUE #1492: Module import fails when auto_reload is active
I have two piece of code (the structure has been simplified for clarity sake)
First in base.py
```
#! /usr/bin/env python
# -*- coding: utf-8 -*-
# module: init0

from abc   import ABCMeta
from sanic import Sanic

class BaseService( metaclass = ABCMeta ):
    def create_app( self ) -> Sanic:
        app = Sanic( __name__ )
        return app

# app = BaseService().create_app()

# app.run( host  = '0.0.0.0',
#      port  = 5000,
#      debug = True,
#     )

print('[DONE]')

```
Second in run_test.py
```
#! /usr/bin/env python
# -*- coding: utf-8 -*-
# module: init0

from init0.base import BaseService

def main():
    app = BaseService().create_app()
    app.run( host  = '0.0.0.0',
         port  = 5000,
         debug = True,
        )
    return

if __name__ == '__main__':
    main()

```

If I were to run `python -m init0.run_test` with `debug = False` then everything works perfectly, however if it's `debug = True`, then it'd throw me `ModuleNotFoundError: No module named 'init0'`

Is it some sort of loading error somewhere that I need to configure beforehand?

Thanks a lot in advance

ashleysommer:
Fixes #1492
Implements a detection routine when determining arguments for launching the child application, it checks if the executable was launched as a module, rather than a script, and if so launch the same module with the -m commandline argument again.

This PR also includes a different auto_reloader script, which passes the current cwd to the subprocess.call() call.

This is a basic workaround to get bugs in the current auto_reloader fixed for the 2019.03 sanic release.
We do intend to replace this simple naive auto_reloader in the future to be more robust one, that will potentially be implemented as a plugin.

### DIFF
--- a/sanic/reloader_helpers.py
+++ b/sanic/reloader_helpers.py
@@ -36,7 +36,15 @@ def _iter_module_files():
 def _get_args_for_reloading():
     """Returns the executable."""
     rv = [sys.executable]
-    rv.extend(sys.argv)
+    main_module = sys.modules["__main__"]
+    mod_spec = getattr(main_module, "__spec__", None)
+    if mod_spec:
+        # Parent exe was launched as a module rather than a script
+        rv.extend(["-m", mod_spec.name])
+        if len(sys.argv) > 1:
+            rv.extend(sys.argv[1:])
+    else:
+        rv.extend(sys.argv)
     return rv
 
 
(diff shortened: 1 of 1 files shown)


ashleysommer:
Can anyone work out what is causing the failure in the py35-no-ext test in travis?
The failure is in logging tests and is not related to this PR.
It seems that for some reason in py35 with no uvloop, the `asyncio` module is emitting a message to the caplog causing the caplog logging test to fail its assertion.

ashleysommer:
Codecov Diff is negative because there are no tests that touch the auto_reloader. So any modifications to the auto_reloader that add additional lines will decrease codecov further.

PR #1502, opened 2019-03-01
Remove current_time prefetch

langma:
Sanic server has a mechanism to prefetch current time, basically to prevent unnecessary `time.time()` calls. With certain installations this leads to leaking `TimerHandle` instances.

Visible at least in Kubernetes installations with continuous health check `GET` calls. 

Fixes https://github.com/huge-success/sanic/issues/1500.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -34,9 +34,6 @@ except ImportError:
     pass
 
 
-current_time = None
-
-
 class Signal:
     stopped = False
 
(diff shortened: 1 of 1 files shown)


vltr:
Thanks, @langma , this was one of the things in my list to remove for good in Sanic. I think it might be a good idea to backport it to version 18.12 as well. @sjsadowski , what do you think?

sjsadowski:
Created #1505 to track against 18.12

PR #1530, opened 2019-03-22
Bump version to 19.03.1

### LINKED ISSUE #1524: Publish 19.3 release to PyPI
Thank you for the release 3 days ago!

https://github.com/huge-success/sanic/releases/tag/19.3

It's missing from PyPI at the moment:

https://pypi.org/project/sanic/#history

Please publish it at your convenience 🙇 

Keep up the awesome work ❤️

seemethere:
Couldn't delete the release on github so we go with the next best thing
which is to just bump the patch version

A re-bump after #1527 

Once again, fixes #1524 

Signed-off-by: Eli Uriegas <seemethere101@gmail.com>

### DIFF
--- a/sanic/__init__.py
+++ b/sanic/__init__.py
@@ -2,6 +2,6 @@ from sanic.app import Sanic
 from sanic.blueprints import Blueprint
 
 
-__version__ = "19.03.0"
+__version__ = "19.03.1"
 
 __all__ = ["Sanic", "Blueprint"]


seemethere:
Going to go ahead and merge this once it goes green and move forward with the release.

seemethere:
There's also a `release.py` script which is what I'm assuming what was used for `19.3` and that needs to be updated to also update the `__version__` variable in `sanic/__init__.py`

harshanarayana:
@seemethere The `release.py` already updates the `sanic/__init__.py`. It uses the `setup.cfg` to hold the file name. But looks like this script wasn't used to do the last release. (Since the `current_version` in `setup.cfg` is still an older value. This would be auto-updated if the `release.py` was used.) But that's all right. That script is just a helper utility.

PR #1539, opened 2019-03-26
Add options to control the behavior of Request.remote_addr

### LINKED ISSUE #801: sanic behind proxy - option to replace IP of request.ip with X-Forwarded-For value

andreymal:
Closes #801 

Using `X-Forwarded-For` to determine real client ip is totally unsafe when proxy is not used, because client can fake this http header. This PR adds `PROXIES_COUNT` option to control the usage of this http header.

When nginx is used, `PROXIES_COUNT` should be `1` and `X-Forwarded-For` will be parsed. When nginx is not used (or does not override `X-Forwarded-For`), `PROXIES_COUNT` should be `0` and `X-Forwarded-For` will be ignored.

This PR also adds `X-Real-IP` that is very useful with nginx, and `FORWARDED_FOR_HEADER` and `REAL_IP_HEADER` options that allow to change the names of these headers.

There are some weird things:

* `PROXIES_COUNT` is `-1` by default, which means using the first ip from `X-Forwarded-For`. It's still totally unsafe, but backwards-compatible. It should be set to `0` for security reasons, but it changes the default behavior and probably breaks the compatibility.

* When `X-Forwarded-For` or `X-Real-IP` are unavailable or untrusted, `request.remote_addr` returns an empty string. This is compatible with the old behavior, but this is weird, I don't know why I have to use `ip = (request.remote_addr or request.ip)` boilerplate.

* English is not my native language, and I use Google Translate, so the documentation may be crappy, sorry for that D:

### DIFF
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -27,6 +27,9 @@ DEFAULT_CONFIG = {
     "WEBSOCKET_WRITE_LIMIT": 2 ** 16,
     "GRACEFUL_SHUTDOWN_TIMEOUT": 15.0,  # 15 sec
     "ACCESS_LOG": True,
+    "PROXIES_COUNT": -1,
+    "FORWARDED_FOR_HEADER": "X-Forwarded-For",
+    "REAL_IP_HEADER": "X-Real-IP",
 }
 
 
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -355,19 +355,38 @@ class Request(dict):
 
     @property
     def remote_addr(self):
-        """Attempt to return the original client ip based on X-Forwarded-For.
+        """Attempt to return the original client ip based on X-Forwarded-For
+        or X-Real-IP. If HTTP headers are unavailable or untrusted, returns
+        an empty string.
 
         :return: original client ip.
         """
         if not hasattr(self, "_remote_addr"):
-            forwarded_for = self.headers.get("X-Forwarded-For", "").split(",")
-            remote_addrs = [
-                addr
-                for addr in [addr.strip() for addr in forwarded_for]
-                if addr
-            ]
-            if len(remote_addrs) > 0:
-                self._remote_addr = remote_addrs[0]
+            if self.app.config.PROXIES_COUNT == 0:
+                self._remote_addr = ""
+            elif self.app.config.REAL_IP_HEADER and self.headers.get(
+                self.app.config.REAL_IP_HEADER
+            ):
+                self._remote_addr = self.headers[
+                    self.app.config.REAL_IP_HEADER
+                ]
+            elif self.app.config.FORWARDED_FOR_HEADER:
+                forwarded_for = self.headers.get(
+                    self.app.config.FORWARDED_FOR_HEADER, ""
+                ).split(",")
+                remote_addrs = [
+                    addr
+                    for addr in [addr.strip() for addr in forwarded_for]
+                    if addr
+                ]
+                if self.app.config.PROXIES_COUNT == -1:
+                    self._remote_addr = remote_addrs[0]
+                elif len(remote_addrs) >= self.app.config.PROXIES_COUNT:
+                    self._remote_addr = remote_addrs[
+                        -self.app.config.PROXIES_COUNT
+                    ]
+                else:
+                    self._remote_addr = ""
             else:
                 self._remote_addr = ""
         return self._remote_addr
--- a/docs/sanic/config.md
+++ b/docs/sanic/config.md
@@ -85,16 +85,19 @@ DB_USER = 'appuser'
 
 Out of the box there are just a few predefined values which can be overwritten when creating the application.
 
-    | Variable                  | Default   | Description                                               |
-    | ------------------------- | --------- | --------------------------------------------------------- |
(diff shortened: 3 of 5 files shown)


sjsadowski:
@huge-success/sanic-core-devs Is this a cleaner way to implement #1465 in your opinions?

sanic/request.py
```diff
@@ -355,19 +355,38 @@ def _get_address(self):
...
+                remote_addrs = [
+                    addr
+                    for addr in [addr.strip() for addr in forwarded_for]
+                    if addr
+                ]
+                if self.app.config.PROXIES_COUNT == -1:
+                    self._remote_addr = remote_addrs[0]
+                elif len(remote_addrs) >= self.app.config.PROXIES_COUNT:
+                    self._remote_addr = remote_addrs[
+                        -self.app.config.PROXIES_COUNT
```
harshanarayana: 👍 I like this way to implementing behavior similar to `set_real_ip_from` provided by Nginx.

sanic/request.py
```diff
@@ -355,19 +355,38 @@ def _get_address(self):
...
         if not hasattr(self, "_remote_addr"):
-            forwarded_for = self.headers.get("X-Forwarded-For", "").split(",")
-            remote_addrs = [
-                addr
-                for addr in [addr.strip() for addr in forwarded_for]
-                if addr
-            ]
-            if len(remote_addrs) > 0:
-                self._remote_addr = remote_addrs[0]
+            if self.app.config.PROXIES_COUNT == 0:
```
harshanarayana: Since this if and the last else block is doing `self._remote_addr = ""`, we can do that outside the `if` block and reduce the number if `if-else` branching.

harshanarayana:
@sjsadowski  I think this is a much cleaner way to implement the `x-forwarded-*` behavior. Looks more in compliance with how things are managed inside the Nginx as well when it comes to how the Real Ip and forwarded IPs are tackled to handle the spoofing.

sanic/request.py
andreymal: In this case I'll have to use more indentation levels or prepend every expression with `if self.app.config.PROXIES_COUNT is not None and self.app.config.PROXIES_COUNT > 0 and [...]` that is more complicated

(Or I'm too stupid and can't see a more obvious and beautiful way to do this)

sanic/config.py
```diff
@@ -27,6 +27,9 @@
     "WEBSOCKET_WRITE_LIMIT": 2 ** 16,
     "GRACEFUL_SHUTDOWN_TIMEOUT": 15.0,  # 15 sec
     "ACCESS_LOG": True,
+    "PROXIES_COUNT": -1,
```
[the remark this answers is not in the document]
andreymal: Perhaps, but the type of this value will be `Optional[int]` instead of just `int`, not sure what is better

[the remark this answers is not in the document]
sjsadowski: Let's leave as submitted. Maybe we'll need to revisit in the future, but I'm fine with this for now.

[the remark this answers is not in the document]
harshanarayana: Works for me. I have no issue retaining this as is.

sanic/request.py
harshanarayana: It's small change that isn't super important. 

I was suggesting to remove the 
```python
if self.app.config.PROXIES_COUNT == 0:
    self._remote_addr = ""
```
and 
```
else:
    self._remote_addr = ""
```
and instead merge them as 
```python
if not hasattr(self, "_remote_addr"):
    self._remote_addr = ""
```

sjsadowski: @andreymal would you be ok with making the above change?

sjsadowski: @harshanarayana when you get a chance, could you look at this again?

harshanarayana: @sjsadowski  @andreymal 
```python
        if not hasattr(self, "_remote_addr"):
            setattr(self, "_remote_addr", "")

        if self.app.config.PROXIES_COUNT != 0:
            if self.app.config.REAL_IP_HEADER and self.headers.get(self.app.config.REAL_IP_HEADER):
                self._remote_addr = self.headers[self.app.config.REAL_IP_HEADER]
            elif self.app.config.FORWARDED_FOR_HEADER:
                forwarded_for = self.headers.get(self.app.config.FORWARDED_FOR_HEADER, "").split(",")
                remote_addrs = [addr for addr in [addr.strip() for addr in forwarded_for] if addr]
                if self.app.config.PROXIES_COUNT == -1:
                    self._remote_addr = remote_addrs[0]
                elif len(remote_addrs) >= self.app.config.PROXIES_COUNT:
                    self._remote_addr = remote_addrs[-self.app.config.PROXIES_COUNT]

        return self._remote_addr
```

This is what I had in mind. But then again this is more of a cosmetic change.

sjsadowski: @harshanarayana I'm going to approve as-is, without this change since it's cosmetic.

PR #1541, opened 2019-03-27
stop number route accepting excess '.'s

cakemanny:
This stops this error

    ValueError: could not convert string to float: '12.34.56'

when passing 12.34.56 as a number route parameter argument.
By accepting ".12" and "12.", this is a non-breaking change. All valid
floats described by [0-9\.]+ are still accepted, just invalid ones are
now rejected.


sjsadowski:
fails tests across the board, probably due to the way the test is written. Please review here: https://travis-ci.org/huge-success/sanic/builds/512038048

PR #1546, opened 2019-04-06
Fixed a docstring typo and simplified code a little

krigar1184:


krigar1184:
> I get the swapping of `list()` for the comprehension. But why the change to `not args`?

This is intended to reduce nesting.

PR #1555, opened 2019-04-18
Fix watchdog reload worker repeatedly if there are multiple changed files

lexhung:
When there are multiple changed files (e.g. git pull, installing new version of a package, etc.), watchdog attempts to reload workers at first detected change and repeat for every changed file.

The patch allow gathering all changes into a single restart action.


lexhung:
> Thanks for the PR.
> 
> Why not break out of the loop when `need_reload` is `True`, instead of continuing to look at every file?

In my case, when I update a dependency package there are about 30 files being touched at the same time which trigger the app to reload 30 times continuously (breaking the loop there will only update the mtime of a single file). This patch allow all modified files' mtime to be updated before trigger reload.

It should have no big performance impact since it have to scan all files in the ordinary case when no file is modified.

ahopkins:
I guess... it does seem like it would potentially have a negative impact though. Especially for larger projects. I am hesitant to accept a change like this that could potentially have a large impact on a number of developers in their daily workflow.

I understand your issue and concern, but I would rather find a solution that does not also impact the existing operation.

ahopkins:
Instead, what if we make [`watchdog`](https://github.com/huge-success/sanic/blob/53f45810ffd38969cd9a24ad9d428ff1dee44378/sanic/app.py#L1093) a configurable param:

```app.run(watchdog=my_custom_watchdog)```

Tronic:
I might have a better look at the auto-reloader in the coming days; its killing of existing processes seems rather broken, and this patch is probably a step in the right direction.

Tronic:
This needs a few changes. First of all, multiprocessing.Process is used to run subprocess ... meaning a chain of two processes. After fixing that so that subprocess is used directly, it can also be terminated reliably, and we no longer need those hacks to kill all child processes.

I worked on this in the Trio branch, but in principle the changes should apply to asyncio implementation too. It appears to be working with single as well as multiple workers, as long as the master server kills its worker processes when it receives SIGTERM (dunno how Windows-compatible that is, might need another solution there).

My current revision: 
https://github.com/huge-success/sanic/blob/53f088e0cf3c5584ece8f61de5082970a9514b78/sanic/reloader_helpers.py

@lexhung could you give it a little bit of polish and testing on current Sanic master?

lexhung:
Hi @Tronic We may want to go through manual review for this to get merged.

I implemented the tests however it is quite tricky to get the patch coverage passed. These are parts that related to OS exceptions and it is not clear to me how to set that up without making the tests brittle.

Tronic:
@ahopkins Are you reviewing this?

ahopkins:
Yes. I am taking a look at it. I will respond with more detailed thoughts when I am back at my workstation and can fire up a few tests on this.

Tronic:
Looks like this is still stuck with no review. Even if the patch isn't quite the ideal solution we might have in our minds, I believe that it is definitely an improvement and it works on my machine, so I suggest for this to be merged as well.

ahopkins:
I am thinking this should be closed once #1827 is finished. Am I mistaken?

Tronic:
@ahopkins #1827 is already finished but it was a bit too late & scary change for 20.3 (needs review now). Closing in favour of #1827 which contains these changes as well.

PR #1560, opened 2019-04-20
Allow to disable Transfer-Encoding: chunked

andreymal:
The `Transfer-Encoding: chunked` is not allowed in some cases:

* HTTP/1.0

* a file streaming, because the `Content-Length` HTTP header is required here

This PR adds an option to disable it.

Closes https://github.com/huge-success/sanic/issues/1194

(I hope that `if self.chunked:` expressions don't affect the performance)

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -59,16 +59,23 @@ class StreamingHTTPResponse(BaseHTTPResponse):
         "status",
         "content_type",
         "headers",
+        "chunked",
         "_cookies",
     )
 
     def __init__(
-        self, streaming_fn, status=200, headers=None, content_type="text/plain"
+        self,
+        streaming_fn,
+        status=200,
+        headers=None,
+        content_type="text/plain",
+        chunked=None,
     ):
         self.content_type = content_type
         self.streaming_fn = streaming_fn
         self.status = status
         self.headers = CIMultiDict(headers or {})
+        self.chunked = chunked
         self._cookies = None
 
     async def write(self, data):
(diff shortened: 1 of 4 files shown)


sanic/response.py
```diff
@@ -59,16 +59,23 @@ class StreamingHTTPResponse(BaseHTTPResponse):
...
     )
 
     def __init__(
-        self, streaming_fn, status=200, headers=None, content_type="text/plain"
+        self,
+        streaming_fn,
+        status=200,
+        headers=None,
+        content_type="text/plain",
+        chunked=None,
```
seemethere: Wouldn't it make more sense in the way you're using this variable to just make it a boolean instead of originally setting it to `None`?

ahopkins: I agree.

And, also instead of this:

```chunked = self.chunked
        self.headers.pop("Content-Length", None)
        if chunked is None:
            chunked = version != "1.0"

        if chunked:
```

Why not something simpler without any variable assignment?

```
if self.chunked and version != "1.0"
```

ahopkins: I agree.

And, also instead of this:

```
        chunked = self.chunked
        self.headers.pop("Content-Length", None)
        if chunked is None:
            chunked = version != "1.0"

        if chunked:
```

Why not something simpler without any variable assignment?

```
if self.chunked and version != "1.0"
```

andreymal: @ahopkins `version != "1.0"` not possible because `version` is not available in the `write` method

andreymal: @ahopkins in general, I think there is a fundamental problem with `HTTPResponse` instances: they don't know an exact HTTP version (and some other info about the request) until the response is sent to the client. It forces me to use weird and complicated code like this assignment inside the `stream` method instead of `__init__`

andreymal: @seemethere @ahopkins ok I tried to simplify it https://github.com/huge-success/sanic/pull/1560/commits/4a7d30ee6c2937e5e6131616306448042254192e

andreymal: @seemethere @ahopkins ok I tried to simplify it https://github.com/huge-success/sanic/pull/1560/commits/7d6e60ab7dbe2ab92a451cf75bb8ca0a4eb69efb

PR #1562, opened 2019-04-23
Testing client

ahopkins:
New `test_client` based upon `request-async` for usage with #1475


ahopkins:
Yeah, I saw that last night after I saw the first test fail. I was thinking of taking a look at it and then hoping that (1) it was a simple fix like F strings and (2) @tomchristie would accept a PR to make it 3.5. If not, then we @huge-success/sanic-core-devs  have a decision to make.

tomchristie:
Great work @ahopkins!

The biggest 3.5 blocker will be async generators syntax. https://www.python.org/dev/peps/pep-0525/

`async for item in ...:`

It looks to me like sanic doens't yet use this syntax in any of it's API, since I can't see any (older style). `__aiter__` or `__anext__`, and since the streaming request/response docs use loop style APIs rather than async iterator styles.

It'd be a bit of a pain to backport to 3.5 for those. It's feasible, but it'd likely be something that I'd still want to drop as soon as possible.

Personally I'd probably advocate for this being a good time to consider a version with 3.6+ support, but that's for you folks to take a call on.

yunstanford:
I'm fine with dropping py3.5 support.

ahopkins:
I too would be in favor of dropping 3.5, but it does seem to account for a decent portion of our downloads. 

This is based off stats from 2019-01-01 thru today.

python_version | downloads |  
-- | -- | --
3.6 | 102,431 | 41%
3.7 | 77,564 | 31%
3.5 | 42,955 | 17%
null | 25,451 | 10%
2.7 | 2,426 | 1%
3.4 | 247 | 0%
3.8 | 49 | 0%
2.6 | 3 | 0%
TOTAL | 251,126 |  



```SELECT
  REGEXP_EXTRACT(details.python, r"([0-9]+\.[0-9]+)") AS python_version,
  COUNT(*) AS downloads
FROM TABLE_DATE_RANGE(
  [the-psf:pypi.downloads],
  TIMESTAMP("2019-01-01"),
  CURRENT_TIMESTAMP()
)
WHERE file.project="sanic"
GROUP BY python_version
ORDER BY downloads DESC```

**Comment by tomchristie** [comment]
Are you able to limit the results to only include downloads against the *current* sanic release?

That'd be more indicative of what Python versions are typically being used by new projects (rather than including deployments that may be pinned to older versions anyways).

**Comment by tomchristie** [comment]
(I'm arguing backwards there, because I think it'd be a good version jump time anyways, but hey. 😆)

**Comment by ahopkins** [comment]
When I filter for `18.12.0` and `19.3.1`, it does drop. You are right.


python_version | downloads |  
-- | -- | --
3.7 | 57852 | 40%
3.6 | 55204 | 38%
null | 17201 | 12%
3.5 | 12368 | 9%
2.7 | 1779 | 1%
3.4 | 151 | 0%
3.8 | 48 | 0%
2.6 | 2 | 0%
TOTAL | 144,605 |  

```
SELECT
  REGEXP_EXTRACT(details.python, r"([0-9]+\.[0-9]+)") AS python_version,
  COUNT(*) AS downloads
FROM TABLE_DATE_RANGE(
  [the-psf:pypi.downloads],
  TIMESTAMP("2019-01-01"),
  CURRENT_TIMESTAMP()
)
WHERE file.project="sanic"
AND (file.version="18.12.0" OR file.version="19.3.1")
GROUP BY python_version
ORDER BY downloads DESC
```

ahopkins:
See further discussion about dropping 3.5 support here: https://community.sanicframework.org/t/should-we-bump-the-minimum-python-required-version-to-3-6/238/6?u=ahopkins

andreymal:
Debian stable don't have py3.6+ yet, so I think this is not a good idea to drop py3.5

ahopkins:
@andreymal The current direction we are leaning towards is to drop 3.5 going forward, but extending our 18.12LTS support thru the [3.5 end of life](https://devguide.python.org/#status-of-python-branches) in September 2020.

seemethere:
Agree with @yunstanford, let's squash these commits then it's a LGTM for me as well.

ahopkins:
Will do. But there are a few more changes I want to make. Some tests use aiohttp directly and need to be refactored or else we will need both libraries, which seems bloated to me. Still WIP for current moment.

ahopkins:
@yunstanford and @seemethere This should be ready to roll assuming all of the tests pass.

ahopkins:
FYI - To help you in your review since everything is squashed now...

The files that changed since the last time are [`tests/test_keep_alive_timeout.py`](https://github.com/huge-success/sanic/pull/1562/files#diff-671cf29b7f90d893d12b72be23cb78db) and [`tests/test_request_timeout.py`](https://github.com/huge-success/sanic/pull/1562/files#diff-29bb84ec085816e2fc0ad70f8b2acca0), with a minor change to [`sanic/testing.py`](https://github.com/huge-success/sanic/pull/1562/files#diff-8ff5a7907a8a3e558aa2868b65263a87).

PR #1574, opened 2019-05-14
Fix Sanic.url_for method

### LINKED ISSUE #1420: url_for() doesn't return a working URI for a route with the trailing slash and strict_slashes=True
**Describe the bug**
Say we have a route with the trailing slash and `strict_slashes=True`:
`@app.route('/endpoint/', strict_slashes=True)`
this way, `url_for()` will always trim the trailing slash, returning a non-working URL.

**Code snippet**
```python
from sanic import Sanic
from sanic.response import html

app = Sanic()

# @app.route('/endpoint', strict_slashes=True)
@app.route('/endpoint/', strict_slashes=True)
async def endpoint(request):
    return html('OK')


@app.route('/')
async def index(request):
    # print(request.app.router.routes_all)
    return html('<a href="' + request.app.url_for('endpoint') + '">endpoint</a>')


if __name__ == '__main__':
    app.run(port=8000)
```

**Expected behavior**
In my example, `request.app.url_for('endpoint')` is expected to return `'/endpoint/'`.

 - Version 0.8.3

**Additional context**
The current implementation of `url_for()` just blindly removes the trailing slash, see [app.py, lines 617-618](https://github.com/huge-success/sanic/blob/7c182f63c8aee1d6cb31aabf6408595b195685ae/sanic/app.py#L617-L618).

5onic:
This is PR fixed #1420 issue.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -783,9 +783,6 @@ class Sanic:
 
                 uri = "{}/{}".format(folder_, filename)
 
-        if uri != "/" and uri.endswith("/"):
-            uri = uri[:-1]
-
         out = uri
 
         # find all the parameters we will need to build in the URL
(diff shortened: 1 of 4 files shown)


ahopkins:
See #1591

ahopkins:
Is this still necessary?

Issue #1577, opened 2019-05-15
Testing fails with ModuleNotFoundError: No module named 'aiohttp'


logileifs:
**Describe the bug**
After installing sanic and setting up tests like in the [documentation](https://sanic.readthedocs.io/en/latest/sanic/testing.html) and running them sanic fails with `ModuleNotFoundError: No module named 'aiohttp'`.

It succeeds after installing 'aiohttp' manually

**Code snippet**
Just do a clean install of sanic and create a test as in the documentation and try running it.

**Expected behavior**
If 'aiohttp' is a hard dependency for testing it should be included when installing sanic

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04
 - Version sanic==19.3.1

**Stack trace**
```
Traceback (most recent call last):
  File "/home/logi/.virtualenvs/lemo3.6/lib/python3.6/site-packages/sanic/testing.py", line 97, in _collect_response
    method, url, *request_args, **request_kwargs
  File "/home/logi/.virtualenvs/lemo3.6/lib/python3.6/site-packages/sanic/testing.py", line 20, in _local_request
    import aiohttp
ModuleNotFoundError: No module named 'aiohttp'
```

harshanarayana:
@logileifs Can you please add the stack trace to the bug description?

logileifs:
`[2019-05-15 10:02:59 +0000] [8698] [INFO] Goin' Fast @ http://127.0.0.1:42101
[2019-05-15 10:02:59 +0000] [8698] [ERROR] Exception
Traceback (most recent call last):
  File "/home/logi/.virtualenvs/lemo3.6/lib/python3.6/site-packages/sanic/testing.py", line 97, in _collect_response
    method, url, *request_args, **request_kwargs
  File "/home/logi/.virtualenvs/lemo3.6/lib/python3.6/site-packages/sanic/testing.py", line 20, in _local_request
    import aiohttp
ModuleNotFoundError: No module named 'aiohttp'
[2019-05-15 10:02:59 +0000] [8698] [INFO] Starting worker [8698]
[2019-05-15 10:02:59 +0000] [8698] [INFO] Stopping worker [8698]
[2019-05-15 10:02:59 +0000] [8698] [INFO] Server Stopped`

harshanarayana:
`aiohttp` is a `test` only package that doesn't get installed by default unless explicitly asked for. 

```bash
# Install sanic with test utilities
$ pip3 install "sanic[test]"
```

[...]

PR #1579, opened 2019-05-16
Update docs regarding sanic[test] and point extensions to awesome-sanic

ahopkins:


ahopkins:
See #1577

docs/sanic/getting_started.md
```diff
@@ -25,6 +25,12 @@ conda config --add channels conda-forge
 conda install sanic
 ```
 
+**IMPORTANT**: The `SanicTestClient` requires some additional dependencies. Therefore, it is **recommended** that you install Sanic in your development environment this way:
```
andreymal: Why is it here? The sanic installation for a end user (system administrator) is not related to development and testing.

ahopkins: I thought it was pretty self-explanatory. If I am developing something, I would go to **Getting Started**. If there is something I should do differently for my dev environment, it makes sense here, no?

Do you have another suggestion? It is a sort of *gotcha* that is not obvious at first sight.

andreymal: >Do you have another suggestion?

Yes of course — `docs/sanic/testing.md` :) This is an obvious place for this information.

ahopkins: [Its there too](https://github.com/huge-success/sanic/pull/1579/files#diff-e5a1a9c6467ae061d903d8d3abf6ad8e)

andreymal: I know, my question is about `gettins_started.md` :) Because it's not a required part to getting started (unless TDD is used)

yunstanford: agree, it probably should not need this as user instead of contributor

yunstanford: This should live in contributing/testing page

ahopkins: Not meant to be towards contributors. But as a user, I may want it to run unit tests. If we do indeed mean for it to **only** be used for contribution, then I would like to revisit the earlier conversation a out whether it belongs in sanic-core at all as opposed to a separate package.

yunstanford: unit tests folder is not packaged and published along with sanic modules. If any user want to run unit tests, `pip install sanic[test]` will not work anyway. They should `git clone sanic-repo` and then install those test deps for running unit tests.

yunstanford: I may misunderstand the context if you're referring to unit tests belongs to user's own repo.

In any case, i'm not sure if that should be in `get_started` page tho.

ahopkins: yes. I was referring to that, not the sanic unit tests.

I'll make the change.

ahopkins:
Closing because of #1595

PR #1600, opened 2019-06-04
Fix #1587: add support for handling Expect Header

### LINKED ISSUE #1587: Request Streaming is extremely slow
**Describe the bug**
I created a simple POC app to accept streaming binary data. The syntax is compliant with the official docs. The app works as expected but the response time is more than 1000 ms for even small data sizes (1.1K). In the snippet below tried uploading 2 files - `tiny.txt` (57B) and `small.txt` (1137B). File `tiny.txt` took 0.021s and `small.txt` took 1.026s on average. Testing was done on the same host via a loopback, so no network delay involved. I think the issue is cause by Sanic not responding with 100-continue so the client wastes time waiting for it.

**Code snippet**
Source code:
```
from sanic import Sanic
from sanic.response import stream, text

app = Sanic('request_stream')

async def detect_handler(request):
    result = bytes()
    while True:
        body = await request.stream.read()
        if body is None:
            break
    return text('done waiting!')

app.add_route(detect_handler, '/api/v1/detect', methods=['POST'], stream=True)

if __name__ == '__main__':
```

curl testing with Sanic:
```
$ time curl -vvv -H "Transfer-Encoding: chunked"  --data-binary @small.txt  http://127.0.0.1:8000/api/v1/detect
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> POST /api/v1/detect HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: curl/7.54.0
> Accept: */*
> Transfer-Encoding: chunked
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
>
* Done waiting for 100-continue
< HTTP/1.1 200 OK
< Connection: keep-alive
< Keep-Alive: 5
< Content-Length: 13
< Content-Type: text/plain; charset=utf-8
<
* Connection #0 to host 127.0.0.1 left intact
done waiting!
real	0m1.028s
```

curl testing with a similar Golang/go-chi based app that returns 100-continue:
```$ time curl -vvv -H "Transfer-Encoding: chunked"  --data-binary @big41K.wav  http://127.0.0.1:8000/detect
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> POST /detect HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: curl/7.54.0
> Accept: */*
> Transfer-Encoding: chunked
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
>
< HTTP/1.1 100 Continue
< HTTP/1.1 200 OK
< Date: Wed, 22 May 2019 22:40:40 GMT
< Content-Length: 38
< Content-Type: text/plain; charset=utf-8
<
{
    "predictions": [0.911111]
* Connection #0 to host 127.0.0.1 left intact
}
real	0m0.040s
```

**Expected behavior**
I'd expect all small POSTs to take less than 50 ms

 - OS: MacOS High Sierra 10.13.6, Python 3.7.3
 - Sanic Version 19.3.1

yunstanford:
Fix https://github.com/huge-success/sanic/issues/1587

I've verified the fix with code-snippet posted in the issue page, and also added unit tests for verifying.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -15,6 +15,7 @@ from httptools.parser.errors import HttpParserError
 from multidict import CIMultiDict
 
 from sanic.exceptions import (
+    HeaderExpectationFailed,
     InvalidUsage,
     PayloadTooLarge,
     RequestTimeout,
@@ -22,7 +23,7 @@ from sanic.exceptions import (
     ServiceUnavailable,
 )
 from sanic.log import access_logger, logger
-from sanic.request import Request, StreamBuffer
+from sanic.request import Request, StreamBuffer, EXPECT_HEADER
 from sanic.response import HTTPResponse
 
 
@@ -314,6 +315,10 @@ class HttpProtocol(asyncio.Protocol):
         if self._keep_alive_timeout_handler:
             self._keep_alive_timeout_handler.cancel()
             self._keep_alive_timeout_handler = None
+
+        if self.request.headers.get(EXPECT_HEADER):
+            self.expect_handler()
+
         if self.is_request_stream:
             self._is_stream_handler = self.router.is_stream_handler(
                 self.request
@@ -323,6 +328,17 @@ class HttpProtocol(asyncio.Protocol):
                     self.request_buffer_queue_size
                 )
                 self.execute_request_handler()
+
+    def expect_handler(self):
+        """
+        Handler for Expect Header.
+        """
+        expect = self.request.headers.get(EXPECT_HEADER)
+        if self.request.version == "1.1":
+            if expect.lower() == "100-continue":
+                self.transport.write(b"HTTP/1.1 100 Continue\r\n\r\n")
+            else:
+                self.write_error(HeaderExpectationFailed("Unknow Expect: {expect}".format(expect=expect)))
 
     def on_body(self, body):
         if self.is_request_stream and self._is_stream_handler:
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -340,7 +340,7 @@ class HttpProtocol(asyncio.Protocol):
             else:
                 self.write_error(
                     HeaderExpectationFailed(
-                        "Unknow Expect: {expect}".format(expect=expect)
+                        "Unknown Expect: {expect}".format(expect=expect)
                     )
                 )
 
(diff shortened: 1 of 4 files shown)


sanic/server.py
```diff
@@ -324,6 +329,21 @@ def on_headers_complete(self):
...
+        Handler for Expect Header.
+        """
+        expect = self.request.headers.get(EXPECT_HEADER)
+        if self.request.version == "1.1":
+            if expect.lower() == "100-continue":
+                self.transport.write(b"HTTP/1.1 100 Continue\r\n\r\n")
+            else:
+                self.write_error(
+                    HeaderExpectationFailed(
+                        "Unknow Expect: {expect}".format(expect=expect)
```
andreymal: Unknown?

tests/test_request_stream.py
```diff
@@ -40,6 +42,53 @@ def get(self, request):
...
+                if body is None:
+                    break
+                result += body.decode("utf-8")
+            return text(result)
+
+    app.add_route(SimpleView.as_view(), "/method_view")
+
+    assert app.is_request_stream is True
+
+    request, response = app.test_client.post("/method_view", data=data, headers={"EXPECT": "100-continue"})
```
ahopkins: Maybe parameterize the test with different capitalization and variation of the header?

yunstanford: Sounds good

sanic/server.py
yunstanford: Good catch

seemethere:
Can we get a more descriptive PR title?

yunstanford:
@seemethere added brief description.

tests/test_request_stream.py
```diff
@@ -40,6 +42,38 @@ def get(self, request):
...
+    assert app.is_request_stream is True
+
+    if not expect_raise_exception:
+        request, response = app.test_client.post("/method_view", data=data, headers={"EXPECT": "100-continue"})
+        assert response.status == 200
+        assert response.text == data
+    else:
+        with pytest.raises(ValueError) as e:
+            app.test_client.post("/method_view", data=data, headers={"EXPECT": "100-continue-extra"})
+            assert "Unknow Expect: 100-continue-extra" in str(e)
```
andreymal: Sorry for the flood but... Unknown? :)

PR #1612, opened 2019-06-23
Add bandit code static analyzer for security.

c-goosen:
This is an additional coverage on testing for static analysis of code.

This is useful for general python security mistakes made during development.

Bandit project is described as:
"Bandit is a tool designed to find common security issues in Python code. To do this, Bandit processes each file, builds an AST from it, and runs appropriate plugins against the AST nodes. Once Bandit has finished scanning all the files, it generates a report."

Added to tox.ini a security environment.

Also updated travis.yml

Open for suggestions.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py36, py37, {py36,py37}-no-ext, lint, check
+envlist = py36, py37, {py36,py37}-no-ext, lint, check, security
 
 [testenv]
 usedevelop = True
@@ -31,10 +31,11 @@ deps =
     flake8
     black
     isort
+    bandit
 
 commands =
     flake8 sanic
-    black --config ./.black.toml --check --verbose sanic
+    black --config ./.black.toml --check --verbose sanic/
     isort --check-only --recursive sanic
 
 [testenv:check]
@@ -47,3 +48,10 @@ commands =
 [pytest]
 filterwarnings =
     ignore:.*async with lock.* instead:DeprecationWarning
+
+[testenv:security]
+deps =
+    bandit
+
+commands =
+    bandit --recursive sanic --skip B404,B101 --exclude sanic/reloader_helpers.py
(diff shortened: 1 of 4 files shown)


c-goosen:
@yunstanford will have a look and update.

c-goosen:
@yunstanford better commit now?

tox.ini
```diff
@@ -31,10 +31,11 @@ deps =
     flake8
     black
     isort
+    bandit
 
 commands =
     flake8 sanic
-    black --config ./.black.toml --check --verbose sanic
+    black --config ./.black.toml --check --verbose sanic/
```
ahopkins: Why the trailing `/` here?

tox.ini
```diff
@@ -47,3 +48,10 @@ commands =
 [pytest]
 filterwarnings =
     ignore:.*async with lock.* instead:DeprecationWarning
+
+[testenv:security]
+deps =
+    bandit
+
+commands =
+    bandit --recursive sanic --skip B404,B101 --exclude sanic/reloader_helpers.py
```
ahopkins: What is the thought behind skipping this file?

c-goosen: So that file uses subprocess, which will make the build fail. 

I can add multiple #nosec comments, but this caused issues with black formatter.

I don't want to blanket exclude the specific error in case it pops up anywhere else.

Any suggestions @ahopkins

tox.ini
c-goosen: Just a force of habit, can remove

tox.ini
ahopkins: Nope. Let's leave it out for now. That file is in some serious need of attention anyway. No need to tackle it here.

c-goosen: Great. I would be willing to have a look at that file if you have any pointers of where to begin. I've had some idea, but don't want to run down a rabbit hole.

c-goosen:
Great! Thanks for the tips and quick turn around time. Hope to contribute lost more! @ahopkins @yunstanford .

Just an FYI, out of the last 3 projects I've been contributing to, the tox/pytest setups you have is really impressive. Learnt a lot from working with it.

PR #1625, opened 2019-07-08
fix: GIT-1623: handle cookie duplication and serialization issue

### LINKED ISSUE #1623: using AGSI the Set-Cookie header is send twice, one correct, one wrong
**Describe the bug**
When using a AGSI server (unicorn) the Set-Cookie header is send twice, one time correctly formatted, one time formatted as a json dict. See curl response below.

```
HTTP/1.1 200 OK
date: Sat, 06 Jul 2019 20:01:13 GMT
server: uvicorn
set-cookie: {'path': '/', 'httponly': True}
content-length: 36
set-cookie: test="It worked!"; Path=/; HttpOnly
```

**Code snippet**
```
import logging

from sanic import Sanic
from sanic.response import json, redirect, text, html, stream, raw

logger = logging.getLogger(__name__)
app = Sanic(__name__)

@app.get("/test", name='api')
async def cookie_test(request):
    response = text("There's a cookie up in this response")
    response.cookies['test'] = 'It worked!'
    response.cookies['test']['httponly'] = True
    
    logger.info(response.headers)
    
    return response

def main():
    uvicorn.run(app, host="0.0.0.0", port=5000, debug=True, access_log=False)

```

**Expected behavior**
The first Set-Cookie header should not be send.
```
HTTP/1.1 200 OK
date: Sat, 06 Jul 2019 20:01:13 GMT
server: uvicorn
content-length: 36
set-cookie: test="It worked!"; Path=/; HttpOnly
```

 - OS: Lunix
 - Version: 19.6.0

**Additional context**
The problem seems to happen when the headers are turned into  bytes in the sanic AGSI module. The Cookies are already in de response headers because the CookieJar is adding the cookies to the headers. When `str()` is called on the Cookie object the object in converted to a json string.

[L293-L296](https://github.com/huge-success/sanic/blob/master/sanic/asgi.py#L293-L296)

**A possible solution**
First asume that the header value is a string or a other object with a `encode` function if not convert it to a string. This is similair to the `_parse_headers` function in the response class. 

[_parse_headers (L32-L46)](https://github.com/huge-success/sanic/blob/master/sanic/response.py#L32-L46)

If this is implemented it won't be needed to add the cookies after the headers are converted to bytes. the lines below can be removed.

[L321-L327](https://github.com/huge-success/sanic/blob/master/sanic/asgi.py#L321-L327)

harshanarayana:
Fixes #1623 

```bash
➜ http :5000/test
HTTP/1.1 200 OK
content-length: 36
date: Mon, 08 Jul 2019 07:38:40 GMT
server: uvicorn
set-cookie: test="It worked!"; Path=/; HttpOnly
set-cookie: testing="It worked!"; Path=/
```

### DIFF
--- a/sanic/asgi.py
+++ b/sanic/asgi.py
@@ -1,7 +1,6 @@
 import asyncio
 import warnings
 
-from http.cookies import SimpleCookie
 from inspect import isawaitable
 from typing import Any, Awaitable, Callable, MutableMapping, Union
 from urllib.parse import quote
@@ -288,11 +287,22 @@ class ASGIApp:
         """
         Write the response.
         """
-
+        headers = []
+        cookies = dict()
         try:
-            headers = [
+            cookies = {
+                v.key: v
+                for _, v in list(
+                    filter(
+                        lambda item: item[0].lower() == "set-cookie",
+                        response.headers.items(),
+                    )
+                )
+            }
+            headers += [
                 (str(name).encode("latin-1"), str(value).encode("latin-1"))
                 for name, value in response.headers.items()
+                if name.lower() not in ["set-cookie"]
             ]
         except AttributeError:
             logger.error(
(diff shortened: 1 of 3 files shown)


sanic/asgi.py
```diff
@@ -288,11 +287,22 @@ def __init__(self) -> None:
         """
         Write the response.
         """
-
+        headers = []
+        cookies = dict()
```
yunstanford: nit: `{}`

harshanarayana: @yunstanford Sure. I will update the PR later today and push the changes.

harshanarayana: @yunstanford Done

PR #1632, opened 2019-07-13
feature: GIT-1631 enable towncrier

### LINKED ISSUE #1631: RFC: Enable Towncrier based change log support
Enable usage of `towncrier` to generate and manage the release change logs automatically.

harshanarayana:
Closes #1631 

## Features
1. Enable `towncrier` support for changelog generation
2. Add Documentation test support via `travis`

## Enhancements
1. Additional make command to generate Documentation and test documentation changes
2. Fix Documentation deprecation warnings and other minor documentation issues
3. Have a common single `CHANGELOG` file by referencing the top-level file in the documentation
4. Have a common single `CONTRIBUTING` file by referencing the top-level file in the documentation
5. `make release` and`make changelog` command support

## Notes
1. Going forward, the auto-generated changelog will have a clean format. To adhere to that standard the changelog for last two releases have been converted to match the same format and the rest of the items have been retained in their previous format.


harshanarayana:
@huge-success/sanic-core-devs This involves a bunch of changes to make our life easier during the release cycle. No more editing document in multiple places. The changelog is a common file for both Git and document now and so is Contribution guidelines.

harshanarayana:
> i believe there may be a problem with builds on Travis by having both a `setup.py` and a `pyproject.toml`

Interesting. Do you happen to remember what was the problem? It worked fine when I did my tests. Wonder if I missed some corner cases that might have to be looked into in case if having `pypropject.toml` alongside the `setup.py` causes a problem.


> How will this change our contribution expectations?

The idea is that every PR will have a file under the `./changelogs` directory to indicate the purpose of the PR and a detailed explanation on what it is about. I am working on a GithubApp to enforce this check and we can make it part of the repo so that each PR is automatically validated against that criteria. The contribution guidelines page has already been updated as part of this PR to indicate the same about having a file under the `./changelogs` in each PR you open.

harshanarayana:
@ahopkins  I think the `[build-system]` tag in the `pyproject.toml` doesn’t work well with others.

ahopkins:
@harshanarayana I do not. But I think we could probably find the errors on Travis. `black` uses `[tool.black]`, so... not sure why. Hang on, I will take a look back.

https://github.com/python/black#configuration-format

ahopkins:
Check out [Build 2593](https://travis-ci.org/huge-success/sanic/builds/525713346)

> ERROR: Error installing 'file:///home/travis/build/huge-success/sanic': editable mode is not supported for pyproject.toml-style projects. pip is processing this project as pyproject.toml-style because it has a pyproject.toml file. Since the project has a setup.py and the pyproject.toml has no "build-backend" key for the "build_system" value, you may pass --no-use-pep517 to opt out of pyproject.toml-style processing. See PEP 517 for details on pyproject.toml-style projects.

harshanarayana:
@ahopkins Never mind, I wrote a small script as a wrapper that will take care of this problem.  Instead of running `towncrier` to generate the logs, now you can run the `./scripts/changelog.py` which is a blanket wrapper to `towncrier` and moved the `pyproject.toml` out from top level directory into the `scripts` path. This should take care of our issues in `travis`

scripts/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[tool.towncrier]
+    package = "sanic"
```
yunstanford: I don't think .toml need indentations..

scripts/release.py
```diff
@@ -301,4 +314,5 @@ def release(args: Namespace):
         required=False,
     )
     args = cli.parse_args()
-    release(args)
+    with Directory() as _:
```
yunstanford: you don't need `as _`

sanic/__init__.py
```diff
@@ -1,7 +1,6 @@
+from sanic.__version__ import __version__
 from sanic.app import Sanic
 from sanic.blueprints import Blueprint
 
 
-__version__ = "19.6.2"
```
yunstanford: Why not just keep it here ?

scripts/release.py
harshanarayana: @yunstanford Force of habit. Fixing it in the next commit I will push today or tomorrow.

scripts/pyproject.toml
harshanarayana: @yunstanford Yes, it does not. However, I personally believe that indenting thigs make it more readable and clean to look at. I can always drop the indentation if you think that should go away

sanic/__init__.py
harshanarayana: Moving it to a dedicated `__version__` file can be handy for a lot of tooling down the line if required. It can be useful if we want to automate the process of generating the version update as well.

harshanarayana: Working on adding a `make release` support that will take care of everything except for creating a PR and tagging it. (Creating a PR is a possible addition I will look into)

harshanarayana:
@yunstanford Changes addressed. I also added the`make release` and `make changelog` command.

`make release` already combines the `make changelog` behavior. So the `make release` . will bump up the versions in all applicable places and then generate the changelog and commit the changes into the branch.

PR #1634, opened 2019-07-16
Abort documentation

SakuraSound:
During development, saw that the abort example was missing the return statement, so I made a change to that. Also in the API docs, I made a quick update to make more clear that abort does have a return object.


SakuraSound:
Closed because I realized my assumptions weren't completely correct

Issue #1636, opened 2019-07-19
RFC: Forwarded header and other proxying improvements


Tronic:
The new Forwarded header provides proxied remote_addr, host etc. more reliably than the X-Forwarded headers do. I am currently implementing this into Sanic following the article of Nginx folks https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/

In particular, I am using the secret="..." parameter to identify the correct and true proxy, instead of PROXIES_COUNT.

So far I have a parser that I've made into a property of the request class. It takes about 8 microseconds to parse an average header, returning a dict of key-value pairs for the matched proxy; I had to use the re module because parse_header cannot handle the comma-delimited format of the header.

The next step would be to modify request properties server_name, server_port, remote_addr and scheme to make use of this header but I'd like to hear your comments on the matter before making a huge pull request.

Ideally the Forwarded header would only be parsed once, and not separately for each of the properties -- this is something that I don't see being done with the current code, which determines the X-Forwarded etc. sources separately on each property call. I believe that there should be a single decision for how to source this information, i.e. pick one of, in order of precedence:

1. config.SERVER_NAME
2. Forwarded (if config.FORWARDED_SECRET is set?)
3. X-Forwarded
4. HTTP Host (host, port)
5. config.SERVER_NAME_FALLBACK
6. socket/ssl scheme/host/port

UPDATED PLAN:  (instead of the above list)

1. Any proxied information i.e. request.forwarded (Forwarded *or* X-headers)
2. If SERVER_NAME is defined, use any information in it to fill in missing scheme, hostname, port and path
3. If no proto/scheme specified yet, infer from Websocket and SSL status
-- if request.forwarded: we have been proxied: stop here.
4. Add remote_addr from socket
5. Use Host header (no port means 80 or 443 depending on scheme)
6. If there is no Host, use host and port from socket

Additionally url_for may change scheme between http* and ws* depending on target handler. If path (i.e. app root) is found in 1-2, it gets appended in front of all paths, even for internal URLs.

Use cases:

- A proxy such as nginx fills in either set of forwarding headers, and in that case any local information won't be useful but server admin may choose to use config.SERVER_NAME to provide parts of the information not necessarily available on headers (proxies and the specific headers that they set are often out of admin control).

[...]

PR #1638, opened 2019-07-20
Forwarded headers and otherwise improved proxy handling

Tronic:
This is a work in progress regarding issue #1636.


ahopkins:
@Tronic Thanks for clearly all the hard work you have put in. I apologize for myself that I have not reviewed sooner, I know how frustrating it can be. Summer craziness. I plan to give this the time it deserves later this week.

Tronic:
This uses quoted-string escaping as specified in RFC, which is apparently mostly ignored by all existing software. Does anyone have examples of software using forwarded or x-forwarded headers  that escape or don't escape quotes in their quoted strings?

Additionally, I'll migrate all this code from forwarded.py to the now-merged headers.py, with the intent of collecting all header parsing utility functions there.

Tronic:
I'd like to have this merged. A number of pending bugfixes to request properties will conflict with this if done on current master, and I don't wish to add more work on this already bloated pull request.

harshanarayana:
@sjsadowski Will get it reviewed today.

sanic/request.py
```diff
@@ -372,86 +379,70 @@ def _get_address(self):
...
 
         :return: the server name without port number
         :rtype: str
         """
-        return (
-            self.app.config.get("SERVER_NAME")
-            or self.headers.get("x-forwarded-host")
-            or self.host.split(":")[0]
-        )
+        server_name = self.app.config.get("SERVER_NAME")
```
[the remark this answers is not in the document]
Tronic: I gather that SERVER_NAME is supposed to be a rarely specified override, not a fallback. Not sure how useful that actually is. One use would be to actively _prevent_ any headers being used for URL construction, but this is also limited in that it cannot handle the same Sanic instance serving multiple hosts. Also worth noting is that this is supposed to be in the same format as Host, and thus includes any non-default port.

Another consideration is what to do if none of these match. One could possibly still match against SSL SNI but if that is missing, I would not go further. In particular, sockname (ip/port or unix socket path) should not be exposed in external URLs. OTOH, all reasonable clients include Host or its HTTP/2 equivalent :authority header.

sanic/headers.py
```diff
@@ -24,14 +24,131 @@ def parse_content_header(value: str) -> typing.Tuple[str, Options]:
...
+
+
+def parse_forwarded(headers, config):
+    """Parse HTTP Forwarded headers.
+    Accepts only the last element with secret=`config.FORWARDED_SECRET`
+    :return: dict with matching keys (lowercase) and values, or None.
+    """
+    header = headers.getall("forwarded", None)
+    secret = config.FORWARDED_SECRET
+    if header is None or not secret:
```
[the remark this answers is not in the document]
Tronic: The alternative is to use PROXIES_COUNT, as is done with X-Forwarded headers in the existing as well as the new implementation. If this same setting was applied to Forwarded as well, any existing installed applications using X-Forwarded would instantly become exploitable (send a request with fake Forwarded header), so if we wanted to use that, we'd have to add a new config option for that.

Using a secret key, as suggested by Nginx folks, solves the entire issue in a smart and secure manner. This key is not part of RFC but such extensions are explicitly allowed by https://tools.ietf.org/html/rfc7239#section-5.5

tests/test_requests.py
```diff
@@ -401,6 +401,115 @@ def test_content_type(app):
...
+def test_standard_forwarded(app):
+    @app.route("/")
+    async def handler(request):
+        return json(request.forwarded)
+
+    # Without FORWARDED_SECRET, X-headers should be respected
+    app.config.PROXIES_COUNT = -1
+    headers = {
+        "Forwarded": (
+            'for=1.1.1.1, for=injected;host="'
```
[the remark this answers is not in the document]
Tronic: Each proxy will append its entry to the end, comma separated or as an additional header row. If proxies are not careful in examining the prior entries for correct syntax (and you can be sure that many of them will just blindly append to anything), malformed results like this may occur. We mitigate such problems by:
- Parsing right-to-left, starting with the more trustworthy most-local proxies.
- Parse errors clear any keys found and skip to the next comma (so that malicious for=injected is not combined with secret=mySecret added by the true proxy and accidentally accepted).

tests/test_requests.py
```diff
@@ -401,6 +401,115 @@ def test_content_type(app):
...
+    headers = {"Forwarded": 'for=test;quoted="\\,x=x;y=\\";secret=mySecret'}
+    request, response = app.test_client.get("/", headers=headers)
+    assert response.json == {
+        "for": "test",
+        "quoted": '\\,x=x;y=\\',
+        "secret": "mySecret"
+    }
+
+    # Secret insulated by malformed field #1
+    headers = {"Forwarded": r'for=test;secret=mySecret;b0rked;proto=wss;'}
```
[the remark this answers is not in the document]
Tronic: Good catch. I should learn to avoid Python's raw strings better anyway (they are broken by design, with things like r"c:\windows\").

tests/test_requests.py
```diff
@@ -401,6 +401,115 @@ def test_content_type(app):
...
+    request, response = app.test_client.get("/", headers=headers)
+    assert response.json == {"for": "test", "secret": "mySecret"}
+
+    # Secret insulated by malformed field #2
+    headers = {"Forwarded": r'for=test;b0rked;secret=mySecret;proto=wss'}
+    request, response = app.test_client.get("/", headers=headers)
+    assert response.json == {"proto": "wss", "secret": "mySecret"}
```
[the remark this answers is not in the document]
Tronic: Parse errors clear any keys found and then continue on the next (to left) correctly formed option as if it was another comma-separated element. This insulation is important because we cannot know if there was supposed to be a comma within what the parse error made us skip. Other options would be:
- Keep skipping until the next comma (somewhat harder to implement in this approach)
- Ignore everything on the left side and quit parsing

The current solution is more resilient against accidental errors and I don't see security issues with it.

sanic/headers.py
```diff
@@ -24,14 +24,131 @@ def parse_content_header(value: str) -> typing.Tuple[str, Options]:
...
+        pos = m.end()
+        val_token, val_quoted, key, sep = m.groups()
+        key = key.lower()[::-1]
+        val = (val_token or val_quoted.replace('"\\', '"'))[::-1]
+        ret[key] = val
+        if secret is not True and key == "secret" and val == secret:
+            secret = True
+        if secret is True and sep != ";":
+            return normalize(ret)
+    return normalize(ret) if secret is True else None
```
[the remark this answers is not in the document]
Tronic: Trying to use backticks in a git commit -m "..." is a bad idea. That was supposed to say "add `found` to avoid re-using the `secret` variable".

Tronic:
I am updating the docs right now.

It is becoming ever more clear that PROXIES_COUNT should default to 0 to avoid a security hole (clients can easily fake their IPs). It has not been previously fixed because of backward compatibility concerns that affect services running behind proxies using the default value. If the default was changed, such services would see their proxy as client IP, which is actually less dangerous than allowing the client to specify ANY IP. Also, they would begin to see the Host header as server_name, instead of X-Forwarded-Host. This may or may not change anything, depending on how their proxy is configured.

My recommendation is to change PROXIES_COUNT default to 0 (disable X-Forwarded-*, X-Real-IP etc), and to remove the feature that allows negative values (because it is always subject to client spoofing). Proxy headers would then be used *only if configured* and only in a secure manner.

Request for comment: how to proceed with this?

sanic/request.py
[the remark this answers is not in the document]
harshanarayana: @abuckenheimer  I think `SERVER_NAME` here should be fine. I see how it might be a bit tricky to look at but this is generally never used like @Tronic mentioned. At least in all of my work so far, never had to use a parameter like this to override anything. 

As for the SSL SNI goes, do we really need to do that in the core of `sanic` at all ? Considering it's a client supported entity and depending on the browsers you are using it might or might not work. I think at this point, we might be better off not worrying about the SNIs. This is just a personal opinion.

sanic/headers.py
```diff
@@ -24,14 +24,132 @@ def parse_content_header(value: str) -> typing.Tuple[str, Options]:
...
+            return normalize(ret)
+    return normalize(ret) if found else None
+
+
+def parse_xforwarded(headers, config):
+    """Parse X-Real-IP, X-Scheme and X-Forwarded-* headers."""
+    proxies_count = config.PROXIES_COUNT
+    if not proxies_count:
+        return None
+    h1, h2 = config.REAL_IP_HEADER, config.FORWARDED_FOR_HEADER
```
harshanarayana: Can we please use a better name for the variables instead of `h1` and `h2`? Sorry for being so picky.

sanic/headers.py
```diff
@@ -24,14 +24,132 @@ def parse_content_header(value: str) -> typing.Tuple[str, Options]:
...
+    addr = h1 and headers.get(h1)
+    forwarded_for = h2 and headers.getall(h2, None)
+    if not addr and forwarded_for:
+        assert proxies_count == -1 or proxies_count > 0, config.PROXIES_COUNT
```
harshanarayana: Would it be a good idea to have a bit of a clearly defined error message with the assert ?

Or would it be easier to handle it the way the `addr` extraction is handled instead? Return `None` even if the `proxies_count` matching fails.

sanic/headers.py
```diff
@@ -24,14 +24,132 @@ def parse_content_header(value: str) -> typing.Tuple[str, Options]:
...
+            ("proto", "x-forwarded-proto"),  # Overrides X-Scheme if present
+            ("host", "x-forwarded-host"),
+            ("port", "x-forwarded-port"),
+            ("path", "x-forwarded-path"),
+        )
+    )
+    return normalize({"for": addr, **{k: v for k, v in other if v}})
+
+
+_ipv6 = "(?:[0-9A-Fa-f]{0,4}:){2,7}[0-9A-Fa-f]{0,4}"
```
harshanarayana: We might need to make this regex a bit stronger if we are to rely on the regex.  (I am sure you have thought of this already) This would be a good to have. Otherwise, we might as well use the `ipaddress` module itself. 

xref: http://vernon.mauery.com/content/projects/linux/ipv6_regex

sanic/headers.py
```diff
@@ -24,14 +24,132 @@ def parse_content_header(value: str) -> typing.Tuple[str, Options]:
...
+
+def bracketv6(addr):
+    return f"[{addr}]" if _ipv6_re.fullmatch(addr) else addr
+
+
+def normalize(fwd: dict) -> dict:
+    """Normalize and convert values extracted from forwarded headers.
+    Modifies fwd in place and returns the same object.
+    """
+    if "proto" in fwd:
```
harshanarayana:
```python
def handle_host(forward):
    host, port = parse_host(forward["host"])
    if host:
        forward["host"] = host
        if port:
            forward["port"] = port
    else:
        del forward["host"]

actions = {
    "proto": {
        "handler": None,
    },
    "by": {
        "handler": bracketv6
    },
    "for": {
        "handler": bracketv6
    },
    "port": {
        "handler": handle_port
    },
    "host": {
        "handler": handle_host
    }
}

for key in fwd:
    try:
        actions[key]["handler"](fwd)
    except:
        fwd.pop("port", None)
```
Can you please check if it make sense to have this section defined this way without all the `if` checks?

sanic/request.py
```diff
@@ -372,86 +379,70 @@ def _get_address(self):
...
-                    self._remote_addr = remote_addrs[0]
-                elif len(remote_addrs) >= self.app.config.PROXIES_COUNT:
-                    self._remote_addr = remote_addrs[
-                        -self.app.config.PROXIES_COUNT
-                    ]
-                else:
-                    self._remote_addr = ""
-            else:
-                self._remote_addr = ""
+            self._remote_addr = self.forwarded.get("for") or ""
```
harshanarayana: 💯

tests/test_requests.py
```diff
@@ -401,6 +401,115 @@ def test_content_type(app):
...
+        "proto": "wss",
+        "by": "[bad::f00d]",
+        "for": "[cafe::1]",
+        "host": "a",
+        "port": 2,
+        "secret": "mySecret"
+    }
+
+    # Field normalization #2 (remove malformed host and keep separate port)
+    headers = {"Forwarded": 'host="a_:2";port=1;secret=mySecret'}
```
harshanarayana: 👍

sanic/headers.py
Tronic: I would use ipaddress but its implementation is awfully slow.

sanic/headers.py
Tronic: There is enough variation in the tests to make this approach somewhat unfeasible. If there wasn't that host split into host and port, it might just do.

harshanarayana: I don't really have a problem with the current implementation. Just that I am more for lookup over if checks that's all. If you think it won't help in speading up things, then what you already have is perfectly fine.

sanic/headers.py
harshanarayana: @Tronic Really? I wasn't aware of that. Internally at work, I have a custom class called `IP`which does most of the things for me and it's based out of RegEx. But since `ipaddress` was part of python core I thought might as well try. But if it's really slow, then we can stick to regex.

Tronic: These are not entirely comparable as they do different things, but for reference:
```
%timeit ipaddress.ip_address("2606:4700:4700::1001")                   
8.76 µs ± 48.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit ipv6_re.match("2606:4700:4700::1001")                            
588 ns ± 1.34 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit parse_forwarded(headers, config)  # returns {'secret': 'mySecret', 'for': '[2606:4700:4700::1001]'}
5.75 µs ± 26.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

sanic/request.py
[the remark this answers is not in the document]
Tronic: I believe that getting an SNI is practically certain at this point, whenever SSL is used, as vhosts would not work without. Getting the value is quite simple with a callback in SSL context. How that would be used in Sanic is another question, especially since it doesn't contain port (which is present in host/:authority, and is presumed to exist in SERVER_NAME as well, whenever a non-standard port is being used).

What ever is going to be done with `SERVER_NAME`, I propose being postponed to another pull request.

Tronic:
I believe that all concerns raised are now addressed.

A rather big and breaking change resulted in form of new REAL_IP_HEADER/PROXIES_COUNT default values and semantics, and all new proxy configuration documentation was written. Both of these aspects require further review.

sanic/headers.py
harshanarayana: @Tronic That's interesting. Well, if we are using the RegEx, can we please have a bit more strict version of this RegEx? This regex can easily return false positives

harshanarayana:
@Tronic Can you please add a file under `changelogs` directory with the name `1638.feature.rst` and a few details about the PR so that it can be pulled into the change log during the next routine release?

xref : https://sanic.readthedocs.io/en/latest/sanic/contributing.html#changelog

sanic/headers.py
Tronic: I don't see any practical way to make it more strict than it already is. Also, this is only meant to tell IPv4 and IPv6 addresses and hostnames apart, and as far as I can tell, there cannot be any false positives there. Neither IPv4 or hostnames can have 2-7 colons in them. The closest that I can think of is a hexadecimal-looking hostname in a host header like `cafe:8000` which still only has one colon, and IPv6 host should anyway be bracketed like `[f00d::cafe]:8000`.

Tronic: A somewhat bigger concern is that IPv6 are not normalized to compressed form. For this it might be justified to use ipaddress module even though running an address through it takes more time than parsing the entire headers.

Tronic:
> @Tronic Can you please add a file under `changelogs` directory with the name `1638.feature.rst` and a few details about the PR so that it can be pulled into the change log during the next routine release?
> 
> xref : https://sanic.readthedocs.io/en/latest/sanic/contributing.html#changelog

I don't really know how to do it, and there don't seem to be any existing examples. It's all documented in git commits and in updated config documentation. Especially I do not know how you wish to announce the security changes that fix gaping security holes of earlier releases but that also may break some existing installations.

Tronic:
All work done for now, and I'm going to be off keyboard. See you next week!

abuckenheimer:
sorry folks was out of town over the long weekend, I see this got merged which I'm fine with. I do have one lingering question here about changing the PROXIES_COUNT default, namely the new claim in the docs: `Earlier Sanic versions had unsafe default settings. From 19.9 onwards proxy settings must be set manually, and support for negative PROXIES_COUNT has been removed.` I buy that forcing the user to evaluate these options explicitly now probably improves the default state of affairs but I am not convinced that there is no place for a -1 option or that the default was even "unsafe" by itself. I think in reality an attacker can spoof their actual `src` IP address anyway so relying on an ip address alone to do security critical work never seemed to be the right solution, whether it was from the headers or spoofed tcp fields.

I can't really speak towards how the community uses this but I assume this feature was their for a reason so I'd be inclined to change the default but let proxies_count=-1 continue to be a feature preserving backwards compatibility. Obviously we should add some documentation on its ability to be manipulated but I think allowing the developer to make their own call seems more right. That said I may be underestimating the security risk here and am fine being overruled if you folks think otherwise.

Tronic:
It is a very common practice to rely on IP addresses for rate limitation (failed logins etc), temporary bans and reporting to police from which address something was posted. This being so easily spoofed is completely irresponsible, as that would also allow creating false records that would get someone else in trouble.

If using PROXIES_COUNT=-1 behind any standard proxy (that append rather than replace), or directly on Internet,
`curl -H "x-forwarded-for: 1.2.3.4"` and [browser extensions](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj) make it too easy to spoof any address.

The easy migration in most cases is `PROXIES_COUNT=1`. Multi-proxy setups aren't that much harder, (e.g. Cloudflare `REAL_IP_HEADER="CF-Connecting-IP"`).

abuckenheimer:
I'm 100% for changing the default to `None`, it just seems like there are plenty of cases where you know you can trust the client so `PROXIES_COUNT=-1` may be a reasonable thing for a developer to explicitly set. This is mostly a "were all consenting adults here" [approach](https://mail.python.org/pipermail/tutor/2003-October/025932.html), I get how this can go wrong in practice but I'm hesitant to go further than a stern warning since this breaks backwards compatibility. @harshanarayana @vltr any thoughts?

harshanarayana:
@abuckenheimer  @Tronic 

Internally at work, we have a custom webserver extended out of `aiophttop` and we use `proxy_count` behavior that defaults to `-1`. In our case, the deployment is completely in a private network and guarded by the internal proxy + firewalls. So using `-1` works perfectly. 

However, in this context, I do like the default value of `None` over `-1`. But keeping the backward-compatibility in the next release does sound like a good idea. That also falls in line with the idea of giving enough deprecation message so that users can migrate things first.

Tronic:
@abuckenheimer @harshanarayana Is there anything in those setups stopping you from using PROXIES_COUNT=1? Sounds like a small setup that cannot have more than one proxy. Or if you have two, use PROXIES_COUNT=2 or REAL_IP_HEADER, which ever suits you better.

It is very hard to think of actual use cases where (1) you need access to remote_addr, (2) it doesn't need to be trusted (or you can be absolutely sure that your proxy deletes any such headers from client), and (3) it wouldn't be trivially solved by setting one or two of these proxy settings

If you have one, I will reconsider, but otherwise I am **strongly** against adding that security hole back.

Tronic:
It might be helpful to add the options needed by Cloudflare, AWS Cloudfront and other big ones to our docs. What others there are that should be included?

abuckenheimer:
It doesn't sound hard to work around the new restriction I'm just arguing on the behalf of those that aren't in this thread that we should preserve backwards compatibility and move it into deprecation mode.

This is mostly predicated on my analysis that calling this a "security hole" is a bit harsh, really the security hole is assuming you can trust the client (when you can't). If this was more definitively a security hole by itself then we could forgo the deprecation process.

PR #1640, opened 2019-07-22
Fix server crash on request.server_port when bound to IPv6.

Tronic:
If no X-Forwarded-Port nor Host headers are present, Sanic uses "sockname"
to determine the port. This expected (host, port) tuple to be returned but
for IPv6 a 4-tuple is returned instead. Changed code so that port is picked
up in either case. Handling of "peername" was already correct in this regard.

_get_address and server_port both still return incorrect data or crash for
other socket types (e.g unix). Socket type should checked before any queries.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -388,7 +388,7 @@ class Request(dict):
         if forwarded_port:
             return int(forwarded_port)
         else:
-            _, port = self.transport.get_extra_info("sockname")
+            port = self.transport.get_extra_info("sockname")[1]
             return port
 
     @property


Tronic:
````python
import sanic
app = sanic.Sanic()
app.get("/")(lambda req: sanic.response.text(
    f"server_name={req.server_name}, server_port={req.server_port}\n"
))
app.run("localhost")
````
Provided that localhost resolves to ::1, run server and test by: ````curl [::1]:8000````
Current git master: Internal Server Error
This pull request: server_name=[, server_port=8000
Pull request #1638: server_name=[::1], server_port=8000

PR #1641, opened 2019-07-23
Socket binding implemented properly for IPv6 and UNIX sockets.

Tronic:
- app.run("::1") for IPv6
- ~~app.run("unix:/tmp/server.sock")~~ app.run(unix="/tmp/server.sock") for UNIX sockets
- app.run("localhost") retains old functionality (randomly either IPv4 or IPv6, or with workers=1 might be both)

Do note that IPv6 and UNIX sockets are not fully supported by other Sanic facilities.
In particular, request.server_name and request.server_port are currently unreliable.


Tronic:
Compatibility with create_server (used when workers=1) remains an issue. uvloop uses quite different approach to binding than Sanic's implementation, making this difficult to manage.

sanic/server.py
```diff
@@ -842,6 +847,36 @@ def serve(
...
+    """Create socket and bind to host.
+    :param host: IPv4, IPv6, hostname or unix:/tmp/socket may be specified
+    :param port: IP port number, 0 or None for UNIX sockets
+    :return: socket.socket object
+    """
+    if host.lower().startswith("unix:"):  # UNIX socket
+        name = host[5:]
+        sock = socket.socket(socket.AF_UNIX)
+        if os.path.exists(name) and os.stat(name) == stat.S_ISSOCK:
+            os.unlink(name)
```
abuckenheimer: is this because we can't clean up the socket on server close?

Tronic: Unlinking of existing socket? Standard practice that I didn't pay much attention to.

Tronic: Unlinking of existing socket? Standard practice that I didn't pay much attention to.

abuckenheimer: I don't really have a preference, its inconsistent with what happens when you try to bind to a port that's in use though so I figured I'd mention it

Tronic: Sockets are left over e.g. after kill -9, so removing them is practical, and apparently asyncio's and uvloop's create_unix_server both will overwrite any existing socket this way. Also, neither of them removes the socket on close, so I've added unlinking code for workers=1 case now too. As a result, both modes of Sanic now (1) remove any existing socket before bind and (2) remove the socket after normal shutdown.

Tronic: @abuckenheimer Thanks for your notes/reviews!

raphendyr:
:+1: for UNIX sockets. In addition, the implementation looked good.

raphendyr:
`Request.server_port` fails with unix sockets :(

https://github.com/huge-success/sanic/blob/master/sanic/request.py#L401

Tronic:
> `Request.server_port` fails with unix sockets :(

This would conflict with #1638, which I expect to be merged first. Normally, however, even unix socket requests should have a Host header, which would then be used. E.g. try `curl -v --unix-socket /tmp/sanic.sock http://fake.domain:1234/`

With #1638 the existence of Host or other forwarding headers imply scheme's default port if not explicitly mentioned, and I think this would be the best fallback if no such headers are present: return 443 for SSL connections (https, wss) and 80 for anything else.

Tronic:
New unix sockets are created with temporary names, set to listen and then moved over the existing socket to atomically replace it. This allows zero-downtime restarts, no connection attempts may fail.

A new server may be started while the old one is still running.  The old instance will not see any new requests but it may finish its pending requests while at the same time the new instance is already servicing new requests.

sanic/server.py
```diff
@@ -888,15 +868,42 @@ def bind_socket(host: str, port: int, backlog=100) -> socket:
...
+    :param path: filesystem path
+    :param backlog: Maximum number of connections to queue
+    :return: socket.socket object
+    """
+    if os.path.exists(path) and not stat.S_ISSOCK(os.stat(path).st_mode):
+        raise FileExistsError(f"Existing file is not a socket: {path}")
+    sock = socket.socket(socket.AF_UNIX)
+    try:
+        # Atomic zero-downtime socket replace
+        tmp_path = f"{path}.{uuid4().hex[:8]}"
```
raphendyr: Would `os.getpid()` be a simpler solution? It is already ensured to be unique identifier (at least inside a pid namespace (e.g. single container)). If the file would exists, it's would be from an already dead process.

Tronic: Stale sockets might still be left from before reboot, causing services fail to start. There are apparently also some security concerns, in particular that if another user creates a symlink in place of the socket, bind on MacOS follows that symlink and creates a socket at symlink's target. Eight random hex digits is far less likely to be hit by these issues than a pid, although it might be necessary to use even more.

A minor issue with this approach is that the initial random name sticks, and is returned as sockname on the server side and as peername on the clients.

Tronic:
Is anyone aware of why request checks peername and sockname from protocol instead of protocol filling these into requests? The only reasons I can think of are (1) it was the fastest way to hack it, or (2) not to consume runtime when these values are not used. Since the values need to be read from socket only once per connection, not per request, I plan to change this to happen right after connection.

Tronic:
Quite many potentially breaking changes again. @abuckenheimer can you have a look and comment?

Tronic:
Current implementation is balancing between backward compatibility and sensible implementation. And I didn't benchmark at all, in particular the ASGI server that creates new ConnInfo for every request. Probably no slowdown still, but needs to be tested.

In any case, this will require further refining before merging.

Tronic:
Also, it'd be great if someone could write tests for unix sockets (I have no idea how to fit that into the testing framework which afaik doesn't even test live servers).

ahopkins:
> Also, it'd be great if someone could write tests for unix sockets (I have no idea how to fit that into the testing framework which afaik doesn't even test live servers).

This could be done. The Testing client does stand up an instance. RIght now it build the socket on each request, but it should not be too hard to refactor to pass a socket into the test for this functionality.

I can work on it this coming week.

ahopkins:
@Tronic I was looking at this more closely today, and reminded myself that the test client allows you to pass arbitrary `kwargs` from the request call through to the app runner.

```python
request, response = app.test_client.get("/get", server_kwargs={...})

// Translates to....

app.run(debug=debug, **server_kwargs)
```

Therefore, you should be able to run tests:

```python
sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
sock.bind(1234)
request, response = app.test_client.get("/get", server_kwargs={"sock": sock})
```

sanic/server.py
```diff
@@ -42,6 +45,40 @@ class Signal:
...
+        "server",
+        "server_port",
+        "client",
+        "client_port",
+        "ssl",
+    )
+
+    def __init__(self, transport):
+        self.ssl = bool(transport.get_extra_info("sslcontext"))
+        self.sockname = a = transport.get_extra_info("sockname")
```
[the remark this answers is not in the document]
Tronic: It's a strictly local variable, a very short name for one in purpose. Otherwise one could just use `self.sockname` and `self.peername` but it makes the code unnecessarily long.

Alike, multiple assignment should be pretty standard, right?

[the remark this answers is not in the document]
mikeyusko: Yes, I understand your point, but in my opinion, it would be better to rename `a` to something a bit more understandable, also don't forget about Python's zen:)
```
Explicit is better than implicit.
```

Naming is a very important thing and on another hand, it always hard to choose right naming:)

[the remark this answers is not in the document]
ahopkins: I think this is explicit enough. As said, it is not exposed outside. I don't have a strong opinion.

sanic/server.py
```diff
@@ -42,6 +45,40 @@ class Signal:
...
+        self.server = self.client = ""
+        self.server_port = self.client_port = 0
```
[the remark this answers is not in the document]
ahopkins: I'm fine with this. We have it a number of times in the code base and it is not overly complex. It is a fairly standard practice.

ahopkins:
@Tronic is this ready for final review and merge?

Tronic:
> @Tronic is this ready for final review and merge?

Yes. If there are any further changes, they should probably go in a separate PR.

Tronic:
@ahopkins Tests are still missing but I won't have time to look at that now.

Tronic:
I'm done with this. Requesting review.

Tronic:
This doesn't need to go in 20.3 but the ConnInfo structure is something I'll be using in the streaming branch as soon as this gets merged.

Tronic:
@huge-success/sanic-core-devs Anyone care to review?

sanic/request.py
```diff
@@ -349,56 +351,55 @@ def cookies(self):
...
-        if not hasattr(self, "_socket"):
-            self._get_address()
-        return self._socket
-
-    def _get_address(self):
-        self._socket = self.transport.get_extra_info("peername") or (
-            None,
-            None,
-        )
-        self._ip = self._socket[0]
```
ahopkins, line 381: Should we be concerned that we are no longer setting `self._ip` and `self._port`. Yes, they are "hidden" properties, but maybe we leave them in with a dep notice?

Tronic: These were just cache for the `request.ip` and `request.port` properties, right? I don't think anyone other than those property accessors should've been touching them, and if someone still did, it is righteous to have their programs crash.

Another open question is whether `remote_addr` should return the local peer address when no proxy is used (it returns None in current master and IIRC on this too). The reasoning for this is that a web service is interested on user IP, irregardless of whether the service is installed behind a proxy or not, and `request.remote_addr or request.ip` isn't very practical. Probably better left for another PR though.

ahopkins: Agreed. Going to merge in and add to `20.6`. Save for another day.

Tronic:
@ahopkins Are you still including this one?

Tronic:
Hmm, looks like some unix socket tests got broken now. I suppose that'll need more investigation.

ahopkins:
Tests were failing. If you can take a look and solve easily, let's push another release.

Tronic:
> Tests were failing. If you can take a look and solve easily, let's push another release.

#1853 made workers use spawn rather than fork on all OSes, and consequently doesn't allow using function-local app objects or even route decorators. However, I would expect that to break far more things than only this...

Do you know what the current behaviour regards pickling should be? Perhaps something from master was not correctly merged to this? I'll try to investigate.

ahopkins:
@ashleysommer ?

Tronic:
This sets "spawn" mode, earlier "fork" was used on Linux and MacOS (Windows only supports "spawn"):
https://github.com/huge-success/sanic/blob/master/sanic/server.py#L930

As a result, with current master, this doesn't work:
```python
from sanic import Sanic

app = Sanic(__name__)

@app.get("/")
def hello(req):
    pass

app.run(workers=2)
```

```
_pickle.PicklingError: Can't pickle <function hello at 0x10ff18160>: it's not the same object as __main__.hello
```

Tronic:
The multiple worker test is now made compatible with either mode of multiprocessing, so that this can be merged. Let's move further discussion about the pickling to issue #1883 or related PR.

ahopkins:

PR #1649, opened 2019-07-29
PEP 594 has cgi module scheduled for deprecation in Python 3.8

Tronic:
Reimplement cgi.parse_header in Sanic, removing the dependency. The new implementation is much faster than either cgi.parse_header or the equivalent werkzeug.parse_options_header, and unlike the two, handles also quoted values with semicolons or \" in them.

### DIFF
--- a/sanic/headers.py
+++ b/sanic/headers.py
@@ -1,0 +1,18 @@
+import re
+
+token, quoted = r"([\w!#$%&'*+\-.^_`|~]+)", r'"((?:[^"]|\\")*)"'
+parameter = re.compile(f';\s*{token}=(?:{token}|{quoted})', re.ASCII)
+
+
+def parse_options_header(value: str):
+    """Parse HTTP header values of Content-Type format."""
+    pos = value.find(';')
+    if pos == -1:
+        options = {}
+    else:
+        options = {
+            m.group(1).lower(): m.group(2) or m.group(3).replace(r'\"', '"')
+            for m in parameter.finditer(value[pos:])
+        }
+        value = value[:pos]
+    return value.strip().lower(), options
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -4,7 +4,6 @@ import json
 import sys
 import warnings
 
-from cgi import parse_header
 from collections import defaultdict, namedtuple
 from http.cookies import SimpleCookie
 from urllib.parse import parse_qs, parse_qsl, unquote, urlunparse
@@ -13,7 +12,7 @@ from httptools import parse_url
 
 from sanic.exceptions import InvalidUsage
 from sanic.log import error_logger, logger
-
+from sanic.headers import parse_options_header
 
 try:
     from ujson import loads as json_loads
@@ -177,7 +176,7 @@ class Request(dict):
             content_type = self.headers.get(
                 "Content-Type", DEFAULT_HTTP_CONTENT_TYPE
             )
-            content_type, parameters = parse_header(content_type)
+            content_type, parameters = parse_options_header(content_type)
             try:
                 if content_type == "application/x-www-form-urlencoded":
                     self.parsed_form = RequestParameters(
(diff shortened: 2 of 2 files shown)


sanic/headers.py
```diff
@@ -0,0 +1,19 @@
...
+    pos = value.find(";")
+    if pos == -1:
+        options = {}
+    else:
+        options = {
+            m.group(1).lower(): m.group(2) or m.group(3).replace('\\"', '"')
+            for m in parameter.finditer(value[pos:])
+        }
+        value = value[:pos]
+    return value.strip().lower(), options
```
abuckenheimer: we should add the existing cpython tests to the test suite for sanic, here they are converted to pytest:

```import pytest
from sanic.headers import parse_options_header


@pytest.mark.parametrize(
    "input, expected",
    [
        ("text/plain", ("text/plain", {})),
        ("text/vnd.just.made.this.up ; ", ("text/vnd.just.made.this.up", {})),
        ("text/plain;charset=us-ascii", ("text/plain", {"charset": "us-ascii"})),
        ('text/plain ; charset="us-ascii"', ("text/plain", {"charset": "us-ascii"})),
        (
            'text/plain ; charset="us-ascii"; another=opt',
            ("text/plain", {"charset": "us-ascii", "another": "opt"})
        ),
        (
            'attachment; filename="silly.txt"',
            ("attachment", {"filename": "silly.txt"})
        ),
        (
            'attachment; filename="strange;name"',
            ("attachment", {"filename": "strange;name"})
        ),
        (
            'attachment; filename="strange;name";size=123;',
            ("attachment", {"filename": "strange;name", "size": "123"})
        ),
        (
            'form-data; name="files"; filename="fo\\"o;bar"',
            ("form-data", {"name": "files", "filename": 'fo"o;bar'})
        ),
    ]
)
def test_parse_headers(input, expected):
    assert parse_options_header(input) == expected

```

the last test misses. I think the issue is the `quoted` string should be a normal string instead of a raw

sanic/headers.py
```diff
@@ -0,0 +1,19 @@
+import re
+
+
+token, quoted = r"([\w!#$%&'*+\-.^_`|~]+)", r'"((?:[^"]|\\")*)"'
+parameter = re.compile(fr";\s*{token}=(?:{token}|{quoted})", re.ASCII)
+
+
+def parse_options_header(value: str):
```
abuckenheimer: can you add the return type hint while your at it? `-> typing.Tuple[str, dict]`

sanic/headers.py
Tronic: A long time ago browsers didn't escape and that became the common practice; to deal with all the existing broken CGI scripts, modern browsers opt to mark it as %22 instead of attempting to use `\"`. Sanic's form decoding code should probably decode these as well (currently it doesn't and the %-escape is in the returned filename too). I believe it is in order to modify or remove that last test.

sanic/headers.py
Tronic: Done.

sanic/headers.py
abuckenheimer: gotcha, I'm fine with removing that last test then, but we should probably remove the `\\"` part of the `quoted` token then so its are clear we explicitly aren't observing quoting

Tronic: It is included in the RFCs and probably actually used in some other headers, so I'd prefer to keep it if it doesn't cause trouble.

Tronic: Okay, I just tried with a file named `test\` (on unix)
``
`
Content-Disposition: form-data; name="foo"; filename="test\"
```
Looks like we need to remove that escape.

Tronic:
This is turning out to be trickier than I thought, at least if one attempts to get the corner cases right. https://paste.zi.fi/p/content-disposition.html/view

So, what next, user-agent detection and fuzzy heuristics?

Tronic:
@abuckenheimer I believe this is now ready to be merged. With the added heuristics, I could restore the last CPython test and now Firefox and Chrome are mostly handled correctly even when unconventional values appear.

PR #1657, opened 2019-08-13
release: 19.6.3

yunstanford:
release: 19.6.3


harshanarayana:
@yunstanford Just a suggestion. You should probably try out the `make release` command.. It generates the changelog and updates app other applicable places in one shot. Cut a new branch and run the make command on that. And that branch can then be merged to master and tagged.

yunstanford:
@harshanarayana oh, i forgot about that, will give it a shot

yunstanford:
@harshanarayana the release script raises a few errors on my side, also looks like it modifies version in `setup.cfg` to be `current_version = 19.08.0` ? ... didn't look deep in that script. And it checks if latest tag matches with current version, which also doesn't work in our workflow, because we only attach tags in master branch after merging.

harshanarayana:
@yunstanford If you are running the `make release` use the following steps

1. Cut a new and clean branch from master
2. `version=19.6.3 make release`

```bash
 ▲ OpenSource/Python/sanic version=19.6.3 make release                                                                                                                                                                                     7s  ⇡ release-19.6.3 :: 1d :: ● :: ⬢
python scripts/release.py --release-version 19.6.3 --generate-changelog
Loading template...
Finding news fragments...
Rendering news fragments...
Writing to newsfile...
Staging newsfile...
Removing news fragments...
Removing the following files:
/Users/harshnar/Work/OpenSource/Python/sanic/changelogs/1631.doc.rst
/Users/harshnar/Work/OpenSource/Python/sanic/changelogs/1631.feature.rst
Done!
```

That should be it. 

So the default version it picks up is the `CalVer` format so the new version it defaults to is `19.08.0`. However, using the `version` arg with `make` will fix that case.

andreymal:
[tag](https://github.com/huge-success/sanic/blob/b3ada6308bc992fd233929e93d58c6af478313e4/.travis.yml#L52)?

PR #1661, opened 2019-08-22
WIP streaming response

abuckenheimer:
This is is not in a merge ready state but it shows a POC for a potential new streaming response api. Motivated as a comparison to #1645

### DIFF
--- a/examples/simple_stream_server.py
+++ b/examples/simple_stream_server.py
@@ -1,0 +1,18 @@
+import asyncio
+from sanic import Sanic
+
+app = Sanic(__name__)
+
+
+@app.stream("/")
+async def index(request, response):
+    await response.write_headers()
+    await asyncio.sleep(1)
+    await response.write("Hello\n")
+    await asyncio.sleep(1)
+    await response.write("World\n")
+    await asyncio.sleep(1)
+
+
+if __name__ == '__main__':
+    app.run(host="0.0.0.0", port=8000, debug=True)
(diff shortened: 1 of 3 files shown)


Tronic:
Experimenting in trio branch:
```python
@app.put("/echo")
async def echo(req):
    # We (or middleware) can do arbitrary things before choosing to accept
    if int(req.headers["content-length"]) > 1_000_000:
        # This makes the protocol reject the request without receiving its body
        raise HeaderExpectationFailed("I'm not gonna take all that!")
    # Send response headers
    res = await req.respond(content_type="text/plain")
    # Stream request body to response body
    async for data in req.stream:
        await res.write(data)
```

`req.stream` is similar to the old StreamBuffer but can also be iterated, as seen in the example. streaming works both ways without any buffering/queues, and thus we get implicit flow control and back-pressure handling.

I didn't yet look at how to handle non-streaming request bodies. Supposedly Sanic already stores a "streaming handler" flag somewhere for this purpose, and if the flag is missing, the entire body is loaded before running the handler.

Response streaming is independent of request streaming. You can have either, both or neither.

`req.respond` makes a `NewStreamingResponse` and sends its headers out before returning the response object, where body chunks may be sent. It is an error to call this multiple times, or to return a response object after calling this.

Non-streaming response can still be done as before, by returning a response object, which ends up with the app.py handler doing `await req.respond(handler_return_value)` -- send to client a normal Sanic response object and don't return anything.

yunstanford:
@abuckenheimer are u still working on it ?

Tronic:
I am still working on mine -- but it'll likely take until December until I can resume working on Sanic stuff.

abuckenheimer:
I'm not still working on this, happy to pick it back up if anyone else finds the API I'm outlining appealing

PR #1666, opened 2019-09-03
Implement dict-like API on request objects for custom data.

### LINKED ISSUE #1309: Design question: why is `Request` a subclass of `dict`?
Hello. I was just wondering: why is the `Request` class a subclass of `dict`? Would it be to enable developers do "store" information into a instance like `request["name"]`?

Thanks in advance!

Tronic:
- Replaces earlier implementation that inherited request from dict
- Closes #1309

### DIFF
--- a/sanic/helpers.py
+++ b/sanic/helpers.py
@@ -153,27 +153,3 @@ def import_string(module_name, package=None):
     if ismodule(obj):
         return obj
     return obj()
-
-
-class ContextObject:
-    """Storage of arbitrary values using attribute syntax.
-
-    >>> context.key_name = 123
-    >>> "key_name" in context
-    True
-    >>> context.key_name
-    123
-    >>> context.no_such_key
-    None
-    """
-
-    def __getattr__(self, key):
-        """Return None for missing keys."""
-        if not key.startswith("_"):
-            return None
-        # Except system attributes
-        raise AttributeError(f"ContextObject has no attribute {key!r}")
-
-    def __contains__(self, key):
-        """Test if a key actually exists."""
-        return key in self.__dict__
--- a/sanic/helpers.py
+++ b/sanic/helpers.py
@@ -153,3 +153,24 @@ def import_string(module_name, package=None):
     if ismodule(obj):
         return obj
     return obj()
+
+class ContextObject:
+    """Storage of arbitrary values using attribute syntax.
+
+    >>> context.key_name = 123
+    >>> "key_name" in context
+    True
+    >>> context.key_name
+    123
+    >>> context.no_such_key
+    None
+    """
+    def __getattr__(self, key):
+        """Return None for missing keys."""
+        # Except system attributes
+        if key.startswith("_"):
+            raise AttributeError(f"ContextObject has no attribute {key!r}")
+
+    def __contains__(self, key):
+        """Test if a key actually exists."""
+        return key in self.__dict__
--- a/sanic/helpers.py
+++ b/sanic/helpers.py
@@ -154,6 +154,7 @@ def import_string(module_name, package=None):
         return obj
     return obj()
 
+
 class ContextObject:
     """Storage of arbitrary values using attribute syntax.
 
--- a/sanic/helpers.py
+++ b/sanic/helpers.py
@@ -166,6 +166,7 @@ class ContextObject:
     >>> context.no_such_key
     None
     """
+
     def __getattr__(self, key):
         """Return None for missing keys."""
         if not key.startswith("_"):
--- a/sanic/helpers.py
+++ b/sanic/helpers.py
@@ -168,9 +168,10 @@ class ContextObject:
     """
     def __getattr__(self, key):
         """Return None for missing keys."""
+        if not key.startswith("_"):
+            return None
         # Except system attributes
-        if key.startswith("_"):
-            raise AttributeError(f"ContextObject has no attribute {key!r}")
+        raise AttributeError(f"ContextObject has no attribute {key!r}")
 
     def __contains__(self, key):
         """Test if a key actually exists."""
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -122,7 +122,7 @@ class Request:
         self.parsed_not_grouped_args = defaultdict(list)
         self.uri_template = None
         self._cookies = None
-        self._storagedict = None
+        self._storagedict = {}
         self.stream = None
         self.endpoint = None
 
@@ -138,19 +138,14 @@ class Request:
 
     def get(self, key, default=None):
         """Request data storage. Arbitrary per-request data may be stored."""
-        return (
-            default
-            if self._storagedict is None
-            else self._storagedict.get(key, default)
-        )
+        return self._storagedict.get(key, default)
 
     def __contains__(self, key):
-        return key in self._storagedict if self._storagedict else False
+        """Request data storage. Arbitrary per-request data may be stored."""
+        return key in self._storagedict
 
     def __getitem__(self, key):
         """Request data storage. Arbitrary per-request data may be stored."""
-        if self._storagedict is None:
-            raise KeyError("Nothing is stored yet.")
         return self._storagedict[key]
 
     def __delitem__(self, key):
@@ -159,8 +154,6 @@ class Request:
 
     def __setitem__(self, key, value):
         """Request data storage. Arbitrary per-request data may be stored."""
-        if self._storagedict is None:
-            self._storagedict = {}
         self._storagedict[key] = value
 
     def body_init(self):
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -7,6 +7,7 @@ import warnings
 from collections import defaultdict, namedtuple
 from http.cookies import SimpleCookie
 from urllib.parse import parse_qs, parse_qsl, unquote, urlunparse
+from types import SimpleNamespace
 
 from httptools import parse_url
 
@@ -17,7 +18,6 @@ from sanic.headers import (
     parse_host,
     parse_xforwarded,
 )
-from sanic.helpers import ContextObject
 from sanic.log import error_logger, logger
 
 
@@ -115,7 +115,7 @@ class Request:
 
         # Init but do not inhale
         self.body_init()
-        self.custom_context = ContextObject()
+        self.custom_context = SimpleNamespace()
         self.parsed_forwarded = None
         self.parsed_json = None
         self.parsed_form = None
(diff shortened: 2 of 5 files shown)


sanic/request.py
```diff
@@ -71,7 +71,7 @@ def is_full(self):
         return self._queue.full()
 
 
-class Request(dict):
+class Request:
```
harshanarayana: This change will break any and all current usage where someone is iterating over the request object like `for _key in request`. Is there a reason why these changes is required?

abuckenheimer:
If we were going to do this we should make a breaking change where we create a a public property like `_storagedict` and tell the user to interact with that directly. This would be less magical and make it more apparent to the user that they are not abusing the request class to keep track of their own information.

sanic/request.py
Tronic: @harshanarayana Is this actually being used? Iteration, or the entire storage feature? Request is not a dictionary, so it shouldn't inherit just to gain those features. This has led to problems as pointed out in #1309.

The iteration APIs of dict could be added, but frankly I'd rather do what @abuckenheimer suggested and make it a public property instead. If anything, I'd expect request[...] to fetch me HTTP headers (but keeping that at request.headers[...] is better, just like request.storage[...]).

sanic/request.py
```diff
@@ -94,6 +94,7 @@ class Request(dict):
         "parsed_json",
         "parsed_forwarded",
         "raw_url",
+        "storage",
```
abuckenheimer: maybe `user_storage`? I really want to hammer home that this is not used by sanic and is for the developer

abuckenheimer: we should update the docs to mention the new property

sanic/request.py
```diff
@@ -134,6 +136,26 @@ def __bool__(self):
             return True
         return False
 
+    def get(self, key, default=None):
+        """Request data storage. Arbitrary per-request data may be stored."""
```
abuckenheimer: I think the `request.get(...)` /  `request[...]` modes of access should be deprecated, if people are on board with that lets add a warning in the docstrings

tests/test_request_data.py
```diff
@@ -12,18 +12,35 @@
 def test_storage(app):
     @app.middleware("request")
     def store(request):
+        try:
```
abuckenheimer: we should copy this test to make another that uses `request.storage` directly

sanic/request.py
harshanarayana: I can't tell you if it is being used or not. I mean the iterable nature of it. This was how it was working. So before we break this, it might be a good idea to give out a deprecation in the next release and then remove it.

sanic/request.py
harshanarayana: @abuckenheimer how about using `custom_context` instead of `storage` or `user_storage` ?

sanic/request.py
harshanarayana: Agreed. If we want to let user add additional request context but not let them use the request object directly as a `dict`, then we are better off moving away from the the `get(k, default)`

Tronic:
I realised that things like `user = request.custom_context["user"]` being repeated in the beginning of every handler would be unbearably ugly, and took the liberty of changing that into `request.custom_context.user`, built on actual Python member variables. For convenience, `None` is returned for missing attributes and one can use `"key_name" in request.custom_context` to check if the key actually exists (even if its value is None).

sanic/helpers.py
```diff
@@ -153,3 +153,25 @@ def import_string(module_name, package=None):
     if ismodule(obj):
         return obj
     return obj()
+
+
+class ContextObject:
+    """Storage of arbitrary values using attribute syntax.
```
abuckenheimer: so I really do think we should use a dictionary instead of a `ContextObject`, this object quacks like a dict so it should just be a dict. A couple problems I can imagine:
1. there is no way to tell if an attribute is set explicitly as None or your just getting the default (while with a dict the `[key]` vs `.get(key)` syntax lets you choose when getting the default is the right choice)
2. this cleans up direct access `request.custom_context.user` but it makes variable access opaque `request.custom_context.__dict__[my_key]` or `getattr(request.custom_context, my_key)`
3. extending a dict to [addict](https://github.com/mewwts/addict) style access is easy and feels natural, where as going back to a dict from this feels like your missing and important decision the framework developers made
```
from addict import Dict

@app.middleware("request")
def extend_custom_context(request):
    request.custom_context = Dict(request.custom_context)
```

Tronic:
Considering the usage of this, presumably to inject `user` or other such variables from request middleware, I don't believe iteration has ever been used, and thus adding a deprecation compatibility function for that would be counter-intuitive. We'd additionally have to add `__missing__`,  `values`, `items` etc. to fully imitate dict interface. Also, the iteration would have to omit from `__dict__` things that begin with underscore.

abuckenheimer:
I totally agree that having requests implement `__iter__`, `values`, `items` is weird! which is why this is a great idea for a PR, but this functionality has been out for almost 2 years so its hard to say what its been used for.

It should be easy to implement these, here is the current functionality:

```
# lightly modified simple_server.py
from sanic import Sanic
from sanic import response

app = Sanic(__name__)


@app.middleware("request")
async def test1(request):
    request['foo'] = 'bar'
    request['__dunder__'] = 'baz'


@app.route("/")
async def test(request):
    print(list(request))
    print(request.values())
    print(request.keys())
    print(request.items())
    print('foo:', request['foo'], '__dunder__:', request['__dunder__'])
    return response.json({"test": True})


if __name__ == '__main__':
```
stdout:

```
[2019-09-04 12:56:20 -0400] [18576] [INFO] Goin' Fast @ http://0.0.0.0:8000
[2019-09-04 12:56:20 -0400] [18576] [INFO] Starting worker [18576]
['foo', '__dunder__']
dict_values(['bar', 'baz'])
dict_keys(['foo', '__dunder__'])
dict_items([('foo', 'bar'), ('__dunder__', 'baz')])
foo: bar __dunder__: baz
[2019-09-04 12:56:33 -0400] - (sanic.access)[INFO][10.150.11.87:56330]: GET http://host:8000/  200 13
```

so all of these act exactly as if they were done on the new `custom_context` property.

I also don't think we need to implement `__missing__`, that would be something that someone would have to subclass for themselves to matter I think.

sanic/helpers.py
Tronic: > so I really do think we should use a dictionary instead of a `ContextObject`, this object quacks like a dict so it should just be a dict. A couple problems I can imagine:

We seem to consider it from very different angles. For you it is fundamentally a dict and thus you consider iteration an issue. For me it is really equivalent to adding member variables to `request`, but namespaced to `custom_context` so that extensions do not collide with Sanic core (in particular, so that we don't need to consider someone already using a variable name whenever we add a new member to `request`).

In the context of variables, the `.name` syntax is correct, and there is no need for iteration (just like you wouldn't iterate all members of the `request` object).

Let's consider a session store, using a request middleware to get session variables from a session cookie, and a response middleware for updating the cookie with new values from custom_context.

The session variables cannot be loaded directly to custom_context because then any other middlewares' variables would also get stored to session -- this is not what we want -- and this is also why I think that custom_context should not be iterated.

If custom_context was a dict, the middleware would have to add its storage as `request.custom_context["session"]` and then the handler would be using `request.custom_context["session"].get("foo")` to get a value from session. Even PHP with its `$_SESSION["foo"]` is not nearly as cumbersome.

As a matter of fact, I am proposing another name change to simply `ctx`. While `custom_context` is explanatory, it really is only a namespace and a very tedious name for such a thing. `request.ctx.session.get("foo")` seems far more reasonable.

> 1. there is no way to tell if an attribute is set explicitly as None or your just getting the default (while with a dict the `[key]` vs `.get(key)` syntax lets you choose when getting the default is the right choice)

You can find out by `"foo" in request.custom_context` (`True` for existing keys, even if their value is `None`), although I would expect most request handlers to be more interested on the truthiness of values than specifically which of *undefined*, `None` or even otherwise *falsy* they are.

Tronic: It turns out that such a session store already exists https://github.com/xen/sanic_session (although storing in memory instead of cookie).

Tronic: Thumb emoji vote: thumb-up for `ctx`, thumb-down for `custom_context`.

Please read the above messages first. If you think that some other very short name would be better, thumb-up and comment which.

abuckenheimer: My intuition here is that `.name` syntax to me implies some sort of predefined schema / api, the whole point of this `custom_context` is we don't know ahead of time how people are going to use this . I think most people will instinctively have different reactions when they get an `AttributeError` and a `KeyError`, the first makes you think "oh I'm using this object wrong" the second makes you think "why didn't that key get set". Because no one really owns the custom context (extensions can add whatever items they want in any order) I think a dict better reflects the reality that these are not predefined `.name` attributes but instead attached at runtime. I get that `request.custom_context["session"]` is strictly 3 more characters than it needs to be but I think there is an intention that is conveyed there that is worth keeping.

I agree that iterating seems like the wrong behavior in most cases! Its super fragile with pluggable extensions. However maybe were both starting with the wrong user story! If you look at the session store that you link to its whole purpose is to create this `"session"` variable that really does feel like it should be a available as a `request.session` attribute. Its kind of an extension of the `Request` class api, and after I `Session(app, interface=InMemorySessionInterface())` I always expect `request.session` to be there just like I expect `request.content_type` to be there and I expect it to be defined in exactly one place. As is using a `dict` or a `CustomContext` both enable a class of bugs where multiple extensions may not know about each other and both try to set their own `"session"` i.e. `request.custom_context['session'] = foo` or `request.custom_context.session = bar`. I think maybe the real answer is to come up with a registry of request attributes where extensions can add their custom logic. i.e. something like:

```
# haven't thought about the signature / names too hard, open to suggestions
@app.request_ctx
def session(request: sanic.request.Request) -> sanic_session.SessionDict:
    # sanic_session logic goes here
    return ...

@app.route("/")
def index(request):
    my_session = request.session
    return ...
```

I think this has a lot of benefits
1. the api for extending sanic looks just like the api for using it
2. we can prevent name clashes at import time
2. we can implement a lot of the exiting request class using this kind of logic and dogfood our api
3. the request_ctx handler can be lazy but cache the result, this makes the order of execution of middle ware unimportant + mimics how we operate in the actual request class preventing unnecessary work.

tldr:
I think with how sanic currently works a dictionary just reflects the reality of the `custom_context` better however I think if we rework how extensions are implemented we end up with a better product and something that makes sense using that sweet sweet syntactical sugar of  `request.name`

abuckenheimer: just realized @ashleysommer may have some thoughts on this as the author of https://github.com/ashleysommer/sanicpluginsframework

PR #1670, opened 2019-09-05
Sanic does not support Python 3.5 and won't need this code.

Tronic:


sanic/request.py
```diff
@@ -1,7 +1,6 @@
 import asyncio
 import email.utils
 import json
```
abuckenheimer: minor tweak, we don't need to import json if ujson doesn't exist, we should make this import in the except branch

Tronic: Good catch!

Tronic:
Fixing bugs while riding the metro, without testing... Doesn't always work!

Tronic:
You can do the squash by pressing the arrow button next to the merge button, if I am not mistaken.

PR #1674, opened 2019-09-11
More robust response datatype handling

Tronic:
Currently `HTTPResponse` takes a `body` argument, always first converted into str and then encoded into bytes, and alternatively `body_bytes` (used when body is None). The easy response helper functions `text` and `html` pass their arguments directly, and `json` passes after dumps to HTTPResponse's `body`, while the `raw` and `file` responses use `body_bytes`.

Additionally, `StreamingHTTPResponse.write` takes only `data`, which according to documentation is bytes-ish but actually gets encoded if its type is anything but `bytes`: bytearray, memoryview, array.array, numpy.array and all other bytes-ish types get converted into strings and then encoded.

In current implementation, `html(b"<!DOCTYPE html>...")` writes literally `b'<!DOCTYPE html>...'` to the client, and while streaming response `write(b"<!DOCTYPE html>")` actually works correctly, `write(bytearray(b"<!DOCTYPE html>"))` sends to client `bytearray(b'<!DOCTYPE html>')`

I would like to remove auto-conversion of anything into string because it will silently convert *any* Python object, even those without string conversion (you get something like `<class '__main__.Foo'>`).

If we limit ourselves a bit, reliable and *fast* autodetection becomes possible by simply
```python
body_bytes = body.encode() if hasattr(body, "encode") else body
```

This should avoid a lot of bugs and while it actually avoids the need for separate `body_bytes` arguments, support for those may be retained to avoid breaking existing applications.

In this pull request I've kept full type conversion in `sanic.response.text` because that is the one place where one might sensibly rely on it. This kind of type detection is slightly slower and ideally should be removed entirely (after a deprecation period).

StreamingHTTPResponse and HTTPResponse will only take encodeable (str) or bytes-ish objects, which may be passed as the first argument.

Additionally, I've added support for `__html__` and `_repr_html_` accessors to `sanic.response.html`. These functions are used by pandas dataframes, various HTML-generating libraries and other modules wishing to provide automatic HTML formatting on their objects, and supported by Jupyter and Qt notebooks. This means that you can `return sanic.response.html(pandas.read_excel("sales.xls"))` and get a pretty HTML table.


Tronic:
This is built on top of headerformat which also rewrote _encode_body.

Tronic:
Anyone up for reviewing this?

ahopkins:
On it ... 

`__html__` support sounds like it makes a lot of sense. I sort of wish that were in a seperate PR.

As you stated yourself, the question is how to handle the potential breaking change here. There has been some discussion in the past about how exactly to handle these items, and I am not sure there ever was a good answer.

Anyway, in the mean time, until we answer that question I will take a look at the substance of the PR itself.

Thanks.

sanic/response.py
```diff
@@ -249,6 +192,15 @@ def text(
...
     """
+    # Type conversions are deprecated and quite b0rked but still supported for
+    # text() until applications get fixed. This try-except should be removed.
+    try:
+        # Avoid repr(body).encode() b0rkage for body that is already encoded.
+        # memoryview used only to test bytes-ishness.
+        with memoryview(body):
+            pass
+    except TypeError:
+        body = f"{body}"  # no-op if body is already str
```
ahopkins: We should think about throwing a `warning` here.

ahopkins: @Tronic can you also take a look at the conflict in `headers.py`?

Tronic: Which types should be deprecated for `text()`?

- [ ] str
- [ ] bytes
- [ ] numeric and others with meaningful `__str__`
- [ ] those that produce `<Foo object at 0x106471210>`

sjsadowski:
@Tronic can you fix conflicts? I think this was largely good to go but didn't get included in 19.12 - I'd like to get it merged for 20.3

sanic/response.py
Tronic: Deprecation warning added for everything but str-derived types.

Tronic:
Too much time wasted on merging, but also some improvements done (hopefully).

Left for other PRs & discussion: full deprecation of HTTP/1.0, removal of version and keep-alive parameters and/or removing the `get_headers` function entirely.

Tronic:
A number of tests now spawn deprecation warnings for trying to response.text something else than strings. Should I also start fixing those, and possibly add one specifically hitting that warning so that coverage doesn't go down?

Otherwise, I consider this ready to be merged.

Tronic:
@sjsadowski Could you merge this and perhaps #1739 which is related?

PR #1676, opened 2019-09-13
Fixes "after_server_start" when using return_asyncio_server.

ashleysommer:
Fixes ability to trigger "after_server_start", "before_server_stop", "after_server_stop" server events when using `app.create_server()` along with `return_asyncio_server=True` to start and manage your own asyncio_server.

See example file `examples/run_async_advanced.py` for a full example.

Related to community forum post: https://community.sanicframework.org/t/triggering-after-server-start-event-when-running-asyncio-server/385

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -634,7 +634,12 @@ def trigger_events(events, loop):
             loop.run_until_complete(result)
 
 
-class AsyncServerCoro:
+class AsyncioServer:
+    """
+    Wraps an asyncio server with functionality that might be useful to
+    a user who needs to manage the server lifecycle manually.
+    """
+
     __slots__ = (
         "loop",
         "serve_coro",
@@ -654,6 +659,8 @@ class AsyncServerCoro:
         before_stop,
         after_stop,
     ):
+        # Note, Sanic already called "before_server_start" events
+        # before this helper was even created. So we don't need it here.
         self.loop = loop
         self.serve_coro = serve_coro
         self._after_start = after_start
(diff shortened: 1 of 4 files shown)


ashleysommer:
Pushed a fix. Hopefully that is all that was required.

sanic/server.py
```diff
@@ -634,6 +634,78 @@ def trigger_events(events, loop):
...
+        # before this helper was even created. So we don't need it here.
+        self.loop = loop
+        self.serve_coro = serve_coro
+        self._after_start = after_start
+        self._before_stop = before_stop
+        self._after_stop = after_stop
+        self.server = None
+        self.connections = connections
+
+    def after_start(self):
```
yunstanford: I'm thinking if we should expose those as async.

ashleysommer: Yes, its definitely possible.

It looks like sanic has two different `trigger_events()` helpers.

One in app.py is async: https://github.com/huge-success/sanic/blob/b7d4121586e779bd40a1cfd2188b8180d4486ab5/sanic/app.py#L1247
and one in server.py is not async: https://github.com/huge-success/sanic/blob/b7d4121586e779bd40a1cfd2188b8180d4486ab5/sanic/server.py#L625 but it uses `loop.run_until_complete()` to execute the events.

Currently using the second one, so that's why these methods aren't async.

One issue with making these async, by using the async version of `trigger_events()`, doing 
`await server.after_start()` will not work until after the loop has started with loop.run_forever() or loop.run_until_complete(), and that doesn't fit with some user's workflow for managing their own asyncio_server.

yunstanford: yeah, managing the stop side might be different, the loop may already be stopped at that moment, it really depends.

I think let's keep it as it is for now.

VMAtm:
@yunstanford Sorry to bother, any information when this code will be published? Looks like it's not on PyPi or conda-forge channels yet, right?

harshanarayana:
@VMAtm This change is part of v19.9.0 release.

PR #1680, opened 2019-09-18
Fix tests for multiprocessing pickle app and pickle blueprint

ashleysommer:
The old tests were not quite checking for the right thing.
Fixing the test does not change Sanic code, does not expose any bugs, or fix any bugs.

### DIFF
--- a/tests/test_multiprocessing.py
+++ b/tests/test_multiprocessing.py
@@ -5,6 +5,7 @@ import signal
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


ashleysommer:
The original tests were contributed by me here: https://github.com/huge-success/sanic/commit/5cf2144b3ff3438afe0c9f7dbcbcb29cb94c48e0 so its my own bad test I'm fixing.

Issue #530, opened 2017-03-08
mypy type annotations


graingert:
As this is a Python >= 3.5 library, it would be easy for callers of this library to support or ignore type annotations if they were added.

subyraman:
Would you be interested in helping to add type annotations?

graingert:
Yup, can get you setup with a simple mypy config

alexyer:
Is there any progress? I can help with it.

jgelens:
I'm slowly adding some annotations for usage in my own project . Do you want to integrate the annotations in the Sanic code itself or leave it as a .pyi stub file (that defines the types in a separate file)

feluxe:
Cython now understands PEP484 and PEP526 type annotations.

That means you can run the Cython compiler on (pure) Python files and get decent speed improvements.

On my machine it takes `9.79` seconds to run `func_a` `5000000` times: 

```python
def f(x):
    return x ** 2 - x

def func_a(a, b, N):

    s = 0
    dx = (b - a) / N

    for i in range(N):
        s += f(a + i * dx)

    return s * dx
```

The same code with some type annotations runs in `2.40` seconds (compiled with cython).

```python
import cython as c


def fd(x:c.double):
    return x ** 2 - x


def func_d(a:c.double, b:c.double, N:c.int):

    i:c.int

    s = 0
    dx = (b - a) / N

    for i in range(N):
        s += fd(a + i * dx)

    return s * dx

```

Maybe this is something sanic could make use of?

ahopkins:
There are two separate things going on here. First, it probably makes sense to create some,stub files to help developers. 

Second, making use of Cython. We've talked about this before and it may be in the table for some of the features in 2019. But I would suggest moving this conversation to its own thread. And since it is not a specific enhancement, I'd suggest starting a conversation on the [community forums](https://community.sanicframework.org/).

ahopkins:
Anyone interested to start working in some stub files? @sjsadowski and @yunstanford, 19.03?

yunstanford:
Yeah, I'd like to optimize it with `Cython`. I've done some work before.

vltr:
Yeah, this might give some performance boost but I think we should be able to have a pure Python implementation as a fallback in case installing binaries is not possible (ex: using pypy (or any other Python implementation that's not CPython); container Linux images without the proper build tools (there's an issue describing this problem [here](https://github.com/docker-library/docs/issues/904)) and perhaps some other scenarios).

[...]

PR #1682, opened 2019-09-20
Add a type checking pipeline

viniciusd:
This pull request aims to add provisional support to type checking via `mypy`.

It has been discussed at #530 

Running mypy against the current master prompts us for 80 issues


viniciusd:
Fixed 45 out of the 80 mypy errors so far.

viniciusd:
Fixed all the errors

PR #1635, opened 2019-07-19
Upgrade websockets, resolve incompatible issue between multidict and websockets

yunstanford:
Fix: https://github.com/huge-success/sanic/issues/1629
Resolve incompatible issue between multidict and websockets.

Issue #1597, opened 2019-05-28
Sanic v19.3 not support websockets v7.0.


jaggerwang:
Sanic v19.3 required websockets version >=6.0 and <7.0. Is there any break change from v6.0 to v7.0? Websockets v6.0 not support to run as a module like `python -m websockets wss://echo.websocket.org/`.

ahopkins:
Yes, that is correct. Sanic requires `websockets` v6 right now because v7 introduced some breaking changes that have not yet been adopted.

See #1214 and #1233. There are probably some others too I am missing.

If you are at all interested in taking this on, I can help point you in the direction needed.

---

If you want to run `websockets` that way, then why do you need Sanic on top of it? Is there a particular feature you are looking for in v7? I also should mention that #1475 is nearing completion and should make it into Sanic 19.6, which will give the ability to run for an ASGI server (which includes websocket support without the `websockets` package).

jaggerwang:
Got it! Right now just because I want to use websockets module as a client to test, but I can install v7.0 outside of the venv.

ahopkins:
:+1: FWIW, I added an alias to my terminal to conveniently allow me to connect using the CLI utility: `ws ws://localhost/feed`

```
ws='/home/adam/.virtualenvs/websockets/bin/python -m websockets'
```

Tronic:
I tried current git master with websockets 8.0.1 and all tests pass OK. The requirement for websockets<7 is causing dependency issues, especially when multiple web frameworks are used in the same project. Issues #1214 and #1233 appear to be about new features. Is there still something that is actually broken by using the latest websockets version?

ahopkins:
@Tronic The last I had tested it, there were some breaking changes in names. And, in running `examples/websocket.py`, it looks like that is still the case.

```
[2019-08-11 10:34:43 +0300] [22005] [ERROR] Exception occurred while handling uri: 'ws://localhost:8000/feed'
Traceback (most recent call last):
  File "/home/adam/Projects/Sanic/sanic/sanic/app.py", line 944, in handle_request
    response = await response
  File "/home/adam/Projects/Sanic/sanic/sanic/app.py", line 491, in websocket_handler
    request, subprotocols
  File "/home/adam/Projects/Sanic/sanic/sanic/websocket.py", line 75, in websocket_handshake
    key = handshake.check_request(request.headers)
[...]

PR #1687, opened 2019-09-30
Support websockets 8.x as well as 7.x

simonchatts:
Sanic currently requires websockets 7.x, but the [changelog for 8.x](https://websockets.readthedocs.io/en/stable/changelog.html) is pretty benign from Sanic's perspective (now that Python 3.5 has been dropped). All that's missing is [a declaration](https://github.com/aaugustin/websockets/blob/master/src/websockets/server.py#L75-L76) whether sanic has the client or server end of the websocket.

Supporting websockets 8.x is required for [nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/sanic/default.nix), since this defines a globally consistent set of Python packages that includes 8.x.

I've tested this with a websocket-based sanic app with websockets versions 7.0 and 8.0.2 and it works fine for me.

Thanks for Sanic by the way!

### DIFF
--- a/sanic/websocket.py
+++ b/sanic/websocket.py
@@ -119,6 +119,9 @@ class WebSocketProtocol(HttpProtocol):
             read_limit=self.websocket_read_limit,
             write_limit=self.websocket_write_limit,
         )
+        # Following two lines are required for websockets 8.x
+        self.websocket.is_client = False
+        self.websocket.side = "server"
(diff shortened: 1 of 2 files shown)


Tronic:
Related to #1597 and #1635. Thanks for your contribution!

PR #1690, opened 2019-10-08
GIT-37: fix blueprint middleware application

### LINKED ISSUE #37: Blueprint middleware applied globally

harshanarayana:
1. If you register a middleware via `@blueprint.middleware` then it will apply only to the routes defined by the blueprint.
2. If you register a middleware via `@blueprint_group.middleware` then it will apply to all blueprint based routes that are part of the group.
3. If you define a middleware via `@app.middleware` then it will be applied on all available routes

Fixes #37

### DIFF
--- a/changelogs/37.bugfix.rst
+++ b/changelogs/37.bugfix.rst
@@ -1,0 +1,11 @@
+Fix blueprint middleware application
+
+Currently, any blueprint middleware registered, irrespective of which blueprint was used to do so, was
+being applied to all of the routes created by the :code:`@app` and :code:`@blueprint` alike.
+
+As part of this change, the blueprint based middleware application is enforced based on where they are
+registered.
+
+- If you register a middleware via :code:`@blueprint.middleware` then it will apply only to the routes defined by the blueprint.
+- If you register a middleware via :code:`@blueprint_group.middleware` then it will apply to all blueprint based routes that are part of the group.
+- If you define a middleware via :code:`@app.middleware` then it will be applied on all available routes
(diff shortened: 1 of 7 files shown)


sjsadowski:
@harshanarayana do we need to add test to prevent reversions for this behavior?

harshanarayana:
@sjsadowski 
> do we need to add tests to prevent reversions for this behavior?

Existing tests have been modified to assert new behavior. So any change in this behavior again will cause those tests to fail. Do you happen to have any specific scenario in mind that I may have missed ?

harshanarayana:
@huge-success/sanic-core-devs  Can we get this one into the next release ? This is sort of a breaking change but in the right direction

harshanarayana:
cc @huge-success/sanic-core-devs

sjsadowski:
I've approved it. I'm going to add a note on sanicframework.org and figure out how to call it out in the generated release notes before merging.

harshanarayana:
> figure out how to call it out in the generated release notes before merging.

@sjsadowski This can be done easily. I've added a command `make release` that will generate the changelog based on the added changelogs snippet.

sjsadowski:
@harshanarayana awesome!

PR #1696, opened 2019-10-10
wraps websocket handler

andribas404:
Add wraps to websocket_handler.
Without it, all routes_names for websockets are identical and do not keep blueprint name.

Signed-off-by: Andrey Petukhov <a.petukhov@drweb.com>


andribas404:
Greetings,
probably we could split this PR to 2 pull requests:

- one that adds first test to @wraps method
- and second with tests for strict_slashes logic:

1. if `strict_slashes` is True, url must end with slash for "/route" and "/route/"?
2. `url_for` should always return valid url that result to response status code 200, not 404.
I think this statements needs clarifications from the maintainer.

Thank you for Sanic project.

Tronic:
I'd like to see a migration path for getting rid of the strict_slashes setting. The only sensible option in this regard is (1) if a route ends with slash, redirect GETs and 404 anything else trying to access without slash, (2) if a route does not end with slash, 404 anything without slash, or (3) if matching routes exist with and without slash, match strictly the one used in request.

Those wishing for non-strict functionality are free to rely on option 3 and explicitly specify two routes. The question is how to get there without breaking existing apps too badly.

ahopkins:
@Tronic I am with you on that. But, I think it needs to be part of a larger discussion that resolves a long list of items that need to be resolved with the router. I know I have been saying this for a while and there has been a lot of tangential talk about it, but I think 2020 should finally be the time when we tackle this item.

andribas404:
Sorry for impatience, changed pull request according to Single Responsibility Principle.

Tronic:
Tagging for 20.3.

ahopkins:
Conflicts with #1853. Need to move the `@wraps`. I will take a look at this later.

ahopkins:
Closing in favor of #1880

Issue #1601, opened 2019-06-05
A mysterious issue with Cython, sanic 19.3.1


heshiming:
I'm seeing a mysterious exception in `_build_endpoint_name` of `app.py`. It occurs only when the app server module is compiled using Cython. It happens to sanic 19.3.1 but not 18.12.0 (or below).

Here is a minimal example to replicate this issue in the docker environment. First, in an empty folder, create `dockerfile` with the following contents:

    FROM python:3.7-stretch
    RUN pip install sanic cython gunicorn

Then, in the same folder, create a very simple server `app.py` like this:

    from sanic import Sanic, response

    app = Sanic()

    @app.route('/')
    async def root(request):
        return response.text('Hello')

Build the docker image:

    $ docker build --rm -f dockerfile -t sanic-test ./

Launch and go inside the container:

    $ docker run --rm --name sanic-test -it -v `pwd`:/app -w="/app" -p 127.0.0.1:5000:5000/tcp sanic-test bash

Run the following command to get `app.py` compiled:

    cython app.py -3
    gcc -shared -pthread `python3-config --includes` -fPIC -fwrapv -O2 -Wall -fno-strict-aliasing -o app.cpython-37m-x86_64-linux-gnu.so app.c

Now launch the compiled module using `gunicorn`:

    $ gunicorn app:app --bind 0.0.0.0:5000 --worker-class sanic.worker.GunicornWorker
    [2019-06-05 03:32:57 +0000] [7] [INFO] Starting gunicorn 19.9.0
    [2019-06-05 03:32:57 +0000] [7] [INFO] Listening at: http://0.0.0.0:5000 (7)
    [2019-06-05 03:32:57 +0000] [7] [INFO] Using worker: sanic.worker.GunicornWorker
    [2019-06-05 03:32:57 +0000] [10] [INFO] Booting worker with pid: 10

The docker container is set up with external port mapping. So on your host machine, you can do:

    $ curl http://localhost:5000

Which generates:

    [2019-06-05 03:32:59 +0000] [10] [ERROR] Exception occurred while handling uri: 'http://localhost:5000/'
    Traceback (most recent call last):
      File "/usr/local/lib/python3.7/site-packages/sanic/app.py", line 906, in handle_request
        handler.__name__
      File "/usr/local/lib/python3.7/site-packages/sanic/app.py", line 1355, in _build_endpoint_name
        return ".".join(parts)
    TypeError: sequence item 0: expected str instance, NoneType found
[...]

PR #1705, opened 2019-10-22
deprecate None value support for app name

harshanarayana:
The `None` value support behavior for the app name can be hard to mange when things change/ the app runner changes. This can lead to problems. It is easier instead to turn the `name` into a mandatory parameter and let the user decide what they want to use. 

This PR marks the `None` values for app names as `deprecated`. 

fixes: #1601

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -51,7 +51,7 @@ class Sanic:
                 "for `name` will be removed in the next release. "
                 "Please use Sanic(name='your_application_name') instead.",
                 DeprecationWarning,
-                stacklevel=2
+                stacklevel=2,
             )
             frame_records = stack()[1]
             name = getmodulename(frame_records[1])
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -46,6 +46,13 @@ class Sanic:
 
         # Get name from previous stack frame
         if name is None:
+            warnings.warn(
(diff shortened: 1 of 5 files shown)


sjsadowski:
Do we want to deprecate the None value with a warning or do we want to change it so that if there is no value we set a default name like "default" and deprecate with a warning that a default name will be set?

Tronic:
I never used app name, didn't even know one could be defined. Seems useless to invent a name not used anywhere, so where is this needed?

sjsadowski:
@Tronic just from quick review, this is cython specific.

It may make more sense just to update the docs to include cython caveats.

harshanarayana:
@sjsadowski  @Tronic 

I personally believe in `explicit is better than implicit` when it come to naming things. The way to derive this info current relies on the `inspect` package. Which is fine, but then again, I personally prefer the explicit way of specifying this as flask does. 

Now coming to the `cython` thing, as I mentioned in the thread, I could not reproduce the problem. So I am not sure if the stack extract and get the module on that is behaving flaky for some reason. 

But if we all believe it is okay to use the `None` default for name and then derive it via the internal mechanism, I am open to that and I will revert these changes and make the necessary changes in the documentation to indicate the specific behavior.

PR #1708, opened 2019-10-23
fix SERVER_NAME enforcement in url_for and request.args documentation

### LINKED ISSUE #1704: Improve documentation in *Accessing values using `get` and `getlist`*
Documentation here should be improved:
https://sanic.readthedocs.io/en/latest/sanic/request_data.html#accessing-values-using-get-and-getlist
It isn't clear how to use `get` and `getlist`

Change
> The request properties which return a dictionary actually return a subclass of dict called RequestParameters.

To
> `request.args` which return a dictionary actually return a subclass of dict called RequestParameters.

### LINKED ISSUE #1707: Using request.url_for throws KeyError: 'SERVER_NAME' in 19.9.0
**Describe the bug**
Running Sanic behind a proxy and trying to generate an url using `request.url_for` throws a KeyError. This used to work but now Sanic suddenly requires a SERVER_NAME in order to do this. I get the following trace:

```
  File "/usr/local/lib/python3.7/site-packages/sanic/app.py", line 942, in handle_request
    response = await response
  File "/app/tsauth/views/activate.py", line 41, in init_activate
    request=request)
  File "/usr/local/lib/python3.7/site-packages/mako/template.py", line 476, in render
    return runtime._render(self, self.callable_, args, data)
  File "/usr/local/lib/python3.7/site-packages/mako/runtime.py", line 883, in _render
    **_kwargs_for_callable(callable_, data)
  File "/usr/local/lib/python3.7/site-packages/mako/runtime.py", line 920, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File "/usr/local/lib/python3.7/site-packages/mako/runtime.py", line 947, in _exec_template
    callable_(context, *args, **kwargs)
  File "transaction_mako", line 60, in render_body
  File "/usr/local/lib/python3.7/site-packages/sanic/request.py", line 522, in url_for
    if "//" in self.app.config.SERVER_NAME:
  File "/usr/local/lib/python3.7/site-packages/sanic/config.py", line 54, in __getattr__
    raise AttributeError("Config has no '{}'".format(ke.args[0]))
AttributeError: Config has no 'SERVER_NAME'
[2019-10-23 12:38:09 +0000] - (sanic.access)[INFO][<redacted>:33328]: GET <redacted>  500 144
10/23/2019 12:38:09 PM ERROR Exception occurred while handling uri: '<redacted>'
  File "/usr/local/lib/python3.7/site-packages/sanic/config.py", line 52, in __getattr__
    return self[attr]
KeyError: 'SERVER_NAME'
```

**Code snippet**
`request.url_for("static",  name='static', filename="badge.png")`

**Expected behavior**
`request.url_for` should generate a full url with the given path, e.g. `https://hostname:port/static/badge.png`. As I am running my server behind a proxy, I do not want to hardcode the hostname in the SERVER_NAME. This used to work with Sanic version 19.6.3.

 - Python 3.7 Docker container, Kubernetes v1.16.2 with nginx-ingress 0.26.1
 - Sanic version 19.9.0

**Additional context**
The calling code to `request.url_for` has not changed since the Sanic version update. 
The documentation also does not speak a word about why the SERVER_NAME variable needs to be configured...

harshanarayana:
fixes #1707

fixes #1704

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -519,8 +519,11 @@ class Request:
         :rtype: str
         """
         # Full URL SERVER_NAME can only be handled in app.url_for
-        if "//" in self.app.config.SERVER_NAME:
-            return self.app.url_for(view_name, _external=True, **kwargs)
+        try:
+            if self.app.config.SERVER_NAME:
+                return self.app.url_for(view_name, _external=True, **kwargs)
+        except AttributeError:
+            pass
 
         scheme = self.scheme
         host = self.server_name
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -520,7 +520,7 @@ class Request:
         """
         # Full URL SERVER_NAME can only be handled in app.url_for
         try:
-            if self.app.config.SERVER_NAME:
+            if "//" in self.app.config.SERVER_NAME:
                 return self.app.url_for(view_name, _external=True, **kwargs)
         except AttributeError:
             pass
--- a/changelogs/1704.doc.rst
+++ b/changelogs/1704.doc.rst
@@ -1,0 +1,3 @@
+Fix documentation for `get` and `getlist` of the `request.args`
+
+Add additional example for showing the usage of `getlist` and fix the documentation string for `request.args` behavior
--- a/changelogs/1707.bugfix.rst
+++ b/changelogs/1707.bugfix.rst
@@ -1,0 +1,4 @@
+Fix `url_for` behavior with missing SERVER_NAME
+
+If the `SERVER_NAME` was missing in the `app.config` entity, the `url_for` on the `request` and  `app` were failing
+due to an `AttributeError`. This fix makes the availability of `SERVER_NAME` on our `app.config` an optional behavior.
--- a/docs/sanic/request_data.md
+++ b/docs/sanic/request_data.md
@@ -204,9 +204,8 @@ The output will be:
 
 ## Accessing values using `get` and `getlist`
 
-The request properties which return a dictionary actually return a subclass of
-`dict` called `RequestParameters`. The key difference when using this object is
-the distinction between the `get` and `getlist` methods.
+The `request.args` returns a subclass of `dict` called `RequestParameters`. 
+The key difference when using this object is the distinction between the `get` and `getlist` methods.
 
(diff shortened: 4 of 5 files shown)


sanic/request.py
```diff
@@ -519,8 +519,11 @@ def url_for(self, view_name, **kwargs):
         :rtype: str
         """
         # Full URL SERVER_NAME can only be handled in app.url_for
-        if "//" in self.app.config.SERVER_NAME:
-            return self.app.url_for(view_name, _external=True, **kwargs)
+        try:
+            if "//" in self.app.config.SERVER_NAME:
+                return self.app.url_for(view_name, _external=True, **kwargs)
+        except AttributeError:
+            pass
```
Tronic: What is this expected to do? I gather that full URL will still be mishandled by the logic that follows.

Tronic: Nevermind my earlier comment; this is OK, although checking in the if instead of try-except could be clearer.

harshanarayana: > Perhaps you could also add a request.url_for test without SERVER_NAME?

I think there are tests around this already and don't think there is a need to add another one. But Let me check the coverage and see if it's missing. In case if missing, let me add one. 

> although checking in the if instead of try-except could be clearer.

I did not want to do an additional check. If this works, great, we go ahead and return and if it doesn't we just ignore and continue. This way we don't add anything extra in the branching structure.

Tronic: `request.server_name` property actually uses `app.config.get("SERVER_NAME")` and tests that for truthiness. That way `SERVER_NAME=""` behaves the same as if it wasn't defined at all.

Tronic: As for tests, I gather that this wasn't tested for because otherwise #1707 would've been caught in testing. I suppose that only `app.url_for` gets thorough testing and that all existing `request.url_for` tests define `SERVER_NAME` to something.

harshanarayana: > this wasn't tested for because otherwise #1707 would've been caught in testing

Agreed. I realized that after I wrote the message and hit submit.

harshanarayana: @Tronic Added an extra test to make sure there is a check when `SERVER_NAME` is not set.

PR #1710, opened 2019-10-23
Add python nightlies to test matrix

seemethere:
The goal of this is to test sanic against the nightly versions of python but still allow failures.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,11 +1,11 @@
 [tox]
-envlist = py36, py37, {py36,py37}-no-ext, lint, check, security, docs
+envlist = py36, py37, pyNightly, {py36,py37,pyNightly}-no-ext, lint, check, security, docs
 
 [testenv]
 usedevelop = True
 setenv =
-    {py36,py37}-no-ext: SANIC_NO_UJSON=1
-    {py36,py37}-no-ext: SANIC_NO_UVLOOP=1
+    {py36,py37,pyNightly}-no-ext: SANIC_NO_UJSON=1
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,11 +1,11 @@
 [tox]
-envlist = py36, py37, {py36,py37}-no-ext, lint, check, security, docs
+envlist = py36, py37, pyNightly, {py36,py37,pyNightly}-no-ext, lint, check, security, docs
```
harshanarayana: Can we restrict the `nightly` build check/support only under `travis`? For `tox` it's better to have a fixed version that developers can use. Nightly might not always work and can mark tests for failure.

seemethere: Is there any reason to have any different local testing environments with `py36` vs `py37`?

Would it be better to just change this to `py3` and let travis handle installing the separate python versions to test.

harshanarayana:
@seemethere The behavior is pretty interesting. In the `no-ext` env of the travis, you have `SANIC_NO_UVLOOP`. However, it is still trying to install `uvloop`. huh! Interesting and weird

ahopkins:
@seemethere What do we need to do to push this over the finish line? It is still marked as `WIP`.

harshanarayana:
@ahopkins uvloop officially has 3.8 support. So this should hopefully be smooth sailing

ahopkins:
Removing `WIP`. Unless I am missing something...

harshanarayana:
@ahopkins  Can you please re-trigger the tests once so that we can do a quick check on the test failure status ?

ahopkins:
Restarted.

PR #1716, opened 2019-11-17
Fix #1714

### LINKED ISSUE #1714: When calling abort() in a stream=True handler, error occurs
**Describe the bug**

When calling abort() inside a request handler with stream=True, this exception occurs in the logs:
```
future: <Task finished name='Task-3' coro=<HttpProtocol.body_append() done, defined at /usr/local/lib/python3.8/site-packages/sanic/server.py:356> exception=AttributeError("'NoneType' object has no attribute 'stream'")>
  File "/usr/local/lib/python3.8/site-packages/sanic/server.py", line 357, in body_append
    if self.request.stream.is_full():
AttributeError: 'NoneType' object has no attribute 'stream'
```
**Code snippet**

```
import cgi

from sanic import Sanic
from sanic.exceptions import abort

app = Sanic('request_stream')


@app.put('/stream', stream=True)
async def handler(request):
    mimetype, options = cgi.parse_header(request.headers['content-type'])
    charset = options.get('charset')

    if charset not in {'utf-8'} and mimetype != 'text/plain':
        abort(400, f'Content-Type must be "text/plain; charset=utf-8')

    async def streaming(response):
        while (body := await request.stream.read()) is not None:
            body = body.decode(charset).upper()
            await response.write(body)

    return stream(streaming, content_type="text/plain; charset=utf-8")
```

**Expected behavior**
On the client side, behavior is as expected, but it seems that request is None in this case.  It should not be

Python3.8 official docker image
Sanic 19.9.0

**Additional context**
/

yunstanford:
Fix https://github.com/huge-success/sanic/issues/1714

`AttributeError: 'NoneType' object has no attribute 'stream'`

It happens mainly because the `cleanup` function call (https://github.com/huge-success/sanic/blob/master/sanic/server.py#L593) cleaned `request` context, while the body_append asyncio.task in event loop still relies on that.

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -354,6 +354,9 @@ class HttpProtocol(asyncio.Protocol):
             self.request.body_push(body)
 
     async def body_append(self, body):
+        if self._request_stream_task is None or self._request_stream_task.cancelled():
+            return
+
         if self.request.stream.is_full():
             self.transport.pause_reading()
             await self.request.stream.put(body)
@@ -598,8 +601,12 @@ class HttpProtocol(asyncio.Protocol):
         self.request = None
         self.url = None
         self.headers = None
-        self._request_handler_task = None
-        self._request_stream_task = None
+        if self._request_handler_task:
+            self._request_handler_task.cancel()
+            self._request_handler_task = None
+        if self._request_stream_task:
+            self._request_stream_task.cancel()
+            self._request_stream_task = None
         self._total_request_size = 0
         self._is_stream_handler = False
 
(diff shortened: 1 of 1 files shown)


seemethere:
I think this would be a good use case for the `any` function to make it a bit more clear on why this is actually doing. 

Also appears as though there’s a merge conflict.

sjsadowski:
@yunstanford can you take a look at the conflict when you get a chance?

yunstanford:
sure, will take a look

PR #1736, opened 2019-12-22
Allow empty body without Content-Type; Introduce response.empty()

LiraNuna:
Sometimes, a 204 response is just enough to signal the client what the backend wants to do. Currently, this is possible but has some awkwardness.
First, an empty response has no `Content-Type`, however all responses require Content-Type header.

This PR allows `content_type` param in `HTTPResponse` to be `None`. It is the simplest solution, however I have to say I am not a big fan.
A big disadvantage is the fact that the content type is now located in two locations: `self.headers` and `self.content_type` and they COULD be out of sync, since `headers` have a higher precedence.
This means that if one passes `text/plain` to `content_type` but `Content-Type: application/json` in `headers`, the final result will be an inconsistant response object with one field saying one thing and the other another.

Since this behavior exists before this PR, I did not set out to fix it in this enhancement PR. I personally like to keep PRs atomic, and therefor will contribute a refactor if requested in a sequetial PR.

Fixes https://github.com/huge-success/sanic/issues/1735

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -153,7 +153,7 @@ class HTTPResponse(BaseHTTPResponse):
         body=None,
         status=200,
         headers=None,
-        content_type="text/plain",
+        content_type=None,
         body_bytes=b"",
     ):
         self.content_type = content_type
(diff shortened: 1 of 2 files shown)


ahopkins:
Looks good to me. I also like the atomic approach, and would welcome a second PR on that point.

Tronic:
Looks good, excellent reasoning and I absolutely agree on a second PR that might also get rid of `body`/`body_bytes` duplication as well as the `content_type` argument (but taking compatibility concerns into account).

LiraNuna:
How  would you address compatibility concerns with a removal of a parameter? It is a breaking change to interface. 

I will gladly merge the `body`/`body_bytes` param in addition to eliminating `content_type` property (I was planning on keeping it in headers and make a getter for compatibility)

ahopkins:
@LiraNuna Let's open this discussion in a new conversation. I have some thoughts to share.

@huge-success/sanic-release-managers this one is good to go.

PR #1738, opened 2019-12-24
Simplify status code to text lookup

LiraNuna:
While working on a followup to https://github.com/huge-success/sanic/pull/1736, I noticed that there's a bizarre check that is redundant. 
Since the map `STATUS_CODES` already contains the staus code of 200, there is no need to check for it specifically. 

This could be done "in the name of performance" but I challenge that fact.

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -130,12 +130,7 @@ class StreamingHTTPResponse(BaseHTTPResponse):
         )
 
         headers = self._parse_headers()
-
-        if self.status == 200:
-            status = b"OK"
-        else:
-            status = STATUS_CODES.get(self.status)
-
+        status = STATUS_CODES.get(self.status, b"UNKNOWN RESPONSE")
         return (b"HTTP/%b %d %b\r\n" b"%b" b"%b\r\n") % (
             version.encode(),
             self.status,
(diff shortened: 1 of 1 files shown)


sanic/response.py
```diff
@@ -130,12 +130,7 @@ def get_headers(
...
-
-        if self.status == 200:
-            status = b"OK"
-        else:
-            status = STATUS_CODES.get(self.status)
-
+        status = STATUS_CODES.get(self.status)
         return (b"HTTP/%b %d %b\r\n" b"%b" b"%b\r\n") % (
             version.encode(),
             self.status,
```
LiraNuna: Please note that unlike line 187, there is no default `UNKNOWN RESPONSE` response here, which means this could be `None` and cause issues. Should this be fixed to mimic the behavior below?

Tronic:
Dict lookup in Python is extremely fast, so there probably isn't any performance gain. Consistently handling with `UNKNOWN RESPONSE` would make sense.

LiraNuna:
@Tronic, changes made. Do you know why the coverage was lowered? I removed lines of code - I never seen this happen before. 

Is this acceptable?

LiraNuna:
It's been a while since this was approved, is there any additional thing I need to do?

ahopkins:
@huge-success/sanic-release-managers

PR #1739, opened 2019-12-24
Deprecate body_bytes to merge into body

LiraNuna:
As suggested in https://github.com/huge-success/sanic/pull/1736#issuecomment-568266049, this PR deprecates `body_bytes` and merges it into `body`.

Please note that I have intentionally removed the `body_bytes` param completely to elicit discussion about how to properly deprecate it (cc @ahopkins who has some thoughts to share).

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -149,6 +149,12 @@ class HTTPResponse(BaseHTTPResponse):
         self.headers = Header(headers or {})
         self._cookies = None
 
+        if body_bytes:
+            warnings.warn(
+                "Parameter `body_bytes` is deprecated, use `body` instead",
+                DeprecationWarning,
+            )
+
     def output(self, version="1.1", keep_alive=False, keep_alive_timeout=None):
         body = b""
         if has_message_body(self.status):
@@ -173,7 +179,7 @@ def empty(status=204, headers=None):
     :param status Response code.
     :param headers Custom Headers.
     """
-    return HTTPResponse(body_bytes=b"", status=status, headers=headers)
+    return HTTPResponse(body=b"", status=status, headers=headers)
 
 
 def json(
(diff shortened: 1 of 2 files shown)


Tronic:
Check out #1674 as there is some overlap, in particular with how bytes vs. other types are detected.

LiraNuna:
Sure, I can upgrade body to bytes but that will keep the weirdness that is the two parameters merging into the same thing. 

Here are the questions I have in a more blunt form:
* How will the developer know to stop using `body_bytes` and to start using `body`?
* When will `body_bytes` be removed?
* What happens if both `body` and `body_bytes` are passed in?
* And what's up with `black` adding comma to line 208?

LiraNuna:
@Tronic, my main questions are "is it okay to deprecate this parameter?" and "how will library users know to avoid code breakage?"

I'm assuming a version bump and a note in the CHANGELOG, but is it this PR's responsibility to do that? 

I think this PR is ready to review assuming the param deprecation is wanted.

ahopkins:
Sorry for not jumping in with thoughts sooner. I was sort of waiting for [the new organizational structure](https://sanicframework.org/2020-release-managers-and-a-new-oraganizational-structure) to kick in and let there be a bit more of a formalized discussion about how to handle deprecations.

I would like to put this one on hold until that conversation takes place. I expect the Steering Council to be in place within the next week or two.

Tronic:
@ahopkins Any updates on that? I'd like to see this merged for 20.3.

LiraNuna:
What is stopping this PR from being merged? I'd love to get it over the finish line

Tronic:
Alternatively, modify this PR so that it still takes body_bytes arguments but gives a warning whenever those are used.

LiraNuna:
@Tronic how does this look? I chose to accept the old param so when the true deprecation comes, there are more options

PR #1752, opened 2020-01-02
Py38 tox env

ahopkins:
Adds Python 3.8 to PyPI classifiers, `tox` environments, and Travis builds


sanic/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "19.12.0"
+__version__ = "19.12.2"
```
seemethere: Does the version need to be bumped for this?

ahopkins: Nope. It was a mistake and was meant to be bumped on a previous commit.

.travis.yml
```diff
@@ -21,23 +21,46 @@ matrix:
       dist: xenial
       sudo: true
       name: "Python 3.7 without Extensions"
+    - env: TOX_ENV=py38
```
seemethere: There has to be a better way of doing this instead of just copy / pase the config block.

Not really a blocker on this PR but just a note for us to look for in the future.

ahopkins:
@huge-success/sanic-core-devs How do you guys want to handle this? This PR really should get merged so we have the Python 3.8 builds. BUT .... it is failing because apparently there is some work to do on Windows on 3.8.

Should we break it off? Mark that one as an accepted failure? Remove Python 3.8 from appveyor tests?

harshanarayana:
@ahopkins Does appveyor support marking some as expected test failures like it was done for travis where the failures are ignored for now?

ahopkins:
> To allow failing jobs add matrix.allow_failures section into appveyor.yml

Yes.

sjsadowski:
Coming back around to this - appveyor is purely windows and builds pass even if appveyor fails. I'm going to merge this in so if we have someone who wants to focus on improving windows support, which is currently best effort, and now noted as such in the documentation per #1778

PR #1760, opened 2020-01-09
Add version parameter to websocket routes

### LINKED ISSUE #1757: Websocket doesn't work when using version parameter
**Describe the bug**
When defining a websocket handler using `version=1` a exceptions will be thrown.

**Code snippet**
Relevant source code, make sure to remove what is not necessary.
```
@app.websocket("/changes", version=1)
```

**Expected behavior**
It should work like without `version=1`

 - OS: Fedora 31
 - Sanic: 19.12.2

ahopkins:
Resolves #1757


sanic/app.py
```diff
@@ -451,7 +451,13 @@ def add_route(
 
     # Decorator
     def websocket(
-        self, uri, host=None, strict_slashes=None, subprotocols=None, name=None
+        self,
+        uri,
+        host=None,
+        strict_slashes=None,
+        subprotocols=None,
+        version=None,
```
seemethere: I believe we should add the `version` parameter after the already existing `name` parameter since some people may be using positional args instead of named args and we don't want to break those users

sanic/app.py
```diff
@@ -522,6 +528,7 @@ def response(handler):
                 methods=frozenset({"GET"}),
                 host=host,
                 strict_slashes=strict_slashes,
+                version=version,
```
seemethere: Same applies here as well

sanic/app.py
```diff
@@ -535,6 +542,7 @@ def add_websocket_route(
         host=None,
         strict_slashes=None,
         subprotocols=None,
+        version=None,
```
seemethere: Here too

sanic/app.py
```diff
@@ -562,6 +570,7 @@ def add_websocket_route(
             host=host,
             strict_slashes=strict_slashes,
             subprotocols=subprotocols,
+            version=version,
```
seemethere: And here as well

sanic/app.py
ahopkins: I did it this way to be consistent with the order of the parameters on the other routes where name is the final.

What's worse? Inconsistent ordering or changing positional order?

danieldaeschle: Maybe nobody really used that feature before. Nobody has complained about that bug. 😅

sjsadowski:
@seemethere any further commentary after @ahopkins response

sjsadowski:
@ahopkins providing there's no further commentary, if you'll resolve the conflicts I'll get this merged.

Tronic:
@ahopkins The conflict still needs resolving.

ahopkins:
@sjsadowski

ahopkins:
@huge-success/sanic-core-devs Can someone review and approve so we can merge?

PR #1764, opened 2020-01-10
Allow route decorators to stack up again

### LINKED ISSUE #1742: Changes to Blueprints application breaks multiply applied routes!!!
Consider the following pattern:

```
@app.get("/url1")
@app.get("/url2")
async def handle_multiple(request):
     return text("some response")
```

The recent change in #1690 breaks that pattern.

It's a common pattern to use multiple `@blueprint.get(...)` and `@app.get(...)` to handle routes that have common behavior with slight variations, like a login process.

The only alternative I can think of is splitting out each of these routes into distinct rump functions that literally forward onto a common function, which inflates from one line declarations to three lines plus the constant common route handler.

ashleysommer:
A PR:https://github.com/huge-success/sanic/pull/1690 broke the ability to stack route decorators.

This PR will allow again route decorators to stack up without causing a function signature inspection crash
Fix #1742

Also adds a test to detect this breakage, and verify its now fixed, and to ensure it stays fixed in the future.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -194,6 +194,10 @@ class Sanic:
             strict_slashes = self.strict_slashes
 
         def response(handler):
+            if isinstance(handler, (tuple, list)):
+                routes, handler = handler
+            else:
+                routes = []
             args = list(signature(handler).parameters.keys())
 
             if not args:
@@ -205,7 +209,7 @@ class Sanic:
             if stream:
                 handler.is_stream = stream
 
-            routes = self.router.add(
+            new_routes = self.router.add(
                 uri=uri,
                 methods=methods,
                 handler=handler,
@@ -214,6 +218,7 @@ class Sanic:
                 version=version,
                 name=name,
             )
+            routes.extend(new_routes)
             return routes, handler
 
         return response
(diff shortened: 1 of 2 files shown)


sanic/app.py
```diff
@@ -194,6 +194,10 @@ def route(
             strict_slashes = self.strict_slashes
 
         def response(handler):
+            if isinstance(handler, (tuple, list)):
```
yunstanford: Can we add some comments here for explaining?

sanic/app.py
```diff
@@ -205,7 +209,7 @@ def response(handler):
             if stream:
                 handler.is_stream = stream
 
-            routes = self.router.add(
+            new_routes = self.router.add(
```
yunstanford: new_routes seems a little bit weird.. maybe we can just extend in place.

sanic/app.py
ashleysommer: Added a comment explaining it

sanic/app.py
ashleysommer: Changed to inplace extend.

ashleysommer:
Pushed some changes, added the same fix for `@websocket()` routes too, added a new test for double-stacked websocket routes.

PR #1767, opened 2020-01-19
[Trio] Quick fixes to make Sanic usable on hypercorn -k trio myweb.app

Tronic:
Quite surprisingly it only took a few lines of changes to make Sanic run on Trio (with Trio-using webapp), on top of Hypercorn. If it is that easy, I'd like to include this in mainline, provided that it can be configured in some smart way (the first commit lacks configuration).

Is there some way that the library could be chosen at module load time, to avoid importing aiofiles when running on trio?

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -3,10 +3,7 @@ from mimetypes import guess_type
 from os import path
 from urllib.parse import quote_plus
 
-#from aiofiles import open as open_async  # type: ignore
-from trio import open_file as open_async
-
-from sanic.compat import Header
+from sanic.compat import Header, open_async
 from sanic.cookies import CookieJar
 from sanic.headers import format_http1
 from sanic.helpers import STATUS_CODES, has_message_body, remove_entity_headers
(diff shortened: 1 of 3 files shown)


sjsadowski:
We're going to need to update the CI tests for this, but I like where it's going. Could you add a quick section in docs/sanic/examples.rst for it?

sanic/response.py
```diff
@@ -3,7 +3,8 @@
 from os import path
 from urllib.parse import quote_plus
 
-from aiofiles import open as open_async  # type: ignore
+#from aiofiles import open as open_async  # type: ignore
+from trio import open_file as open_async
```
yunstanford: The easy way'd be defining those incompatible imports in `compat.py` module, then import things from compat.

Tronic: Somewhat better now but detection of environment by argv is far from optimal...

sjsadowski:
@Tronic linter is complaining but everything else looks good now

sjsadowski:
LGTM.

Tronic:
Okay. FWIW, this is a bit hacky and probably not 100 % functional/problem-free in Trio mode yet.

sjsadowski:
We'll just note in the docs that it's 'experimental support'

Issue #1766, opened 2020-01-19
sanic.exceptions.NotFound: Requested URL /favicon.ico not found


mystic-cg:
**Describe the bug**
like as
https://github.com/RasaHQ/rasa/issues/4129

**Code snippet**
```python
from sanic import Sanic
from sanic.response import json

app = Sanic("blog")


@app.route("/")
async def test(request):
    return json({"hello": "world"})


if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8000)
```

**Expected behavior**
Without any exception

**Environment (please complete the following information):**
 - OS: `win10.1709`
 - Python: `3.8.1`
 - Sanic: `19.12.2`

**Additional context**
> [2020-01-19 12:00:10 +0800] [31552] [INFO] Goin' Fast @ http://0.0.0.0:8000
[2020-01-19 12:00:10 +0800] [31552] [WARNING] Sanic tried to use loop.add_signal_handler but it is not implemented on this platform.
[2020-01-19 12:00:10 +0800] [31552] [WARNING] Sanic tried to use loop.add_signal_handler but it is not implemented on this platform.
[2020-01-19 12:00:10 +0800] [31552] [INFO] Starting worker [31552]
[2020-01-19 12:00:26 +0800] - (sanic.access)[INFO][127.0.0.1:51041]: GET http://127.0.0.1:8000/  200 17
[2020-01-19 12:00:26 +0800] [31552] [ERROR] Exception occurred while handling uri: 'http://127.0.0.1:8000/favicon.ico'
Traceback (most recent call last):
  File "C:\Users\someone\.virtualenvs\LearningSanic-Q13QjVkW\lib\site-packages\sanic\app.py", line 940, in handle_request
    handler, args, kwargs, uri, name = self.router.get(request)
  File "C:\Users\someone\.virtualenvs\LearningSanic-Q13QjVkW\lib\site-packages\sanic\router.py", line 411, in get
    return self._get(request.path, request.method, "")
  File "C:\Users\someone\.virtualenvs\LearningSanic-Q13QjVkW\lib\site-packages\sanic\router.py", line 475, in _get
    raise NotFound("Requested URL {} not found".format(url))
sanic.exceptions.NotFound: Requested URL /favicon.ico not found
[2020-01-19 12:00:26 +0800] - (sanic.access)[INFO][127.0.0.1:51041]: GET http://127.0.0.1:8000/favicon.ico  404 43

mystic-cg:
[...]

PR #1768, opened 2020-01-20
No tracebacks on normal errors and prettier error pages

Tronic:
Refactored default exception handling and error pages.

- Tracebacks are logged for uncaught SanicExceptions now only if status=500 or quiet=False.
    * No traceback for 404 NotFound, 405 MethodNotAllowed, 503 ServiceUnavailable etc.
- Uncaught Internal Server Errors and other exceptions still cause a traceback to be logged.
- Override by exception object's flag quiet=True/False (if not defined or None, the above rules apply).
- New default error pages (errorpages.py):
    * Production mode used to be text/plain, now all responses are HTML5.
    * Now only logged (non-quiet) exceptions are shown, still only in debug mode.


ashleysommer:
Oh, you beat me to it.
I just finished writing some code which does a very similar thing to fix #1766, I was about to open a PR.

My implementation took a different approach. It added the ability to flag each subclass of SanicException with a boolean whether or not to log the traceback. And by default the only one it sets to false for now is NotFound.

Tronic:
Refactoring HTML formatting to a separate module, and traceback pages at least escape their strings before formatting into HTML. Ideally the traceback code would be entirely rewritten, possibly using https://github.com/tronic/niceback as basis. But this PR has already bloated way beyond its original purpose...

@huge-success/sanic-release-managers review?

vltr:
Erm ... Can't we make use of the `DEBUG` flag to show (or not) these errors? I would like to see them in development but not in production mode ... But what to show on production? Keep the same behavior as original?

Tronic:
It still doesn't make sense to print tracebacks for 404 or most other errors even if in debug mode because the only thing that traceback is going to show is Sanic router internals. But if you have a bug in your code and some unexpected exception flies out, you do get a traceback in log (even in production mode) and if in debug mode, also in browser.

vltr:
> It still doesn't make sense to print tracebacks for 404 or most other errors even if in debug mode because the only thing that traceback is going to show is Sanic router internals.

Indeed.

> But if you have a bug in your code and some unexpected exception flies out, you do get a traceback in log (even in production mode) and if in debug mode, also in browser.

That makes sense to me.

If all these scenarios are covered, then I think it'll be a great addition. Just update the PR description (I really took a very quick look at the code and thought it was not taking `DEBUG` into consideration ... Anyway).

I really like your project for tracebacks and code errors, [niceback](https://github.com/tronic/niceback). I was also thinking that we (Sanic) lack some development-friendly tools ... Well, I plan on creating a topic on the forums about that soon :wink:

Tronic:
@vltr Description updated.

PR #1769, opened 2020-01-21
Code cleanup in file responses

Tronic:
Avoids ugly hack of earlier commit and cleans up the other code as well.

Minor difference with streaming file response: headers are now sent without opening the file, so any errors will terminate the HTTP connection. Old behaviour was an uncaught exception causing a 500 response to be sent.

Ideally those errors would be handled and proper code (404, 403) be issued if opening the file fails.

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -293,22 +293,20 @@ async def file(
     headers = headers or {}
     if filename:
         headers.setdefault(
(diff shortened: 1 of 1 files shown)


Tronic:
Be damned fixed line lengths, black uglified the code.

Tronic:
This can be merged. I'll make another PR if there is further development.

Issue #1772, opened 2020-01-26
Using list with host parameter in blueprint.route() does not work.


damianj:
**Describe the bug**
As the title implies, when I try to set the host parameter to a list on a blueprint route, the app does not start and instead gives an error like:

```
Traceback (most recent call last):
  File "test.py", line 16, in <module>
    parent_app.blueprint(child_bp)
  File "/usr/lib/python3.8/site-packages/sanic/app.py", line 757, in blueprint
    blueprint.register(self, options)
  File "/usr/lib/python3.8/site-packages/sanic/blueprints.py", line 146, in register
    route_names = [route.name for route in routes]
  File "/usr/lib/python3.8/site-packages/sanic/blueprints.py", line 146, in <listcomp>
    route_names = [route.name for route in routes]
AttributeError: 'NoneType' object has no attribute 'name'
```

**Code snippet**
To reproduce that error:
```
from sanic import Sanic, Blueprint
from sanic.response import text

child_bp = Blueprint(name='child_blueprint', url_prefix='/')
parent_app = Sanic(__name__)


@child_bp.route('/', host=["localhost:8000", "127.0.0.1:8000"])
async def hello_world(request):
    return text("Hello world!")

parent_app.blueprint(child_bp)
parent_app.run()
```
Note that if host is set to a string (i.e., `@child_bp.route('/', host="127.0.0.1:8000")`) or it's not even specified in the route (i.e., `@child_bp.route('/')`) then the app runs just fine.

**Expected behavior**
I would expect that the host parameter would function similar to that described in [https://sanic.readthedocs.io/en/latest/sanic/routing.html#http-request-types](https://sanic.readthedocs.io/en/latest/sanic/routing.html#http-request-types)

> There is also an optional host argument (which can be a list or a string). This restricts a route to the host or hosts provided. If there is a also a route with no host, it will be the default.

**Environment (please complete the following information):**
 - OS: Arch Linux (Kernel 5.4.13)
 - Version: 19.12.2

**Additional context**
This was working previously so I don't know if it's a bug or some intentional change in functionality.

damianj:
I've opened #1776 to address this issue.

ahopkins:
Closing in favor of PR. Let's continue the conversation there if needed.

ahopkins:
Thanks for bringing this up.

myusko:
@ahopkins @ashleysommer probably we should close the issue since we already merged PR with a fix.

Issue #1775, opened 2020-01-28
Add tests for route decorators


sjsadowski:
Add tests to prevent reversions on route decorators.

ashleysommer:
@sjsadowski 
I added some basic tests for the multiple-@route decorators, when I submitted my fix:
https://github.com/huge-success/sanic/commit/b565072ed9792debb017ba3dc2b3de67f8f03514#diff-4591e684af0cbdb11ee7409ef6f9aa4aR554

Do you know which other decorators don't have proper tests?

sjsadowski:
@ashleysommer not yet - that's why I assigned it to myself. I may close it, but my biggest concern was not running into another reversion with decorators.

PR #1776, opened 2020-01-30
Bug fix for host parameter issue with lists

damianj:
As explained in #1772 there is an issue when using a list as an argument for the host parameter in the Blueprint.route() decorator. I've traced the issue back to this line, and the if conditional should ensure that the name attribute isn't accessed when route is None.


damianj:
@ahopkins I think the tests are being addressed in #1775? Unless that is not addressing Blueprint.route() tests.

damianj:
@ahopkins I added some unit tests for using lists with the host parameter. There were already some for testing strings with the host parameter.

sjsadowski:
@ahopkins can you resolve your change request if you think that we've met your threshhold?

damianj:
Just following up to see if there is anything else I can or need to do in order to get this merged?

Tronic:
> This seems like an okay change. But, it might be nice to have a test with host routing with a string and a list of strings.
> 
> Thanks for the PR.

#1775 was closed (stale) and this now has tests. Can we resolve this change request and get this merged?

PR #1777, opened 2020-01-30
create a separate make_websoket_handler method

mortbauer:
This allows the following pattern:
```
@app.route('/ws/devices/<device_id>')
@auth_required
@app.make_websocket_handler
async def websocket_connection(request,ws,device_id):
```
Which means that the authentication can be checked before the websocket upgrade is started, and thus also responding with a proper 401  Unauthorized.

The current behaviour is not altered by this change.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -457,6 +457,49 @@ class Sanic:
         )(handler)
         return handler
 
+    def make_websocket_handler(self,handler,subprotocols=None):
+        self.enable_websocket()
+        async def websocket_handler(request, *args, **kwargs):
+            request.app = self
+            if not getattr(handler, "__blueprintname__", False):
+                request.endpoint = handler.__name__
+            else:
+                request.endpoint = (
+                    getattr(handler, "__blueprintname__", "")
+                    + handler.__name__
+                )
+
+                pass
+
+            if self.asgi:
+                ws = request.transport.get_websocket_connection()
+            else:
+                try:
+                    protocol = request.transport.get_protocol()
+                except AttributeError:
+                    # On Python3.5 the Transport classes in asyncio do not
+                    # have a get_protocol() method as in uvloop
+                    protocol = request.transport._protocol
+                protocol.app = self
+
+                ws = await protocol.websocket_handshake(
+                    request, subprotocols
+                )
+
+            # schedule the application handler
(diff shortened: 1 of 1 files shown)


sanic/app.py
```diff
@@ -457,6 +457,49 @@ def add_route(
...
+
+                pass
+
+            if self.asgi:
+                ws = request.transport.get_websocket_connection()
+            else:
+                try:
+                    protocol = request.transport.get_protocol()
+                except AttributeError:
+                    # On Python3.5 the Transport classes in asyncio do not
```
ahopkins: Sanic does not aim to support Python 3.5

sjsadowski:
A few minor linting issues to clean up and please add a test around this given that we're restructuring websockets here with a separate function.

sanic/app.py
```diff
@@ -457,6 +457,49 @@ def add_route(
         )(handler)
         return handler
 
+    def make_websocket_handler(self,handler,subprotocols=None):
```
ahopkins: I want to think on this name a little. We might be able to find something more approachable and understandable.

ahopkins:
I put some more thought in this...

Rather than add a new handler, I think a better solution is either to allow this on the `@app.websocket` handler (probably more difficult) or add a `@app.route(websocket=True)` argument.

mortbauer:
The problem with doing that via the `@app.websocket` is that `@app.websocket` does expect the uri as first argument, which would be not used if one just wants to turn the request handler into a websocket handler -> possible but in my opinion not good

The problem with integrating it with a `@app.route(websocket=True)` is the same as the original one since it would wrap and add the route in one function call, however what I want is to split this functionality.

1. wrap to produce a websocket handler
... wrap with other decorators
2. add a route for the handler

sanic/app.py
mortbauer: what about websockify ?

ahopkins:
I think this will need to be a decision with some more discussion and weight from @huge-success/sanic-core-devs since it is a bit of an API change. 

Thoughts from anyone else?

sjsadowski:
I'm ultimately fine with it so long as the previous method is implemented to be compatible. I'm also fine with making a breaking change so long as it is documented. If we want to discuss in depth we should start a thread on the forum.

PR #1789, opened 2020-02-20
Fix #1788 incorrect url_for for routes with hosts, added tests.

### LINKED ISSUE #1788: url_for() doesn't return a working URI for a blueprint route with host
**Describe the bug**
When i use blueprints with `host` argument and try to get url with `url_for` i get an incorrect url - blueprint's host is used as a path, and not as a hostname

**Code snippet**

```python
from sanic import Sanic, Blueprint

bp = Blueprint('bp_app', host='bp.example.com')


@bp.route('/', name='index')
async def bp_index(request):
    url = request.app.url_for('bp_app.index')
    return text(url)


@bp.route('/internal', name='internal')
async def bp_index(request):
    url = request.app.url_for('bp_app.internal')
    return text(url)


@bp.route('/external', name='external')
async def bp_index(request):
    url = request.app.url_for('bp_app.external', _external=True)
    return text(url)


app = Sanic('app_name')
app.blueprint(bp)


if __name__ == '__main__':
    app.run(port=8000)
```

Output

```console
# 1
$ curl -H "Host:bp.example.com" http://127.0.0.1:8000/
bp.example.com

# 2
$ curl -H "Host:bp.example.com" http://127.0.0.1:8000/internal
bp.example.com/internal

# 3
$ curl -H "Host:bp.example.com" http://127.0.0.1:8000/external
http:///bp.example.com/external
```

`1` example returns hostname by like a path
`2` example returns hostname and path but it's like a path
`3` example return full url, but with no hostname (3 slashes)

**Expected behavior**
I'm expecting correct urls.
For internal urls - only path returned as stated in route.
For external urls - fully qualified domain name with full path

`1` example - `/`
`2` example - `/internal`
`3` example - `http://bp.example.com/external`

 - OS: macOS
 - Version: 19.12.2

**Additional context**
Can't get correct place, but found some places, where it can be
[app.py#L829-L832](https://github.com/huge-success/sanic/blob/v19.12.2/sanic/app.py#L829-L832),
[app.py#L848](https://github.com/huge-success/sanic/blob/v19.12.2/sanic/app.py#L848) (perhaps, blueprint's host should be used, and config's server name as default)

Tronic:
The router allows specifying different hostname for specific routes, and handles this internally by string concatenation of host and path into uri, confusing `app.url_for` into creating invalid URLs.

Ultimately it would be more sensible to keep URL and path separately in Router objects, but as a workaround to the mentioned bug report, this is now handled in `url_for` which uses any route-provided host where `SERVER_NAME` would otherwise be used.

Adds `test_url_for.py` to avoid regressions, and to allow for more comprehensive testing of the said function, which doesn't seem to be currently tested properly.

Fixes issue #1788

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -832,7 +832,7 @@ class Sanic:
 
         # If the route has host defined, split that off
         # TODO: Retain netloc and path separately in Route objects
-        host = uri.find('/')
+        host = uri.find("/")
         if host > 0:
             host, uri = uri[:host], uri[host:]
         else:
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -929,7 +929,7 @@ class Sanic:
         query_string = urlencode(kwargs, doseq=True) if kwargs else ""
         # scheme://netloc/path;parameters?query#fragment
         out = urlunparse((scheme, netloc, out, "", query_string, anchor))
-        print(out)
+
         return out
 
     # -------------------------------------------------------------------- #
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -830,6 +830,14 @@ class Sanic:
                 "Endpoint with name `{}` was not found".format(view_name)
             )
 
+        # If the route has host defined, split that off
+        # TODO: Retain netloc and path separately in Route objects
+        host = uri.find('/')
+        if host > 0:
+            host, uri = uri[:host], uri[host:]
+        else:
+            host = None
+
         if view_name == "static" or view_name.endswith(".static"):
             filename = kwargs.pop("filename", None)
             # it's static folder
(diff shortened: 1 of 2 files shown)


sjsadowski:
@Tronic gonna get this in ASAP - can you leave the branch up after merge, I think this needs to go in to 19.12 LTS which I'm still trying to fix. Thx!

Tronic:
I cannot reproduce that test failure on my own system, and it is not quite obvious from the output what went wrong. Is this related to this patch or a problem with the CI pipeline?

sjsadowski:
@Tronic test failure is py38 on windows. It's included for forward compat but it's not required because windows is best-effort support. You're golden.

Issue #1785, opened 2020-02-17
Error handlers are synchronous


lovasoa:
**Describe the bug**
The [documentation](https://sanic.readthedocs.io/en/latest/sanic/exceptions.html) says that error handlers can be asynchronous.
However, asynchronous error handlers do not seem to be supported everywhere.

**Code snippet**
```py
@app.exception(Exception)
async def handle_base_exception(_request: Request, exception: Exception):
    return response.json("too bad")
```

**Expected behavior**
Then, for example when I exceed the `REQUEST_MAX_SIZE` limit, I would expect to receive `"too bad"` as a response.

Unfortunately, sanic contains code like https://github.com/huge-success/sanic/blob/91f6abaa81248189fbcbdf685e8bdcbb7846609f/sanic/server.py#L577-L579

which expects the error handler to be synchronous.

**Actual behavior**

Here are the (verbose) logs of the actual results:

```
2020-02-17 19:03:15,514:INFO:
[2020-02-17 19:03:24 +0100] [37783] [ERROR] Transport closed @ ('127.0.0.1', 58698) and exception experienced during error handling
[2020-02-17 19:03:24 +0100] - (sanic.access)[INFO][UNKNOWN]: nil  0 -1
2020-02-17 19:03:24,145:ERROR:Transport closed @ ('127.0.0.1', 58698) and exception experienced during error handling
2020-02-17 19:03:24,145:INFO:
.../venv/lib/python3.7/site-packages/sanic/server.py:297: RuntimeWarning: coroutine 'handle_base_exception' was never awaited
  self.write_error(PayloadTooLarge("Payload Too Large"))
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

Tronic:
This one seems to be hard to fix because of how asyncio protocols work. The `server.write_error` function is non-async and thus cannot await on the response, and the function is also called from non-async callbacks that cannot be made async because of how `asyncio` is built.

I suppose this could be worked around by using `asyncio.create_task(...)` but doing so might introduce some unexpected effects and since the protocol code is in any case quite convoluted, I am not quite confident enough to just go changing it.

Anyone willing to weigh in, or should I just modify the code and see if tests break?

Tronic:
I've confirmed that this is now fixed in the streaming branch.

lovasoa:
This should probably not be closed  until the fix is merged

wei-hai:
+1

lovasoa:
[...]

PR #1790, opened 2020-02-21
Protocol errors did not support async error handlers

Tronic:
Fix async error handlers which were not being awaited when errors occurred within protocol code (e.g. Payload Too Large).

This appears to be working but I did not review what consequences does it have that `server.write_error` is now in fact asynchronous.

Issue #1785

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -567,7 +567,7 @@ class HttpProtocol(asyncio.Protocol):
                 self.cleanup()
 
     def write_error(self, exception):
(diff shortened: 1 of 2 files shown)


Tronic:
A potential problem: since protocol does not await for error response being sent, it might be so that another request gets handled in the middle and the responses get mixed up. Could someone familiar with the protocol code have a look at that?

Tronic:
Closing in favour of #1791 where this is properly fixed.

PR #1791, opened 2020-02-21
WIP: Streaming

### LINKED ISSUE #1722: curl is not agree wit return in chunk mode
for uploading large file  > 200 MB I write:

@app.post('/data', stream=True)
async def handler(_request):
    async def streaming(_response):
        total_size = 0
        while True:
            body = await _request.stream.read()
            if body is None:
                break
            total_size += len(body)
            print("body " + str(len(body)) + "/" + str(total_size))
    return response.stream(streaming)
```
And send the file with 
```
curl  -F 'file=@Totally_Spies.mp4' -H 'transfer-encoding:chunked' 127.0.0.1:15080/data -X POST -O
```

The curl send #95MB befor stoping with the error:
```
curl: (56) Illegal or missing hexadecimal sequence in chunked-encoding
```

I do not understand what is wrong!

### LINKED ISSUE #1730: ASGI Streaming response contains extra content

Here is a test case:

```
from sanic import Sanic
from pathlib import Path
import aiofiles
import aiohttp

app = Sanic()

async def test(request):
  tf=__file__

async def test2(request):

    while True:
      if not chunk:
        break
      await response.write(chunk)

  return stream(stream_file)

```

```
33d
from sanic import Sanic
from pathlib import Path
...TRUNCATED MIDDLE OF OUTPUT
  return stream(stream_file)
0
```

### LINKED ISSUE #1785: Error handlers are synchronous

### LINKED ISSUE #1804: Why define an exception handler needed async?
with tests/test_exceptions_handler.py define an exception handler like this:

> @exception_handler_app.exception(NotFound, ServerError)
> def handler_exception(request, exception):
>     return text("OK")

but in docs [https://sanic.readthedocs.io/en/latest/sanic/exceptions.html](https://sanic.readthedocs.io/en/latest/sanic/exceptions.html)

> @app.exception(NotFound)
> async def ignore_404s(request, exception):
>     return text("Yep, I totally found the page: {}".format(request.url))

what is the difference? And is async def needed?

### PR DESCRIPTION
**Pull request by Tronic**
This is an effort the clean up server code and make streaming requests and responses first class citizens, with non-streaming implemented on top of them instead of having duplicate code and separate modes for them.

So far request streaming is looked at, and it seems to work fairly well. 7 tests fail, which might be due to expectation that all of request body would be read prior to sending response because the new code sends a response as soon as possible.

A possible bug discovered: I think the old code would get confused if more than one request was sent without waiting for the prior ones be responded first. Additionally, such parallel streaming responses could get entirely mixed up. Although pipelining in such a way is non-standard, it should now be fixed.

This is NOT complete or meant to be merged yet, but the idea is to finish the request handling, and then try to attempt streaming responses without callbacks as prototyped earlier in the Trio branch.

Testing for the advantageous:
```
pip install git+https://github.com/Tronic/sanic@streaming
```

Bug reports resolved in this:
Fix #1722 
Fix #1730 
Fix #1785 
Fix #1790 
Fix #1804

### CODE REVIEW

**Comment by Tronic** [comment]
Status update: rewriting the server with async instead of a lot of callbacks made the code a lot easier to work with and actually faster despite also removing the MagicHttp stack (because it doesn't support async). The current pure Python parser pushes 50000 req/s where the master branch only does 44000 req/s.

Quite a bit of work remains to be done, especially if all corner cases are to be covered. A new parser means that it is very difficult to make it behave identically to the old one. At this point I've got 6 failing tests and 636 passing.

**Comment by vltr** [comment]
Wow! You took out `httptools` and was still able to get better performance from Sanic using plain Python? That's quite unexpected but really interesting ... I'll take a closer look on the code when I can. Great job!

**Comment by vltr** [comment]
So ... Since we might be moving away from `httptools`, one thing comes to mind: we are now parsing all data by ourselves and there are some specific cases that `httptools` (or, more specifically, `http-parser`) would handle for us (and have that available with its own [unit tests](https://github.com/nodejs/http-parser/blob/master/test.c)), like buffer overflows, well formed but incomplete requests and so on. I believe that we should tackle at least a handful of these tests into Sanic to avoid introducing new bugs at the cost of the void left by `httptools | http-parser`.

**Comment by Tronic** [comment]
Agreed, we definitely need to comprehensive testing of request handling. There are also interesting cases around streaming, like if a handler finishes its response before request body is even (fully) sent. I'd appreciate it if someone else was willing to write tests because I'll probably overlook the assumptions done in my implementation.

Next milestones:
- Maybe get rid of StreamingBuffer class as it is already degraded to a simple wrapper.
- Implement `request.respond(...)` for spawning streaming responses
- Implement StreamingHttpResponse and simple responses on top of that

**Comment by Tronic** [comment]
All **responses** made streaming now, too. That was quite a bit more work than anticipated, and there is still work to be done but this is taking good shape now, and doesn't break existing interfaces much. Looks like this could be aimed for Sanic 20.06 release, possibly with some deprecations to prepare for it in 20.03.

Next milestones:
- Resolve remaining test failures
- Improve request.respond usability?
- Fix response middleware handling (how exactly, remains an issue)
- Try to simplify and de-duplicate error handling (especially so in app.py)
- Fix #1722 and #1785

**Comment by vltr** [comment]
> Agreed, we definitely need to comprehensive testing of request handling. There are also interesting cases around streaming, like if a handler finishes its response before request body is even (fully) sent. I'd appreciate it if someone else was willing to write tests because I'll probably overlook the assumptions done in my implementation.

Awesome work, @Tronic ! Regarding the parser (yet), I see that you already transferred most of the logic to the `http.py` module, but to create tests regarding HTTP parsing _only_ we would need to break it down a little bit more (more specifically take out the parsing logic to not be intertwined with the `protocol` logic). Do you think that's "doable" - even if this requires a plethora arguments in function calls?

**Comment by Tronic** [comment]
@vltr The accessing protocol members from Http is only a temporary workaround until tests are passing and the code can be cleaned up some more. I also considered sans-I/O implementation but async I/O calls allow for automatic flow control without extra tasks for copying data, and presumably higher performance.

After cleanup you should be able to insert mock functions `send` and `receive_more` for unit testing the Http class, without dependencies to Protocol (or to asyncio, for that matter).

**Comment by Tronic** [comment]
Due to the nature of changes done, I believe that custom protocols can no longer be supported, at least not preserving compatibility with the earlier versions. Some specific parts like `write_response` could probably be emulated but I wonder what is the real use of this feature and whether such emulation would suffice.

https://sanic.readthedocs.io/en/latest/sanic/custom_protocol.html

I would much rather not have applications directly writing raw HTTP/1 bytes to the transport, as that will necessarily fail if the connection is HTTP/2, HTTP/3 or ASGI (assuming that the former two are at some point to be supported in Sanic).

**Comment by Tronic** [comment]
This is closer to final shape now, although I am still expecting discussion about the API and some refactoring to further reduce code duplication (e.g. between asgi and protocol/http).

At this point most original functionality should be working.

- Timeouts reset every time a byte is sent or received (i.e. handler is doing something), which needs to be documented but shouldn't cause issues. 

- Request maximum size now affects receive buffer size and as before, the limit applies to request header size, but requests are no longer rejected because of body size until the body is received. For non-streaming handlers this happens automatically and the handler is never run, but streaming handlers have a chance to change `request.stream.request_max_size`. Due to these changes, and for overall resilience against DoS attacks, I would recommend reducing the default value to 1 MB, instead of the current limit 100 MB.

- The request handling no longer depends on asyncio events or other facilities in any significant manner, so ASGI Trio clients can work better. The new `http.py` protocol could also work with a Trio server implementation, if anyone ever wrote one.

Deprecations:

- Anything that writes HTTP bytes (custom protocol, HTTPResponse write_output).
- Request `body_init, body_push, body_finish` in favour of `receive_body`
- StreamingHTTPResponse (still supported) in favour of callback-free API:

```python
@app.post("/echo", stream=True)
async def echo(request):
    # Lift the size limit
    request.stream.request_max_size = float("inf")
    # Spawn a streaming response
    response = request.respond(content_type=request.content_type)
    # Stream request body back to sender
    async for data in request.stream:
        await response.send(data)
```

Tronic:
Discussion at https://community.sanicframework.org/t/new-streaming-api-and-other-major-changes/518

Tronic:
I believe that test_payload_too_large failing on Windows is a platform bug instead of a bug in my code. Omitting transport.close() after the response has been sent avoids the test failure, but according to asyncio docs it sends any buffered outgoing data asynchronously, so this should make no difference. The test client httpx raises an exception about connection being closed immediately after the close() call.

Tronic:
Next milestones:
- Call response middleware on the first chunk of data sent
  - Execute middleware also when the handler uses `request.respond`.
  - What about streaming responses, how could middleware mangle those?
- Refactor response spawning/sending code and error handling
  - Problem: error handling needs to be outside of protocol code so that header errors etc. can be handled and responded to with HTML error pages.
  - Avoid duplication in http.py and asgi.py
  - Less jumping between request, response, stream and app (current code is hard to understand)
- Naming and API issues
  - `async for data in request.stream` vs. `async for data in request`
  - read/write vs. receive/send vs. `.stream(data)` (the current implementation mixes both)
  - what parameters does request.respond take (response object or status/headers/content_type, or something else)?

Tronic:
Some lost performance on this one that'll require optimisation later. In particular, it is too slow to encode header fields and values individually, so I'll have to revert to whole header encode as before, and take that penalty only in ASGI where it cannot be avoided.

My first streaming benchmarks are quite promising: a single-core single-request streaming echo handler processes data at about 600 MB/s (or twice that because it echoes it back), and I didn't even check whether curl or other tools used to benchmark are limiting it more than Sanic itself. This is certainly fast enough for typical network connections, and without any optimisation on that yet.

The middleware remains a big problem. The original plan of running it on first send is not working because all new response objects can be produced (and cannot be returned from send), and it would be quite clumsy anyway. Unfortunately the obvious solution of making `request.respond` async and running from there isn't quite that simple either. Existing middleware should return promptly, and in that case a streaming request handler gets ahold of whatever object the middleware returned, which might be OK, and allows for the middleware to override what the send method on its object does. Another minor problem is that Sanic runs response middleware even on error responses that might be triggered without a request object (in particular, after bad request or payload too large).

Tronic:
4.6 GB/s streaming response. Looks like I'm not gonna be using nginx for `/static` anymore.

Tronic:
Response middleware issues more or less resolved but other code still needs a lot of cleanup and some refactoring. And **tests** as I've already found a number of issues not caught by the test suite.

Tronic:
Looks like I don't have permissions to edit "linked issues" but here is the current list of open bugs that are fixed in this PR #1722 #1730 #1785 #1790 #1804.

sanic/http.py
```diff
@@ -0,0 +1,406 @@
...
+        """Receive and parse request header into self.request."""
+        # Receive until full header is in buffer
+        buf = self.recv_buffer
+        pos = 0
+        while True:
+            pos = buf.find(b"\r\n\r\n", pos)
+            if pos != -1:
+                break
+            pos = max(0, len(buf) - 3)
+            if pos >= self.request_max_size:
```
ahopkins: Perhaps this should be a separate config. A max size for just the headers. [Nginx defaults to 1k](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_header_buffer_size).

sanic/http.py
```diff
@@ -0,0 +1,406 @@
...
+        try:
+            raw_headers = buf[:pos].decode(errors="surrogateescape")
+            reqline, *raw_headers = raw_headers.split("\r\n")
+            method, self.url, protocol = reqline.split(" ")
+            if protocol == "HTTP/1.1":
+                self.keep_alive = True
+            elif protocol == "HTTP/1.0":
+                self.keep_alive = False
+            else:
+                raise Exception
```
ahopkins: Probably should place an explicit exception here.

sanic/http.py
```diff
@@ -0,0 +1,406 @@
...
+        calling this function multiple times will just alter the response to be
+        given."""
+        if self.stage is not Stage.HANDLER:
+            self.stage = Stage.FAILED
+            raise RuntimeError("Response already started")
+        self.response, response.stream = response, self
+        return response
+
+    @property
+    def send(self):
```
ahopkins: What is the purpose of this? compat?

sanic/asgi.py
```diff
@@ -255,142 +238,54 @@ def __init__(self) -> None:
...
-                is_stream_handler = sanic_app.router.is_stream_handler(
-                    instance.request
-                )
-                if is_stream_handler:
-                    instance.request.stream = StreamBuffer(
-                        sanic_app.config.REQUEST_BUFFER_QUEUE_SIZE
-                    )
-                    instance.do_stream = True
+            instance.request.stream = instance
+            instance.request_body = True  # FIXME: Use more_body?
```
ahopkins: I'd advocate for this change. I originally thought there was a problem with the `read` method, until I went back and realized what this was doing in the iterator. `more_body` makes more sense to me.

sanic/request.py
```diff
@@ -127,14 +108,32 @@ def __repr__(self):
             self.__class__.__name__, self.method, self.path
         )
 
-    def body_init(self):
```
ahopkins: See comment on [forums](https://community.sanicframework.org/t/new-streaming-api-and-other-major-changes/518/8?u=ahopkins)

sanic/http.py
Tronic: I've also considered separate setting for receive buffer size, as there is for websockets. However, for simplicity and ease of config, the number of options really should be kept to bare minimum, i.e. what is actually required in practical applications. Apache limits them to 8 kB and that doesn't seem to bother anyone. Already implemented in this branch: send buffer size (in asyncio protocol) is hardcoded 64 KiB.

sanic/http.py
Tronic: Yes, and hacking in progress. Should probably store that function pointer directly in self.send, or better yet, do the larger refactoring where the users don't directly call anything on this class. Until this week the send function had some logic of its own in addition to calling response_func.

sanic/http.py
Tronic: This triggers a Bad Request response a few lines down in the same function. I've added a comment to clarify this.

sanic/request.py
Tronic: Deprecated in 20.3 patch already merged.

This PR now raises `NotImplementedError` if a custom request class is used where any of the body functions are implemented but `receive_body` isn't. For cross-version compatibility, it is allowed to implement both APIs.

Tronic:
For some reason on Windows the payload too large test fails because the test client cannot read the response. I suspected a deadlock between writing request and reading response blocking but increasing buffer sizes didn't help, and I can't currently think of other reasons why this might be happening.

Tronic:
I've traced how that goes: the HTTP handler function writes its response to asyncio protocol and exits cleanly, after which transport.close() is called. According to docs this function should write out any pending data before actually closing.

My bet is that this failure occurs because httpx is still trying to send the request and doesn't notice that there is a response coming. Adding a delay allows for the request to flow into Sanic's receive buffer, after which httpx manages to read the response without any error. I do wonder how this is supposed to be solved...

sanic/http.py
ahopkins: I am fine also hard coding it. But tying it to the `request_max_size` could be problematic if the dev sets that extremely high to handle large uploads, etc. I say let's just pick something reasonable.

Tronic:
Yep, confirmed working, although that is not really a fix, just a hack around httpx in this particular test where the request happens to fit in network buffers despite being larger than request maximum size.

Tronic:
Deadlocks are possible in bidirectional streaming handlers if the HTTP client does not support streaming but insists on sending all of request before reading any response. This will lead to request timeout but Sanic could also detect when this happens and issue a proper diagnostic. Unfortunately the problem itself cannot be avoided but it is up to the application whether or not to implement this kind of handlers.

This does not affect:
* Normal handling (because all of request is read first)
* Streaming in most cases

Non-streaming clients fail:
* When request is too large: we won't receive infinite amounts of data
   - Sanic instantly sends a 413 PayloadTooLarge and disconnects (all-ok)
   - Client gets a network error on send and never reads the response
* Streaming handlers with large request and large response
   - Small request or small response fits in network buffers and avoids the deadlock.
   - Client and server both stuck trying to send until response timeout

The problems are not new to this branch, and also occur with any other HTTP server implementing this functionality. Notably, browsers don't do bidirectional streaming but they have large network buffers to avoid any issues with moderately sized bodies. Curl, Nginx, Express and others implement proper bidirectional streaming.

ahopkins:
@Tronic Can you squash these commits?

Tronic:
@ahopkins You mean rebase and force-push? And all of them or just the merge commits? There's further work to be done on this prior to merge to master, so probably such merging should only be done then; and I believe this can be done more cleanly in Github UI while merging.

Tronic:
Since the new policy is not to introduce any breaking changes in .09 and .12 releases, I guess this should target 20.06 (due to the delay with 20.03 I would been more comfortable with 20.09). To make that happen, **I'll need help** with proper code review, API discussion and writing comprehensive tests. Is anyone up for that?

Also, can I expect 20.06 to be delayed at least until the *end* for June, or better yet, July?

ahopkins:
1. I will help with the review
1. Ideally we'd be able to publish notice with some warnings. Not really applicable for a change like this.
1. Our choice is to make it for this next release, or wait until March.
1. Given the nature of this past spring release, and the short notice on declaring depreciation schedule, I'm okay with releasing in July. @huge-success/sanic-release-managers? If there is consensus, I'll post to the Sanic front page with an announcement.

sanic/asgi.py
Tronic: I'm not quite sure what to do with this anymore. Could you jog my memory or change it yourself?

Tronic:
Closing PR as development is moved to huge-success/sanic:streaming

PR #1793, opened 2020-02-24
Use f-strings instread str.format()

mikeyusko:
Hey,

Per a discussion on Sanic's community forum, it will be cool to use `f-strings` instead of `str.format()`.

1.  It will improve our performance since `f-strings` are faster than `str.format()`
2. A bit easier to read.

That's the first phase and it only applies for `tests`, if it looks good, next PR will be applied to core functionality.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -103,7 +103,7 @@ def sanic_router():
         for method, route in route_details:
             try:
                 router._add(
-                    uri="/{}".format(route),
+                    uri=f"/{route}",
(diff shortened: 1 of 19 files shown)


Tronic:
LGTM, barring a few overlong lines (but black's line lengths are not enforced on tests).

sjsadowski:
Only issues are windows python 3.8 which probably needs to get yanked back out of appveyor since it makes it look like there's a problem and there is not.

PR #1795, opened 2020-02-26
Allow verify parameter in SanicTestClient requests

eric-spitler:
Removes the hard-coded **`verify=False`** from `app.test_client` usage.
- Allows **`verify`** argument to be used by `Sanic` applications in **`app.test_client`** requests where there is a custom certificate in place and actually validate against it
- Resolves the `HTTPX` warnings about unverified request certificates in tests

Also improved the use of self-signed certs in tests. Tests will generate a new self-singed certificate before run to ensure it remains valid instead of being expired. This required the addition of `tox-run-before` dependency.

The existing certificate is removed from the repository since it is generated during test execution. It has been replaced by a CSR configuration file.


sjsadowski:
Looks like it's not finding the generated cert-related info for the tests here. I believe the tests are run from the root of the project so it may need to look in {path}/tests/certs instead of just {path}/certs

eric-spitler:
Yeah I noticed that and tried to do some more local testing to figure out why. My guess is something different is happening with openssl on the build machine than on mine.

However, even locally the generated certificate wasn't working with Python 3.7 or 3.8 (only worked in 3.6), so it seems there is more to look at from that process.

Can probably get around the need for auto-generated certs with an updated copy, if you can provide one or the steps you used before.

eric-spitler:
I just realized this is a Windows build machine - therefore OpenSSL would be unlikely to be available

Tronic:
@eric-spitler Could you merge master to this and resolve the conflict?

eric-spitler:
Sorry, life got in the way of fixing the tests surrounding validation of a certificate. I'll see if I can get a better certificate solution in place to allow these to pass (hopefully soon).

I'm thinking it will need 2 files - one that is a CA and is used to sign the other that acts as a server certificate. Tests should use the CA to validate the server cert, but currently its trying to use the same file for both to keep the original single-file test usage.

ahopkins:
Closing this. Please remake the PR here.

https://github.com/huge-success/sanic-test

PR #1798, opened 2020-03-01
Allow multiple workers on MacOS with Python 3.8

Tronic:
The app object cannot be pickled, apparently due to it containing decorated route functions and the decorator wrapper cannot be passed to worker processes. This became a bigger issue with Python 3.8 that made spawn mode the default of multiprocessing on MacOS, causing workers > 1 to stop working.

This patch forces fork mode on POSIX systems (*nix, MacOS) because it doesn't pickle, and on other systems (Windows) issues a warning and sets `workers=1`.

The proper fix would be to rewrite process spawning so that app object is initialised separately in each worker, not passed via pickling, but I doubt anyone will have time to implement that.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1096,7 +1096,7 @@ class Sanic:
         stop_event: Any = None,
         register_sys_signals: bool = True,
         access_log: Optional[bool] = None,
-        **kwargs: Any
+        **kwargs: Any,
     ) -> None:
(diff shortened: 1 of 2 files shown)


Tronic:
@sjsadowski This needs to go in 19.12LTS.

sjsadowski:
@Tronic thanks for the heads up

PR #1803, opened 2020-03-09
Update config docs to match DEFAULT_CONFIG

KevinGDialpad:
The [docs](https://sanic.readthedocs.io/en/latest/sanic/config.html) indicate that the default value of REAL_IP_HEADER is "X-Real-IP" but the code never sets it to that value.

### DIFF
--- a/docs/sanic/config.rst
+++ b/docs/sanic/config.rst
@@ -115,15 +115,25 @@ Out of the box there are just a few predefined values which can be overwritten w
 +---------------------------+-------------------+-----------------------------------------------------------------------------+
 | KEEP_ALIVE_TIMEOUT        | 5                 | How long to hold a TCP connection open (sec)                                |
 +---------------------------+-------------------+-----------------------------------------------------------------------------+
+| WEBSOCKET_MAX_SIZE        | 2^20              | Maximum size for incoming messages (bytes)                                  |
++---------------------------+-------------------+-----------------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


Tronic:
The default value was changed to None for security reasons. Otherwise an attacker could send any spoofed addresses to Sanic servers that are not behind proxies but that are using the default config. Also, each service provider seems to use their own names for this header, so a default value would not work for all. For more information, see https://sanic.readthedocs.io/en/latest/sanic/config.html#proxy-configuration

However, the table in the documentation lists wrong values for many variables, and should be updated to match with current codebase. Good catch! Could you change this PR to do that?

andreymal:
When I made #1539 I meant that `REAL_IP_HEADER` will work only if `PROXIES_COUNT` has non-zero value. It seems the logic is changed now, so the docs should be updated instead

Tronic:
#1638 is the relevant PR for changes in semantics and default values. Prior to that `PROXIES_COUNT` defaulted to -1 (i.e. use the last proxy), so in effect the real IP header was also always enabled.

KevinGDialpad:
> The default value was changed to None for security reasons. Otherwise an attacker could send any spoofed addresses to Sanic servers that are not behind proxies but that are using the default config. Also, each service provider seems to use their own names for this header, so a default value would not work for all. For more information, see https://sanic.readthedocs.io/en/latest/sanic/config.html#proxy-configuration
Ah, fair enough.
I'll make this PR a config change then!

KevinGDialpad:
I noticed that the `WEBSOCKET_` default config values weren't documented, so I added them. Let me know if they were left out on purpose.

PR #1808, opened 2020-03-18
Fix Ctrl+C and tests on Windows.

### LINKED ISSUE #1753: Any plans to support Ctrl+C on Windows?

Tronic:
Alternative signal handling logic for Windows (fix #1753) and allow green CI pipelines again.
- Adds `app.is_stopping` for determining whether `app.stop` has been called, which is different from app not running.
- Most tests use random ports because on Windows CI they would otherwise fail to bind.
- Fixed deprecation warnings in tests (Sanic missing name argument).

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -81,6 +81,7 @@ class Sanic:
         self.sock = None
         self.strict_slashes = strict_slashes
         self.listeners = defaultdict(list)
+        self.is_stopping = False
         self.is_running = False
         self.is_request_stream = False
         self.websocket_enabled = False
@@ -1209,7 +1210,9 @@ class Sanic:
 
     def stop(self):
         """This kills the Sanic"""
-        get_event_loop().stop()
+        if not self.is_stopping:
+            self.is_stopping = True
+            get_event_loop().stop()
 
     async def create_server(
         self,
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1096,8 +1096,6 @@ class Sanic:
         stop_event: Any = None,
         register_sys_signals: bool = True,
         access_log: Optional[bool] = None,
-        auto_reload = None,
-        asyncio_server_kwargs = None,
         **kwargs: Any,
     ) -> None:
         """Run the HTTP Server and listen until keyboard interrupt or term
@@ -1128,9 +1126,6 @@ class Sanic:
         :type register_sys_signals: bool
         :param access_log: Enables writing access logs (slows server)
         :type access_log: bool
-        :param asyncio_server_kwargs: key-value arguments for
-                                      asyncio/uvloop create_server method
-        :type asyncio_server_kwargs: dict
         :return: Nothing
         """
         if "loop" in kwargs:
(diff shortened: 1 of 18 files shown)


Tronic:
The code adds a background task that wakes up 10 times a second so that Python can receive signals (otherwise the signal handler is only called once Sanic receives network traffic). This is of course not needed on other operating systems.

The signal handler cannot call `app.stop` directly because it is in signal handling context (on other OSes `loop.add_signal_handler` is used but this of course is not available on Windows). So, the handler sets a flag, which will be seen by the background task, causing graceful app exit. Additionally, in case of more than one Ctrl+C being received, it falls back to good old KeyboardInterrupt termination.

It took too many hours to figure all that crap out, because it is not in the docs.

Windows still cannot support multiple workers, so that side has not been considered in this fix.

Tronic:
Okay, I think I'm finished with this. Care to review?

sanic/testing.py
```diff
@@ -203,7 +207,7 @@ def __init__(
 
         self.app = app
 
-        dispatch = SanicASGIDispatch(app=app, client=(ASGI_HOST, PORT))
+        dispatch = SanicASGIDispatch(app=app, client=(ASGI_HOST, PORT or 0))
```
yunstanford: Why 0 ?

sanic/server.py
```diff
@@ -929,15 +929,11 @@ def serve(
...
-        _singals = (SIGTERM,) if run_multiple else (SIGINT, SIGTERM)
-        for _signal in _singals:
-            try:
-                loop.add_signal_handler(_signal, loop.stop)
-            except NotImplementedError:
-                logger.warning(
-                    "Sanic tried to use loop.add_signal_handler "
-                    "but it is not implemented on this platform."
-                )
+        if os.name == "nt":
```
yunstanford: Let's define a constant and use that instead of string literal directly

sanic/app.py
```diff
@@ -1177,6 +1178,7 @@ def run(
 
         try:
             self.is_running = True
+            self.is_stopping = False
```
yunstanford: What's the diff between `is_running` and `is_stopping` ?

sanic/testing.py
```diff
@@ -114,6 +115,9 @@ def _collect_request(request):
             url = "{scheme}://{host}:{port}{uri}".format(
                 scheme=scheme, host=host, port=port, uri=uri
             )
+        # Tests construct URLs using PORT = None, which means random port not
+        # known until this function is called, so fix that here
+        url = url.replace(":None/", f":{port}/")
```
yunstanford: a little bit concern on this, probably should avoid this kinda of hacky pattern.

sanic/testing.py
Tronic: Typing says it needs to be int, so `None` causes a failure. Either the type hint should be changed to `Optional[int]` or 0 be used instead but I did not look at ASGI code to see how this is actually handled. In any case, 0 is the standard way to request for a free random port.

sanic/app.py
Tronic: When the app has not yet started, both are False. After calling `app.stop` the app may keep running for a while, so in that period both are True. Asyncio implementation uses `loop._stopping` in a similar fashion but uvloop doesn't have that and depending on private parts of the implementation is ugly in any case.

sanic/testing.py
Tronic: Avoiding this would require major restructuring of `testing.py` which I am not quite up to, or rewriting/disabling all tests that depend on this.

sanic/server.py
Tronic: I've added a constant in server.py but did not change this in `skipif` of OS-dependent tests. Do note that `os.name` is the more coarse platform definition where all Windows are `"nt"`, where `sys.platform` uses many names for Windows (and quite confusingly, 64 bit Python on 64 bit Windows is still called `"win32"`).

sanic/testing.py
Tronic: Actually all of Sanic should probably be changed to accept 0 port. Currently entering zero gives port 8000. But that is a potentially breaking change, and for another PR even if implemented.

Issue #1452, opened 2019-01-02
Enforce Integer only value for Max-Age Cookie


harshanarayana:
# Description
As per the `HTTP` standard guidelines, `Max-Age` for a `Set-Cookie` can only be an integer. 

> Max-Age=<number> Optional
> Number of seconds until the cookie expires. A zero or negative number will expire the cookie 
> immediately. Older browsers (ie6, ie7, and ie8) do not support max-age. For other browsers, if both > (Expires and Max-Age) are set, Max-Age will have precedence.

Current `Cookie` class provided by `sanic` has a fallback catch block that uses any value that you provide to `max-age` as a valid item and sets it. 

# Reference
1. `encode` method in `Cookie` [class](https://github.com/huge-success/sanic/blob/096c44b910443e74c821dbf21adf2f86ccda8611/sanic/cookies.py#L111-L115)
2. Community Discussion [Thread](https://community.sanicframework.org/t/housekeeping-items/91)

# Expected Behavior
If you pass a valid `interger` value to the `max-age` parameter, it needs to be set. Otherwise, it needs to be defaulted to `-1` or `0` causing it to auto expire.

chenjr0719:
Related test:
https://github.com/huge-success/sanic/blob/cea1547e08230b6ad49eb7777fd8db5335382b7a/tests/test_cookies.py#L141-L156

PR #1809, opened 2020-03-18
hard error for bad max-age

jamesstidard:
See: #1810

Thanks

### DIFF
--- a/sanic/cookies.py
+++ b/sanic/cookies.py
@@ -1,4 +1,5 @@
 import re
+import math
 import string
 
 from datetime import datetime
@@ -106,12 +107,25 @@ class Cookie(dict):
             raise KeyError("Unknown cookie property")
         if value is not False:
             if key.lower() == "max-age":
-                is_str_int = isinstance(value, str) and value.isdigit()
-                is_integer = isinstance(value, int) or is_str_int
-                if not is_integer:
+                if not isinstance(value, (str, int, float)):
                     raise TypeError(
-                        "Cookie 'max-age' property must be a int"
+                        "Cookie 'max-age' property must be a str, int, or float"
                     )
+
+                try:
+                    value = float(value)
+                except ValueError:
+                    raise ValueError(
+                        "Cookie 'max-age' property must be a number. "
+                        "Got: {}".format(value)
+                    )
+
+                if value < 0 or value == float("inf"):
+                    raise ValueError(
+                        "Cookie 'max-age' property must be >= 0 and not infinate"
+                    )
+
+                value = math.floor(value)
             elif key.lower() == "expires":
                 if not isinstance(value, datetime):
                     raise TypeError(
--- a/sanic/cookies.py
+++ b/sanic/cookies.py
@@ -115,8 +115,6 @@ class Cookie(dict):
                     raise TypeError(
                         "Cookie 'max-age' property must be a integer value"
                     )
-                else:
-                    value = int(value)
             elif key.lower() == "expires":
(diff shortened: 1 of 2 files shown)


Tronic:
Would you care to elaborate on the choice of type checking logic, and on the change to make non-integer float in tests a failure?

jamesstidard:
Hey @Tronic, sure, sorry.

First, here's a bit of context that I maybe didn't cover in my issue very clearly. 

I had some code that set this attribute as a `float`, and this worked in previous versions of Sanic. I then updated Sanic, and the `float` value of `604,800.0` was being silently coerced into `0`.

It actually took me quite a while to figure out this was the problem for multiple reasons:

1. The code worked before
2. Debugging why cookies are not being set in browsers is generally quite hard. Browsers don't seem to log a reason of why they choose to ignore invalid cookies.
3. I was moving domains at the same time and the whole `samesite` attribute was new to me, so I went off down a rabbit hole thinking it had to do with that. Not spotting that actually now my max age was being set to 0.

1. choice of type checking logic

> I think generally, it's better practice for code to error, then make implicit error corrections. So a hard error here would have let me catch this right away.
> 
> The reason this logic was changed in Sanic in the first place #1452 was to ensure that the value was an integer value and not some float. So I wrote this to allow any kind of integer looking thing (string integer, int, or float integer). Otherwise throw an error.
> 
> In retrospect I think that allowing float integers is also probably bad. If people are programatically generating their `max-age` length based on some dynamic values, then they might only hand in a `float` with integer values during their testing. However, later in production some requests may try and set non-integer `float` values. So it's probably better to error if a `float` and force the user to explicitly `math.ceil`/`math.floor` it into an `int`. I will update my pull request with that. I think I was just trying to be a little forgiving (like how the code currently accepts `str` type).

2. make non-integer float in tests a failure

> For the reasons above for having a hard error. It prevents 'unexpected behaviour' for someone that accidentally hands in a float value >0 being silently converted to a 0 value. 
>
> This is similar to how there is a hard error for handing in a non `datetime.datetime` object for `expiry`; it forces the programmer to make a simple fix and not missing a subtle bug as I did.
>
> I updated the `pytest` test to reflect the above. Though now I'll update it again to just not allow any float.

I hope that makes some sense. Would you agree?

Sorry if this pull request comes across pedantic, I just felt like I lost quite a lot of time to this and that I could save others the same trouble.

Tronic:
Python uses float seconds as its basic time unit, so any time calculation like `max_age = expiry_time - time.time()` produces a float that is generally not an integer but that can be safely rounded down. Doing so would not affect your original problem, where float `604_800.0` should turn into `max-age=604800` if I am not mistaken.

Negative values should probably also be considered errors (raise ValueError?).

jamesstidard:
Ah yes, hadn't considered negatives, I can add a check for that.

your example is exactly my original code that produced this problem in the first place for me :).

I had considered doing a floor or ceiling automatically on the value, but felt that is a similar problem as the original error. The Sanic implementation making an implicit choice for you that you may not expect. 

Though I'm not too picky, and the `Expires` works in the same way (flooring to the closet second), so it will be consistent:

```python
expires = datetime(2000, 1,1,3, 2, 1, 999999)
expires.strftime("%a, %d-%b-%Y %T GMT")
'Sat, 01-Jan-2000 03:02:01 GMT'
```

I'll add those changes. Thanks

Tronic:
Related #1457 with similar intention.

jamesstidard:
I’ll choose this seeing that this has been addressed in another issue

PR #1814, opened 2020-03-20
[Doc] Update getting_started.rst

koug44:
Replacing command to install Sanic without uvloop as the provided one is not working (at least in my case)

I've took the working from your Pypi page

### DIFF
--- a/docs/sanic/getting_started.rst
+++ b/docs/sanic/getting_started.rst
@@ -8,14 +8,14 @@ syntax, so earlier versions of python won't work.
 1. Install Sanic
 ----------------
 
->   If you are running on a clean install of Fedora 28 or above, please make sure you have the ``redhat-rpm-config`` package installed in case if you want to use ``sanic`` with ``ujson`` dependency.
+If you are running on a clean install of Fedora 28 or above, please make sure you have the ``redhat-rpm-config`` package installed in case if you want to use ``sanic`` with ``ujson`` dependency.
 
 .. code-block:: bash
 
     pip3 install sanic
 
 To install sanic without `uvloop` or `ujson` using bash, you can provide either or both of these environmental variables
-using any truthy string like `'y', 'yes', 't', 'true', 'on', '1'` and setting the `SANIC_NO_X` (`X` = `UVLOOP`/`UJSON`)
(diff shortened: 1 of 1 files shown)


docs/sanic/getting_started.rst
```diff
@@ -20,7 +20,9 @@ to true will stop that features installation.
 
 .. code-block:: bash
 
-    SANIC_NO_UVLOOP=true SANIC_NO_UJSON=true pip3 install sanic
+    $ export SANIC_NO_UVLOOP=true
+    $ export SANIC_NO_UJSON=true
+    $ pip3 install --no-binary :all: sanic
```
Tronic: I'd prefer a oneliner to avoid littering the environment. If multiple lines, maybe remove the dollar signs so that it can be copy&pasted?

yunstanford: I believe previous one liner doesn't work for `fish`

koug44: Updated commit as oneliner

Tronic: @yunstanford Support for this syntax was added in Fish 3.1, probably because it was a major incompatibility with other shells.

Tronic:
Could you also fix the example app on this page? It should use `Sanic("App Name")` to avoid the deprecation warning about missing name. I suggest also removing that paragraph about truthy values as irrelevant for getting started.

yunstanford:
hmm.. it's still waiting for travis-ci  to report back results, looks like hanging there forever. Maybe it's easy to just bring up another pr

Tronic:
There were hangs like this with Travis on other PRs too, about two weeks ago. Making any new change on this will re-run it, probably with success. Most current PRs display various test failures although in reality they are all-OK, because of this and other test problems that should now be all fixed.

This PR in particular only affects docs in a way that could hardly cause any test failures.

yunstanford:
Yeah, but the merge button is disabled, and i can't merge it until travis-ci reports results back

Tronic:
@koug44 Could you push something (anything) to this PR so that tests can be run again and this can be merged? *sigh*

koug44:
> @koug44 Could you push something (like merge master) to this PR so that tests can be run again and this can be merged? _sigh_
Done

PR #1815, opened 2020-03-20
Initialise StreamingHTTPResponse.protocol = None

Tronic:
Avoids mypy errors from asgi.py setting that variable.


Tronic:
This is merged to #1808.

PR #1816, opened 2020-03-21
Avoid py38 test failure on Windows

Tronic:
Reduces test data size to 3 MB instead of 30 MB. This should not affect what the tests are trying to accomplish but makes them run faster, which apparently was an issue on Windows until now. Each test includes a large number of requests each of which include this data, and I could get all of them passing if I commented out some of the others, without reducing data size.

Reducing data size makes the tests run much faster and now they pass on my Windows virtual machine with all requests enabled.


Tronic:
This is merged to #1808 and maintaining two branches is too troublesome. Closing this in favour of the other one.

PR #1818, opened 2020-03-26
Deprecation and test cleanup

Tronic:
Drop `app.remove_route` and `request.raw_args` that have been deprecated for a long time. Fix all warnings in tests (mainly asyncio py36 vs. py38 incompatibility). On py38 we'll need the next yet unreleased version websockets>8.1 to get rid of the few remaining deprecated loop warnings.


Tronic:
Tagging for 20.3.

PR #1827, opened 2020-03-31
Fix reloader on OSX py38 and Windows

Tronic:
Simplified a bit and this now appears to work on all OS. On Windows we don't get graceful shutdowns though, due to lack of SIGTERM.

Also removes `app.run(..., **kwargs)` that was silently consuming invalid arguments, and forces using keywords for all args but host and port (potentially breaking some apps).


Tronic:
This includes #1555 but due to bigger and potentially breaking changes will need more testing. One known issue is that if the server does not exit, the autoreloader will get stuck waiting until another Ctrl+C, after which it will probably abandon the server process and any worker processes that it may have, because this no longer uses external tools to kill all children. This should be fixed in server code, making it exit and terminate its children in a more reliable manner on SIGTERM (for instance, making non-graceful exit if not completed within a timeout). SIGKILL on the server is not a good idea, especially so in multiple worker mode, where the workers would be left running.

I believe that eventually autoreloader should be merged to the main process of multiworker mode but that will require a lot more work. Another consideration is that with workers=1 possibly the main process should run the server and the autoreloader be its child, to allow normal debugging etc.

sanic/app.py
```diff
@@ -1067,7 +1069,7 @@ def run(
         stop_event: Any = None,
         register_sys_signals: bool = True,
         access_log: Optional[bool] = None,
-        **kwargs: Any,
```
ahopkins, line 1070: Why remove this? Breaking change, no?

Tronic: It was only used for `auto_reload`, which is now a named argument, and it was causing invisible bugs as any typoed kwargs would be silently ignored. As far as I know, there are not other named parameters that should be caught here, so it will only break applications with such typos in their args. Notice that the `*` is also a potentially breaking change for anyone who tried to provide `debug` as positional rather than by keyword.

PR #1833, opened 2020-04-22
docs: Fix doc build

### LINKED ISSUE #1832: Document build failed
Hey guys, I just noticed the document build on readthedocs has failed since 1 month ago. It caused by the `conda` again. I would like to fix this issue but I have one question, do we still need to use `conda` for creating the environment when building the document? Or, we can just use `pip install .[docs]`? Here is the [doc](https://docs.readthedocs.io/en/stable/config-file/v2.html#python) explain how to build the document on readthedocs with extra requirements. Let me know your idea. 😃

chenjr0719:
This PR aims to fix the doc build filed on readthedocs.

close #1832


readthedocs.yml
```diff
@@ -1,2 +1,9 @@
-conda:
-    file: environment.yml
\ No newline at end of file
+version: 2
+python:
+   version: 3.6
```
Tronic, line 3: Maybe 3.8 instead, or is there some problem with the latest release?

chenjr0719: Sure, I think use 3.8 should be fine. 😃

chenjr0719:
BTW, the CI failed because of the new release of `pytest-asyncio`. Do you think we need to use the previous version instead? The `pytest-asyncio` is defined in `tox.ini`:
https://github.com/huge-success/sanic/blob/ae40f960ff0aa1c0ce24c50411faae6f16790acf/tox.ini#L12

Tronic:
> BTW, the CI failed because of the new release of `pytest-asyncio`. Do you think we need to use the previous version instead?

I wonder what else should be upgraded for the new version to work. In any case, if locking an older version in tox.ini gets the job done, it's OK.

Since this is the only PR hit by the test problem so far, I suggest that you do the change on this same PR despite it being a separate problem (because otherwise this bogus test failure will prevent merging).

Thanks for keeping the plumbing working! :)

yunstanford:
Do we need pytest-asyncio ?

chenjr0719:
> Do we need pytest-asyncio ?

We have a lot of tests are implemented in `async` so I think yes we need it. For example:
https://github.com/huge-success/sanic/blob/ae40f960ff0aa1c0ce24c50411faae6f16790acf/tests/test_asgi.py#L245-L258

And, this issue is caused by the unmeet required version of `pytest`. The `pytest-asyncio` is required `pytest>=5.4.0` but we pin it to `5.2.1`. However, if using `pytest>=5.4.0`, then the `pytest-sugar` will cause another error. 🤔  They also discuss about this dependency issue at Teemu/pytest-sugar#187.

So, I would suggest that we can bump to `pytest>=5.4.0` and remove `pytest-sugar` to fix current CI failed first. And if we really need `pytest-sugar`, we can add it back when they fix the dependency issue. What do you think?

chenjr0719:
OK, then I think the `pytest-asyncio` is not necessary anymore. After removing it, the CI now working. Thanks, @yunstanford. 😃

PR #1805, opened 2020-03-12
Update docs for order of listeners

pwalimbe:
Added comments in the middleware listener section to highlight the order. 
https://github.com/huge-success/sanic/issues/1610

Issue #1610, opened 2019-06-23
Add docs for execution order of listeners


Ronserruya:
I noticed an undocumented behavior with listeners, where some of them run in reverse order of registration

1. What is the rationale for that?
2. Can this be added to the docs?

**Additional context**
Relevant lines of code from https://github.com/huge-success/sanic/blob/master/sanic/app.py#L1356
```python
 # -------------------------------------------- #
        # Register start/stop events
        # -------------------------------------------- #

        for event_name, settings_name, reverse in (
            ("before_server_start", "before_start", False),
            ("after_server_start", "after_start", False),
            ("before_server_stop", "before_stop", True),
            ("after_server_stop", "after_stop", True),
        ):
            listeners = self.listeners[event_name].copy()
            if reverse:
                listeners.reverse()
            # Prepend sanic to the arguments when listeners are triggered
            listeners = [partial(listener, self) for listener in listeners]
            server_settings[settings_name] = listeners
```

Tronic:
This convention comes from Python `with` statements, C++ RAII and similar constructions of scope bound resource management. The rule is that you should deconstruct in exact reverse order of construction.

For instance, if your first before_server_start handler setups a database connection, ones registered after it can rely on that connection being alive both when they are started and stopped, because stopping is done in reverse order, and the database connection is torn down last.

Perhaps it should be documented. Pull requests are welcome. Look at docs/sanic/

tomaszdrozdz:
@ahopkins @Tronic  Can we close this issue ?

PR #1834, opened 2020-04-23
Update docs for order of listeners #1805

pwalimbe:
Added comments in the middleware listener section to highlight the order.
#1610


Tronic:
Now this got hit by the same problem as #1833 ffuuu... A bit of bad timing, but the CI will need fixing again.

pwalimbe:
Adding 5 comment lines in .rst file. CI needs to handle this, right ??

PR #1835, opened 2020-04-24
Wait for delayed tasks before stopping test server

mymedia2:
This is handy for automatic tests of asynchronous handlers that return a result before finishing their work. That is possible with asyncio.create_task.

### DIFF
--- a/sanic/testing.py
+++ b/sanic/testing.py
@@ -1,3 +1,4 @@
+import asyncio
 from json import JSONDecodeError
 from socket import socket
 
@@ -127,6 +128,11 @@ class SanicTestClient:
             except Exception as e:
                 logger.exception("Exception")
                 exceptions.append(e)
+
+            our = asyncio.current_task()
+            their = filter(lambda elem: elem is not our, asyncio.all_tasks())
+            # Wait all other pending tasks before stopping the event loop.
+            await asyncio.gather(*their)
             self.app.stop()
 
         self.app.run(debug=debug, **server_kwargs)


ahopkins:
I can see how this would be helpful, but I also can see how it might not be. Seems like it should be configurable. Can you add a `wait_for_all_tasks` (or some other) kwarg?

Tronic:
Since users cannot possibly handle worker processes' pending background tasks themselves, all tasks should **always** be run to completion by Sanic, until `GRACEFUL_SHUTDOWN_TIMEOUT`, after which I suppose they should be cancelled (by exception) but never just abandoned. Abandoning an async task easily leads to data corruption, orphaned processes left running and other severe problems.

ahopkins:
@yunstanford and @Tronic , I agree with you both. which is why if it will be added it should be configurable and not required.

yunstanford:
I don't see why user can not handle pending background tasks themselves, when launching background tasks, can always stored the task as app context for proper tracking... Personally, i don't think it's good practice to just launch sth. there and don't care, but in the very end assume the framework to handle that. Also shutting down those background tasks may also involve customized logic, e.g. the order could matter, etc.

Tronic:
@yunstanford Do you mean waiting for remaining tasks in a custom `after_server_stop` listener? After that the loop gets closed and the worker process terminates. All the rules break for any tasks still running: execution just stops, `finally` and `with __exit__` are not run, functions don't return etc. Bad stuff.

One of the big points of Trio is to ensure that this can never happen. Any nursery blocks and eventually the `trio.run()` call won't exit until all tasks within have completely finished execution.

ahopkins:
Closing. Please remake PR here.

https://github.com/huge-success/sanic-test

PR #1837, opened 2020-04-30
update asgi.py replace the latin-1 with utf-8.

SchopenhauerZhang:
for avoid another unicode  warn or error,please use utf-8.Thanks!
latin-1 with utf-8 difference.：https://stackoverflow.com/questions/7048745/what-is-the-difference-between-utf-8-and-iso-8859-1/39109074


Tronic:
Yes, the `path` property of ASGI scope is in UTF-8 but I think this re-encoding should be avoided entirely by using something like:

```python
url_bytes = b"%b?%b" % (scope["raw_path"], scope["query_string"])
```

I believe that `root_path` being included in existing code is a mistake, so I left that out (which may be a breaking change). Any comments from other devs or ASGI users?

ahopkins:
I'll do another survey of other implementations and the spec. But I believe that is correct.

Tronic:
> Yes, the `path` property of ASGI scope is in UTF-8 but I think this re-encoding should be avoided entirely by using something like:
> 
> ```python
> url_bytes = b"%b?%b" % (scope["raw_path"], scope["query_string"])
> ```
> 
> I believe that `root_path` being included in existing code is a mistake, so I left that out (which may be a breaking change). Any comments from other devs or ASGI users?

@SchopenhauerZhang could you implement and test this change, so that we could get this merged?

ahopkins:
Ping @SchopenhauerZhang

ashleysommer:
I agree with @Tronic [here](https://github.com/huge-success/sanic/pull/1837#issuecomment-621674418).
We probably don't need to re-encode the url to bytes if we already have the raw_path available in the correct encoding anyway.
The only thing I'm not sure about, is the current implementation does url-encode the `scope['path']` using `quote(path)` before adding it to the root_path, I'm not familiar enough with ASGI to know if/why that is required.

ahopkins:
I am closing this. If someone thinks there is something we still need here, please reopen a new PR.

Issue #1845, opened 2020-05-13
Misleading documentation for response middlewares execution order


kirilyer:
Hi,

The [documentation](https://sanic.readthedocs.io/en/stable/sanic/middleware.html#modifying-the-request-or-response) states that middlewares (request and response) are executed in the order they are defined but this is not the case for response middlewares. Response middlewares are executed in reverse order as a result of them being appended to the **left** of the app's `response_middleware` collection, [app.py:639](https://github.com/huge-success/sanic/blob/e7001b00747b659f7042b0534802b936ee8a53e0/sanic/app.py#L639):
```python
def register_middleware(self, middleware, attach_to="request"):
    ...
    if attach_to == "response":
        if middleware not in self.response_middleware:
            self.response_middleware.appendleft(middleware)
```

It seems that the documentation needs to be fixed to reflect that.

**Additional context**
This was already reported a while ago (Dec 2017) in #1046 and received a response by @ashleysommer, but the issue was closed without the documentation being fixed.

ahopkins:
Thanks for bringing this back up. Closing in favor of the PR.

PR #1846, opened 2020-05-13
Update docs to clarify response middleware execution order

kirilyer:
Added a clarification regarding the execution order of the request and response middlewares.

Addresses #1845

### DIFF
--- a/docs/sanic/middleware.rst
+++ b/docs/sanic/middleware.rst
@@ -14,8 +14,8 @@ There are two types of middleware: request and response. Both are declared
 using the `@app.middleware` decorator, with the decorator's parameter being a
 string representing its type: `'request'` or `'response'`.
 
-* Request middleware receives only the `request` as argument.
-* Response middleware receives both the `request` and `response`.
+* Request middleware receives only the `request` as an argument and are executed in the order they were added.
+* Response middleware receives both the `request` and `response` and are executed in *reverse* order.
 
 The simplest middleware doesn't modify the request or response at all:
 
@@ -64,12 +64,12 @@ this.
 
     app.run(host="0.0.0.0", port=8000)
 
-The three middlewares are executed in order:
+The three middlewares are executed in the following order:
 
 1. The first request middleware **add_key** adds a new key `foo` into request context.
 2. Request is routed to handler **index**, which gets the key from context and returns a text response.
-3. The first response middleware **custom_banner** changes the HTTP response header *Server* to say *Fake-Server*
-4. The second response middleware **prevent_xss** adds the HTTP header for preventing Cross-Site-Scripting (XSS) attacks.
+3. The second response middleware **prevent_xss** adds the HTTP header for preventing Cross-Site-Scripting (XSS) attacks.
+4. The first response middleware **custom_banner** changes the HTTP response header *Server* to say *Fake-Server*
 
 Responding early
 ----------------


docs/sanic/middleware.rst
```diff
@@ -14,8 +14,8 @@ There are two types of middleware: request and response. Both are declared
 using the `@app.middleware` decorator, with the decorator's parameter being a
 string representing its type: `'request'` or `'response'`.
 
-* Request middleware receives only the `request` as argument.
-* Response middleware receives both the `request` and `response`.
+* Request middleware receives only the `request` as an argument and are executed in the order they were added.
```
ahopkins, line 17: Change  "and are" to "and is"

docs/sanic/middleware.rst
```diff
@@ -14,8 +14,8 @@ There are two types of middleware: request and response. Both are declared
...
+* Response middleware receives both the `request` and `response` and are executed in *reverse* order.
```
ahopkins, line 18: same

ashleysommer:
Just looking into this again to verify this change is correct.
I believe the code used to have a `list.reverse()` function in the run_response_middleware routine, which would make the execution order in the old documentation correct. But seems that was removed at some point, so for at least the last 2 years they are executed in the manner @kirilyer describes.

However during this checking I believe I've uncovered another bug introduced by PR https://github.com/huge-success/sanic/pull/1690. This line: https://github.com/huge-success/sanic/blob/e7001b00747b659f7042b0534802b936ee8a53e0/sanic/app.py#L656 in `register_named_middleware` should have `appendleft()` rather than `append()` to match the behavior of the normal `register_middleware` method.

docs/sanic/middleware.rst
ashleysommer: I think 'middleware' is used as a plural in this context. See "the order _they_ were added", implying plural. So "and are" is more correct in this sentence.

For this reason I've personally started using the term _middlewares_ to mean plural, and _middleware_ to mean singular, to reduce the ambiguity.

ashleysommer:
@kirilyer As part of this PR, can you please add a new example file in the `examples` directory, named `middleware_execution_order.py` with the following content (or something similar).

```
from sanic import Sanic
'''
Demonstrates that request middleware are executed in the order they are added
And response middleware are executed in _reverse_ order.
On a valid request, it should print "1 2 3 6 5 4" to terminal
'''

app = Sanic()

@app.middleware('request')
def request_middleware_1(request):
    print('1')

@app.middleware('request')
def request_middleware_2(request):
    print('2')

@app.middleware('request')
def request_middleware_3(request):
    print('3')

@app.middleware('response')
def resp_middleware_4(request, response):
    print('4')

@app.middleware('response')
def resp_middleware_5(request, response):
    print('5')

@app.middleware('response')
def resp_middleware_6(request, response):
    print('6')

@app.get('/')
def pop_handler(request):
    return text('hello world')

app.run(host="0.0.0.0", port=8000, debug=True, auto_reload=False)
```

ahopkins:
Did you take a look at history to see if there was in fact `reverse`?

ashleysommer:
Looks like there was never `reverse`. I think I was misremembering that from this section of code relating to Listeners: https://github.com/huge-success/sanic/blame/e5c7589fc057ec2900ecca68110e6b4d2169eae2/sanic/app.py#L1320

Anyway, I'm pretty sure we can all agree the correct order is how it currently implemented, and how @kirilyer has described in this PR.

Tronic:
@ahopkins I think you need to mark this resolved and accepted, which appears to be already agreed in the discussion.

PR #1852, opened 2020-05-14
fix-#1851: correct step name

### LINKED ISSUE #1851: Test name against nightly without extensions is incorrect
It is "Python nightly Extensions" instead of "Python nightly without Extensions"

DavidBord:
Fixes #1851

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ matrix:
       name: "Python nightly with Extensions"
     - env: TOX_ENV=pyNightly-no-ext
       python: 'nightly'
-      name: "Python nightly Extensions"
+      name: "Python nightly without Extensions"
 install:
   - pip install -U tox
   - pip install codecov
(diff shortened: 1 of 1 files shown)


ashleysommer:
I noticed this discrepancy in Travis a few days ago. Thanks for fixing.

.travis.yml
```diff
@@ -78,7 +78,7 @@ matrix:
       name: "Python nightly with Extensions"
     - env: TOX_ENV=pyNightly-no-ext
       python: 'nightly'
-      name: "Python nightly Extensions"
+      name: "Python nightly without Extensions"
```
ahopkins, line 81: Need to also make the change to the name up on line 74 so it does not break the build. This fails Travis on this build, but it should allow failure.

Tronic: @ahopkins Fix implemented, can you finish the review?

PR #1842, opened 2020-05-07
Fix static _handler pickling error.

ashleysommer:
Moves the subfunction _handler out to a module-level function, and parameterizes it with functools.partial().
Fixes the case when picking a sanic app which has a registered static route handler. This is usually encountered when attempting to use multiprocessing or auto_reload on OSX or Windows.
Fixes #1774

PR #1853, opened 2020-05-14
Fix pickle error when attempting to pickle an application which contains websocket routes

ashleysommer:
Similar fix to PR #1842, this fixes the same problem but for websocket routes.
Moves the `websocket_handler` subfunction out to a class-level method, which can be more easily pickled by the built-in python Pickler.
Also includes a small addition of the same fix for the `app.add_task()` deferred task scheduler subfunction.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -117,24 +117,12 @@ class Sanic:
         :param task: future, couroutine or awaitable
         """
         try:
-            if callable(task):
-                try:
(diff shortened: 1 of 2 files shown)


ashleysommer:
@ahopkins just a heads up, I don't think the change in server.py was supposed to be part of this fix, I might've been an accidental inclusion. I'll double check.

ahopkins:

PR #1857, opened 2020-05-16
fix-#1856: adjust websockets version to setup.py and make nightly (py39) tests pass

### LINKED ISSUE #1856: Nightly build fails due to websockets version not matching setup.py
on setup.py: >=0.7.0,<0.9
on tox.ini: >=0.7.0,<0.8

DavidBord:
Fixes #1856

### DIFF
--- a/sanic/websocket.py
+++ b/sanic/websocket.py
@@ -113,7 +113,7 @@ class WebSocketProtocol(HttpProtocol):
 
         # hook up the websocket protocol
         self.websocket = WebSocketCommonProtocol(
-            timeout=self.websocket_timeout,
+            close_timeout=self.websocket_timeout,
             max_size=self.websocket_max_size,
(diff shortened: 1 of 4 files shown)


DavidBord:
@ahopkins you want me to change the minimal supported websockets version to be 8.1?

ahopkins:
Yes. let's see what it breaks. I think there was a change with the handling of subprotocols.

DavidBord:
@ahopkins to accomplish what you wrote in your last comment I think we'll have to either have non matching websockets versions between setup.py & tox.ini or keep py3.9 tests failing.

Tronic:
Unrelated test failure, possibly a random mishap in test_reloader that'll need to be addressed in another PR. This one is good to go if only the test failure can be bypassed/ignored.

PR #1864, opened 2020-05-30
Fallback to psutil when /proc/.../children isn't present

Kamori:
I'm developing a project using Ubuntu 18.04 on Windows WSL2. For some reason `/proc/{pid}/task/{pid}/children` isn't available. When this happened, we return with no good explanation why.

To make sure we don't break current users, I added a fallback to psutil to discover the children and kill them and move on. 

Hope this helps others who use WSL2

Edit: This affects the live reloading feature

### DIFF
--- a/sanic/reloader_helpers.py
+++ b/sanic/reloader_helpers.py
@@ -2,7 +2,6 @@ import os
 import signal
 import subprocess
 import sys
-
 from multiprocessing import Process
 from time import sleep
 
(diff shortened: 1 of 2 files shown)


Tronic:
The code patched here is removed entirely in #1827 which rewrites the autoreloader kill logic (i.e. the autoreloader signals the main server, which then signals its children, avoiding the need to kill all children).

@Kamori Would you mind testing that PR to see if it also works for you?

ahopkins:
@Kamori Thanks for the PR. I am placing on hold until we see if this is still necessary once #1827 is merged.

Kamori:
First test seems to work well! I'm gonna do a bit of development today to make sure I don't run into any issues. I'll close this ticket out if all looks good. Thanks everyone!

PR #1865, opened 2020-05-30
Fixing rst format issue

keul:


keul:
The bash command was invisible on final document. See https://sanic.readthedocs.io/en/latest/sanic/deploying.html#running-via-command

PR #1866, opened 2020-06-02
Add sanic as an entry point command

ahopkins:
One thing that I have always thought missing was this:

```
$ sanic server.app --port=5000 --workers=4
```

Since we already have...

```
$ python -m sanic server.app --port=5000 --workers=4
```

... this PR just goes one step further to adding the entrypoint to run the Sanic server with the `sanic` command.

### DIFF
--- a/sanic/__main__.py
+++ b/sanic/__main__.py
@@ -1,3 +1,5 @@
+import os
+import sys
 from argparse import ArgumentParser
 from importlib import import_module
 from typing import Any, Dict, Optional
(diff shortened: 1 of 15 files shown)


ashleysommer:
@yunstanford 
>probably better to expose entry_points via "entry_points/console_scripts" in setup.py

I might be wrong, but I believe that is what this change provides.

ahopkins:
Exactly.

yunstanford:
ahh, sorry, didn't read that carefully.. just saw that

```
if __name__ == "__main__":
    main()
```
and thought differentlly.

PR #1876, opened 2020-06-21
Streaming Server

### LINKED ISSUE #1653: In ASGI, headers is output as body
simple test code
```Python3
@app.get('')
async def test(request):
    return await file_stream(...)
```

The HTTP response will output twice headers.
One of them is in the headers (✅), the other time in the body (❎)

### LINKED ISSUE #1722: curl is not agree wit return in chunk mode

### LINKED ISSUE #1730: ASGI Streaming response contains extra content

### LINKED ISSUE #1749: Sanic 19.12 breaks some middleware usages
The commit _GIT-37: fix blueprint middleware application_ (https://github.com/huge-success/sanic/commit/a6077a179067812766293f9ab4391b5e1499a9c6#diff-c4444a35dbd5e37ee480b0e8888e0880R939) changes how the middleware are evaluated. 

Before this commit the middleware where evaluated before the route was determined. This allowed for example to rewrite to requested url. In my case I was using that to remove a url-prefix that is used for routing purposes.

I do think this is a legitimate use case for middlewares. According to the documentation:

> Middleware are functions which are executed before or after requests to the server. They can be used to modify the request to or response from user-defined handler functions.

modifying the requested url should be possible.

### LINKED ISSUE #1785: Error handlers are synchronous

### LINKED ISSUE #1804: Why define an exception handler needed async?

### LINKED ISSUE #1824: How to increase timeout on specific method?
Is there any option to increase timeout and request size on the top of method such as decorators?
Sometimes we need to increase our time of request such as uploading files.
Also, I asked a question on [Stackoverflow](https://stackoverflow.com/questions/60915546/how-to-increase-timeout-while-uploading-on-the-sanic-framework) about this issue.

### LINKED ISSUE #1875: Response timeout error from all routes after returning several timeouts from a specific route
**Response timeout error**
Return "Response timeout" error from all routes after returning several timeouts from a specific route.

**Codes**
Set response timeout to 5 sec in config:
`RESPONSE_TIMEOUT = 5`

Define funcs:
```
@app.get('test')
async def test_func(request):
    """Timeout func"""
    await asyncio.sleep(10)
    ...
@app.get('test2')
async def test_func2(request):
    """Func without timeout"""
    return response.text('OK')
```
Make several timeout responses:
```
for i in range(20):
    resp = requests.get('/test')
```
Then if I request all other endpoints, it returns "Response timeout" error:
`resp = requests.get('/test2')`

**Environment:**
 - OS: Ubuntu
 - Python: 3.8.2
 - Sanic: 20.3.0

### LINKED ISSUE #1988: Handling of safe methods with body

Currently, Sanic will allow a request body to be accepted (and parsed) even on safe methods (GET, HEAD, DELETE, OPTIONS). This is _technically_ accurate [since the specs do not require](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) that a server disallow this behavior.

According to Mozilla:

> ... while not prohibited by the specification, the semantics are undefined.

It is is, however, somewhat counter-intuitive.

Add a flag (`ignore_body`) to the save method decorators that defaults to `True`. If it receives a safe request with a body while `ignore_body == True`, then it should **not** parse the body.

Tronic:
Moved from https://github.com/huge-success/sanic/pull/1791

A plain Python reimplementation of Sanic's built-in server and related restructuring also in ASGI backend and in response handling. Includes first-class support for streaming, improved performance and better ground for further development.

By the sheer volume of it, it also brings some breaking changes like the removal of custom protocol support but typical applications that do not do anything special with the server's internals should remain mostly unaffected.

Installation:
```
pip install git+https://github.com/huge-success/sanic@streaming
```

Issues of current git master, resolved in this:
Fix #1653
Fix #1722
Fix #1730
Fix #1749 
Fix #1785
Fix #1790
Fix #1804
Fix #1824
Fix #1875
Fix #1988


ahopkins:
I am sure there are already a number of conflicts here. Before it gets too wildly out of sync, do you want to try to do a rebase (or merge, not picky) to bring this back current to `master`? I am thinking that I want to keep this at the forefront so it is ready to merge asap after December's LTS, with some of the other major changes that are waiting to follow after this.

Tronic:
@ahopkins Help with that would be nice. I'm busy with work and academia pretty much 7 days every week until Nov/Dec, also you are probably better familiar with all the patches merged to master meanwhile. I believe it would be easier to merge them to this branch than rebase the whole thing.

ahopkins:
@Tronic. Sounds good. :metal: 

After the release this week ...

ahopkins:
I merged in recent changes to master. I did not spend any time on the tests. Looks like about 12 need some attention. I can take a look at these this week.

ahopkins:
@huge-success/sanic-core-devs Please take a look at this one. I'd like to get this merged by the end of the weekend. Even if it is not 100% perfect and release worthy, I want to get this merged because of the large number of changes that it entails. Having it merged into `master` will allow us to continue making other PRs without resolving too many conflicts. Release it not set until March, but I want to make sure there is **plenty** of time to test this in `master`.

ahopkins:

https://community.sanicframework.org/t/new-streaming-api-and-other-major-changes/518/15

ahopkins:

https://community.sanicframework.org/t/2021-release-managers-and-updates/781/1

sanic/asgi.py
```diff
@@ -257,161 +241,57 @@ async def create(
                 instance.transport,
                 sanic_app,
             )
+            instance.request.stream = instance
+            instance.request_body = True  # FIXME: Use more_body?
```
ahopkins, line 245: Needs resolution.

sanic/request.py
```diff
@@ -135,36 +110,43 @@ def __repr__(self):
...
-        """.. deprecated:: 20.3
-        To be removed in 21.3"""
-        self.body.append(data)
-
-    def body_finish(self):
-        """.. deprecated:: 20.3
-        To be removed in 21.3"""
-        self.body = b"".join(self.body)
+    async def respond(
+        self, response=None, *, status=200, headers=None, content_type=None
```
ahopkins, line 114: Need to take a closer look to confirm that this is actually the correct place for this.

sanic/response.py
```diff
@@ -28,50 +27,51 @@ def _encode_body(self, data):
...
-        else:
-            self.headers["Connection"] = "close"
+    @property
+    def processed_headers(self):
+        """Obtain a list of header tuples encoded in bytes for sending.
 
-        if self.status in (304, 412):
+        Add and remove headers based on status and content_type.
+        """
+        # TODO: Make a blacklist set of header names and then filter with that
```
ahopkins, line 42: Resolve.

sanic/response.py
```diff
@@ -81,63 +81,28 @@ def __init__(
         status=200,
         headers=None,
         content_type="text/plain; charset=utf-8",
-        chunked=True,
+        chunked="deprecated",
```
ahopkins, line 84: Remove?

sanic/response.py
```diff
@@ -224,7 +173,15 @@ def text(
         raise TypeError(
             f"Bad body type. Expected str, got {type(body).__name__})"
         )
-
+    # Type conversions are deprecated and quite b0rked but still supported for
+    # text() until applications get fixed. This try-except should be removed.
+    try:
+        # Avoid repr(body).encode() b0rkage for body that is already encoded.
+        # memoryview used only to test bytes-ishness.
+        with memoryview(body):
```
ahopkins, line 181: Let's benchmark this.

ahopkins (changes requested):
I'll handle these. When done, I want to merge. @sanic-org/sanic-core-devs any other thoughts before I do so?

ahopkins (approved):

ashleysommer:
Just looking over these merged changes now, to learn how Sanic works at a low level now.

I have noticed it looks like there is a possibility for the Request Middleware to run twice on a request.
Both the `app.handle_request` and `app.handle_exception` methods execute the Request middleware.

If the main handler encounters an exception _after_ running the middleware, then the exception handler will execute and run the request middleware _again_.

ahopkins:
Yeah. That is the purpose of it.  I think eventually I want to get rid of it and not have to carry that state trigger (there is not a similar one for response).

ahopkins:

https://community.sanicframework.org/t/highly-different-cpu-usage-when-proxying-m2tp-udp-unicast-stream-to-a-streamed-sanic-response/795/10

PR #1880, opened 2020-06-28
Add handler names for websockets for url_for usage

ahopkins:
Accomplishes what #1696 set out to do, but was unable after change to #1853

### DIFF
--- a/tests/test_reloader.py
+++ b/tests/test_reloader.py
@@ -1,8 +1,8 @@
 import os
 import secrets
 import sys
-from contextlib import suppress
 
+from contextlib import suppress
 from subprocess import PIPE, Popen, TimeoutExpired
 from tempfile import TemporaryDirectory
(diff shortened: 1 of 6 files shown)


tests/test_reloader.py
```diff
@@ -23,16 +23,20 @@
...
         proc.terminate()
 
+
 def write_app(filename, **runargs):
     text = secrets.token_urlsafe()
     with open(filename, "w") as f:
-        f.write(dedent(f"""\
+        f.write(
+            dedent(
+                f"""\
```
Tronic, line 39: I suppose this was by linter because the original formatting was better. Maybe reindent the rest of the string literal too to match the extra level at opening?

ahopkins: Yes. `make fix-import`

Oh trust in the almighty `black`.

PR #1884, opened 2020-06-29
Revert change to multiprocessing mode

ashleysommer:
Revert change to multiprocessing mode accidentally included in https://github.com/huge-success/sanic/pull/1853

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -927,7 +927,7 @@ def serve_multiple(server_settings, workers):
 
     signal_func(SIGINT, lambda s, f: sig_handler(s, f))
     signal_func(SIGTERM, lambda s, f: sig_handler(s, f))
-    mp = multiprocessing.get_context("spawn")
+    mp = multiprocessing.get_context("fork")
 
     for _ in range(workers):
         process = mp.Process(target=serve, kwargs=server_settings)


sanic/server.py
```diff
@@ -927,7 +927,7 @@ def sig_handler(signal, frame):
 
     signal_func(SIGINT, lambda s, f: sig_handler(s, f))
     signal_func(SIGTERM, lambda s, f: sig_handler(s, f))
-    mp = multiprocessing.get_context("spawn")
+    mp = multiprocessing.get_context("fork")
```
Tronic, line 930: This should restore prior functionality. Do note that it won't work on Windows but Sanic doesn't support multiple workers on Windows at this time anyway (although that really ought to be fixed, PRs welcome!).

ahopkins:
Should we include this in `20.6`?

Tronic:
> Should we include this in `20.6`?

Yes, absolutely, and I believe that with 20.6.1 people will get major b0rkage in their apps because not even the standard Hello World will work with multiple workers. Perhaps roll another release today with this and the unix socket PR? The pickling issue is perhaps important enough to add specific tests to avoid it happening again (actually I am surprised that the current multiple worker tests do not hit it at all), but then that is probably better left for a later release.

Great work with getting the release rolled out despite all the difficulties. In future hopefully we can get PRs merged in good time prior to release, to have releases rolled out in a more orderly fashion.

ahopkins:
Github is throwing 500's right now on merges. Will try again in a little bit to see if they resolved.

ahopkins:

Issue #1878, opened 2020-06-24
WebsocketConnection does not send subprotocol in accept


darkvariantdivine:
**WebsocketConnection does not send subprotocols during accept**
Client sent graphql-ws subprotocol during websocket handshake, Sanic replied with an empty string instead of populating data extracted from scope

**Code snippet**

```python
class WebSocketConnection:

    # TODO
    # - Implement ping/pong

    def __init__(
        self,
        send: Callable[[ASIMessage], Awaitable[None]],
        receive: Callable[[], Awaitable[ASIMessage]],
    ) -> None:
        self._send = send
        self._receive = receive

    ...

    async def accept(self) -> None:
        await self._send({"type": "websocket.accept", "subprotocol": ""})
```
In the accept function, the subprotocol field is hardcoded as an empty string

**Expected behavior**
Allow the subprotocol field to be populated from an ASGI client or external scope

**Environment (please complete the following information):**
 - OS: Fedora 30
 - Python: 3.7.7
 - Sanic: 19.12.2, 20.3.0 

**Additional context**
Sanic is a simple and elegant framework, I would love to help out with a PR but will need some direction. :)

ahopkins:
You are running ASGI server? Websockets in ASGI mode do still need some work. I believe when I added ASGI support, they were not yet supported in the ASGI spec. It looks like they are there now, so it should be a relatively small PR if you would like to give it a shot.

Without looking too deeply at this right now, I believe you need to grap `scope['subprotocols']` in here:

https://github.com/huge-success/sanic/blob/master/sanic/asgi.py#L236

darkvariantdivine:
@ahopkins thanks for the guidance, I'll be following up with a PR on this.

PR #1887, opened 2020-07-01
Websocket subprotocol

darkvariantdivine:
Fix for issue #1878, @ahopkins for your review

### DIFF
--- a/sanic/websocket.py
+++ b/sanic/websocket.py
@@ -5,7 +5,7 @@ from typing import (
     Dict,
     MutableMapping,
     Optional,
-    Union,
+    Union, List,
 )
 
 from httptools import HttpParserUpgrade  # type: ignore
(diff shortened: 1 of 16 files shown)


sanic/websocket.py
```diff
@@ -137,9 +137,11 @@ def __init__(
         self,
         send: Callable[[ASIMessage], Awaitable[None]],
         receive: Callable[[], Awaitable[ASIMessage]],
+        subprotocols: List[str]=[],
```
ahopkins, line 140: This should default to `None`. If there needs to be a default list, that should be handled inside the function. [See rationale](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

```
subprotocols: Optional[List[str]]=None

...

subprotocols = subprotocols or []
```

tests/test_asgi.py
```diff
@@ -208,6 +208,21 @@ async def test_websocket_receive(send, receive, message_stack):
...
+    subprotocols = ['graphql-ws', 'test']
+
+    ws = WebSocketConnection(send, receive, subprotocols)
+    await ws.accept()
+
+    assert len(message_stack) == 1
+
+    message = message_stack.popleft()
+    assert message["type"] == "websocket.accept"
+    assert message['subprotocol'] == "graphql-ws,test"
```
ahopkins, line 222: Let's also add a test where there is only 1 subprotocol.

sanic/websocket.py
darkvariantdivine: Updated, thanks for the advice!

tests/test_asgi.py
darkvariantdivine: Added test for single subprotocol and no subprotocols

darkvariantdivine:
@ahopkins I fixed all the linter errors and made all the changes you requested :)

PR #1890, opened 2020-07-06
Adjust isort options and invocation to work with isort 5.0.0

ashleysommer:
isort 5.0.0 removed command line option `recursive` and removed config option `not_skip`.

This fixes #1889

PR #1891, opened 2020-07-07
Fix imports and isort to remove from Makefile deprecated options

ahopkins:


ahopkins:
on the tail of #1890

ahopkins:
Linting is passing, but looks like it is complaining about `trio` annotations. :man_facepalming:

ashleysommer:
I added back in the missing `#  type: ignore` hint after the `from trio import ...` line was split in `compat.py`, hope that fixes it.

ashleysommer:
Yep, travis is passing now.

PR #1894, opened 2020-07-09
Test mode

ahopkins:
One thing that I plan to add to the [`sanic-test`](https://github.com/huge-success/sanic-test) module is the ability for Sanic app to be aware of whether or not it is being tested. This adds a variable to the Sanic app that is set by the internal test clients.


Tronic:
What's up with the failing tests, is it related to this one?

ahopkins:
Ummm.... this is super weird. At one point I had added a check [here](https://github.com/huge-success/sanic/blob/050a563e1d5b0bba6f5e037152ef3229b3fe52be/sanic/asgi.py#L120) to check against `test_mode` and not throw that warning. Which would make the test that failed fail. We would have needed to change it. But... I decided against that since if someone wants to suppress that warning it would be simple enough, and it might be good for them to see it in their test suite and ask themselves why that is.

But, none of that happened. I left the test the way it is. 

I am going to resubmit the test suite.

ahopkins:
Ignore the ridiculous number of commits. I was going too fast.

ahopkins:
Ready for review and merge.

PR #1897, opened 2020-07-25
feat: fixes exception due to unread bytes in stream

imgurbot12:
Hi there! I noticed that the stream handlers for sanic seem to raise an exception when stream data is not read entirely or the request is cancelled for whatever reason, which was annoying me slightly so I found an easy fix and am hoping to get it merged into master.

I have a test script to replicate the error. There are two functions in the example server instance.
1. `works` will run just fine without any tracebacks because it reads all of the stream data
2. `errror` will raise a traceback due to the fact that the request object gets deleted after request completion, but the byte-reader is still active because there is still data in the buffer

**Fix Description:**
I fixed it by simply adding an additional check that will ignore any request operations if the request object is no longer present
allowing the remaining bytes in the buffer to simply be discarded.

**Example Traceback:**
```
future: <Task finished name='Task-6' coro=<HttpProtocol.stream_append() done, defined at /home/andrew/Desktop/code/github/sanic/sanic/server.py:416> exception=AttributeError("'NoneType' object has no attribute 'stream'")>
  File "/home/andrew/Desktop/code/github/sanic/sanic/server.py", line 419, in stream_append
    if self.request.stream.is_full():
AttributeError: 'NoneType' object has no attribute 'stream'
```

**Sample Script for Replication of Error:**
```python
from sanic import Sanic

app = Sanic(__name__)

#NOTE: this will raise a traceback with the stream left unread
@app.route('/error', methods=['POST'], stream=True)
async def error(request):
    return json({'Hello': 'World!'})

#NOTE: this works just fine because the stream content is read completley
@app.route('/works', methods=['POST'], stream=True)
async def works(request):
    while True:
        body = await request.stream.read()
        if body is None:
            break
    return json({'Hello': 'World!'})

```

**Sample Requests Made w/ Curl:**
```bash
curl -d 'this is a test' -v 'http://127.0.0.1:8000/error'
curl -d 'this is a test' -v 'http://127.0.0.1:8000/works'
```


Tronic:
Note: this is already fixed in #1876 (streaming branch) which prints a proper diagnostic and no traceback:

```
[2020-07-26 13:50:46 +0300] [17097] [INFO] Goin' Fast @ http://127.0.0.1:8000
[2020-07-26 13:50:46 +0300] [17097] [INFO] Starting worker [17097]
[2020-07-26 13:50:56 +0300] - (sanic.access)[INFO][127.0.0.1:49252]: POST http://127.0.0.1:8000/error  200 -1
[2020-07-26 13:50:56 +0300] [17097] [ERROR] <Request: POST /error> body not consumed.
[2020-07-26 13:50:56 +0300] - (sanic.access)[INFO][127.0.0.1:49253]: POST http://127.0.0.1:8000/works  200 -1
```

This PR is still useful because the streaming branch is postponed until Sanic 21.03 or so.

ahopkins:
I'm working on a fix for the current branch to patch between 20.9 and 20.12 since that branch is set for 21.3

Tronic:
@imgurbot12 Could you add tests so that this has 100 % patch coverage?

imgurbot12:
Hi @Tronic, very sorry about the delay. I'm afraid I'm not very familiar with this sort of testing coverage benchmark. What tests do I need to improve/write to add coverage for the single line that I've added? I can't seem to find any tests in `sanic/tests` that actually run `stream_append` directly.

Tronic:
Can you add another function in `tests/test_request_stream.py` that would hit this case? I think the `test_request_stream_handle_exception` function is similar to what you need to test, just give yours another name and modify so that it hits the bug and is handled in your patch.

Coverage comes automatically from all lines of code actually hit by at least a single test case.

imgurbot12:
hrm. Well it seems like the unit-tests passed. I'm not sure whats wrong with the linter that's causing it to error now

ahopkins:
Run `make fix-import` commit that and we should be good. I am going to bring this current to `master.

ahopkins (approved):

tests/test_request_stream.py
```diff
@@ -336,6 +338,20 @@ async def post(request, id):
...
+    err = None
+    protocol = HttpProtocol(loop=asyncio.get_event_loop(), app=app)
+    try:
+        protocol.request = None
+        protocol._body_chunks.append('this is a test')
+        await protocol.stream_append()
+    except AttributeError as e:
+        err = e
+
+    assert err is None and not protocol._body_chunks
```
Tronic: Please use `app.test_client` to test, instead of directly touching Sanic internals (like `_body_chunks`). Or did that not work for some reason (I know, streaming can be problematic with it).

ahopkins (approved):

ahopkins (approved):

imgurbot12:
@ahopkins That should hopefully be the last of the automated changes lol. Thanks

ahopkins:
> @ahopkins That should hopefully be the last of the automated changes lol. Thanks

Thanks for this. Will get it merged once the tests are done.

imgurbot12:
> > @ahopkins That should hopefully be the last of the automated changes lol. Thanks
> 
> Thanks for this. Will get it merged once the tests are done.

Yeah no problem! Glad to help out :)

Tronic (changes requested):
Tests should use the test client whenever possible, so that they don't break when server's internal implementation is modified as long as the external functionality stays the same.

tests/test_request_stream.py
imgurbot12: I wasn't sure how to access `steam_append` to do a proper test, so i just opted for accessing the object directly. I can try replacing references to body_chunks and instead just pass in a basic request-object into it using `on_headers_complete` to generate the request and then `on_body`. would that work?

ahopkins: For this test, I am not sure that is necessary. At issue is the non-existence of the request object. If we want to add an additional end-to-end test, that works as well. But I think this is a valid solution.

Tronic: I suppose so. In this case I won't be taking this test to the streaming branch at all but it should suffice in master for now.

Tronic (approved):

tests/test_request_stream.py
ahopkins: > I wasn't sure how to access `steam_append` to do a proper test, so i just opted for accessing the object directly. I can try replacing references to body_chunks and instead just pass in a basic request-object into it using `on_headers_complete` to generate the request and then `on_body`. would that work?

You can get access through the `request` object.

```
request.transport.get_protocol()
```

PR #1770, opened 2020-01-21
WIP: add initial signals class definitions

harshanarayana:
Implementation for #1630

Issue #1630, opened 2019-07-11
RFC : Signals Support for Sanic Proposal


harshanarayana:
# Proposal : Signals Support for Sanic

- [Proposal : Signals Support for Sanic](#proposal---signals-support-for-sanic)
  * [Introduction](#introduction)
  * [Background](#background)
  * [Proposal](#proposal)
    + [Signal Description](#signal-description)
    + [Implementation Details](#implementation-details)
      - [Helper Utilities](#helper-utilities)
      - [Implementation Requirements](#implementation-requirements)
      - [Impacted Areas](#impacted-areas)
    + [Implementation Guidelines](#implementation-guidelines)
  * [To Do](#to-do)

## Introduction

This proposal defines a brief set of guidelines and behaviors with regards to a new set of entities in the `sanic` modules called `signals`. These entities will serve as a standalone entities during the lifecycle of a `sanic` request and will be used as a mechanism to signal a certain kind of even being taken place under the respective core components or any `sanic` extension that wants to adopt the signaling behavior.

## Background

Currently, `sanic` framework provides you a way to bind into the `request` and  `response` lifecycle of an incoming API request via the `middleware` support as `@app.middleware("request")` or `@app.middleware("response")`. However, this adds a strict binding between your code and how you want to process these events.

This can be cumbersome when it comes to adding additional `sanic` extensions that can help in monitoring and metrics tracking purpose. This introduces the following set of drawbacks.

1. The process required to handle events is coupled with business logic. But the way to handle these events is very generic. So, there is no reason for them to be co-located with the business logic. Not even in the form of a `decorator`
2. This adds an additional delay to each `request-response` cycle and this is not always a preferred behavior. All you wanted to do was to monitor the events being triggered in the lifecycle of an event. But tracing behavior should not introduce additional delay

So this proposal suggests a `signal` as a way to bypass this constraint. This will let you decouple the way the signals are handled and that process can reside in it's own context without impacting the business logic. 

This proposal in it's basic form is a simple `pub-sub` model implementation where the extensions and the `sanic` core components will generate certain events and the subscribers to those events can handle them the way they all see fit. 

## Proposal

A signal in it's most simple form is just an `event` like in any of your `pub-sub` models. And the possible behavior is as follow.

1. `sanic` core components will be responsible for generating a certain set of events described below. 
   1. Server Lifecycle Event
      1. starting_server
      2. server_started
      3. stopping_server
      4. server_stopped
   2. Request Lifecycle Events
[...]

PR #1902, opened 2020-08-02
Rfc/git 1630 signals

ahopkins:
Take over from #1770 re: #1630

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -3,7 +3,6 @@ import logging.config
 import os
 import re
 import warnings
-
 from asyncio import CancelledError, Protocol, ensure_future, get_event_loop
 from collections import defaultdict, deque
 from functools import partial
@@ -21,16 +20,12 @@ from sanic.config import BASE_LOGO, Config
 from sanic.constants import HTTP_METHODS
 from sanic.exceptions import SanicException, ServerError, URLBuildError
 from sanic.handlers import ErrorHandler
+from sanic.helpers import publish, subscribe
 from sanic.log import LOGGING_CONFIG_DEFAULTS, error_logger, logger
 from sanic.response import HTTPResponse, StreamingHTTPResponse
 from sanic.router import Router
-from sanic.server import (
-    AsyncioServer,
-    HttpProtocol,
-    Signal,
-    serve,
-    serve_multiple,
-)
+from sanic.server import AsyncioServer, HttpProtocol, Signal, serve, serve_multiple
+from sanic.signals import Namespace
 from sanic.static import register as static_register
 from sanic.testing import SanicASGITestClient, SanicTestClient
 from sanic.views import CompositionView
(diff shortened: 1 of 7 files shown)


ashleysommer:
@ahopkins 
Thanks for taking this over. I'm very interested in this Signals work, because it is integral for the future of Sanic-Plugins-Framework.
I'll read over the proposed implementation and see if I can offer feedback.

sanic/app.py
```diff
@@ -146,7 +152,7 @@ def register_listener(self, listener, event):
         :param event: when to register listener i.e. 'before_server_start'
         :return: listener
         """
-
+        subscribe(event_name=event, callback=listener, signals=self.signals)
```
ashleysommer, line 155: If self.listener() decorator calls `subscribe()` (on line 141 above) then why does this helper function also need to call `subscribe()`? Wouldn't this result in two subscribes with the same listener and signals?

sanic/app.py
```diff
@@ -219,9 +225,7 @@ def response(handler):
         return response
 
     # Shorthand method decorators
-    def get(
-        self, uri, host=None, strict_slashes=None, version=None, name=None
-    ):
+    def get(self, uri, host=None, strict_slashes=None, version=None, name=None):
```
ashleysommer, line 228: Looks like quite a lot of the changes in this file are formatting changes, was that just a result of merging in latest master? Or should these formatting changes be filtered out?

sanic/helpers.py
```diff
@@ -154,3 +188,44 @@ def import_string(module_name, package=None):
     if ismodule(obj):
         return obj
     return obj()
+
+
+def _extract_signal_namespace(event_name, signals):
+    global _CLASSIC_EVENT_ALIAS
```
ashleysommer, line 194: can this be done without using a global variable?

sanic/helpers.py
```diff
@@ -3,6 +3,9 @@
 from importlib import import_module
 from inspect import ismodule
 
+from sanic.signals import Namespace
```
ashleysommer, line 6: Looks like `Namespace` is only imported for Type checking, so has no purpose at runtime.

This import can changed to:
```
if typing.TYPE_CHECKING:
    from sanic.signals import Namespace
```
Then in type declarations, change `Namespace` to `'Namespace'` (string form)

sanic/app.py
```diff
@@ -92,6 +87,18 @@ def __init__(
         self.go_fast = self.run
         self.test_mode = False
 
+        # Signal Handlers and Registry
+        self._signals = {
+            "server": Namespace(namespace="server", owner=self, classic_event=True,),
+            "request": Namespace(namespace="request", owner=self),
+            "response": Namespace(namespace="response", owner=self),
+            "middleware": Namespace(namespace="middleware", owner=self),
+        }
```
ashleysommer, line 96: Looks like 4 Namespaces of signals are defined, but signal emitters are only implemented for `server` signals. Are the others planned to be implemented?

ahopkins:
I think this is a really great potential win. I've been playing with the branch, and have some ideas on how to refine the implementation. I'll post on the RFC thread some changes I'm thinking about. One is to scrap the additional dependency and just roll our own signaling solution.

ahopkins:
Coming 2021

PR #1905, opened 2020-08-05
making default reponse text, with optional formatting to be passed

ProjitB:
The default error handling should not return html text. Sanic itself can be run as a headless server just for API consumption (ex. CLIs). It would make sense for errors to be propagated with the text they are originally supplied with rather than with surround html tags.  While custom error handling is of course possible, I do believe that this should be the default behavior.

Additionally, custom formatting can be passed as a string as before, just pass `{{body}}`:
`
f"<!DOCTYPE html><meta charset=UTF-8><title>{title}</title><style>{TRACEBACK_STYLE}</style>\n <h1>⚠️ {title}</h1><p>{{body}}\n {_render_traceback_html(request, exception)}"
`

### DIFF
--- a/sanic/errorpages.py
+++ b/sanic/errorpages.py
@@ -35,22 +35,9 @@ def exception_response(request, exception, debug):
     text = escape(text)
 
     if debug and not getattr(exception, "quiet", False):
-        return html(
-            f"<!DOCTYPE html><meta charset=UTF-8><title>{title}</title>"
-            f"<style>{TRACEBACK_STYLE}</style>\n"
-            f"<h1>⚠️ {title}</h1><p>{text}\n"
-            f"{_render_traceback_html(request, exception)}",
-            status=status,
-        )
+        return html(f"{text}", status=status)
 
-    # Keeping it minimal with trailing newline for pretty curl/console output
-    return html(
-        f"<!DOCTYPE html><meta charset=UTF-8><title>{title}</title>"
-        "<style>html { font-family: sans-serif }</style>\n"
-        f"<h1>⚠️ {title}</h1><p>{text}\n",
-        status=status,
-        headers=headers,
-    )
+    return html(f"{text}", status=status, headers=headers)
 
 
 def _render_exception(exception):
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -253,18 +253,24 @@ def raw(
     )
 
 
-def html(body, status=200, headers=None):
+def html(body, status=200, headers=None, formatting=None):
     """
     Returns response object with body in html format.
 
     :param body: str or bytes-ish, or an object with __html__ or _repr_html_.
     :param status: Response code.
     :param headers: Custom Headers.
+    :param formatting: Custom formatting (string) containing {body} for insertion point
     """
     if hasattr(body, "__html__"):
         body = body.__html__()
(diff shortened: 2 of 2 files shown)


Tronic:
A custom error handling middleware is probably the best way to go for machine-readable error messages. Different APIs require vastly different error messages ranging from empty documents (only status code) to plain text message and JSON in various formats.

I think that changing this properly needs more work than simply this patch, which would still show internal server errors as HTML if I am not mistaken (tracebacks etc. are HTML formatted).

ProjitB:
@ahopkins thank you for your comments. I agree with the proposal of having a configurable option. It should be something which is configurable by a user though right? I couldn't see a clean way in the current design to pass user preferences til that point (shouldn't be a global config for the reasons @Tronic mentioned..of different APIs consuming different types of errors).

@Tronic Yes, the current patch would only address the successful responses. I could extend it to the other responses, but I think a configurable option throughout would make sense.

To handle the multiple responses, should a parameter be added to `SanicException`? 
Also, what would be a clean way of getting the plain text response without modifying the code in the current released version? The exceptions like `raise ServerError("<some error")` were being consumed directly before. Upgrading Sanic versions broke that compatibility due to the new html formatted errors.

Tronic:
Error handling middleware has the advantage that it can also contain the logic for determining which type of response is required. Some apps may determine this by request content-type, others might use path such as `/api/` or specific blueprint for custom style responses, some might even need XML or other very custom responses. Overall, this appears to vary a lot from one webapp to another.

ahopkins:
Clearly exception handlers and middleware are the recommended methodology in a production server. However, when prototyping something or getting something started, it is not convenient and often annoying that the exceptions come back in HTML format.

I am not suggesting that we offer full blown solutions since there will clearly need to be considerations that the individual developer needs to make. But, having text and JSON options will be considerably more user-friendly than the current state.

Furthermore, I disagree and think it absolutely should be globally defined. We are talking about the default fallback handler.

```python
if request.app.config.DEFAULT_ERROR_FORMAT == "text":
    resp = text(...)
elif request.app.config.DEFAULT_ERROR_FORMAT == "json":
    resp = json(...)
else:
    resp = html(...)
```

myuz:
Yes, agree with @ahopkins that's the `default error output` should be defined globally, and also do we consider dynamic changing of `default error output`?

ahopkins:
> Yes, agree with @ahopkins that's the `default error output` should be defined globally, and also do we consider dynamic changing of `default error output`?

I am generally not a fan of this. Or at least trying to support this from within Sanic. The problem is that running services like Sanic _generally_ will be scaled horizontally. Syncing config values is tricky and undoubtedly would need another service to achieve that.

**HOUEVER**, I am working on a new user guide with tutorials. A good place to put something like this would be there as a "how could you achieve this" type of reference material.

ashleysommer:
My opinion on this, with highest preference first:
1) Leave it as-is. The default exception handler will output HTML formatted errors, and users can add custom handlers if they want other formats
2) As @ahopkins suggested [above](https://github.com/huge-success/sanic/pull/1905#issuecomment-669798474), have a config variable like `request.app.config.DEFAULT_ERROR_FORMAT = "text"` but keep "html" the default fallback.

ahopkins:
Closing in favor of #1937

PR #1909, opened 2020-08-13
fix websocket ping variables issues

brooklet:
fixed the problem that the websocket ping_timeout and ping_interval parameter settings did not take effect

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -6,7 +6,6 @@ import socket
 import stat
 import sys
 import traceback
-
 from collections import deque
 from functools import partial
 from inspect import isawaitable
@@ -14,7 +13,7 @@ from ipaddress import ip_address
 from signal import SIG_IGN, SIGINT, SIGTERM, Signals
 from signal import signal as signal_func
 from time import time
-from typing import Type
+from typing import Dict, Type, Union
 
 from httptools import HttpRequestParser  # type: ignore
 from httptools.parser.errors import HttpParserError  # type: ignore
@@ -32,7 +31,6 @@ from sanic.exceptions import (
 from sanic.log import access_logger, logger
 from sanic.request import EXPECT_HEADER, Request, StreamBuffer
 from sanic.response import HTTPResponse
-
 
 try:
     import uvloop  # type: ignore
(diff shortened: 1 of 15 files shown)


ahopkins:
@brooklet Can you take a look at the failing tests?

sanic/server.py
```diff
@@ -955,14 +955,14 @@ def serve(
 def _build_protocol_kwargs(
     protocol: Type[HttpProtocol], config: Config
 ) -> dict:
-    if hasattr(protocol, "websocket_timeout"):
+    if (dir(protocol).__contains__("websocket_handshake")):
```
ahopkins, line 958: Why are you checking for properties this way? Please avoid using `dir()` here.

sanic/server.py
```diff
@@ -955,14 +955,14 @@ def serve(
...
         return {
-            "max_size": config.WEBSOCKET_MAX_SIZE,
-            "max_queue": config.WEBSOCKET_MAX_QUEUE,
-            "read_limit": config.WEBSOCKET_READ_LIMIT,
-            "write_limit": config.WEBSOCKET_WRITE_LIMIT,
-            "ping_timeout": config.WEBSOCKET_PING_TIMEOUT,
-            "ping_interval": config.WEBSOCKET_PING_INTERVAL,
+            "websocket_max_size": config.WEBSOCKET_MAX_SIZE,
+            "websocket_max_queue": config.WEBSOCKET_MAX_QUEUE,
```
ahopkins, line 961: Why are these being renamed `websocket_`?

ashleysommer:
@brooklet we want to try to get this merged before the Sanic 2020-09 release. Are you able to make those changes suggested?

sanic/server.py
ahopkins: reverted to `hasattr`, but I might circle back on this to see why we are not using `isinstance` here, which seems more logical. Do we need duck typing for some reason I am unaware of?

sanic/server.py
ahopkins: Left this change in, but I updated the tests so that the change was carried through consistently.

ahopkins (approved):

PR #1913, opened 2020-08-22
Update static.py

NuclearVGA:
Auto return "index.html" for dirs.


sjsadowski:
@huge-success/sanic-core-devs we probably need a policy about how and when to close PRs that die on the vine. I'm closing this one due to lack of response from the submitter.

ahopkins:
We do: https://probot.github.io/apps/stale/

PR #1931, opened 2020-09-23
Delete asyncio_python37.rst

nitipit:
deprecate since there's an example at https://sanic.readthedocs.io/en/latest/sanic/examples.html


ashleysommer:
@nitipit do you think we should remove documentation pages if there is a code snippet in the examples? I'm not sure I agree.

nitipit:
@ashleysommer 
I think an example here which is more easier and more accurated to python 3.7 docs about how to use sanic with asyncio event loop.
https://sanic.readthedocs.io/en/latest/sanic/examples.html#logging-enhancements

However, that example point out different topics. I agree that just removing `asyncio_python37.rst` is not a good way to do.
So I think this pull request should be rejected.

ahopkins:
I think this is probably something that ultimately will live in the new Sanic guide.

ahopkins:
This was a mistaken reference. Ignore.

PR #1850, opened 2020-05-14
WIP: Remove test clients

ahopkins:
In order to [remove `httpx`](https://github.com/huge-success/sanic/issues/1826) from Sanic core, I have moved the `sanic.testing` module to its [own package](https://github.com/huge-success/sanic-test).

Currently it is a near identical implementation of the current testing module. I would like to add some further functionality to it, as well as help make some convenience utilities to make the transition seemless for existing users.

PR #1935, opened 2020-09-27
Upgrade httpx

ahopkins:
I had thought pretty long about whether to get the PR for removing `httpx` requirement (#1850), but that didn't seem quite appropriate just yet.

Because `httpx` is a fairly fast moving project, there were a number of changes needed to be made to some tests as well as the test client. Hopefully, we will get `sanic-testing` completed and then we will not need to deal with these changes here.

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -249,10 +249,7 @@ def raw(
     :param content_type: the content type (string) of the response.
     """
     return HTTPResponse(
-        body=body,
-        status=status,
-        headers=headers,
-        content_type=content_type,
+        body=body, status=status, headers=headers, content_type=content_type,
     )
 
 
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -249,7 +249,10 @@ def raw(
     :param content_type: the content type (string) of the response.
     """
     return HTTPResponse(
-        body=body, status=status, headers=headers, content_type=content_type,
+        body=body,
+        status=status,
+        headers=headers,
+        content_type=content_type,
     )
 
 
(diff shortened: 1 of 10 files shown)


ashleysommer (commented):
Looks good to me

ashleysommer (dismissed):

ashleysommer:
ReuseableSanicTestClient and DelayableHTTPConnection are some pretty hairy hacks I put together a couple of years ago in order to test the HTTP Keep-Alive mechanism, and Request Timeouts respectively.
They're a pain to maintain, but I don't know of any other way to test these features (unless httpx gets these features natively).
It might be worth considering pulling these out of the sanic test code and into the Sanic-Testing project, to keep all of that stuff together.

ashleysommer:
Black test failing:
```
would reformat /home/travis/build/huge-success/sanic/sanic/testing.py
would reformat /home/travis/build/huge-success/sanic/sanic/response.py
```

ahopkins:
I agree. I have written and rewritten those several times because of the fast pace at which underlying libraries reorganize themselves. I do have a thought on how to handle, and this is actually THE reason I think we need to hold off on sanic-testing for now. 

While I understand the desire to move them out to keep them closer, I think the test themselves need to stay here because it is Sanic functionality we need to worry about. 

Therefore, I want to add the functionality into the core of sanic-testing so the tests can stay, but the hack clients can go.

ahopkins:
Had to upgrade `black` and `isort` with the particular env I had running. Should be ready when done building.

ashleysommer (approved):

PR #1937, opened 2020-09-29
Add text and json fallback error handlers

ahopkins:
Replaces #1905 

PR #1768 added HTML default exception pages. This PR adds two alternative fallback exception pages: text and JSON.

*This is **NOT** meant to replace custom error handlers, but is meant to be for **fallback***.

This is configurable with:

```python
app.config.FALLBACK_ERROR_FORMAT = "text"
# or
```

The **default** will still be **HTML**. And, the HTML page is untouched (except for some fixes to make valid HTML.

Just like in HTML, there are two handlers: one for debuggable exception responses, the other for minimal responses.

### Text

```
⚠️ 500 — Internal Server Error
==============================
foobar

Exception: foobar while handling path /hello
Traceback of __BASE__ (most recent call last):

  Exception: foobar
    File /path/to/sanic/app.py, line 938, in handle_request
    response = await response

    File server.py, line 12, in foo
    raise Exception("foobar")
```

```
⚠️ 500 — Internal Server Error
==============================
The server encountered an internal error and cannot complete your request.

```

### JSON

```json
{
  "description": "Internal Server Error",
  "status": 500,
  "message": "foobar",
  "path": "/hello",
  "args": {
    "name": [
      "world"
    ]
  },
  "exceptions": [
    {
      "type": "Exception",
      "exception": "foobar",
      "frames": [
        {
          "file": "/path/to/sanic/app.py",
          "line": 938,
          "name": "handle_request",
          "src": "response = await response"
        },
        {
          "file": "server.py",
          "line": 12,
          "name": "foo",
          "src": "raise Exception(\"foobar\")"
        }
      ]
    }
  ]
}
```

```json
{
  "description": "Internal Server Error",
  "status": 500,
  "message": "The server encountered an internal error and cannot complete your request."
}

```

### DIFF
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -32,6 +32,7 @@ DEFAULT_CONFIG = {
     "REAL_IP_HEADER": None,
     "PROXIES_COUNT": None,
     "FORWARDED_FOR_HEADER": "X-Forwarded-For",
+    "FALLBACK_ERROR_FORMAT": "html",
 }
 
 
--- a/sanic/errorpages.py
+++ b/sanic/errorpages.py
@@ -4,10 +4,262 @@ from traceback import extract_tb
 
 from sanic.exceptions import SanicException
 from sanic.helpers import STATUS_CODES
-from sanic.response import html
+from sanic.response import html, json, text
 
 
-# Here, There Be Dragons (custom HTML formatting to follow)
+FALLBACK_TEXT = "The server encountered an internal error and cannot complete your request."
+FALLBACK_STATUS = 500
+
+
+class BaseRenderer:
+    def __init__(self, request, exception, debug):
+        self.request = request
+        self.exception = exception
+        self.debug = debug
+
+    @property
+    def headers(self):
+        if isinstance(self.exception, SanicException):
+            return getattr(self.exception, "headers", {})
+        return {}
+
+    @property
+    def status(self):
+        if isinstance(self.exception, SanicException):
+            return getattr(self.exception, "status_code", FALLBACK_STATUS)
+        return FALLBACK_STATUS
+
+    @property
+    def text(self):
+        if self.debug or isinstance(self.exception, SanicException):
+            return str(self.exception)
+        return FALLBACK_TEXT
+
+    @property
+    def title(self):
+        status_text = STATUS_CODES.get(self.status, b"Error Occurred").decode()
+        return f"{self.status} — {status_text}"
+
+    def render(self):
+        output = (
+            self.full
+            if self.debug and not getattr(self.exception, "quiet", False)
+            else self.minimal
+        )
+        return output()
+
+    def minimal(self):
+        raise NotImplementedError
+
+    def full(self):
+        raise NotImplementedError
+
+
+class HTMLRenderer(BaseRenderer):
+    TRACEBACK_STYLE = """
+        html { font-family: sans-serif }
+        h2 { color: #888; }
+        .tb-wrapper p { margin: 0 }
+        .frame-border { margin: 1rem }
+        .frame-line > * { padding: 0.3rem 0.6rem }
+        .frame-line { margin-bottom: 0.3rem }
+        .frame-code { font-size: 16px; padding-left: 4ch }
+        .tb-wrapper { border: 1px solid #eee }
+        .tb-header { background: #eee; padding: 0.3rem; font-weight: bold }
+        .frame-descriptor { background: #e2eafb; font-size: 14px }
+    """
+    TRACEBACK_WRAPPER_HTML = (
+        "<div class=tb-header>{exc_name}: {exc_value}</div>"
+        "<div class=tb-wrapper>{frame_html}</div>"
+    )
+    TRACEBACK_BORDER = (
+        "<div class=frame-border>"
+        "The above exception was the direct cause of the following exception:"
+        "</div>"
+    )
+    TRACEBACK_LINE_HTML = (
+        "<div class=frame-line>"
+        "<p class=frame-descriptor>"
+        "File {0.filename}, line <i>{0.lineno}</i>, "
+        "in <code><b>{0.name}</b></code>"
+        "<p class=frame-code><code>{0.line}</code></p>"
+        "</p></div>"
+    )
+    OUTPUT_HTML = (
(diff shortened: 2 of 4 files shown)


ahopkins:
I am still working on the tests, but I wanted to get this out there for some feedback ASAP so we can try and include it in upcoming release. It is **not** a breaking change, and a relatively minor change.

ashleysommer (commented):
Looks great!

FWIW, the old HTML responses _are_ actually valid HTML5.
The `<!DOCTYPE html>` hint tag puts the document in HTML5-mode. In HTML5 the `<html>` document wrapper tag is not needed (though most people put it in anyway for html4 backward compatibility). Also the `<head>` tag is _technically_ optional, you can just put your meta, script, title, and style tags before your content. Along the same lines, it doesn't need a `<body>` tag if there's no head before it.
`<p>` tags never need to be closed (some linters even generate an error for a closing `</p>` tag) and finally `</body>` and `</html>` closing tags don't need to be present at the end of the file (they are assumed present if the HTML just ends).

ahopkins:
> Looks great!
> 
> 
> 
> FWIW, the old HTML responses _are_ actually valid HTML5.
> 
> The `<!DOCTYPE html>` hint tag puts the document in HTML5-mode. In HTML5 the `<html>` document wrapper tag is not needed (though most people put it in anyway for html4 backward compatibility). Also the `<head>` tag is _technically_ optional, you can just put your meta, script, title, and style tags before your content. Along the same lines, it doesn't need a `<body>` tag if there's no head before it.
> 
> `<p>` tags never need to be closed (some linters even generate an error for a closing `</p>` tag) and finally `</body>` and `</html>` closing tags don't need to be present at the end of the file (they are assumed present if the HTML just ends).

I put it in a linter and it yelled at me 🤷‍♂️

vltr:
long time no see, but I really like this PR. I was just wondering: what if we read the `Content-Type` header from the request (in case it exists) and provide the error properly? let's say, `text/plain` would return the text error, `application/json` would return the JSON error and anything else not declared would return the default fallback? it's just an initial thought, it's open to ideas :sweat_smile:

ahopkins:
> long time no see, but I really like this PR. I was just wondering: what if we read the `Content-Type` header from the request (in case it exists) and provide the error properly? let's say, `text/plain` would return the text error, `application/json` would return the JSON error and anything else not declared would return the default fallback? it's just an initial thought, it's open to ideas :sweat_smile: 

I like this idea. Maybe this can be the behavior for

```
app.config.FALLBACK_ERROR_FORMAT = "auto"
```

I would suggest that would be a sensible default. But maybe since *that* would be a breaking change, maybe we don't make that default until after LTS.

vltr:
> ```
> app.config.FALLBACK_ERROR_FORMAT = "auto"
> ```

hey, I like that as well! :sunglasses: I don't know if this will add more complexity to the code (after all, "_gotta go fast_"), but this is just in case of an error, which is not what any developer is looking on their app (except during development, sure), so ... but, **even so**, I'm pretty sure that could be passed to a constructor / `partial` to eliminate any unnecessary `if`s and / or `else`s from the code (not that I'm concerned, but I had to bring this up before this could become a concern).

ahopkins:
@vltr, I think the main usage of this is really for development purposes. They **should** be safe enough to not leak in production, but this is not intended to be a production-ready replacement for error handling. Primarily because (as mentioned in the other thread), there are too many considerations for us to be able to account for. With that said, I will keep this in mind.

vltr:
@ahopkins yeap, being just for development I think it's fine ... I just mentioned about performance because, well, never mind, that should not be the case :+1:

ahopkins:
> Looks great!
> 
> FWIW, the old HTML responses _are_ actually valid HTML5.
> The `<!DOCTYPE html>` hint tag puts the document in HTML5-mode. In HTML5 the `<html>` document wrapper tag is not needed (though most people put it in anyway for html4 backward compatibility). Also the `<head>` tag is _technically_ optional, you can just put your meta, script, title, and style tags before your content. Along the same lines, it doesn't need a `<body>` tag if there's no head before it.
> `<p>` tags never need to be closed (some linters even generate an error for a closing `</p>` tag) and finally `</body>` and `</html>` closing tags don't need to be present at the end of the file (they are assumed present if the HTML just ends).

I discovered the issue. Needed a single closing `</div>`.

vltr:
> I discovered the issue. Needed a single closing `</div>`.

"[The missing `</div>`](https://thecodinglove.com/when-i-find-the-missing-div)". Sorry, but I had to ... :rofl:

ashleysommer (dismissed):

ashleysommer:
Ok, I think we can still push this out in this release, with default format as "html", to keep compatibility. But we'll change it to default to "auto" in v21.03.

ashleysommer:
There is still one type-hinting mypy error
```
error: "BaseRenderer" not callable
```

ashleysommer (approved):
Go ahead and merge

Tronic:
If I understood the discussion and the code correctly, only the request `Content-Type` header is used for determining which format to return in `auto` mode. `Content-Type` on request specifies the format of the request itself, and is normally not provided in GET requests which carry no body.

It might be a good idea to use the `Accept` request header if available, and see if it explicitly lists `text/html` (I think all browsers put that at the very beginning). Curl and wget by default send `Accept: */*` in which case a text response is more appropriate. Some JSON APIs only work when `Accept: application/json` is [specified on the request](https://stackoverflow.com/questions/43209924/rest-api-use-the-accept-application-json-http-header), so despite the widespread misuse of `Content-Type` for this, it might be enough to only examine the `Accept` header.

ahopkins:
I thought about that, but as you said, most browsers list a bunch of different options in Accept headers.

Tronic:
I just tested Chrome, Firefox, Safari, Edge, IE11, Android-Chrome, Android-Firefox, iOS-Safari and Opera Mini, which all can be handled by `accept.startswith("text/html")`, really fast with no need for further parsing. The fetch API on all these browsers (except IE11 which doesn't support fetch) uses `*/*` only, as do wget and curl.

The results and a quickly hacked testing page at https://browser.zi.fi/

ahopkins:
Nice. @Tronic can you move this to a new issue for further development?

PR #1939, opened 2020-09-30
Document test client's allow_redirect option

### LINKED ISSUE #1933: Sanic test client follows redirects
**Describe the bug**
When using the test client to test a route that returns a `sanic.response.redirect`, the test client follows the actual redirect and then returns the response instead of simply returning the redirect response. This isn't handy to make assertions about the redirect response. It also generates unnecessary network traffic when the redirect points to an actual domain. It makes tests rely on network status, making them change their behaviour if network is down.

**Code snippet**
```python
from sanic.response import redirect
from sanic import Sanic
from unittest import TestCase
import unittest

bp = Blueprint('test')

@bp.route('/redir/', methods=['GET'])
async def redir(request):
    return redirect('https://example.com/')

app = Sanic('test')
app.register_blueprint(bp)

class TestTest(TestCase):
    def test_test(self):
        request, response = app.test_client.get('/redir/')
        print(response.status)

unittest.main()
# Shows 200
```

**Expected behavior**
The test client should return the actual redirect response, to be able to make assertions about it and avoid unnecessary network interactions when running tests. This means the snippet above should show the redirect status instead of `200`.

 - OS: Pop!_OS 20.04 LTS
 - Sanic Version 20.6.3
 - Python Version 3.8.2

**Additional context**

ohemelaar:
closes #1933

### DIFF
--- a/docs/sanic/testing.rst
+++ b/docs/sanic/testing.rst
@@ -33,6 +33,7 @@ The `test_client` methods accept the following arguments and keyword arguments:
   response is returned.
 - `server_kwargs` *(default `{}`)* a dict of additional arguments to pass into `app.run` before the test request is run.
 - `debug` *(default `False`)* A boolean which determines whether to run the server in debug mode.
+- `allow_redirects` *(default `True`)* A boolean which indicates whether the test client should follow redirections before returning a response (`True`) or return the redirect response as-is (`False`).
 
 The function further takes the `*request_args` and `**request_kwargs`, which are passed directly to the request.
 


ahopkins (changes requested):
I might also suggest adding a comment and a link to `httpx` to mention that any other lwargs accepted there will be passed through.

ahopkins:
ping

alexpirine (approved):

sjsadowski:
Should this have been approved without the requested changes?

ahopkins:
@sjsadowski I think that documentation is needed before merge.

ahopkins:
This is no longer relevant and should be made in https://github.com/sanic-org/sanic-guide

PR #1941, opened 2020-09-30
Improve FileNotFound exception

tomaszdrozdz:
@ahopkins @myusko 

Without this changes my web browser shows only
> File not found

With this changes it shows:
> File not found: path=./some/thing relative_url=/something

### DIFF
--- a/sanic/exceptions.py
+++ b/sanic/exceptions.py
@@ -67,7 +67,9 @@ class URLBuildError(ServerError):
 
 class FileNotFound(NotFound):
     def __init__(self, message, path, relative_url):
-        super().__init__(f"{message: path={path}, relative_url={relative_url}")
+        super().__init__(message)
+        self.path = path
+        self.relative_url = relative_url
 
 
 @add_status_code(408)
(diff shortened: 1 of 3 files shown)


ahopkins:
I think this might be an instance where it is better to leave it. If a message like that is desired, then there should probably be something to catch and reraise. We do not want to add something that could potentially leak information not intended.

tomaszdrozdz:
OK.
How about another idea ?

ahopkins:
```python
@app.exception(FileNotFound)
def handle_missing_files(request, exception):
    message = f"No file found at {exception.path=} looking at {exception.relative_url=}"
    return text(message, status=exception.status_code)
```

tomaszdrozdz:
OK - You are right  - now I understand :)

---------

And now I understand i should have describe my issue more precisselly.

So i wanted to:

    app.static("/css", "./css", "css")

But going to web browser under: /css/main.css showed me only 
> File not found

There was also no any information what file, what path, ... in command line.

(My fault, I provided wrong directory path name).

So I just thought it would help me (and I think in such case it could also help other developers writing Sanic app) if i had more detailed information about what is not found in terminal (command line) where I run my Sanic app in debug mode.

Hope You understand my concern that in this case (and when running Sanic in debug mode) we could be more verbose and give more hint what is wrong.

ahopkins:
Perhaps a `log.error` [here](https://github.com/huge-success/sanic/blob/master/sanic/static.py#L43) would help that.

PR #1944, opened 2020-10-01
Fix #1943

### LINKED ISSUE #1943: Methods of classes cannot be used as blueprint routes
**Describe the bug**
Unable to define routes in classes

**Code snippet**
```python
class MyApp:
    def __init__(self):
        self.bp = Blueprint("app")
        self.bp.add_route(self.route, "/route")

    def route(self, request):
        return text("hi")
```

**Expected behavior**
Accessing /route says hi

**Actual behavior**
Including the blueprint fails with because trying to assign `__blueprintname__` on a method, but methods cannot have arbitrary attributes assigned.

19.12.2, and master

**Additional context**
Introduced in #1445

penn5:
I've tested it, it fixed the bug. I haven't run tests though, so there could be unintended consequences, but I doubt it.

I've intentionally left some references to `__blueprintname__` in case some projects are hacking it, but I've not fully maintained compatibility (it isn't assigned by default any more)

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -924,12 +924,12 @@ class Sanic:
                 else:
                     if not getattr(handler, "__blueprintname__", False):
                         request.endpoint = self._build_endpoint_name(
-                            handler.__name__
+                            name or handler.__name__
                         )
                     else:
                         request.endpoint = self._build_endpoint_name(
                             getattr(handler, "__blueprintname__", ""),
-                            handler.__name__,
+                            name or handler.__name__,
                         )
 
                 # Run response handler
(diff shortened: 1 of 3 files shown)


ashleysommer:
Hi @penn5 
Thanks for the PR.
This looks like a pretty clunky hack to work around a problem that you've created yourself.

Sanic was never designed to use methods as route handlers, and adding this feature will likely open the gates for bugs and breakages in unexpected places.

ahopkins:
I agree. It is a pretty big breaking change. I know of a number of plug-ins that use that api. I like @ashleysommer solution in the issue.

PR #1945, opened 2020-10-02
Making static route more verbose if file not found

tomaszdrozdz:
@ahopkins something like that ?  (contnuation of #1941)


ahopkins (approved):

ahopkins:
Can we get these passing so we can include in the upcoming release?

ahopkins (approved):

PR #1948, opened 2020-10-06
Improving "quiet" login in Sanic exceptions

tomaszdrozdz:
Logic for setting quiet attribute for exception instance may be wrong,
or I do not understand intention behind it (if so please explain it to me).

What I mean by wrong:

!500 and !None

| quiet provided to decorator | quiet provided to init | quiet in exception instance ||
|-|-|-|-|
| True | False | True | But because I explicitly provided False when instantiating exception I expect quiet in this instance to be False |
| False | None | True | But because I explicitly provided False to decorator I expect quiet in this instance  to be False |

I can plot table with more examples like that.

So I propose a more straightforward logic for setting quiet attribute.
Please take a look, comment or If I am wrong explain.

### DIFF
--- a/sanic/exceptions.py
+++ b/sanic/exceptions.py
@@ -11,9 +11,16 @@ def add_status_code(code, quiet=None):
 
     def class_decorator(cls):
         cls.status_code = code
-        if quiet or quiet is None and code != 500:
+
+        if quiet is not None:
+            cls.quiet = quiet
+        elif code == 500:
+            cls.quiet = False
+        else:
             cls.quiet = True
+
         _sanic_exceptions[code] = cls
+
         return cls
 
     return class_decorator
@@ -27,8 +34,14 @@ class SanicException(Exception):
             self.status_code = status_code
 
         # quiet=None/False/True with None meaning choose by status
-        if quiet or quiet is None and status_code not in (None, 500):
+        if quiet is not None:
+            self.quiet = quiet
+        elif hasattr(type(self), "quiet"):
+            pass
+        elif status_code not in (None, 500):
             self.quiet = True
+        else
+            self.quiet = False
 
 
 @add_status_code(404)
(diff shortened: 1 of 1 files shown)


sanic/exceptions.py
```diff
@@ -38,10 +38,10 @@ def __init__(self, message, status_code=None, quiet=None):
             self.quiet = quiet
         elif hasattr(type(self), "quiet"):
             pass
-        elif status_code not in (None, 500):
-            self.quiet = True
-        else
+        elif status_code in (None, 500):
```
ahopkins, line 41: Why this change?

I'd rather see it without the multiple conditionals.

```python
else:
    self.quiet = status_code not in (None, 500)
```

ahopkins (changes requested):

sanic/exceptions.py
tomaszdrozdz: OK.

tomaszdrozdz:
Are there any more obstacles, issues that can be introduced by this change ?
Can we merge it to master ?

sanic/exceptions.py
```diff
@@ -11,9 +11,14 @@ def add_status_code(code, quiet=None):
 
     def class_decorator(cls):
         cls.status_code = code
-        if quiet or quiet is None and code != 500:
-            cls.quiet = True
+
+        if quiet is not None:
+            cls.quiet = quiet
+        else
```
AntonDnepr, line 17: This and line 39 are triggering `SyntaxError: invalid syntax`, you missed `:`

ahopkins:
Breaking changes need to be tabled until after LTS.

ahopkins:
@tomaszdrozdz Are you able to make these changes?

ahopkins:
I am closing this. If someone things there is something we still need here, please reopen a new PR.

PR #1949, opened 2020-10-06
Introspection.

tomaszdrozdz:
I think that kind of "tools" would be nice to have in core functionality.
Now two simple, but maybe some day we can have more.
If this is bad idea to have such tools in core functionality please share with me why.

### DIFF
--- a/sanic/introspection.py
+++ b/sanic/introspection.py
@@ -1,0 +1,13 @@
+async def get_uri(app):
+        return {'uris': app.router.routes_all}
+
+
+
+async def get_middlewares(app):
+    middlewares = {
+        'request': app.request_middleware,
+        'response': app.response_middleware,
+        'named_request': app.named_request_middleware,
+        'named_response': app.named_response_middleware }
+
+    return {'middlewares': middlewares}


sanic/introspection.py
```diff
@@ -0,0 +1,13 @@
+async def get_uri(app):
```
ashleysommer, line 1: Why is this fn named `get_uri` (singular) but returns `'uris'` (plural)?

sanic/introspection.py
```diff
@@ -0,0 +1,13 @@
...
+        return {'uris': app.router.routes_all}
```
ashleysommer, line 2: Why return a dict? Shouldn't it be `return app.router.routes_all`?

sanic/introspection.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+async def get_middlewares(app):
+    middlewares = {
+        'request': app.request_middleware,
+        'response': app.response_middleware,
+        'named_request': app.named_request_middleware,
+        'named_response': app.named_response_middleware }
+
+    return {'middlewares': middlewares}
```
ashleysommer, line 13: Why return a dict? Shouldn't it be `return middlewares`?

ashleysommer:
Hi @tomaszdrozdz 
Thanks for this PR. I've left some in-line comments above.

Overall, I'm struggling to see how this would be useful or helpful.
Why would a user want to do this:
```
from sanic import introspection
uris = introspection.get_uri(app)
```
when you can already simply do:
```
uris = app.router.routes_all
```
The second option is faster to execute, one fewer lines, and easier to read and understand.

Similarly, why would a user want to do:
```
from sanic import introspection
request_middlewares = introspection.get_middlewares(app)['request']
```
when you can already simply do:
```
request_middlewares = app.request_middleware
```
Again, thats faster to execute, fewer lines, and easier to read and understand.

tomaszdrozdz:
Thank You for comments.

Your comments about returning dicts, ... are right.
I Just wanted to get opinion about this idea, and if You like it then implement it, but if You have reasons not to then here why not.
It had to dig a bit in code to know how to get this info so I thought we can have kind of "introspection" tools that will provide informations about Sanic inner state.
I know these are simple examples but we can have more difficult in future possibly ?

Or maybe in this case adding documentation that would explain how to to it would be enough ?

tomaszdrozdz:
And off course for end user it would be as easy as:

    app.get_uris()
   app.get_middlewares()

Just what do You think about this idea, or do You prefer perhaps some documentation section like "introspection" ?

ahopkins:
It sounds more like a documentation issue.

Issue #1942, opened 2020-10-01
documentation bugs ?


tomaszdrozdz:
https://sanic.readthedocs.io/en/latest/sanic/static_files.html

chunk_size = 1&emsp;&emsp;

means 1 Byte ???

If so then

chunk_size = 1024 * 1024 * 8

Shouldn't it be 8MiB ???
Not 8KB, not 8MB but just 8MiB

ashleysommer:
Yes, I think you're right. That is a mistake in the documentation.

tomaszdrozdz:
Also what I found in

    sanic/exceptions.py:

in docstring for

    def abort(status_code, message=None):
        """...
        :param message: The HTTP response body. Defaults to the messages in response.py for the given status code."""

But in code it is:

    from sanic.helpers import STATUS_CODES
    
    if message is None:
        message = STATUS_CODES.get(status_code)

tomaszdrozdz:
And please correct me if I am wrong but:
https://sanic.readthedocs.io/en/latest/sanic/blueprints.html#blueprint-middleware
> Using blueprints allows you to also register middleware globally.

But it does not seams so, and even next paragraph
https://sanic.readthedocs.io/en/latest/sanic/blueprints.html#blueprint-group-middleware
states something different:
> 'applied on Blueprint : bp1 Only'

tomaszdrozdz:
Also we could write tiny paragraph for:
register_named_middleware

tomaszdrozdz:
And also in middlewares documentation in listeners section for listeners execution order:
> The listeners are deconstructed in the reverse order of being constructed.

I think it would be more clear to describe it like that:

"before_server_start", "after_server_start" listeners are executed in order they were registered
but "before_server_stop", "after_server_stop" listeners are executed in reverse order.

ashleysommer:
@tomaszdrozdz 
As you can see there is a clear problem we are facing that code changes over time, but the documentation doesn't.
Eg:
> :param message: The HTTP response body. Defaults to the messages in response.py for the given status code."""
This behavior changed when the standard HTTP response messages were refactored into "sanic.helpers" but the docstring for `abort()` (and possibly others) was not updated. 
Similarly:
>Using blueprints allows you to also register middleware globally.
This paragraph in the documentation was not updated when "named" middleware was introduced, so now blueprint middleware is only executed on the routes in that blueprint.

[...]

PR #1951, opened 2020-10-19
Improving documentation.

tomaszdrozdz:
@ashleysommer @ahopkins 
As promised and discussed in #1942 - tiny documentation improvements.
Deliberately not touching middlewares, listeners because there is another discussion pending around them.

Would it be good idea to keep this branch open so anyone anytime can improve documentation ?


tomaszdrozdz:
And how about "introspection" section (#1951) ?
Should we have a new file for it in docs/sanic/introspection.rst ??

ahopkins (approved):

PR #1957, opened 2020-10-25
Ignore writing headers when in ASGI mode

### LINKED ISSUE #1653: In ASGI, headers is output as body

### LINKED ISSUE #1730: ASGI Streaming response contains extra content

ahopkins:
This bug is fixed in #1876. In the interim, this resolves the issue and should be applied to both 19.12 and 20.9.

This resolves #1730 and also resolves #1653. 

This branch can be tagged and pushed as `v20.9.1`. For 19.12, I will submit a seperate PR to the LTS branch.


Tronic (approved):
A very cursory review: looks good. Note: this does not solve the double response header problem in some cases where exceptions are thrown after a response has already started, for which there is a separate issue open.

ahopkins:
> for which there is a separate issue open.

@Tronic Which issue is that? Trying to replicate the issue.

ahopkins:
After approvals and passing tests, I will tag and push. After that I will "update branch", which will invalidate the approvals, so they will need to be submitted again.

ashleysommer:
Just curious, why did Multidict need to be updated in order to pass tests? I saw the same change in #1956.

ahopkins:
It has to do with the new change to PIP that will not let you install conflicting versions of packages. Something else was using a higher pinned version so PIP now fails.

ahopkins:
Luckily upgrading to v5 had no impacts and tests all ran fine.

ahopkins:
FYI - the appveyor tests will start passing again once I pull master back in here before merge to master.

ashleysommer (approved):

ashleysommer:
Go ahead and merge master, double check tests, then merge and release 20.9.1
Looks like the only other change from 20.9.1 is https://github.com/huge-success/sanic/pull/1954 that is fine to include in the 20.9.1 release.

ahopkins:
OK.... pulling master in.

Tronic:
> > for which there is a separate issue open.
> 
> @Tronic Which issue is that? Trying to replicate the issue.

This is something that I came across during the streaming development, and frankly I cannot quite remember for sure if there was a bug in mainline or only in an earlier version of the streaming branch. Since I cannot find such issue now (open or closed), I'm inclined to think it was the latter.

ahopkins:
:+1: I tried testing it pretty hard, and with the two additional tests, I think the coverage is pretty decent there. Soon enough we will be able to merge `streaming`.

tests/test_response.py
```diff
@@ -235,13 +235,29 @@ def test_chunked_streaming_returns_correct_content(streaming_app):
     assert response.text == "foo,bar"
 
 
+@pytest.mark.asyncio
+async def test_chunked_streaming_returns_correct_content_asgi(streaming_app):
+    request, response = await streaming_app.asgi_client.get("/")
+    assert response.text == "4\r\nfoo,\r\n3\r\nbar\r\n0\r\n\r\n"
```
ashleysommer, line 241: So turns out this is wrong.
The response.text should be "foo,bar" just like in "test_chunked_streaming_returns_correct_content" non-asgi test.
The bug #1964 is causing the extra lengths and `\r\n` to be added by Sanic _and_ by the ASGI handler.

ashleysommer: Actually, in the test this is correct because the `asgi.py` MockTransport does not differentiate between chunked or unchunked encoding, so we need this in out response.text to verify that chunked mode was used correctly by the sanic-side stream fn.

ashleysommer: Continuing conversation in #1964

PR #1969, opened 2020-11-16
remove upper bound for multidict

all2ham:
Is there a reason this had a hard pin? It would be beneficial to set this range as wide as possible to avoid conflicts with Sanic users.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,13 +80,13 @@ requirements = [
     ujson,
     "aiofiles>=0.6.0",
     "websockets>=8.1,<9.0",
-    "multidict==5.0.0",
+    "multidict>=5.0,<6.0",
     "httpx==0.15.4",
 ]
 
 tests_require = [
     "pytest==5.2.1",
-    "multidict==5.0.0",
+    "multidict>=5.0,<6.0",
     "gunicorn",
     "pytest-cov",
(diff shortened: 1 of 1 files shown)


ahopkins:
As long as the tests pass, I have no problem with this.

ahopkins (approved):

ashleysommer:
I know in the past we've had to pin Multidict to specific versions because major version changes brought in API-level incompatibilities (specifically in CIMultiDict, that we use for Headers).
Eg, Sanic 19.3.x, 19.6.x, and 19.9.x each had `multidict>=4.0,<5.0` to avoid the incompatibilities introduced in 5.0.

For Sanic 19.12.x we adapted to the Multidict 5.0 changes, and updated the dependency, but pinned it to `==5.0.0` to avoid further changes.
Thankfully it looks like Multidict is stabilised enough now that changing the requirement to `>=5.0.0` seems not too scary. Maybe we could make it `>=5.0,<6.0`, just in case?

ashleysommer (approved):

ashleysommer:
@ahopkins 
There is a potential argument for backporting this change to 19.12LTS: https://github.com/huge-success/sanic/issues/1977

PR #1970, opened 2020-11-18
Add py.typed file

tbekolay:
I was happy to see that Sanic has some type annotations in the codebase! But `mypy` couldn't find those annotations when I ran it on my project. This is because Sanic hasn't added a `py.typed` file; see [PEP-561](https://www.python.org/dev/peps/pep-0561/#packaging-type-information) for details. This PR adds the file and exposes it through `package_data`.

### DIFF
--- a/changelogs/1970.misc.rst
+++ b/changelogs/1970.misc.rst
@@ -1,0 +1,1 @@
+Adds py.typed file to expose type information to other packages.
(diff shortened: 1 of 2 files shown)


ahopkins (approved):
Awesome. Thanks, this has been on my list.

PR #1972, opened 2020-11-22
Name endpoints at startup

ahopkins:
I was messing around with `Sanic.handle_request` and i realized there was this fairly simple optimization to be made. There is no reason that we calculate the `request.endpoint` at runtime. This can be at startup and cached on the `Route` object.

After looking closer at _how_ this is working, it is a prime example for further refactoring in the future because I do not think that it is doing a good job in handling names of nested objects (app, blueprints, CBV, etc). I have some different methods on how this can be achieved, but I want to hold off on adding it until after the first of the year.

In total, this is a fairly minor change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,8 +26,8 @@ Sanic | Build fast. Run fast.
    :target: https://gitter.im/sanic-python/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
 .. |Codecov| image:: https://codecov.io/gh/huge-success/sanic/branch/master/graph/badge.svg
     :target: https://codecov.io/gh/huge-success/sanic
-.. |Build Status| image:: https://travis-ci.org/huge-success/sanic.svg?branch=master
-   :target: https://travis-ci.org/huge-success/sanic
+.. |Build Status| image:: https://travis-ci.com/huge-success/sanic.svg?branch=master
+   :target: https://travis-ci.com/huge-success/sanic
 .. |AppVeyor Build Status| image:: https://ci.appveyor.com/api/projects/status/d8pt3ids0ynexi8c/branch/master?svg=true
    :target: https://ci.appveyor.com/project/huge-success/sanic
 .. |Documentation| image:: https://readthedocs.org/projects/sanic/badge/?version=latest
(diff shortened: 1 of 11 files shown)


ahopkins:
Here are some benchmarks before/after. Take them as you will. I do not claim them to be scientifically accurate representation of this change.


```
Running 30s test @ http://127.0.0.1:3000
  12 threads and 400 connections
    Latency     4.40ms    5.02ms 100.84ms   89.17%
    Req/Sec     9.94k     2.08k   25.15k    73.10%
  3561967 requests in 30.10s, 400.84MB read
Requests/sec: 118357.37
Transfer/sec:     13.32MB

Running 30s test @ http://127.0.0.1:3000
  12 threads and 400 connections
    Latency     4.25ms    4.93ms 117.93ms   90.01%
    Req/Sec    10.16k     2.60k   26.17k    74.75%
  3640517 requests in 30.05s, 409.68MB read
Requests/sec: 121137.64
Transfer/sec:     13.63MB
```

vltr (approved):
awesome! good catch on these changes :sunglasses:

LordOdin:
Just out of curiosity what is your method of benchmarking this?

> Here are some benchmarks before/after. Take them as you will. I do not claim them to be scientifically accurate representation of this change.
> 
> ```
> Running 30s test @ http://127.0.0.1:3000
>   12 threads and 400 connections
>   Thread Stats   Avg      Stdev     Max   +/- Stdev
>     Latency     4.40ms    5.02ms 100.84ms   89.17%
>     Req/Sec     9.94k     2.08k   25.15k    73.10%
>   3561967 requests in 30.10s, 400.84MB read
> Requests/sec: 118357.37
> Transfer/sec:     13.32MB
> 
> Running 30s test @ http://127.0.0.1:3000
>   12 threads and 400 connections
>   Thread Stats   Avg      Stdev     Max   +/- Stdev
>     Latency     4.25ms    4.93ms 117.93ms   90.01%
>     Req/Sec    10.16k     2.60k   26.17k    74.75%
>   3640517 requests in 30.05s, 409.68MB read
> Requests/sec: 121137.64
> Transfer/sec:     13.63MB
> ```

ahopkins:
Tough to say exactly what the endpoint I was using looked like. I simply do not remember. But I usually use something like this:

```
wrk -t12 -c400 -d30s http://127.0.0.1:8000
```

Issue #1968, opened 2020-11-05
Sanic Server instance 


listaction:
Does Sanic have any singleton type patterns which have a map of all the created servers and can it return the one that was created ?

basically:
`Sanic.getServer('some_name')`

ahopkins:
Would be interested to see a PR of something you have in mind.

PR #1979, opened 2020-12-10
App registry

ahopkins:
See #1968 

The purpose of this is to add a convenience for access an application instance.

```python
from sanic import Sanic

app = Sanic.get_app("my_awesome_server")
```

### DIFF
--- a/docs/sanic/getting_started.rst
+++ b/docs/sanic/getting_started.rst
@@ -60,3 +60,21 @@ Open the address `http://0.0.0.0:8000 <http://0.0.0.0:8000>`_ in your web browse
 the message *Hello world!*.
 
 You now have a working Sanic server!
+
+5. Application registry
+-----------------------
+
+When you instantiate a Sanic instance, that can be retrieved at a later time from the Sanic app registry. This can be useful, for example, if you need to access your Sanic instance from a location where it is not otherwise accessible.
+
+.. code-block:: python
+
+    # ./path/to/server.py
+    from sanic import Sanic
+
+    app = Sanic("my_awesome_server")
+
+    # ./path/to/somewhere_else.py
+    from sanic import Sanic
+
+    app = Sanic.get_app("my_awesome_server")
+
--- a/docs/sanic/getting_started.rst
+++ b/docs/sanic/getting_started.rst
@@ -78,3 +78,8 @@ When you instantiate a Sanic instance, that can be retrieved at a later time fro
 
     app = Sanic.get_app("my_awesome_server")
 
+If you call ``Sanic.get_app("non-existing")`` on an app that does not exist, it will raise ``SanicException`` by default. You can, instead, force the method to return a new instance of ``Sanic`` with that name:
+
+.. code-block:: python
+
+    app = Sanic.get_app("my_awesome_server", force_create=True)
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -2,7 +2,6 @@ import logging
 import logging.config
 import os
 import re
-import warnings
 
 from asyncio import CancelledError, Protocol, ensure_future, get_event_loop
 from collections import defaultdict, deque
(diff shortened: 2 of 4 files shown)


sanic/app.py
```diff
@@ -1453,9 +1452,34 @@ async def __call__(self, scope, receive, send):
...
         Please refer to config.py::Config.update_config for documentation."""
 
         self.config.update_config(config)
+
+    # -------------------------------------------------------------------- #
+    # Class methods
+    # -------------------------------------------------------------------- #
+
+    @classmethod
+    def register_app(cls, app: "Sanic") -> None:
```
vltr, line 1468: These methods would be great candidates to the new [`final`](https://docs.python.org/3/library/typing.html#typing.final) typing hint, but that's only available for Python 3.8+ ... I think we should, at least, provide some kind of warning in the docs to avoid overriding these methods - unless you're **absolute sure** of what you're doing. This way we won't need to even _enforce_ this being `final` (I don't like enforcing things in Python) ...

docs/sanic/getting_started.rst
```diff
@@ -60,3 +60,21 @@ Open the address `http://0.0.0.0:8000 <http://0.0.0.0:8000>`_ in your web browse
 the message *Hello world!*.
 
 You now have a working Sanic server!
+
+5. Application registry
```
vltr, line 64: Should we not at least mention the workflow needed for this to work properly? I mean, you should never run `get_app` prior to `Sanic("...")` ... Or, perhaps, make _only_ the `Sanic("...")` call, where if there's no instance called `foo`, it will create one, otherwise return that instance ? This would avoid the workflow problem I mentioned ...

vltr (commented):
I left some comments ... They're just ideas - it's not my intention to start a big discussion over it.

sanic/app.py
ahopkins: There's a long list if we want to go down that route. I think this is not the place for that.

docs/sanic/getting_started.rst
ahopkins: > When you instantiate a Sanic instance, that can be retrieved at a later time ...

We do. I was playing with the idea of allowing `get_app()` to create and return an instance. But, I like the explicit nature because it would avoid a potential bug. What if I created `Sanic("my_app")`, but accidentally ran `Sanic.get_app("myapp")`? I think allowing this to return a new instance would lead to hard to find bugs. Getting an explicit `SanicException` is pretty clear.

I then thought maybe we do this:

```python
try:
    return cls._app_registry[name]
except KeyError:
    if len(cls._app_registry) == 0:
        return cls(name)
    raise SanicException(f'Sanic app name "{name}" not found.')
```

It would be a nice fallback. Perhaps even throw in a warning. This one I am more open to, but still I think it suffers a problem because now the method does not always behave the same way. What about a flag (`strict=True`)? By the time we complicate this so much, it still doesn't get around the fact that this is a convenience, and should not be abused, or mistakenly implemented.

---

In the end, I think a simpler logic with exception leads to more explicit implementation with fewer pitfalls for the dev.

sjsadowski: @vltr any thoughts on the response to this? I'm okay with @ahopkins proposal for both adding the change in behavior along with the flag to manage it.

sanic/app.py
sjsadowski: Let's circle back to the typehint later, I think this is fine for now.

docs/sanic/getting_started.rst
ahopkins: @sjsadowski I had a conversation with @vltr offline. We decided that perhaps the clearest thing would be to raise an exception when it does not exist, but allow the method to force a new instance.

```python
    assert isinstance(
        Sanic.get_app("does-not-exist", force_create=True), Sanic
    )
```

vltr (approved):
All good!

PR #1981, opened 2020-12-14
Cleanup and remove some deprecated code

ahopkins:
Mainly housekeeping. Stuff that has been deprecated for a long time is being removed.


vltr (approved):
This looks great. It's also great to see some things I added in the past as some kind of workaround being finally removed because there's a proper solution in place :wink:

PR #1991, opened 2020-12-30
Fix error handling

ddc67cd:
According to docs, error handler should be a coroutine. Its used correctly in app.py, but not there.


ahopkins:
Looks like same issue as #1785. If we are not going to backport this to the LTS, we can close this since it is handled in #1876.

Thanks for the PR.

ahopkins:
This is not needed as it is already in `master`.

Issue #1989, opened 2020-12-29
20.12.0 causes the parallel unittest fails with 'app ... already in use' 


atezerjani:
After updating to the latest sanic 20.12.0, my pytest tests are failing with following error: 
`sanic.exceptions.SanicException: Sanic app name 'xxxxxx' already in use.`

I reverted back to 20.9.0, and the issue was fixed. What has changed in the recent version that has caused this issue to happen?

ahopkins:
The simplest way to fix this would be to give each of the apps in your test a unique name. Newest build adds a feature where the instance is stored in an internal class registry for later retrieval:

```
Sanic.get_app("my_app")
```

We should probably add a patch that will disable this feature. It should be available both as a config:

```
SANIC_REGISTER=false
```

and on the instance itself

```
Sanic(..., register=False)
```

atezerjani:
Thanks a lot @ahopkins for your prompt response.

ahopkins:
https://github.com/huge-success/sanic/pull/1993/files

ahopkins:
v20.12.1 will have this feature.

CheneyZeng:
Let let us know if this issue fixed! Thanks, team!

ahopkins:
Yes.

In your `conftest.py`:

```python
Sanic.test_mode = True
```

Or, set an env variable in your test suite: `SANIC_REGISTER=False`

I will be releasing [`sanic-test`](https://github.com/sanic-org/sanic-test) incoming days that will handle this automatically.

riderx:
Thanks @ahopkins worked for my issue

ashleysommer:
@ahopkins When do you expect `sanic-test` to be released? I have a project using Sanic 20.12 that I need to write some tests for, and ideally I'd like to use sanic-test to do that.

ahopkins:
I was planning on doing that tonight, but didn't quite get to it. I am shooting for an initial release tomorrow, and getting #1850 ready to merge to `master`.

ahopkins:
released. Changes to come, but it's out there.

PR #1993, opened 2021-01-05
#1989 Add disable app registry

ahopkins:
- Adds a new config value `REGISTER` that disables registering an app.
- Adds `Sanic(..., register=False)` to disable registering a single app instance

See #1989


vltr (approved):
Makes sense to me - for legacy purposes, mostly.

PR #1994, opened 2021-01-05
#1993 Disable registry

ahopkins:
Bring #1993 to v20.12


vltr (approved):
Same comment as #1993.

PR #1903, opened 2020-08-04
Update config

tomaszdrozdz:

PR #1999, opened 2021-01-10
Remove config deprecations and resolve compat

ahopkins:
Removes the deprecated config methods from #1903, and resolves some compatibility with the old API.


ahopkins:
Lots more unit testing needed here.

ahopkins:
See #2007

PR #2001, opened 2021-01-11
Raise ValueError when cookie max-age is not an integer

### LINKED ISSUE #1810: Hard error on invalid max-age cookie
**Describe the bug**
Currently when setting the `max-age` cookie value, it's possible for a valid value to not be set as expected, as well as an invalid value from raising a hard error. In both cases the values are replaced by a `0` `max-age`.

**Code snippet**
```python
response.cookie["my-cookie"]["max-age"] = 10.0  # max-age is set to 0
response.cookie["my-cookie"]["max-age"] = 10.5  # max-age is set to 0
response.cookie["my-cookie"]["max-age"] = "ten"  # max-age is set to 0
response.cookie["my-cookie"]["max-age"] = "10"  # max-age is set to 10
response.cookie["my-cookie"]["max-age"] = 10  # max-age is set to 10
```

**Expected behavior**
Here's what I think the expected behaviour should be (akin to how the `expires` cookie attribute is handled; raising an error if not a `datetime.datetime`).

```python
response.cookie["my-cookie"]["max-age"] = 10.0  # max-age is set to 10
response.cookie["my-cookie"]["max-age"] = 10.5  # raise ValueError
response.cookie["my-cookie"]["max-age"] = "ten"  # raise ValueError
response.cookie["my-cookie"]["max-age"] = "10"  # max-age is set to 10
response.cookie["my-cookie"]["max-age"] = 10  # max-age is set to 10
```

 - OS: macOS
 - Version 19.12.2

**Additional context**
I've created a pull request for this here #1809. Here's the issue relating to the original implementation #1452.

Creating this issue so I can have an issue number for the changelog.

ahopkins:
This fixes #1810. It is slightly different than the proposed solution which stated that a whole number float should be valid.

This change will now raise a ValueError whenever a non-integer is passed as a max-age. It still accepts `"10"`.

### DIFF
--- a/examples/run_asgi.py
+++ b/examples/run_asgi.py
@@ -19,21 +19,34 @@ def handler_text(request):
     return response.text("Hello")
 
 
+@app.route("/cookie")
+def handler_cookie(request):
+    resp = response.text("Hello")
+
+    resp.cookies["my-cookie"] = "foo"
+    resp.cookies["my-cookie"]["max-age"] = 10.0  # max-age is set to 10
+    # resp.cookies["my-cookie"]["max-age"] = 10.5  # raise ValueError
+    # resp.cookies["my-cookie"]["max-age"] = "ten"  # raise ValueError
+    # resp.cookies["my-cookie"]["max-age"] = "10"  # max-age is set to 10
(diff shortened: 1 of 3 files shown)


ashleysommer (approved):
Looks good to me

PR #2003, opened 2021-01-11
Replace ujson with orjson

ahopkins:
There are lots of benchmarks showing `orjson` to be faster than `usjon`. And, now is the time to make breaking changes if we will this year. This is a fairly trivial test just to see what it would look like. On large responses, I actually saw a HUGE degradation in performance. So, perhaps we need to look closer to see if this is even worthwhile in Sanic.

See #1479

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -10,7 +10,7 @@ from sanic.helpers import has_message_body, remove_entity_headers
 
 
 try:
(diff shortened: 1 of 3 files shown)


ahopkins:
I am closing this. Maybe for another day if/when someone can convince me that `orjson` is actually faster.

PR #2005, opened 2021-01-18
Add request.id

ahopkins:
This adds a convenience property on the `Request` instance for tracking: `request.id`. The property will default to the value of the `X-REQUEST-ID` header if it is set. If not, it will fallback to a UUID.

There is a class method that will allow you to override the default (`Request.generate_id`).


ashleysommer:
Is this in response to a Feature request? I don't really see the benefit of doing this, but since the UUID is lazy and only generated on the first access to `request.id`, its probably not a problem to merge in.

In the past, when I need to have a unique ID for a request, I use the built-in python identity fn `id(request)`.

ahopkins:
I had seen something in a somewhat older issue which triggered the idea. And, I had seen some related questions recently on SO. I thought about using `id` or some variant on `itertools.count`, but I think using a UUID is a well understood pattern.

As for the lazy, that was 100% the point. I did not want this to start throwing IDs where they are not needed. It feels like a somewhat nice compliment to `request.token`, which also could be made cacheable like `request.id`.

ashleysommer (approved):

alextsil:
I've been looking for a way to scope my sqlalchemy sessions to a per-request basis ([docs here](https://docs.sqlalchemy.org/en/13/orm/contextual.html?highlight=scopefunc#sqlalchemy.orm.scoping.scoped_session.params.scopefunc)).
I've used request.__hash__() but i'm not sure it's safe. 
This new ID property solves the issue neatly 👍

PR #2007, opened 2021-01-18
21.3 deprecations

ahopkins:
Removes remaining items that were deprecated and slated for removal in 21.3.

This branch had originally branched off `streaming`, which was squash merged. That is why there are a ton of commits here. We can ignore them and squash merge this when ready.


ashleysommer (approved):
Great!

PR #2006, opened 2021-01-18
API Docs & Annotations

ahopkins:
Alongside the release of [the new user guide](https://github.com/sanic-org/sanic-guide), this is a removal of all the documentation we no longer need. It also goes through and adds missing type annotations and docstrings.

Requires #2010 to be merged.

Freshen up the look just a little with logo and removing the blue.

image

PR #2008, opened 2021-01-19
Allow Pathlib Path objects to be passed to `app.static()` helper 

ashleysommer:
This change allows Pathlib Path objects to be passed to the `app.static()` file endpoint register helper fn.

This was a feature request from a colleague. He intuitively passed in a `pathlib.Path` object to the `app.static()` helper, and was surprised when it didn't work as expected.

This also might be a good time to talk about how difficult it can be sometimes for new users to troubleshoot a problem in Sanic.

This is the error my colleague was seeing:
```
  File "/home/code/sanic/app.py", line 1002, in handle_request
    response = await response
  File "/home/code/sanic/static.py", line 44, in _static_request_handler
    file_path = path.abspath(unquote(file_path))
  File "/usr/lib/python3.8/urllib/parse.py", line 634, in unquote
    if '%' not in string:
TypeError: argument of type 'PosixPath' is not iterable
```

This was emitted on the static file response handler, at runtime. It is caused by the use of `unquote()` on file_path. The `unquote()` fn is not pathlib aware, so when `file_path` is a `pathlib.Path` object, we get this error. My colleague does not have a lot of experience with Python, so it was not obvious to him that it was caused by his use of pathlib.Path when registering the Static endpoint.


ahopkins:
This is a good idea, and makes total sense.

As to your second point, that is 100% something I want to spend this whole year focusing on. By the time we get to the next LTS, I think we should feel comfortable dropping the `beta` classifier. Also, with the idea of #2006, I am trying to comb over the entire codebase jotting notes for places we should add some more type validation (like what you are pointing out), reduce DRY (would love to get `Sanic` and `Blueprint` inheriting from a common base), and other improvements. I will post a list when I am done.

ahopkins:
Also... as displayed by the all passing tests except for linting, we should: (1) cleanup `Makefile` and (2) get some required precommit hooks.

ashleysommer:
Merged in new commits from master, and fixed the isort lint error

ahopkins (approved):
Looks good. Do we want to throw an exception in there if after checking to stringify and we still do not have a good value?

ashleysommer:
Yeah, I thought about that. Its probably a good idea. Its better to throw the error in the register helper, than to get a weird unexpected error at runtime.

ahopkins:
Whenever possible, we should always favor extra checks, conditionals, etc at startup over runtime.

ashleysommer:
I added a new feature to raise a ValueError if the static file path is not an acceptable type.
Also added more tests to improve coverage on the new type checks.

ahopkins (approved):

PR #2009, opened 2021-01-19
Remove test client

ahopkins:
This is a more recent implementation of #1850. This is to take out the testing module (and httpx dependency) in favor of `sanic-testing`.

Still some broken tests to fix.


tests/test_keep_alive_timeout.py
```diff
@@ -8,10 +8,11 @@
 import httpx
```
ashleysommer, line 8: Can we remove httpx from here?

ashleysommer (commented):
Looking great so far.

ahopkins:
Yes. There are two tests that are failing and I'm not really sure why. The clients are nearly identical. It seems when we do a websocket call on the test_client, the server doesn't completely shutdown. I'm trying to see if it is a problem with the test, the client, or the websockets implementation. 🤔

I created a third client in my local reo that opens the server in a subprocess instead of using listeners. I can get the test passing with that. 

I hoe to have this

tests/test_keep_alive_timeout.py
ahopkins: Maybe at some point, but I don't think we are there yet. We really need a new way to test timeouts. I think I have an idea, but there are higher priorities that are on my list to accomplish before the next release.

With that said, if anyone wants to take on these tests, I would be happy to help.

vltr: I don't see a problem in using third-party requirements on our testing, since they are actually required in our own testing package, unless we remove `httpx` from `sanic_testing` in the future.

ahopkins: I meant that I really don't like the tests themselves that require us to hack httpx here. I have no intention of removing httpx from the testing client because the ASGI reach in is awesome.

ashleysommer (approved):
I'm happy for this to be merged.

sanic/app.py
```diff
@@ -1032,11 +1033,21 @@ async def handle_request(self, request):
 
     @property
     def test_client(self):
-        return SanicTestClient(self)
+        if self._test_client:
+            return self._test_client
+        from sanic_testing.testing import SanicTestClient  # type: ignore
```
ashleysommer, line 1038:
```suggestion
        elif hasattr(self, "_test_manager"):
            return self._test_manager.test_client  # type: ignore
        from sanic_testing.testing import SanicTestClient  # type: ignore
```

sanic/app.py
```diff
@@ -1032,11 +1033,21 @@ async def handle_request(self, request):
...
+
+        self._test_client = SanicTestClient(self)
+        return self._test_client
 
     @property
     def asgi_client(self):
-        return SanicASGITestClient(self)
+        if self._asgi_client:
+            return self._asgi_client
+        from sanic_testing.testing import SanicASGITestClient  # type: ignore
```
ashleysommer, line 1047:
```suggestion
        elif hasattr(self, "_test_manager"):
            return self._test_manager.asgi_client  # type: ignore
        from sanic_testing.testing import SanicASGITestClient  # type: ignore
```

ashleysommer (commented):
Should add have these too?

sanic/app.py
ahopkins: Let's add that in another PR. I want to get this one merged.

Issue #1671, opened 2019-09-10
Content-Type


Lagicrus:
**Origin Story**
I was testing my site via an online tester and it brought up the fact that my site did not specify the charset via the `Content-Type` header. Luckily I discovered after a quick bit of research that changing the `content_type` of the static file was added back in https://github.com/huge-success/sanic/pull/1267
As such I have currently added `content_type="text/html; charset=utf-8")` to my `app.static`'s so that it fixes that issue.
However, having to constantly keep adding that to all my static routes when I add more is not the best thing in the world.

**Solution**
Change the default content_type of `app.static` from `text/html` to `text/html; charset=utf-8`.
Which I believe is altering https://github.com/huge-success/sanic/blob/master/sanic/static.py#L111

Tronic:
I just made a commit regarding this yesterday, changing the defaults of HTTPResponse and StreamingHTTPResponse to include charset. Looks like I missed one!

Another concern is that unknown types shouldn't be treated as `text/plain` but they should have either
- no content-type header [RFC-7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) -- making browsers sniff by content but then they get the charset wrong (ISO-8859-1 unless the format specifies otherwise e.g. by HTML meta charset element)
- `application/octet-stream`, making browsers download as file.

Nginx does the latter for files without extension or otherwise unrecognised, and it appears to work better in practice. Thus, HTTPResponse, StreamingHTTPResponse and static.py line 111 should probably default to `application/octet-stream`.

Tronic:
Another consideration. Since the `content_type` argument is part of an easy API, and because the charset selection is Sanic internals, it might make sense to auto-append the charset whenever a text-like type is found:

```python
if 'content-type' not in headers and content_type:
    if 'charset=' not in content_type and (
        content_type.startswith('text/') or content_type == 'application/javascript'
    ):
        content_type += '; charset=utf-8'
    headers['content-type'] = content_type
```

What do you think, is this reasonable? Note that if specified via headers instead of the content_type argument, no such magic is applied.

Lagicrus:
> I just made a commit regarding this yesterday, changing the defaults of HTTPResponse and StreamingHTTPResponse to include charset. Looks like I missed one!
> 
> Another concern is that unknown types shouldn't be treated as `text/plain` or `text/html` but they should have either
> 
[...]

PR #2010, opened 2021-01-27
Sanic routing

ahopkins:
Implementation of [sanic-routing](https://github.com/sanic-org/sanic-router). Currently, it is written as a dependency. Whether it stays that way or is merged into this repo is up for discussion at a later date.

Besides overhauling the Router to use the compiled version, this also hits the following:

- Resolve bug with middleware in Blueprint Group when passed callable
- Moves common logic between `Blueprint` and `Sanic` into mixins
- Will ~likely~ touch on some consistency with naming of routes, ~I am divided on this but if we are going to break it, then we should do it now~
- Some new convenience decorators:
    - `@app.before_server_start`
    - `@app.after_server_start`
    - `@app.before_server_stop`
    - `@app.after_server_stop`
    - `@app.on_request`
    - `@app.on_response`
- fixes `Allow` header that did not include `HEAD`
- using "name" keyword in url_for for a "static" route where name does not exist
- Cannot have multiple app.static() without using the named param
- using "filename" keyword in url_for on a file route
- request endpoint is the route name, should mark that as deprecated
- route names are fully namespaced
- no handler type name prefixes (static, websocket, etc)
- unquote in route def (not automatic)
- routes_all is tuples
- handler arguments are kwarg only
- `request.match_info` is now a cached (and not computed) property
- unknown static file mimetype is sent as `application/octet-stream`
- `_host` keyword in `url_for`
- add charset default to utf-8 for text and js content types if not specified
- versionfor a route can be `str`, `float`, or `int`
- route has `ctx` property
- app has `routes_static`, `routes_dynamic`, `routes_regex`

[See discussion](https://community.sanicframework.org/t/a-fast-new-router/649)


ahopkins:

https://community.sanicframework.org/t/a-fast-new-router/649/38

ashleysommer:
Looking great!

ahopkins:
Requires: https://github.com/sanic-org/sanic-testing/pull/13

ahopkins:
Relies upon:
- https://pypi.org/project/sanic-routing/0.1.0/
- https://pypi.org/project/sanic-testing/0.3.0/

ahopkins:
Passing tests except for Python 3.6 with uvloop. See #2024

ahopkins:
Fixes #1671

ashleysommer (approved):
Looking great

PR #2016, opened 2021-02-01
Add samesite documentation to cookies.rst

ruckc:
Document samesite in cookies.


ahopkins:
These docs are being deprecated. RTD will only be sphinx auto doc (see #2006).

This PR should be make on this repo instead: https://github.com/sanic-org/sanic-guide

PR #2020, opened 2021-02-05
Remove prefix from websocket handler name

laggardkernel:
Update: opened for wrong branch, reopened another pr.

---

Remove the websocket prefix `websocket_handler_` introduced in
761eef7. It makes `url_for()` unable to find the corresponding ws route.

```python
@app.websocket('/ws')
async def websocket(request, ws):
    return ""
```

The key for the above handler used in `Router.route_names` is `websocket_handler_websocket`, not `websocket`. Removing the `websocket_handler_` prefix fixes this bug. Searching in the source code, this very prefix is not used in anywhere.

I guess the contributor of commit 761eef7 tried to mimic the registration behavior of static file handler, which prepend the handler name with prefix `_static_`, in order to store static file handler separately in `Router.routes_static_files`, not in `Router.routes_names`.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -524,9 +524,7 @@ class Sanic:
             websocket_handler = partial(
                 self._websocket_handler, handler, subprotocols=subprotocols
             )
-            websocket_handler.__name__ = (
-                "websocket_handler_" + handler.__name__
-            )
+            websocket_handler.__name__ = handler.__name__
             websocket_handler.is_websocket = True
             routes.extend(
                 self.router.add(


ahopkins:
Thanks, this is awesome and I appreciate it. But there's another PR that is a complete overhaul of the Router and the api for url_for is also getting an update. This is included.

ahopkins:
Maybe you can change this as a PR to the 20.12 LTS branch and we can include it backwards there.

laggardkernel:
During reading the source code. I've found inconsistent behaviors among route storing of normal http handler and websocket, static file handlers.

Firstly, the ways used to indicate a websocket handler and static file handler are different.

`_static_` is prepend into `static_file_handler.__name__` in `Sanic.static()`, and the flag is only used in `Route._add()`. `websocket.is_websocket` is set in `Sanic.websocket()`, and the flag is only used in `Sanic.handle_request()`.

Secondly, the ways to pass blueprint names of http handler, websocket handler, static file handler, to be used in `Router._add()` are different.

Formerly, only `handler.__blueprintname__` is used. The bp name is prepend with handler name as the route name (key used in `Router.routes_name`, `Router.routes_static_files`) in `Router._add()`.

For blueprint websocket handler (`Blueprint.websocket()`), not the original handler but a handler wrapper is passed to `Router._add()`. (Check `Sanic._websocket_handler()` and `static._static_request_handler()`) Seems later contributors (938c49b) didn't copy the `__blueprintname__` attr to the handler wrapper. They chose to pass `bp_name + handler` to `Router._add()` directly. Similar thing happens on static file handler `Blueprint.static()` registration.

laggardkernel:
> Maybe you can change this as a PR to the 20.12 LTS branch and we can include it backwards there.

Thx. Opened another PR for 20.12LTS.

vltr:
2020 was never going to happen, anyway :rofl:

ahopkins:
😷

PR #2021, opened 2021-02-05
[20.12LTS] Remove prefix from websocket handler name

laggardkernel:
Formerly try to open for master branch GH-2020. Cause the Router is overhauled on master, open this pr for 20.12LTS.

---

Remove the websocket prefix `websocket_handler_` introduced in
761eef7. Makes `url_for()` unable to find the corresponding ws route for non-bp websocket view.

```python
@app.websocket('/ws')
async def websocket(request, ws):
    return ""
```

The key for the above handler used in `Router.route_names` is `websocket_handler_websocket`, not `websocket`. Removing the `websocket_handler_` prefix fixes this bug. Searching in the source code, this very prefix is not used in anywhere.

I guess the contributor of commit 761eef7 tried to mimic the registration behavior of static file handler, which prepend the handler name with prefix `_static_`, in order to store static file handler separately in `Router.routes_static_files`, not in `Router.routes_names`.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -12,6 +12,7 @@ from ssl import Purpose, SSLContext, create_default_context
 from traceback import format_exc
 from typing import Any, Dict, Optional, Type, Union
 from urllib.parse import urlencode, urlunparse
+from warnings import warn
 
 from sanic import reloader_helpers
 from sanic.asgi import ASGIApp
@@ -494,9 +495,7 @@ class Sanic:
             websocket_handler = partial(
                 self._websocket_handler, handler, subprotocols=subprotocols
             )
-            websocket_handler.__name__ = (
-                "websocket_handler_" + handler.__name__
-            )
+            websocket_handler.__name__ = handler.__name__
             routes.extend(
                 self.router.add(
                     uri=uri,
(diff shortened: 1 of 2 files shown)


laggardkernel:
During reading the source code. I've found inconsistent behaviors among route storing of normal http handler and websocket, static file handlers.

Firstly, the ways used to indicate a websocket handler and static file handler are different.

`_static_` is prepend into `name` param from `Sanic.static(name="static")`. It's the same `name` param passed to  `Router._add(..., name)`. The `_static_` is used as a flag in `Router._add()` to indicate current handler being registered is a static file handler, not used in anywhere else.

`websocket.is_websocket` is set in `Sanic.websocket()`, and the flag is only used in `Sanic.handle_request()`.

Secondly, the ways to pass blueprint names of http handler, websocket handler, static file handler, to be used in `Router._add()` are different.

Formerly, only `handler.__blueprintname__` is used. The bp name is prepended with handler name as the route name (key used in `Router.routes_name`, `Router.routes_static_files`) in `Router._add()`.

For blueprint websocket handler (`Blueprint.websocket()`), not the original handler but a handler wrapper is passed to `Router._add()`. (Check `Sanic._websocket_handler()` and `static._static_request_handler()`) Seems later contributors (938c49b) didn't copy the `__blueprintname__` attr to the handler wrapper. They chose to pass `bp_name + handler` to `Router._add()` directly. Similar thing happens on static file handler `Blueprint.static()` registration.

ahopkins:
The new branch I am working on does not use any special prefixes for static or websockets. Also, there is a change to how routes are named in general. But I'll leave rhat discussion for #2010.

ahopkins:
Since we are merging this into the LTS, I am hesitant to cause breaking changes. While I agree there is the issue with `url_for`, potentialling the name with the prefix are in use in some applications. Therefore I am thinking that we probably need to add support for both with and without the handler here.

ahopkins (changes requested):
Needs to handle fetching routes with and without so that it is not a breaking change.

sjsadowski:
@laggardkernel are you able to implement @ahoplins request?

laggardkernel:
@sjsadowski Just pushed an update with backward support, in case anyone find the bug and use `url_for("websocket_handler_<view_name>")` in his code.

ahopkins (changes requested):
Looks great @laggardkernel. Can you add a test case to `test_url_building.py` for both cases?

Something like this. Also, one for Blueprints.

```python
def test_url_for_with_websocket_handlers(app):
    @app.websocket("/ws")
    async def my_handler(...):
        ...
    
    assert app.url_for("my_handler") = "..."
    assert app.url_for("websocket_handler_my_handler") = "..."
```

ahopkins (approved):

laggardkernel:
@ahopkins Thanks for pointing out the correct place to add the test. I've updated and squashed the commits.

ahopkins:
Approved and will merge when Travis is done 🐢

I'll get this pushed out to the LTS tonight.

ahopkins:
Thanks again for this.

ahopkins:
@laggardkernel Can you address the CI issues? Linting issues can be fixed by running:

```
make pretty
```

laggardkernel:
@ahopkins Sorry, fixed and passed the CI.

ahopkins (approved):

PR #2024, opened 2021-02-15
Fix uvloop to 0.14.0

ahopkins:
uvloop is [dropping python 3.6](https://github.com/MagicStack/uvloop/releases/tag/v0.15.0) support. I think [we should also](https://community.sanicframework.org/t/dropping-python-3-6/793/4), but I am not quite ready to do that in v21.3. In the mean time, let's fix uvloop to 0.14. Also making a PR for LTS.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,7 @@ env_dependency = (
     '; sys_platform != "win32" ' 'and implementation_name == "cpython"'
 )
 ujson = "ujson>=1.35" + env_dependency
-uvloop = "uvloop==0.14.0" + env_dependency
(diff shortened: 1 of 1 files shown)


ashleysommer:
Why fix it to exactly 0.14? That could be an upper limit:
`uvloop>=0.5.3,<=0.14.0`
That might be more compatible for users who have require uvloop pinned to other versions.

ahopkins:
Fair enough. That is what I did for the two LTS branches: #2025, #2026

ashleysommer (approved):

PR #2031, opened 2021-02-17
Router tweaks

### LINKED ISSUE #1420: url_for() doesn't return a working URI for a route with the trailing slash and strict_slashes=True

### LINKED ISSUE #1525: Routing is incorrect with some special characters (e.g. dot)
I want to use something like this:

```python
@app.route('/path/to/file.<ext>')
async def generate_some_file(request, ext):
    ...
```

It works, but the dot is interpreted as a special regex character that represents ANY character, so there urls are working:

- `/path/to/file.txt`
- `/path/to/fileQtxt`
- `/path/to/file txt`
- `/path/to/file%3Ftxt` etc.

When I try to escape the dot:

```python
@app.route(r'/path/to/file\.<ext>')
async def generate_some_file(request, ext):
    ...
```

then this route works as expected, but `app.url_for('generate_some_file', ext=ext)` returns broken url:

```
/path/to/file\.txt
```

Is also affects other regex characters e.g. `^`, `$`, `[`, `]` etc. For example,

```python
@app.route('/path/to/fi[abcl]e.<ext>')
async def generate_some_file(request, ext):
    return response.text(app.url_for('generate_some_file', ext=ext))
```

When I open `GET /path/to/file.txt` it returns `/path/to/fi[abcl]e.txt` that is totally incorrect.

Note that Flask does not interpret these characters as special and it works as expected (`werkzeug/routing.py` uses `re.escape` everywhere).

I use Sanic 18.12 (because 19.3 is not yet uploaded to PyPI)

ahopkins:
- Branches off #2006, so merge that first.
- Fixes #1420 (`url_for` where `strict_slashes` are on for a path ending in `/`)
- Fixes #1525 to allow for partial matching
- Adds `Request.route`
- Allow for partial regex matching

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -568,25 +568,27 @@ class Sanic(BaseSanic):
         # Define `response` var here to remove warnings about
         # allocation before assignment below.
         response = None
-        name = None
         try:
             # Fetch handler from router
             (
+                route,
                 handler,
                 kwargs,
-                uri,
-                name,
-                ignore_body,
             ) = self.router.get(request)
-            request.name = name
+
             request._match_info = kwargs
+            request.route = route
+            request.name = route.name
+            request.uri_template = f"/{route.path}"
+            request.endpoint = request.name
 
             if (
                 request.stream
                 and request.stream.request_body
-                and not ignore_body
+                and not route.ctx.ignore_body
             ):
-                if self.router.is_stream_handler(request):
+
+                if hasattr(handler, "is_stream"):
                     # Streaming handler: lift the size limit
                     request.stream.request_max_size = float("inf")
                 else:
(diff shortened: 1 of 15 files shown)


sjsadowski (approved):
No real quibbles. Appreciate the increased internal consistency.

ahopkins:
> No real quibbles. Appreciate the increased internal consistency.

How about fake ones?

PR #2032, opened 2021-02-17
Add raw header info to request object

### LINKED ISSUE #1630: RFC : Signals Support for Sanic Proposal

### LINKED ISSUE #1963: In some cases, I need to get the raw request header
Hi,

I find that all request's headers had do a casefold transform.  I know that's right. But I think we should keep save raw headers instead of drop it. I haven't find any property or function to get it yet. In web security, I need to find something from the raw request that is not the same as the real human request.

I know we can add a raw_headers property to solve it. But i think you have a better idea.

ahopkins:
Resolves #1963

This will add raw data from the header. Unit tests and final naming to come. I have been back and forth in my head about trying to keep it consistent with `request.body`. But, I also do not want to have to `decode` more than once. I might change my mind 4 more times before the PR is ready for review.


ahopkins:
I pulled #2006 into this one. So, after that is merged this will be a _much_ smaller PR.

ashleysommer (approved):

PR #2033, opened 2021-02-17
Add BufferedProtocol support

### LINKED ISSUE #1873: The AsyncIO Server should use Python's 3.7 BufferedProtocol when possible ?
Not sure how much performance it can add, but might be worth exploring:
https://bugs.python.org/issue32251

It can be done in a backward compatible way with Python 3.6 like this:
https://github.com/huge-success/sanic/blob/bedf68a9b2025618a94cb8044f495a0abd87a134/sanic/server.py#L47

Instead of inheriting asyncio.Protocol, it can be done:
```python
try:
    from asyncio import BufferedProtocol as BaseProtocol
# Python 3.6 support
except ImportError:
    from asyncio import Protocol as BaseProtocol
```

And inside the implementation, it can be written for both Python 3.6 and 3.7+ by defining the 3 methods (where the first 2 use a preallocated buffer):
```python
    def get_buffer(self, sizehint):
    def buffer_updated(self, nbytes):
    # Python 3.6 support
    def data_received(self, data):
```

ahopkins:
Resolves #1873

This makes use of the [`asyncio.BufferedProtocol`](https://docs.python.org/3/library/asyncio-protocol.html#buffered-streaming-protocols).


ahopkins:
for reference: https://bugs.python.org/issue32251

sanic/server.py
```diff
@@ -295,6 +309,29 @@ def data_received(self, data):
...
+    # Python 3.7+
+    # -------------------------------------------- #
+
+    def get_buffer(self, sizehint=-1):
+        return self._buffer
+
+    def buffer_updated(self, nbytes: int) -> None:
+        data = self._buffer[:nbytes]
+        self._time = current_time()
+        self.recv_buffer += data
```
ashleysommer, line 322: Do you need to clear `self._buffer` here?

what happens if `buffer_updated()` is called a second time with a new nbytes, should it still read `[:nbytes]` from the start of the buffer?

Does it need to keep track of buffer position like this implementation?:
https://gist.github.com/1st1/1c606e5b83ef0e9c41faf21564d75ad7#file-get_buffer_bench-py-L27-L43

ashleysommer: I have lots more questions here, (does get_buffer() get called again after buffer_updated() and before buffer_updated() is called again?, should you clear `self._buffer` before returning it in `get_buffer()`) I should probably just test it locally to clarify to myself exactly how this sequence works.

ashleysommer: Ok, I understand it better now. Resolving this thread.

sanic/server.py
```diff
@@ -295,6 +309,29 @@ def data_received(self, data):
...
+
+        if len(self.recv_buffer) > self.app.config.REQUEST_BUFFER_SIZE:
```
ashleysommer, line 324: After playing with this new code, I wonder if this line should be
```
if len(self.recv_buffer) >= self.app.config.REQUEST_BUFFER_SIZE
```

Heres one thing I observed:
- Set REQUEST_BUFFER_SIZE to 128 bytes.
- BufferedProtocol chooses `self._buffer` with size 128 (self.recv_buffer is unsized/unlimited).
- Send a request to the app with 384 byte request size
- buffer_updated is called with nbytes 128
- 128 bytes is writren to recv_buffer
- `if nbytes > REQUEST_BUFFER_SIZE` is False, because 128 is not greater than 128
- data_received is set, and http headers are read
- http protocol calls _receive_more(), tries to unpause the transport, but its not paused
- buffer_updated() is called again with nsize 128
- a new 128 bytes of data is written to self.recv_buffer
- len(self.recv_buffer) is now 256, so self.transport.pause_reading() is now executed.
- data_received is set, and further http headers are read
- http protocol calls _receive_more() again which _unpauses_ the transport.
- repeats until all of the request is read and parsed by the HTTP protocol

So I'm wondering, should the transport be paused every time the recv_buffer size matches REQUEST_BUFFER_SIZE, or only when its _greater_ than that?
What effect does pausing the transport even have?

ahopkins: The point of pausing would be to allow for it to drain. It is more of an issue with streaming then sending complete responses. It is a protection from exhausting memory limits.

ahopkins: With that said, you are probably correct that it should be `>=`.

sanic/app.py
```diff
@@ -3,7 +3,14 @@
 import os
 import re
 
-from asyncio import CancelledError, Protocol, ensure_future, get_event_loop
+
+try:
```
yunstanford, line 7: maybe we can have this in compat.py module, and don't need try/except blob in multiple places

sanic/server.py
```diff
@@ -295,6 +309,29 @@ def data_received(self, data):
         except Exception:
             logger.exception("protocol.data_received")
 
+    # -------------------------------------------- #
+    # Python 3.7+
+    # -------------------------------------------- #
+
+    def get_buffer(self, sizehint=-1):
```
yunstanford, line 316: what is `sizehint` used for here ?

ahopkins: > sizehint is the recommended minimum size for the returned buffer. It is acceptable to return smaller or larger buffers than what sizehint suggests. When set to -1, the buffer size can be arbitrary. It is an error to return a buffer with a zero size.

[Source](https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BufferedProtocol.get_buffer)

ahopkins: We are ignoring it here.

ashleysommer:
I'm going to do some further testing on this, to determine a recommendation to include it in 21.3 or not.

ahopkins:
Awesome 🤘

ashleysommer:
Ok.. like last time I tested it, the results are again inconclusive.

Speeds are basically the same whether using BufferedProtocol or Protocol. So similar I thought there was something wrong with my testing methodology, but all seems to be good.

I'm testing a basic hello-world Sanic application, with one server worker (no multiprocessing).
I tried different scenarios too, specifically with uvloop and without uvloop, and with an artificial receive-buffer restriction (to trigger stream pause and resume) and without a restriction.
Using `wrk` with 8 connections across 8 threads to hammer the server for 30 seconds per test.

My matrix was (BufferedProtocol,Protocol)*(with_uvloop, without_uvloop)*(with_restriction,without_restriction) giving a total of separate 8 tests.

Every test gave 4400-4500 requests per second. Even with_uvloop and without_uvloop made no difference in my testing on my laptop with Python 3.8.

Repeating all 8 tests above when restricting `wrk` to 1 connection at a time on 1 thread. Again the results were all within margin of error. Results were all 2200-2300 requests per second.

So I don't know. I like the BufferedProtocol implementation, its clean and correct and works well. It _should_ be faster, but in the Sanic server it doesn't provide any speed advantage.

ahopkins:
Did you try with a large payload? What happens if the response is a large blob? Maybe throw in a couple tests at different sizes?

However, these results pretty much confirm exactly my own results. And, I agree with your conclusion. It should be better, it seems like it. But, I think @Tronic streaming solution pretty much doubles up the effort, so we are already seeing that performance increase. So the Bufferedprotocol is doing the same thing except with two methods.

ahopkins:
Continuing the conversation from #2183.

`BufferedProtocol` seems to work for no body, single body, and chunked body requests when sent individually or pipelined.

One thing that probably needs to be reexamined if we are going this direction is whether we want to continue writing to `recv_buffer` the same way, or use a different sync primitive for control.

```python
        if (
            len(self.recv_buffer) > self.app.config.REQUEST_BUFFER_SIZE
        ) and self.transport:
            self.transport.pause_reading()
```

sanic/http.py
```diff
@@ -492,7 +492,7 @@ async def read(self) -> Optional[bytes]:
                 self.request_body = None
                 # Because we are leaving one CRLF in the buffer, we manually
                 # reset the buffer here
-                self.recv_buffer = bytearray()
+                self.recv_buffer = self.protocol.recv_buffer = bytearray()
```
Tronic, line 495: This is still incorrect, and in general you should not recreate the buffer. Use `del self.recv_buffer[:2]` or so to keep the existing buffer but to consume the data that you want removed (note: as discussed earlier, removing all or the first two bytes here might be invalid anyway).

sanic/server.py
```diff
@@ -320,23 +325,26 @@ def pause_writing(self):
...
 
-            if self._data_received:
-                self._data_received.set()
-        except Exception:
-            error_logger.exception("protocol.data_received")
+    def buffer_updated(self, nbytes: int) -> None:
+        data = self._buffer[:nbytes]
+
+        self._time = current_time()
+        self.recv_buffer += data
```
Tronic, line 335: Shouldn't the `get_buffer` function be returning a memoryview to `recv_buffer` directly, rather than separate `_buffer`, to avoid this copying of data here?

Tronic: However, since we are concurrently receiving more data (so cannot alter the buffer) and handling a request (which needs to consume bytes from buffer), this cannot easily work. Which begs the question: why use BufferedProtocol in the first place?

Tronic:
This could offer performance benefits if a single buffer was used. In this case `get_buffer` needs to return a memoryview to a location just past the end of the space already used, and `recv_buffer` should be another memory from the currently unconsumed position (zero if no bytes currently in buffer have been consumed yet) to the end of data received (i.e. the same position where the memoryview of `get_buffer` begins).

The tricky part is that at some point the consumed bytes need to be *actually* removed from buffer and both memoryviews recreated accordingly, so that everything appears intact. I suppose that this update could be done within the `get_buffer` function.

sanic/server.py
```diff
@@ -301,6 +303,9 @@ def connection_made(self, transport):
             self.transport = transport
             self._task = self.loop.create_task(self.connection_task())
             self.recv_buffer = bytearray()
+            self._buffer = memoryview(
+                bytearray(self.app.config.REQUEST_BUFFER_SIZE)
+            )
```
Tronic, line 308: This is a waste of RAM. The buffer should be kept as small as possible, not always the maximum permitted size. This makes a difference when handling a lot of parallel requests.

sanic/http.py
ahopkins: I still am not sure I see how. If we do not, and even if we account for the http loop, you end up with `\r\n` as the first two bytes on the next request.

sanic/server.py
ahopkins: Yes, this was just my proof of concept to see if the BufferedProtocol would play more nicely with streaming requests.

sanic/server.py
ahopkins: I am not sure where this value is coming from, but it is the size limit that `get_buffer` seems to want :thinking: 
```
>> get_buffer sizehint=65536
```

Tronic: The sizehint is hardcoded in libuv (and probably in a similar manner in plain asyncio). We can safely ignore the hint because libuv will then simply call `recv_into` with a smaller buffer. For the sake of simplicity, it might be best to keep our buffer hardcoded to 64 KiB as well, at least until it can be evaluated whether BufferedProtocol offers those performance benefits that we want.

PR #2035, opened 2021-02-19
grammer on docs homepage.

russellballestrini:


sjsadowski:
This changes the meaning of the sentence. "Highly performant" is not the same as "high performance," though they are similar. Possibly picking nits here, but I'm going to close. @sanic-org/sanic-core-devs feel free to reopen if you disagree.

As a side note, I also got a good chuckle out of a grammar correction including a misspelled word in the PR.

ahopkins:
Thanks @russellballestrini. I do appreciate stepping up to submit a PR on this. I do agree with @sjsadowski that there is nothing wrong, and the phrase "highly performant" was what I was after. I know there are some more typos in the new docs, and over the coming month I hope to correct those and fill in a lot of missing information.

PR #2041, opened 2021-02-28
GIT-1630: add basic signal infra

harshanarayana:
@ahopkins / @sanic-org/sanic-core-devs  This is in no way an attempt to discard the PR you already have in place, but during my break from all the community, I had a different implementation in mind than my original DRAFT. The migration to `sanic-router` made that way more easier. The whole idea of Mixin is coming to the resume incredibly well. 

This is still an early draft of the changes and if you don't mind, I would like to pick this back up in the coming days as I am trying to get back to the community and a familiar implementation requirement might be a good thing to start with. 

```python
from sanic import Sanic
from sanic.signals import SignalData

from httptools import parse_url

s = Sanic(name="test")

@s.signal("app.route.before")
async def route_before(app, loop, signal, signal_data: SignalData):
    # Having some event generation before the route resolution can help you automatically route your request to a different URL in case of deprecation/Removal of Old APIs
    signal_data.request._parsed_url = parse_url(b"/tests")
    print(f"Route Event {signal_data}")

@s.signal("app.route.missing")
async def invalid_route_event(app, loop, signal, signal_data):
    print(f"Missing {signal_data}")

@s.signal(signal="foo.bar.baz")
async def event_handler(app, loop, signal, signal_data):
    print(f"Signal Received with {signal_data}")

@s.signal(signal="foo.bar.baz")
async def event_handler1(app, loop, signal, signal_data):
    print("Signal Received1 ")

@s.signal(signal="foo.bar.baz")
async def event_handler2(app, loop, signal, signal_data):
    print("Signal Received2")

@s.middleware("request")
async def middleware(reques):
    print("HERE")

@s.route("/get")
async def get(request):
    await s.publish(signal="foo.bar.baz", data={"email": "something@gmail.com"})
    return json({"test": "test"})

s.freeze()
s.run(port=1245, debug=True)
```
## TODO

- [ ] Handle Existing Event Infra for listeners
- [ ] Performance Test to make sure this doesn't add a noticeable overhead
- [ ] Clean exception handling (Possibly indicate what signal dispatch failure can be ignored)
- [ ] If the Signal Dispatch doesn't need to happen in a sequence, run them as async tasks ?
- [ ] Enable Support on the middleware, blueprint and blueprint groups
- [ ] Possible Router level events ?
- [ ] Handle Signal freeze ?
- [ ] Better cache support for signal registry (lru_cache)
- [ ] Async support for `Router.get` might enable much better route even generation?
- [ ] Finalise Custom Event Generation API names

### DIFF
--- a/sanic/mixins/signals.py
+++ b/sanic/mixins/signals.py
@@ -1,18 +1,8 @@
 import typing as t
 
 from sanic.exceptions import InvalidSignalFormatException
-from sanic.signals import SignalRegistry, SignalContext, SignalData
-
-
-class Singleton(type):
-    _instances = {}
-
-    def __call__(cls, *args, **kwargs):
-        if cls not in cls._instances:
-            cls._instances[cls] = super(Singleton, cls).__call__(
-                *args, **kwargs
-            )
-        return cls._instances[cls]
+from sanic.signals import SignalRegistry, SignalContext
+from sanic.utils import Singleton
 
 
 class SignalMixin:
--- a/sanic/mixins/signals.py
+++ b/sanic/mixins/signals.py
@@ -38,7 +38,7 @@ class SignalMixin:
 
         return _wrapper
 
-    def get_signal_context(self, signal: str) -> SignalContext:
+    def get_signal_context(self, signal: str) -> t.Union[None, SignalContext]:
         return self._ctx_mapper.get(signal)
 
     def freeze(self, signal: t.Union[None, str] = None):
--- a/sanic/mixins/signals.py
+++ b/sanic/mixins/signals.py
@@ -21,7 +21,7 @@ class SignalMixin:
     def __init__(self, *args, **kwargs) -> None:
         self._signal_registry = SignalRegistry()
         self._ctx_mapper = {}  # type: t.Dict[str, SignalContext]
-        self.name = None
+        self.name = ""
 
     def signal(self, signal: str):
         def _wrapper(handler: t.Callable):
--- a/sanic/mixins/signals.py
+++ b/sanic/mixins/signals.py
@@ -1,0 +1,60 @@
+import typing as t
+
+from sanic.exceptions import InvalidSignalFormatException
+from sanic.signals import SignalRegistry, SignalContext, SignalData
+
+
+class Singleton(type):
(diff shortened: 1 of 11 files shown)


sanic/mixins/signals.py
```diff
@@ -0,0 +1,60 @@
+import typing as t
+
+from sanic.exceptions import InvalidSignalFormatException
+from sanic.signals import SignalRegistry, SignalContext, SignalData
+
+
+class Singleton(type):
```
harshanarayana, line 7: Need to refactor this into a common one for reuse.

sanic/mixins/signals.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __init__(self, *args, **kwargs) -> None:
+        self._signal_registry = SignalRegistry()
+        self._ctx_mapper = {}  # type: t.Dict[str, SignalContext]
+        self.name = None
+
+    def signal(self, signal: str):
+        def _wrapper(handler: t.Callable):
+            parts = signal.split(".")
+            if len(parts) < 2 or len(parts) > 3:
```
harshanarayana, line 29: May be we can avoid checking and raise an Index error and be done with it?

sanic/signals.py
```diff
@@ -0,0 +1,150 @@
...
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super(Singleton, cls).__call__(
+                *args, **kwargs
+            )
+        return cls._instances[cls]
+
+
+class SignalContext:
+    __slots__ = ["_namespace", "_context", "_action"]
```
harshanarayana, line 25: Add typedef for slots

ahopkins:
I have a branch that implements this using the new router. Some slight changes in API design, but I think a similar approach. I will push the branch for you to look at and take a closer look at this PR later. Ping me on the DIscord server, it might be nice to find a chance to hash out some details in realtime.

ahopkins:
See [signals-api](https://github.com/sanic-org/sanic/tree/signals-api)

ahopkins (changes requested):
Need to discuss implementation strategy.

PR #2042, opened 2021-02-28
RFC/1630 Signals

### LINKED ISSUE #1630: RFC : Signals Support for Sanic Proposal

ahopkins:
Resolves #1630 

Adds the following API:

- `Sanic.signal`
- `Sanic.event`
- `Sanic.dispatch`
- `Blueprint.signal`
- `Blueprint.event`
- `Blueprint.dispatch`
- `Blueprint.app` (I have been wanting this one for a long time. After we register, it should be available)

See: https://community.sanicframework.org/t/a-fast-new-router/649/41?u=ahopkins and https://github.com/sanic-org/sanic/issues/1630#issuecomment-777373467


ahopkins:
Replaces #1902 
Alternative to #2041

ahopkins:
A current limitation is having a signal on both an app and a blueprint

```python
@app.signal("foo.bar.baz")
def app_signal():
    ...

@bp.signal("foo.bar.baz")
def bp_signal():
    ...
```

This is related to [this issue](https://github.com/sanic-org/sanic-router/issues/2). Since signals is being released as a BETA API, I am not too concerned about this limitation for 21.3. I will try and work on a solution, but even if there is not one, I think we can still include this. This is an absolute must for 21.6.

@sanic-org/sanic-core-devs  thoughts?

ashleysommer:
What actually happens in the blueprint case you described above?

When triggering the signal `"foo.bar.baz"`, does it trigger the `bp_signal()` handler, but not the `app_signal()` handle, because the `bp_signal()` has additional requirements, or something different? What is the expected outcome?

ahopkins:
I think the desired outcome should be that both are executed. Instead it comes up with a NotFound, and neither are executed. I have a patch I will push over the weekend.

ahopkins:
The solution was simple. Will push this soon.

```python
@pytest.mark.asyncio
async def test_dispatch_signal_triggers_on_bp(app):
    bp = Blueprint("bp")

    app_counter = 0
    bp_counter = 0

    @app.signal("foo.bar.baz")
    def app_signal():
        nonlocal app_counter
        app_counter += 1

    @bp.signal("foo.bar.baz")
    def bp_signal():
        nonlocal bp_counter
        bp_counter += 1

    app.blueprint(bp)
    app.signal_router.finalize()

    await app.dispatch("foo.bar.baz")
    assert app_counter == 1
    assert bp_counter == 1

    await bp.dispatch("foo.bar.baz")
    assert app_counter == 1
    assert bp_counter == 2
```

ashleysommer:
I'm not so sure about including `Blueprint.app` feature in this PR.
Is it required for signals to work?

The issue being that a blueprint by its nature is one step removed from the app. It should be possible (and is possible) to register the blueprint on many apps, and multiple apps at the same time. The blueprint by concept shouldn't have any way of knowing what app it is registered on. Having functionality to ask the blueprint instance "what app are you registered on?" doesn't make sense given the design pattern that Blueprints implement.

The best we can offer is perhaps answering the question "what was the _most recent_ app this blueprint was registered on" where the blueprint instance can keep a reference to the last `app` that called `register()` on it.

ahopkins:
You do raise a good point. The necessity for this stemmed from needing access to the `signal_router` in the `event` method. From the instance (app or bp) we should be able to wait on an event.

The problem with allowing a single BP on multiple apps is how should it know what signal to emit? Or, for that matter, to which app should it bind its signals? If more than one, should it be all of them?

ashleysommer:
Yep, easy solution that I think still fits with the design pattern, is to have a List of `apps`, when the blueprint is registered it puts the app in the list. When it emits a signal, it emits it to _all_ the apps in the List.

ahopkins:
It is not _quite_ so simple.

```python
await bp.event("foo.bar.baz")
```

What is this waiting for? If we wait for `all` of the apps, it will block. I suppose we _could_ wait for any of them and then run them in a loop with a timeout for each.

ashleysommer:
Ok. I'm not familiar enough with how the signals are implemented to see why that would block if sending the event to _all_ of the apps. I'll look into it.

ahopkins:
@ashleysommer 

the solution was to use `asyncio.wait(..., FIRST_COMPLETED)` pattern:

```python
        if events:
            return asyncio.wait(
                [event.wait() for event in events],
                return_when=asyncio.FIRST_COMPLETED,
            )
```

This way the blueprint will be listening to any app that dispatches.

ahopkins:

https://community.sanicframework.org/t/https-redirection-with-sanic/810/3

ahopkins:
I cannot figure out why that test sometimes stops responding:
https://travis-ci.com/github/sanic-org/sanic/builds/219739784

sanic/mixins/signals.py
```diff
@@ -0,0 +1,65 @@
...
+
+            if apply:
+                self._apply_signal(future_signal)
+
+            return future_signal
+
+        return decorator
+
+    def event(self, event: str):
+        signal = self.signal_router.name_index.get(event)  # type: ignore
```
harshanarayana, line 62: The `self.signal_router` property doesn't really belong to this mixin directly. Can we do the same thing we did with `_apply_signal` for this as well and implement this on the `Sanic` app instead ? Makes it more easy to leverage the code completion support as well.

sanic/models/futures.py
```diff
@@ -50,3 +51,9 @@ class FutureStatic(NamedTuple):
     host: Optional[str]
     strict_slashes: Optional[bool]
     content_type: Optional[bool]
+
+
+class FutureSignal(NamedTuple):
+    handler: SignalHandler
+    event: str
```
harshanarayana, line 58: If possible can we change the order for handler and event? The way to register them using a decorator would alway have the event in first place. Would be more intuitive.

sanic/signals.py
```diff
@@ -0,0 +1,135 @@
...
+    async def _dispatch(
+        self,
+        event: str,
+        *fields,
+        context: Optional[Dict[str, Any]] = None,
+        where: Optional[Dict[str, str]] = None,
+    ) -> None:
+        if fields:
+            try:
+                event = self.delimiter.join([event, *fields])
```
harshanarayana, line 55: Is there any advantages to supporting this ?  Can we not just enforce you send the full even as a single string for the first argument instead ? It will also let us avoid and entire check

harshanarayana, line 55: Is there any advantages to supporting this ?  Can we not just enforce you send the full even as a single string for the first argument instead ? It will also let us avoid and entire check

tests/test_signals.py
```diff
@@ -0,0 +1,255 @@
...
+    counter = 0
+
+    @app.signal("foo.bar.baz")
+    def sync_signal(amount):
+        nonlocal counter
+        counter += amount
+
+    app.signal_router.finalize()
+
+    task = await app.dispatch("foo.bar.baz", {"amount": 9})
```
harshanarayana, line 145: I think this should still be an acceptable way to dispatch a signal. We can avoid the `*fields` in the `dispatch` function and allow this way to sending the context instead? This is also more user friendly.

harshanarayana (commented):
I am totally digging this new implementation of `signals` using the Router mixin behavior. Except for a few minor requests and comments, this is a 👍 👍  from me.

sanic/blueprints.py
```diff
@@ -208,3 +237,24 @@ def register(self, app, options):
         self.middlewares = middleware
         self.exceptions = exception_handlers
         self.listeners = dict(listeners)
+
+    async def dispatch(self, *args, **kwargs):
+        where = kwargs.pop("where", {})
```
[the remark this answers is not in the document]
ahopkins: Makes sense. The name sort of stuck from the underlying implementation that is meant to be more generalized. I agree we could do a better job here with variable names.

sanic/blueprints.py
```diff
@@ -208,3 +237,24 @@ def register(self, app, options):
...
+        where.update({"blueprint": self.name})
+        kwargs["where"] = where
+        for app in self.apps:
+            await app.dispatch(*args, **kwargs)
```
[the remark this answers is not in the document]
ahopkins: Nope, that's better.

sanic/mixins/signals.py
ahopkins: Yeah. Especially since app and blueprints need different implementations.

sanic/models/futures.py
ahopkins: I think I used this order to mirror Routes. I'll double check.

sanic/signals.py
ahopkins: 🥴 We talked about it months ago. I fee like there was a good use case, but I can't remember. You may be right though that we should just let the dev handle this if they need it.

tests/test_signals.py
ahopkins: If we get rid of the exploded strings, then yes. We can support this.

ahopkins:
Great feedback, thanks.

sanic/models/futures.py
ahopkins: Keeping this:

```python
class FutureRoute(NamedTuple):
    handler: str
    ...
```

tests/test_signals.py
ahopkins: I think that maybe force kwarg here:

```python
    def dispatch(
        self,
        event: str,
        *,
        where: Optional[Dict[str, str]] = None,
        context: Optional[Dict[str, Any]] = None,
    ):
```

In this example:

```python
     task = await app.dispatch("foo.bar.baz", {"amount": 9}) 
```

It is ambiguous what was intended and easy to mixup.

harshanarayana: I see what you did there. I am good with what you have done now. :)

harshanarayana (approved):
This looks great. I don't have any more change requests on this. I did a few tests locally using the branch and it works great. 

/lgmt.

ashleysommer:
I'm nearly finished my review. Nothing extra from me at this stage.

ahopkins:
👍 I will aim to merge it tomorrow. Then finish some more documentation, and we should be on track for 2021-03-21 release.

sanic/blueprints.py
```diff
@@ -209,3 +237,27 @@ def register(self, app, options):
...
+            *[app.dispatch(*args, **kwargs) for app in self.apps]
+        )
+
+    def event(self, event: str):
+        events = set()
+        for app in self.apps:
+            signal = app.signal_router.name_index.get(event)
+            if not signal:
+                raise NotFound("Could not find signal %s" % event)
+            events.add(signal.ctx.event)
```
ashleysommer, line 255: When this BP is registered on multiple apps, does every app need to have a signal in the router? In the current code this will throw an error if _any_ of the apps doesn't have the signal. I think this an editing error, I think the correct logic is:
```suggestion
        events = set()
        for app in self.apps:
            signal = app.signal_router.name_index.get(event)
            if signal:
                events.add(signal.ctx.event)
```

sanic/app.py
```diff
@@ -309,6 +311,28 @@ def _apply_middleware(
...
+            event,
+            context=context,
+            where=where,
+        )
+
+    def event(self, event: str):
+        signal = self.signal_router.name_index.get(event)
+        if not signal:
+            raise NotFound("Could not find signal %s" % event)
+        return signal.ctx.event.wait()
```
ashleysommer, line 334: Can we have the ability to add a timeout to Event?
```suggestion
    def event(self, event: str, timeout: Optional[Union[Int,Float]]):
        signal = self.signal_router.name_index.get(event)
        if not signal:
            raise NotFound("Could not find signal %s" % event)
        return asyncio.wait_for(signal.ctx.event.wait(), timeout=timeout)
```

ashleysommer (changes requested):

ashleysommer:
One thing thing that confused me when I was doing some testing, I was trying to get this to work:
```
async def after_start1(app, loop):
    log.info("waiting on event")
    await app.event("test.me.event")

async def after_start2(app, loop):
    log.info("Sending event")
    await app.dispatch("test.me.event")
```

When you run that, `after_start1` blocks, and `after_start2` never runs. Even though the server is started and `loop.running` is `True`.
It is because listeners are executed like this:
```python3
def trigger_events(events, loop):
    if events:
        for event in events:
            result = event(loop)
            if isawaitable(result):
                loop.run_until_complete(result)
```
So the listeners are  `run_until_complete` _individually, one at a time_.

I think we can change `trigger_events` to:
```
    if events:
        awaitables = []
        for event in events:
            result = event(loop)
            if isawaitable(result):
                awaitables.append(result)
        if awaitables:
            loop.run_until_complete(asyncio.gather(*awaitables))
```
I've tested this and it works.

ashleysommer:
Also just noticed `after_server_start` listeners are firing _twice_ in current master branch. (Someone has added an extra `trigger_events(after_start)` in `app.serve()`. That's a regression that needs to be fixed, I'll create an Issue.

ashleysommer (changes requested):
Need to add this to asgi.py `startup()`:
```python3
@                self.asgi_app.sanic_app.router.finalize()
+                if app.signal_router.routes:
+                    app.signal_router.finalize()
```

ahopkins:
If we make that change we just need to call out that we will not guarantee the order of listeners. Effectively it means sync gets run all before async. Not rhat it's bad, but it's a breaking change to call out.

sanic/blueprints.py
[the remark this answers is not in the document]
ahopkins: :thinking: They are both set to `where` right now. 

Let me first backup and explain. The `requirements` attribute on the `signal` is akin to what the router is using under the hood. Initially, its purpose is to help facilitate virtual host routing:

```python
app.get("/", host="example.com")()
```

At somepoint, I would like to expand its usage to also include routing based upon cookies, headers, etc. But there also might be a need for requirements to hold more structured meta data. I do not want to get too far ahead, but I also want to allow for flexibility to grow. Hence, it is a dictionary.

For now, we can simply pass through that whole object. At some point, maybe we need to add some logic.

In the current usage in signals, it adds the name of the blueprint to the object both with `signal` and `dispatch`.

So, `requirements` is a vestige of the underlying API. I agree that we can expose another term that makes more sense. However, whatever we choose I want to make sure is clear and unambiguous enough that someone can easily pick up the code and let it be obvious what is happening.

---

And this is where I am torn. Originally my thought was that they needed to be two terms since the definition on the signal would potentially carry more details and meta information. On the `dispatch` it is just passing matching criteria.

On the other hand, having two terms can ultimately be confusing. Especially if they are both beginning with `w`. I like `where` because it is a familiar construct in query languages. I think `matches` and `provides` would be the types of keywords that someone would forget and need to look up every time they want to use it. _Was it `match` or `matches`? Does that go in `signal` or `dispatch`?_

I definitely am not a fan of a verb here. I still think `where` is easiest to remember. And, while I originally like having two names for future flexibility, I do not want to make a design decision on a feature that may or may not come into existence in the future.

---

I guess that was a long way to say that I am in favor of a single argument still. If I had to rank the proposals:

1. where
1. condition
1. when
1. extra
1. with

- ~needs~ 
- ~provides`
- ~matches`
- ~arguments~ (confusing what this means, especially since we have another keyword that will provide arguments to the handler

sanic/app.py
```diff
@@ -1031,11 +1055,18 @@ def _helper(
...
-            if not Sanic.test_mode:
-                raise e
+        async def finalize(app, _):
+            try:
+                app.router.finalize()
+                if app.signal_router.routes:
+                    app.signal_router.finalize()  # noqa
+            except FinalizationError as e:
+                if not Sanic.test_mode:
+                    raise e  # noqa
```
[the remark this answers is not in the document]
ahopkins: I think you mean `staticmethod` here. 

:+1:

sanic/blueprints.py
ahopkins: When you register a BP, all of its signals should be on the app. In the case where there are multiple apps, it should be on all of them. If it is missing from one, well, likely something went wrong somewhere.

I am fine making this change, but it still feels to me like something bad just happened that should be handled here.

sanic/app.py
ahopkins: How does `trio` play with `wait_for`?

ahopkins:
> I've tested this and it allows the above example to work as expected.

I think running like this sort of misses the point. I would expect and want that to block. Making your change will make both of them run, but I think that is incorrect. In your use case, you should be using `create_task` or `add_task` to put that waiter into its own Task.

As I pointed out, your change would prioritize all sync functions over async for a a given listener with no way to control that.

sanic/app.py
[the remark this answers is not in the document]
ashleysommer: I usually use `@classmethod`, because I feel its more tightly coupled to the class, but `@staticmethod` works fine too.

sanic/blueprints.py
ashleysommer: Ok, if it was your intention to do it that way, thats fine. It just looked like a copy&paste error when I was looking through the code, like that exception wasn't supposed to be there. (Eg, if the exception is thrown in the `for` loop, there is no way for the `NotFound` exception at the end of the function to be hit).

sanic/app.py
ashleysommer: I haven't tested it on trio, but I expect it to work the same as uvloop and asyncio.

ashleysommer (approved):

ashleysommer:
ok, happy to merge now.

ashleysommer:
Example of an Event-Triggered-Task that I mentioned in #2059. Good idea?
```
    @staticmethod
    async def _event_task_runner(_event, f, *args, **kwargs):
        await _event
        result = f(*args, **kwargs)
        if isawaitable(result):
            result = await result

    def event_task(self, event: str, f):
        _event = self.event(event)
        self.add_task(self._event_task_runner(_event, f))
```

ahopkins:
What would be a use case for this? Why not just create a signal handler? That would run in a task.

Well, I suppose more specifically _ALL_ signal handlers execute in a task.

sanic/blueprints.py
ahopkins: I went back and forth on this myself a couple times. I think maybe you are right though that the final exception doesn't quite fit.

That exception would only be hit if there are no registered apps, in which caste the exception `NotFound` doesn't make sense.

ahopkins: And, because we are accessing `self.apps` as a property, you would get an exception earlier than that before it ever reaches there.

I suppose I was thinking that we needed to catch when `events` was empty, but I am not seeing how it could be.

Anyone see something else?

sanic/blueprints.py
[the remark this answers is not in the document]
ahopkins: @sanic-org/core-devs What if we go with `condition` for both right now:

```python
@app.signal("foo.bar.baz", condition={"one": "two"})
await app.dispatch("foo.bar.baz", condition={"one": "two"})
```

Then, _IF_ in the future we want to add more logic to the matching, we can add another kwarg:

```python
await app.dispatch("foo.bar.baz", match={
    "conditions": [{"one": "two"}, {"one": "three"}],
    "rules": RouteRules.MATCH_ANY
}
```

ashleysommer:
Regarding my helper suggestion:

First I was thinking of how to make `app.event()` more useful. In its most basic usage, `app.event()` will block, and only some future thing on the running event loop (like a scheduled task, or an action on a route handler) can unblock it. So it will be a common design pattern that a user will start a new task and run `app.event()` in there and have that task blocked, if they don't want to deal with the block in their code.

So if that's going to be a pretty common usage pattern, then a helper to do that would be useful. The helper takes a function, and puts it on the new task, to execute after the event unblocks.

Then I was thinking generally a helper that takes a callback function as an argument is better implemented as a function decorator. Hence my second suggestion.

But I agree that might be going a bit too far, as then its actually just replicating the functionality of the Signal Handlers.

ahopkins:
Let's maybe keep this idea in our back pocket as something to implement in round 2 for 21.6. Maybe once it is in the wild and we get to play with it we will see if this might be helpful or not.

ashleysommer (approved):
All good from me

PR #2043, opened 2021-03-02
Add __str__ and __repr__ to Sanic and Blueprint

ahopkins:
I am not sure why we never added something like this earlier.

```
assert str(app) == "<Sanic my_app>"
assert repr(app) == 'Sanic(name="my_app")'
assert str(bp) == "<Blueprint my_bp>"
assert repr(bp) == 'Blueprint(name="my_bp", url_prefix=None, host=None, version=None, strict_slashes=None)'
```

### DIFF
--- a/sanic/base.py
+++ b/sanic/base.py
@@ -33,4 +33,8 @@ class BaseSanic(
     ExceptionMixin,
     metaclass=Base,
 ):
-    ...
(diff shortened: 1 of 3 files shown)


vltr (approved):
this looks simple enough :wink:

PR #2047, opened 2021-03-06
GIT-2045: enable versioning and strict slash on BlueprintGroup

### LINKED ISSUE #2045: Versioning in the Blueprint Group
I use sanic to make api, and I am using it.

Currently, the url is managed through a blueprint group, if use single blueprints can be versioned, but blueprints cannot be versioned in the case of a group.

If versioning is possible, maintenance will be much more convenient.

Can use versioning in the blueprint group.

**Additional context**
```py
Blueprint.group(example_blueprint, example_blueprint_two, url_prefix="api", version="v1")
```

harshanarayana:
closes #2045 

- [x] Unit Tests
- [x] Update Documentation


sanic/blueprint_group.py
```diff
@@ -130,6 +151,15 @@
         :param item: New `Blueprint` object.
         :return: None
         """
+        if self._version and item.version and self._version != item.version:
```
ahopkins, line 154: I would add an `@property` for `BlueprintGroup.version` and `BlueprintGroup.strict_slashes`

sanic/blueprint_group.py
```diff
@@ -130,6 +151,15 @@
...
+            raise APIVersionMismatchException(
```
ahopkins, line 155: I don't think this should be an exception. Maybe a warning, but probably not even that. We do not raise such mismatches when you drilldown Blueprint to handler. Indeed, it is a feature to allow the greater the specificity to win out.

sanic/blueprint_group.py
```diff
@@ -130,6 +151,15 @@
...
+                f"API Version Mismatch. Blueprint {item.name} has version {item.version} "
+                f"while Blueprint Group has {self._version}"
+            )
+        if self._version and not item.version:
+            item.version = self._version
+        if self._strict_slashes is not None:
```
ahopkins, line 161: This should be the same as the version logic. Only apply if it is not set in the `Blueprint`.

ahopkins (changes requested):
Don't forget to add test coverage :sunglasses:

sanic/blueprint_group.py
harshanarayana: Done.

sanic/blueprint_group.py
harshanarayana: All righty. If not an exception, just better to ignore it completely and let the user decide how they want to manage it.

sanic/models/futures.py
```diff
@@ -1,39 +1,46 @@
...
-        "stream_large_files",
-        "name",
-        "host",
-        "strict_slashes",
-        "content_type",
-    ],
-)
+from pathlib import PurePath
+from typing import NamedTuple, List, Union, Coroutine, Any
+
```
harshanarayana, line 3: Hope you don't mind the rewrite of this entire file. Makes it much easier to auto complete and lint things this way.

sanic/blueprint_group.py
```diff
@@ -59,14 +79,33 @@ def url_prefix(self) -> str:
...
+    def blueprints(self) -> List["sanic.Blueprint"]:
         """
         Retrieve a list of all the available blueprints under this group.
 
         :return: List of Blueprint instance
         """
         return self._blueprints
 
+    @property
+    def version(self) -> Union[None, str, int]:
```
ahopkins, line 91: Version can also be a `float`

sanic/blueprints.py
```diff
@@ -106,10 +108,18 @@ def chain(nested):
...
+            version=version,
+            strict_slashes=strict_slashes,
+        )
         for bp in chain(blueprints):
             if bp.url_prefix is None:
                 bp.url_prefix = ""
+            if bp.version is None and version:
+                bp.version = version
+            if bp.strict_slashes is None:
+                bp.strict_slashes = strict_slashes
```
ahopkins, line 122: This needs to mirror the `for _attr in ["version", "strict_slashes"]` pattern. More specifically, I am not a fan of the fact that this logic is duplicated. Ideally there is only one way to onboard a blueprint into a group, even if it can be done from either side.

sanic/models/futures.py
ahopkins: Works for me. Was on the mental list of things to clean up.

sanic/models/futures.py
```diff
@@ -1,39 +1,46 @@
...
+    name: str
+    ignore_body: bool
+    websocket: bool
+    subprotocols: Union[None, List[str]]
+    unquote: bool
+    static: bool
+
+
+class FutureListener(NamedTuple):
+    listener: Callable
```
ahopkins, line 22: We do have defined types that are more specific than just `Callable`. I think they are currently defined in `./handlers.py`, but could probably find a nicer home.

sanic/models/futures.py
```diff
@@ -1,39 +1,46 @@
...
+    event: str
+
+
+class FutureMiddleware(NamedTuple):
+    middleware: Callable
```
ahopkins, line 27: Same.

sanic/models/futures.py
```diff
@@ -1,39 +1,46 @@
...
+    attach_to: str
+
+
+class FutureException(NamedTuple):
+    handler: Callable
```
ahopkins, line 32: Same

sanic/models/futures.py
```diff
@@ -1,39 +1,46 @@
...
+    exceptions: List[Exception]
```
ahopkins, line 33: This should probably be [`BaseException`](https://docs.python.org/3/library/exceptions.html#BaseException).

sanic/models/futures.py
```diff
@@ -1,39 +1,46 @@
...
-    ],
-)
+from pathlib import PurePath
+from typing import NamedTuple, List, Union, Callable
+
+
+class FutureRoute(NamedTuple):
+    handler: str
+    uri: str
+    methods: Union[None, List[str]]
```
ahopkins, line 8: I am not sure that this is always a `List`? :thinking: Don't we also have it as a `frozenset`? Maybe a more generic `Iterable` would be better?

tests/test_blueprints.py
```diff
@@ -893,3 +893,64 @@ def third(request):
...
+        return json({"from": "bp2/r2"})
+
+    @bp2.get("/r3", version=3)
+    async def r3(request):
+        return json({"from": "bp2/r3"})
+
+    group2.insert(0, bp5)
+    app.blueprint([group, group2])
+
+    print(f"{app.router.routes_all}")
```
ahopkins, line 946: Remove

tests/test_blueprints.py
```diff
@@ -893,3 +893,64 @@ def third(request):
 
     assert app.test_client.get("/f1")[1].status == 200
     assert app.test_client.get("/f1/")[1].status == 200
+
+
+def test_blueprint_group_versioning_and_strict_slash():
```
ahopkins, line 898: I think versioning and strict_slashes should really be tested separately. Will make refactoring and maintenance easier.

ahopkins (changes requested):
Thanks for your work on this.

sanic/blueprints.py
harshanarayana: Let me merge these two into a common way using the group handlers

sanic/models/futures.py
harshanarayana: Agreed. `Iterable` is way more well suited for this case than `List[str]`

tests/test_blueprints.py
harshanarayana: 👍

sanic/models/futures.py
harshanarayana: Relocating this into a common model under `sanic.models.handler_types` so that we can extend them as required. This comes in like with how we are managing the `protocol_types` as well.

examples/versioned_blueprint_group.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+app.blueprint(bpg)
+
+print(f"{app.router.routes_all}")
+if __name__ == '__main__':
+    app.run(host="0.0.0.0", port=8000)
+
+from urllib import parse
+parse.urlparse(url=b'/v2/sentient/robot/vision/name')
```
ahopkins, line 39: ??

sanic/blueprint_group.py
```diff
@@ -121,7 +160,30 @@ def __len__(self) -> int:
...
-    def insert(self, index: int, item: object) -> None:
+    def _sanitize_blueprint(self, bp: "sanic.Blueprint") -> "sanic.Blueprint":
+        """
+        Sanitize the Blueprint Entity to override the Version and strict slash behaviors as required.
+
+        :param bp: Sanic Blueprint entity Object
+        :return: Modified Blueprint
+        """
+        item_prefix = bp.url_prefix or ""
+        bp.url_prefix = f"{self._url_prefix}{item_prefix}"
```
ahopkins, line 171: Doesn't this have the potential of `/foo//bar` or `/foobar`, instead of the intended: `/foo/bar`?

tests/benchmark/test_route_resolution_benchmark.py
```diff
@@ -13,6 +16,34 @@
 sanic.router.ROUTER_CACHE_SIZE = 0
 
 
+# noinspection PyBroadException
+class TestURLParseBenchmark:
```
ahopkins, line 20: I think we can keep this out. I do not think there is a need to keep this in the core repo right now. If you want to keep it somewhere safe for later usage, I would suggest a gist, or open an new repo under `sanic-org`.

tests/test_blueprints.py
```diff
@@ -893,3 +893,75 @@ def third(request):
...
+    async def r3(request):
+        return json({"from": "bp2/r3"})
+
+    app.blueprint([group, group2])
+
+    assert app.test_client.get("/v1/group1/bp1/r1/")[1].status == 200
+    assert app.test_client.get("/v2/group1/bp2/r2")[1].status == 200
+    assert app.test_client.get("/v1/group1/bp1/pre-group")[1].status == 200
+    assert app.test_client.get("/v3/group1/bp2/r3")[1].status == 200
+    assert app.test_client.get("bp3/r1")[1].status == 200
```
ahopkins, line 936: What's happening here? Is the test client adding the first `/`?

tests/test_blueprints.py
```diff
@@ -893,3 +893,75 @@ def third(request):
...
+    group = Blueprint.group(
+        [bp1, bp2], url_prefix="/slash-check", version=1.3, strict_slashes=True
+    )
+
+    app.blueprint(group)
+
+    assert app.test_client.get("/v1.3/slash-check/r1")[1].status == 200
+    assert app.test_client.get("/v1.3/slash-check/r1/")[1].status == 200
+    assert app.test_client.get("/v3/slash-check/bp2/r1")[1].status == 200
+    assert app.test_client.get("/v3/slash-check/bp2/r1/")[1].status == 404
```
ahopkins, line 967: Can you also add

```python
@bp2.get("/r2/")
async def bp2_r2(request):
    return json({"from": "bp2/r2"})

...

assert app.test_client.get("/v3/slash-check/bp2/r2")[1].status == 404
assert app.test_client.get("/v3/slash-check/bp2/r2/")[1].status == 200
```

ahopkins (changes requested):
Looks good. A couple more minor changes.

examples/versioned_blueprint_group.py
harshanarayana: Oh man! What was I thinking!  Something from the benchmark thing got copied wrong!

tests/test_blueprints.py
harshanarayana: Yes. The test Cline will prefix it. `uri = uri if uri.startswith("/") else f"/{uri}"` But then again that is beyond the scope of this change. Let me prefix `/` in the test itself.

sanic/blueprint_group.py
```diff
@@ -121,7 +160,33 @@ def __len__(self) -> int:
...
+        Sanitize the Blueprint Entity to override the Version and strict slash behaviors as required.
+
+        :param bp: Sanic Blueprint entity Object
+        :return: Modified Blueprint
+        """
+        if self._url_prefix:
+            merged_prefix = "/".join(
+                u.strip("/") for u in [self._url_prefix, bp.url_prefix or ""]
+            ).rstrip("/")
+            bp.url_prefix = f"/{merged_prefix}"
```
ahopkins, line 174:
```python
bp.url_prefix = (
        f"/{merged_prefix}"
        if not merged_prefix.startswith("/")
        else merged_prefix
    )
```
```

File: tests/conftest.py [hunk 20]
```diff
@@ -1,3 +1,5 @@
+import json
+import os
```
ahopkins, line 2: Do we need these imports?

ahopkins (changes requested):

ahopkins (approved):

PR #2049, opened 2021-03-07
Setup: Find packages

elis-k:
Error when installing from master because some packages ('models' & 'mixins') are not being included.
When installing master like this `pip3 install git+https://github.com/sanic-org/sanic`


ahopkins (approved):

ahopkins:
Nice catch,  thanks for the PR.

PR #2050, opened 2021-03-07
GIT-2023: Enable GitHub Actions support

### LINKED ISSUE #2023: Migrate to GitHub Actions
Both [sanic-routing](https://github.com/sanic-org/sanic-router) and [sanic-testing](https://github.com/sanic-org/sanic-testing) are now using GitHub Actions for running the test suite and also handling deployments. It has proved itself super simple, convenient, and (most importantly) responsive/speedy.

I suggest we move over the CI from Travis to GH.

harshanarayana:
# FOR RELEASE AFTER 21.3

closes #2023 

### Possible Issues

1. Unable to Retry individual test from a workflow. (https://github.com/actions/runner/issues/432)


ahopkins:
looks like a good start. I'll pop in with some of my ideas later.

ahopkins (approved):
Looks good to me.

ahopkins (changes requested):
Can you add a release workflow that copies the one in `sanic-router`? Release to PyPI on creation of a new release tag in GH.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,38 @@
...
+      run: |
+        python --version
+
+    - name: Install Dependencies Required for Tests to be run
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox tox-gh-actions
+
+    - name: Run Tests
+      run: tox || true
```
ahopkins, line 37: Rather than running everything here, maybe we should run each of the tests separately so that it is easy at a glance to see what failed.

ahopkins (changes requested):

harshanarayana:
> Can you add a release workflow that copies the one in `sanic-router`? Release to PyPI on creation of a new release tag in GH.

Yes. I will do that. I am actually considering creating an entirely new action dedicated just for sanic repo in general. Let me see how that effort goes.

.github/workflows/pull-request.yml
harshanarayana: I am good with that. You mean run one tox env Per run and run it over n matrix ? I was actually considering splitting the test tasks and sanity check like type check and linter to a different file so that I can easily manage them. Would that be ok ?

ahopkins:
What do you mean?

harshanarayana:
> What do you mean?


Instead of defining the github actions in each repo, I would rather create a series of reusable action and leverage them across for all `sanic-org` projects. 

```yaml    - name: Run ${{ matrix.action }} on ${{ matrix.python-version }}      uses: sanic-org/sanic-actions@v2      with:        python-version: ${{ matrix.python-version }}        action: ${{ matrix.action }}
```
This is all you need to include in your github workflow with right set of python Version + os matrix. We can have a reusable set of actions in `sanic-org/sanic-actions` and reuse them everywhere.

ahopkins:
@harshanarayana What is left to get this one going?

.github/workflows/pull-request.yml
ahopkins: You mean a different `tox.ini`?

What I meant was having something like this:

```
- name: Run Tests 
  run: tox -e py38
- name: Run Tests 
  run: tox -e py38
- name: Type checking
  run: tox -e type-checking
...  
```

ahopkins: Then there is not a ridiculously  long list of checks on a PR, but when you go into a specific workflow, it is easy to drilldown and see (for example) that it was a linting issue that caused a failure (instead of needing to look through the entire `tox` output.

harshanarayana:
> @harshanarayana What is left to get this one going?

I just need to restructure this to use the new actions setup and then some minor workflow naming. A bit held up on office work for today and tomorrow. I will get this done and update the PR in another 2 days,.

ahopkins:
Sounds great.

ahopkins (approved):

harshanarayana:
@ahopkins once we merge this, let us start by setting all the test from workflow as an optional test and run it for a while to see if there are any other kinks that need to be sorted before we take travis out and then configure the mandatory test pass criteria. Does that sound okay to every @sanic-org/core-devs

harshanarayana:
For now, I have added an optional way to configure the test retires and mark certain test for ignoring the failure like we did in Travis. BUt I am yet to find a clean way to mark the failed but ignored test as warning.

ahopkins:
> @ahopkins once we merge this, let us start by setting all the test from workflow as an optional test and run it for a while to see if there are any other kinks that need to be sorted before we take travis out and then configure the mandatory test pass criteria. Does that sound okay to every @sanic-org/core-devs

Yup. That was the plan.

Issue #1667, opened 2019-09-04
How to rewrite `response.json()` with middleware?


ndaidong:
Hello,

I have a router as below:

```python
@app.get('/api/feed/entries')
async def load_feed_entries(request):
    return list_entries(request)
```

And the handler:

```python
from bson.json_util import dumps
from json import loads

from sanic.response import json

from helpers.MongoStore import mFeed
from helpers.Utils import simplify


def list_entries(req):
    # extract and process args
    # before loading data from mongodb
    cursor = mFeed.find() # with more options

    for item in cursor:
        entries.append(loads(str((dumps(item, ensure_ascii=False)))))

    if len(entries) > 0:
        entries = list(map(simplify, entries))

    result = dict(
        query=dict(
            skip=skip,
            limit=limit,
            language=lang,
            category=category
        ),
        entries=entries
    )

    return json(result)
```

Then I got the output with encoded non-ascii characters:

![Screenshot_2019-09-04_07-55-13](https://user-images.githubusercontent.com/1809619/64217881-5cbb6e80-cee9-11e9-9904-64d4c9ec7eec.png)

In order to get these unicode display as they are, I specify `dumps` option of `json()` method as below:

```python

    def utf8_dumps(data):
        return dumps(data, ensure_ascii=False)

    return json(result, dumps=utf8_dumps)
``` 

Result now looks like below:

![Screenshot_2019-09-04_08-01-13](https://user-images.githubusercontent.com/1809619/64218040-321de580-ceea-11e9-93cf-6907a91e58d7.png)

This approach resolves the problem, however I have many endpoints like that. How to avoid repeating the same code? I think of middleware and try:

```python

from json import dumps

from sanic import Sanic
from sanic.response import json

@app.middleware('response')
async def modify_json_response(request, response):
    def utf8_dumps(data):
        return dumps(data, ensure_ascii=False)

    def json2(data):
        json(data, dumps=utf8_dumps)

    response.json = json2

```

This doesn't work. What's wrong? How to fix it? Thanks.

Tronic:
Response middleware runs after your handler, when the JSON is already written to `response.body`. JSON is converted ty bytes as soon as you call sanic.response.json() with your data. Thus, the only way this can work is if you can read and write back the body in correct format.

Another approach would be to derive from response.json with your own class with custom __init__ and then use that in all code.

Tronic:
[...]

PR #2051, opened 2021-03-08
Add response serializer

elis-k:
Serialize the value returned by a route handler with a custom function. 
A handler can return any value (i.e string, dict) and the serializer function builds the HTTPResponse. 
`app.response_serializer(custom_serializer_func)`
Example in issue #1667

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -104,6 +104,7 @@ class Sanic(BaseSanic):
         self.config = Config(load_env=load_env)
         self.request_middleware: Deque[MiddlewareType] = deque()
         self.response_middleware: Deque[MiddlewareType] = deque()
+        self._response_serializer = None
         self.blueprints: Dict[str, Blueprint] = {}
         self._blueprint_order: List[Blueprint] = []
         self.configure_logging = configure_logging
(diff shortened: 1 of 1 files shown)


ahopkins:
Thank you for submitting a PR. 

What problem are you trying to achieve? How do you intend for this to be used?

elis-k:
When working on API's the response must be on certain format.

It would be quicker to return a string from the route handler like:
`"User registered successfully."`

And then the serializer builds the response like:
`{
"message": "User registered successfully.",
"status": "success"
...add other references 
}`

Yea this could be done in other ways but a simple serializer method is nice.

ahopkins:
> It would be quicker to return a string from the route handler like:
> `"User registered successfully."`

You mean this?

```python
@app.get("/")
def foo(request):
   return "hello, world."
```

That pattern was rejected for Sanic because of the performance implications.

ahopkins:
I did put some more thought into this. I certainly do see the utility, but I would do a few things different.

The most important change is that I do _not_ want to add another check for a somewhat small use case in the response cycle. Therefore, this should wait until at least 21.6 with some of the startup time optimizations where we can write this into the cycle only if it exists.

PR #2052, opened 2021-03-10
Fix some examples and docs

### LINKED ISSUE #1165: Inconsistency in documentation and examples
There is an example of how to [implement request id logging](https://github.com/channelcat/sanic/blob/master/examples/log_request_id.py) The [documentation ](https://sanic.readthedocs.io/en/latest/sanic/deploying.html#asynchronous-support) says that running the Sanic like it is done in this example is highly discouraged.
Is the documentation old and now it's fine to use that approach with multiple workers or the example is misleading?
Cheers.

### LINKED ISSUE #1189: vhost using blueprints with nonstandard port does not work as documented
I noticed that when I followed the example given in the sanic repo for vhosts [here](https://github.com/channelcat/sanic/blob/master/examples/vhosts.py). Taking some exerpts from this example below, I modified my ```/etc/hosts``` file to point ```bp.example.com``` to ```127.0.0.1``` for testing.

I've shortened that example to illustrate the issue:

```
app = Sanic()
bp = Blueprint("bp", host="bp.example.com")

@bp.route("/question")
async def hello(request):
    return response.text("What is the meaning of life?")

@bp.route("/answer")
async def hello(request):
    return response.text("42")

app.blueprint(bp)

if __name__ == '__main__':
```

What I found was, this example didn't work. I was getting errors like this:

```
Error: Requested URL /answer not found
```

Digging deeper into the code, apparently the code is looking at the ```Host:``` header to decide if it has a route for a particular host. In the case of nonstandard ports (not 80 or 443), the ```Host:``` header has the ports appended. What would make this example work is changing one line:

```
bp = Blueprint("bp", host="bp.example.com:8000")
```

It is worth documenting this. A better fix might be perhaps added a ```port``` option to the Blueprint constructor when the port is non-standard.

### LINKED ISSUE #1380: Port not provided in URLs produced with `url_for`

ahopkins:
Fixes #1165
Fixes #1189 
Fixes #1380

Closing some old `documentation` tag items and making some of the examples more current.

### DIFF
--- a/examples/Dockerfile
+++ b/examples/Dockerfile
@@ -1,11 +1,8 @@
-FROM python:3.5
-MAINTAINER Channel Cat <channelcat@gmail.com>
+FROM sanicframework/sanic:LTS
 
-ADD . /code
-RUN pip3 install git+https://github.com/channelcat/sanic
+RUN mkdir /srv
+COPY . /srv
 
-EXPOSE 8000
+WORKDIR /srv
 
-WORKDIR /code
-
-CMD ["python", "simple_server.py"]
+CMD ["sanic", "simple_server.app"]
--- a/examples/log_request_id.py
+++ b/examples/log_request_id.py
@@ -1,86 +1,75 @@
-'''
-Based on example from https://github.com/Skyscanner/aiotask-context
-and `examples/{override_logging,run_async}.py`.
-
-Needs https://github.com/Skyscanner/aiotask-context/tree/52efbc21e2e1def2d52abb9a8e951f3ce5e6f690 or newer
-
-$ pip install git+https://github.com/Skyscanner/aiotask-context.git
-'''
-
-import asyncio
-import uuid
 import logging
-from signal import signal, SIGINT
 
-from sanic import Sanic
-from sanic import response
-
-import uvloop
 import aiotask_context as context
+
+from sanic import Sanic, response
+
 
 log = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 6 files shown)


ahopkins:
Come on Travis!

image
image

PR #2055, opened 2021-03-10
JSON encoder change via app

### LINKED ISSUE #2054: Universal JSON encoder change
By default, Sanic uses `ujson` for serializing JSON data.

This can be changed by _not_ installing it in your environment. Or, you can change it in the response:

```python
from json import dumps

json({"foo": "bar"}, dumps=dumps)
```

But, this means that you would need to change it everywhere or monkeypatch if you want to change the default encoder, but still have `ujson` available in your environment. :unamused: 

---

Proposal to add:

```python
Sanic(..., dumps=dumps)
```

ahopkins:
Resolves #2054 

---

This one has come up a few times in the past. The current method for setting this at the default level by **only** looking at the environment is not ideal.

Adds the following API to set the encoder globally.

```python
Sanic(..., dumps=my_favorite_json_encoder)
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @sanic-release-managers 
+*  @sanic-org/sanic-release-managers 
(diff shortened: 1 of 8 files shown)


harshanarayana (approved):
Looks good to me. Thanks for taking care of some of the import sorting issues.

ahopkins:
Thanks `isort` 🤘

PR #2059, opened 2021-03-13
[FIX] Allow async trigger handlers to run together on the loop

ashleysommer:
Allow async trigger handlers to run together on the loop without blocking each other

This is more important now because Signals introduce blocking behavior.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -3,7 +3,7 @@ import logging.config
 import os
 import re
 
-from asyncio import CancelledError, Protocol, ensure_future, get_event_loop
+from asyncio import CancelledError, Protocol, ensure_future, gather, get_event_loop
 from asyncio.futures import Future
 from collections import defaultdict, deque
 from functools import partial
@@ -969,10 +969,13 @@ class Sanic(BaseSanic):
         :param events: one or more sync or async functions to execute
         :param loop: event loop
         """
+        awaitables = []
         for event in events:
             result = event(loop)
             if isawaitable(result):
-                await result
+                awaitables.append(result)
+        if awaitables:
+            await gather(*awaitables)
 
     async def _run_request_middleware(self, request, request_name=None):
         # The if improves speed.  I don't know why
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -336,10 +336,13 @@ def trigger_events(events: Optional[Iterable[Callable[..., Any]]], loop):
     :param loop: event loop
     """
     if events:
+        awaitables = []
         for event in events:
             result = event(loop)
             if isawaitable(result):
-                loop.run_until_complete(result)
+                awaitables.append(result)
+        if awaitables:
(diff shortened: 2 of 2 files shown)


ashleysommer:
Travis is failing on: `test_signal_handlers` where `uvloop == False`. Is that expected?

ahopkins:
I am not sure that I like this change. It changes existing behavior, and I am not sure that it has a good reason for doing so.

```python
@app.listener("before_server_start")
async def first(*args):
    ...

@app.listener("before_server_start")
def second(*args):
    ...
```

I now have no way to run these as expected; `second` will always before `first` regardless of the order in which they are applied.

As I pointed out, I think the fact that `event` is blocking: (a) is a good thing, and (b) is expected.

To achieve your example, I believe this should be the requirement:


```python
async def wait_for_event(app):
    log.info("waiting on event")
    await app.event("test.me.event")

async def after_start1(app, loop):
    app.add_task(wait_for_event(app))

async def after_start2(app, loop):
    log.info("dispatching event")
    await app.dispatch("test.me.event")
```

Keeping this requirement allows the developer to still maintain the ability to control the order of execution. Even when you have multiple `async` functions (and removing the `sync` preference), when using `gather` there is no guarantee that one listener will complete before the next. This is potentially a very large breaking change, and one that would be very hard for a developer to compensate for.

ahopkins:
As an example:

```python
@app.listener("before_server_start")
async def first(*args):
    print("first")


@app.listener("before_server_start")
def second(*args):
    print("second")


@app.listener("before_server_start")
async def primary(*args):
    print("primary start")
    await asyncio.sleep(0)
    print("primary stop")


@app.listener("before_server_start")
async def secondary(*args):
    print("secondary start")
    print("secondary stop")
```

With current:

```
first
second
primary start
primary stop
secondary start
secondary stop
```

With this change:

```
second
first
primary start
secondary start
secondary stop
primary stop
```

ashleysommer:
Ok, I agree this is a fairly big breaking change for some users, and I did think about that while writing the PR. We can go ahead and close this.

As for reasoning, I think I was misunderstanding the use case of `app.event()`. If the user puts it into a function and schedules it as a Task on the loop, then that is fine, but there needs to be documentation demonstrating that use case (I know that's coming). Or even an "Event-Triggered-Task" helper that does that for you.

I'm now wondering if it makes sense for listener handlers have the ability to be `async` at all? They're executed sequentially and within them you cannot await anything without blocking, so they're not really async and could lead to confusion. As you mentioned, handlers can use `app.add_task()` to put tasks on the loop, and handlers even have access to `loop` to do other loop operations.

ahopkins:
True they are not really async, and that's probably fine. Perhaps should be documented. But they need to have the ability because people will want to call other async functions. Many async dB libs use an async initializer (which I think is lame and annoying, but I digress). Having the the handlers return coroutines makes the syntax easier on the user.

I'll add something to the docs about this though.

PR #2060, opened 2021-03-14
Add signal reservations

ahopkins:
It occurred to me while writing documentation for #1630/#2042 that there were certain patterns of names that we probably do not want people to start declaring on their own. We do not know exactly what those patterns are yet, but I thought it might make sense to lock in some reserved names so people do not start implementing them now, and then end up with weird behaviors down the road.

Basically, that means we are adding some code right now that will be temporary. Or, at the very least will change.

Perhaps I am being overly cautious. Thoughts? Do we need this? Is the list complete enough for now?

Current reservations:

- Anything starting `server.`
- Anything starting `http.`
- ~Anything starting `sanic.` that is not `sanic.notice.` or `sanic.log.`~


ashleysommer:
> that is not `sanic.notice`. or `sanic.log`.

Is the thought that the user might want to handle potential future `sanic.notice.*` and `sanic.log.*` events themselves?

sanic/signals.py
```diff
@@ -131,3 +132,43 @@ def finalize(self, do_compile: bool = True):
...
+        if (
+            len(parts) != 3
+            or parts[0].startswith("<")
+            or parts[1].startswith("<")
+        ):
+            raise InvalidSignal("Invalid signal event: %s" % event)
+
+        current = Reservation(*parts)
+        for reservation in RESERVED_EVENTS:
+            if reservation.namespace == "*":
```
ashleysommer, line 147: None of the Reservations in `RESERVED_EVENTS` have namespace `*`. Why do we check that here?

Should it be `if current.namespace == "*"`? If so, should that go outside the for-loop?

ashleysommer: Or is the intention that if we put `*.foo.bar` in `RESERVED_EVENTS` then all added events are invalid?

sanic/signals.py
```diff
@@ -131,3 +132,43 @@ def finalize(self, do_compile: bool = True):
...
+                    reservation.reference == "*"
+                    or current.reference
+                    not in reservation.reference.split(",")
+                ):
+                    raise InvalidSignal(
+                        "Cannot declare reserved signal event: %s" % event
+                    )
+
+                if reservation.reference is None or reservation.action is None:
+                    continue
```
ashleysommer, line 165: `reservation.reference` cannot be None here. We call `reservation.reference.split(",")` above, so it must be a string. It it was None Line 158 would throw and exception.
Do we need to move `if reservation.reference is None: continue` to above line 155?

sanic/signals.py
ahopkins: Yes, the idea would be that `*.foo.bar` would kill the ability to have any custom namespace.

ahopkins:
> > that is not `sanic.notice`. or `sanic.log`.
> 
> Is the thought that the user might want to handle potential future `sanic.notice.*` and `sanic.log.*` events themselves?

I think these might be names we want to control. Probably also `sanic.exception.*`. But, perhaps this is too much limitation. Maybe we just disallow them from defining anything that is `server`, `http` , or `sanic` and leave it at that.

sanic/signals.py
ahopkins: Good catch. Again, maybe I am overcomplicating and the whole thing should just be:

```python
parts = event.split(",")
if parts[0] in ("server", "http", "sanic"):
    raise InvalidSignal
```

And then worry about any more complicated logic when we get to it.

ashleysommer: Yep. I'd go with the simpler solution for now. We can think about reservation rules and how they are matched against signals, in the future.

ahopkins: Okay, I updated it to only be reserved namespaces `http` and `server`.

I want to push people (via the docs) to get used to using `sanic.notice.something` when they have a simple use case and a full blown namespaced event doesn't make sense.

ashleysommer (approved):

PR #2061, opened 2021-03-14
Add add_signal method

ahopkins:
We have funcitonal methods for adding routes, middleware, and listeners:

- `app.add_route()`
- `app.register_middleware()`
- `app.register_listener()`

It seems natural to have `add_signal` as well.


ashleysommer:
Good catch. I didn't notice that while looking at the Signals PR.

ashleysommer (approved):

PR #2063, opened 2021-03-15
RFC/1684 Context objects

### LINKED ISSUE #1684: RFC: User context in app, connection and request
Sanic allows adding user data to app and request objects (latter via `request.ctx` after #1666 got merged). Additionally, per-connection (across requests in HTTP pipelining) data could be added too, and be useful e.g. for middlewares implementing user tracking or slightly more sophisticated request rate limiting.

According to our current test suite, app middleware would add their things directly as app member variables, although I am not aware of middlewares or other software actually making use of this. Adding directly to Sanic core objects is problematic, as any future additions to the Sanic object will need to consider name collisions with any existing software.

Therefore, I am suggesting to add `app.ctx` and possibly `request.transport.ctx` (or something alike) for holding these data in the same manner as the per-request `request.ctx`.

It needs to be noted that `app` object is duplicated in each worker, and thus changes when the server is running would only affect that particular worker. Furthermore, any data added before the server forks will in general need to be picklable, although I'd like to change this so that all initialization, including that of user's main module and app definition, would be done locally in each worker instead of passing from main process by pickling/forking.

Are you aware of any such use or need for it, or care to share any other thoughts on this matter?

ahopkins:
Resolves #1684 

The proposal is to add a `request.ctx` like API to Sanic applications, blueprints, and connection objects.

A common practice is to add arbitrary objects on the Sanic app instance. This introduces a `__fake_slots__` that is meant to be replaced by a real `__slots__` in the future. In the meantime, we will check and raise a warning to direct the developer to using `app.ctx`.

---

Perhaps the more exciting introduction is `request.connection.ctx`.

This object will remain alive so long as a single transport is kept open. Since the `KEEP_ALIVE` will not close a transport layer, we can leverage that to place a context object on it.

A practical benefit of this might be only needing to check authentication or perform some expensive operation once for several calls.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -17,6 +17,7 @@ from inspect import isawaitable
 from socket import socket
 from ssl import Purpose, SSLContext, create_default_context
 from traceback import format_exc
+from types import SimpleNamespace
 from typing import (
     Any,
     Awaitable,
@@ -32,6 +33,7 @@ from typing import (
     Union,
 )
 from urllib.parse import urlencode, urlunparse
+from warnings import warn
 
 from sanic_routing.exceptions import FinalizationError  # type: ignore
 from sanic_routing.exceptions import NotFound  # type: ignore
@@ -76,6 +78,42 @@ class Sanic(BaseSanic):
     The main application instance
     """
 
+    __fake_slots__ = (
+        "_asgi_client",
+        "_blueprint_order",
+        "_future_routes",
+        "_future_statics",
+        "_future_middleware",
+        "_future_listeners",
+        "_future_exceptions",
+        "_future_signals",
+        "_test_client",
+        "_test_manager",
+        "asgi",
+        "blueprints",
+        "config",
+        "configure_logging",
+        "ctx",
+        "debug",
+        "error_handler",
+        "go_fast",
+        "is_running",
+        "is_stopping",
+        "listeners",
+        "name",
+        "named_request_middleware",
+        "named_response_middleware",
+        "request_class",
+        "request_middleware",
+        "response_middleware",
+        "router",
+        "signal_router",
+        "sock",
+        "strict_slashes",
+        "websocket_enabled",
+        "websocket_tasks",
+    )
+
     _app_registry: Dict[str, "Sanic"] = {}
     test_mode = False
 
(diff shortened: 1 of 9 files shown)


vltr:
> Perhaps the more exciting introduction is `request.connection.ctx`.
> 
> This object will remain alive so long as a single transport is kept open. Since the `KEEP_ALIVE` will not close a transport layer, we can leverage that to place a context object on it.
> 
> A practical benefit of this might be only needing to check authentication or perform some expensive operation once for several calls.

This is awesome! More interestingly would be, in some cases (I can see myself doing that at least), making `request.ctx` to actually be `request.connection.ctx` ... I don't know if by some future option or even a custom implementation where needed. Either way, this is awesome.

sanic/app.py
```diff
@@ -104,31 +144,33 @@ def __init__(
...
-        self.request_class = request_class
-        self.error_handler = error_handler or ErrorHandler()
-        self.config = Config(load_env=load_env)
-        self.request_middleware: Deque[MiddlewareType] = deque()
-        self.response_middleware: Deque[MiddlewareType] = deque()
         self.blueprints: Dict[str, Blueprint] = {}
-        self._blueprint_order: List[Blueprint] = []
+        self.config = Config(load_env=load_env)
         self.configure_logging = configure_logging
+        self.ctx = SimpleNamespace()
```
vltr, line 155: This could later be an option, like the `router`, such as:

```python
self.ctx = ctx or SimpleNamespace()
```

No need to do that now, but I would like to open that discussion for later :wink:

sanic/blueprints.py
```diff
@@ -81,6 +115,18 @@ def __repr__(self) -> str:
...
+        # setting attributes on the app instance can be removed and deprecated
+        # with a proper implementation of __slots__
+        if name not in self.__fake_slots__:
+            warn(
+                "Setting variables on blueprint instances is deprecated and "
+                "will be removed in version 21.9. You should change your "
+                f"blueprint to use bp.ctx.{name} instead."
+            )
+        super().__setattr__(name, value)
+
```
vltr, line 129: Should we add this to a base class (and do the same with `Application`) to map deprecated methods on one specific place? I don't know if that's actually a good idea, but I'll leave here for a thought.

vltr (approved):
I made some comments to some parts, nothing too fancy neither blocking, so I'll leave them there in case you want but approve the PR already, as this seems awesome :sunglasses:

ahopkins:
The PR is ready, and we could potentially squeeze it in. My only concern is if there are any edge cases we are not considering where keep alive would apply, but that connection object is recreated. 🤔 Maybe we merge as "beta" feature and finalize it in 21.6?

sanic/app.py
ahopkins: I think this is actually sort of unnecessary. The whole purpose of the context object, and using `SimpleNamespace` in particular is that you can set anything you want on it.

vltr (approved):
awesome :sunglasses:

PR #2072, opened 2021-03-20
fix?: recursion error on Sanic subclass initialisation

### LINKED ISSUE #2073: RecursionError on Sanic subclass initialisation
Discovered at https://github.com/sanic-org/sanic/issues/2071

version: latest sanic master (`8a2ea626c6d04a5eb1e28d071ffa56bf9ad98a12`)

description:

RecursionError occurs when initialising Sanic subclass

minimal code to reproduce:

```python
from sanic import Sanic


class Custom(Sanic):
    pass

custom = Custom("custom")
```

Potential fix: https://github.com/sanic-org/sanic/pull/2072

artcg:
Fixes #2073

Needs someone to take a look at it, not super familiar with `__new__`,

but it seems to fix the error and work as expected

### DIFF
--- a/sanic/base.py
+++ b/sanic/base.py
@@ -16,9 +16,7 @@ class Base(type):
             nonlocal init
             nonlocal name
 
-            bases = [
-                b for base in type(self).__bases__ for b in base.__bases__
-            ]
+            bases = BaseSanic.__bases__
 
             for base in bases:
                 base.__init__(self, *args, **kwargs)
--- a/sanic/base.py
+++ b/sanic/base.py
@@ -8,35 +8,18 @@ from sanic.mixins.routes import RouteMixin
 from sanic.mixins.signals import SignalMixin
 
 
-class Base(type):
-    def __new__(cls, name, bases, attrs):
-        init = attrs.get("__init__")
-
-        def __init__(self, *args, **kwargs):
-            nonlocal init
-            nonlocal name
-
-            bases = BaseSanic.__bases__
-
-            for base in bases:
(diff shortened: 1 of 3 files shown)


artcg:
Would also add a test case for this if subclassing `Sanic` is something that is officially supported...

ahopkins:
@artcg Thanks. Can you add a test? You can drop it at the end of `test_app.py`. It is something that I know is done, so we should have that in the test suite.

sanic/base.py
```diff
@@ -16,9 +16,7 @@ def __init__(self, *args, **kwargs):
             nonlocal init
             nonlocal name
 
-            bases = [
-                b for base in type(self).__bases__ for b in base.__bases__
-            ]
+            bases = BaseSanic.__bases__
 
             for base in bases:
```
ahopkins, line 21: Since we no longer need the comprehension, no need to have `bases` as a variable.

```python
for base in BaseSanic.__bases__:
```

ahopkins (changes requested):
This works and makes sense. Nice catch.

But, I want to play with this a little more because I think it might not yet cover all the edge cases.

Let's add the test, and I will swing back on this tomorrow with some more thoughts.

artcg:
Is it an optimisation not just to define `BaseSanic.__init__ ` directly?

ahopkins:
I am going to push a change to this branch and merge it to get rid of the metaclass here.

ahopkins:
I am pushing a change directly to this branch to remove the metaclass.

ahopkins (approved):

PR #2074, opened 2021-03-21
Perf improv

ahopkins:
I was playing around with some benchmarking and started to notice that we were not realizing the performance bump that I was expecting.

`v20.12`
```
Running 10s test @ http://localhost:3000
  12 threads and 400 connections
    Latency     3.60ms    3.62ms  77.03ms   92.43%
    Req/Sec    10.73k     2.71k   26.83k    79.17%
  1281439 requests in 10.05s, 144.20MB read
Requests/sec: 127460.90
Transfer/sec:     14.34MB
```

`master`
```
Running 10s test @ http://localhost:3000
  12 threads and 400 connections
    Latency     3.63ms    4.25ms  87.71ms   92.34%
    Req/Sec    11.33k     2.63k   52.29k    84.13%
  1343733 requests in 10.08s, 131.99MB read
Requests/sec: 133339.69
Transfer/sec:     13.10MB
```

I realized that this is mainly due to some minor issues in `handle_request`. With these small changes:

`perf-improv`
```
Running 10s test @ http://localhost:3000
  12 threads and 400 connections
    Latency     3.58ms    4.12ms  88.78ms   92.44%
    Req/Sec    11.49k     2.40k   50.72k    77.39%
  1376168 requests in 10.10s, 135.18MB read
Requests/sec: 136254.12
Transfer/sec:     13.38MB
```

There are some more gains to be had here, but not until 21.6 and a more thorough overhaul of this method.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -752,9 +752,9 @@ class Sanic(BaseSanic):
     ):
         request.app = self
         if not getattr(handler, "__blueprintname__", False):
-            request.endpoint = handler.__name__
+            request._name = handler.__name__
         else:
-            request.endpoint = (
+            request._name = (
                 getattr(handler, "__blueprintname__", "") + handler.__name__
             )
 
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -676,27 +676,23 @@ class Sanic(BaseSanic):
         response = None
         try:
             # Fetch handler from router
-            (
-                route,
-                handler,
-                kwargs,
-            ) = self.router.get(request)
+            route, handler, kwargs = self.router.get(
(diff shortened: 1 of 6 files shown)


ashleysommer (approved):
Looks like a good change, makes the code easier to follow and is a bit faster.

I still want to look further into some performance comparisons, there's some other things I want to check out (eg, last time I tested the streaming server changes, the performance was exactly the same between uvloop and plain asyncio).

ahopkins:
Agreed. I think a large focus for the next three months is to scrutinize every line that is executed in the request/response cycle.

PR #2077, opened 2021-03-21
deprecate abort

artcg:
non-internal changes:

- `SanicException` allows to be constructed without arguments

this is more in-line with other builtin python Exceptions, I was surprised that "message" was required before, and think users would be too.

- add_status_code decorator is removed straight away

I didn't think a warning was necessary here since really nobody should be using this, but can add if desired...

- abort now raises SanicException with status_code instead of the appropriate subtype

Since abort is slated for removal I am not too worried about this, but can add back old behaviour during transition also if it's desired


ahopkins (changes requested):
Looks good. I am fine with these changes/deprecations as is.

Let's add a test case to catch the `abort` warning. If you `grep` I am sure you can find some other instances to copy.

Also, one tests that implement both the fallback message and a custom message. Looks like custom `status_code` is already covered.

artcg:
Sounds good,

btw this is completely wrong place to discuss but just curious, whats the plan with sanic-routing? Keep it as a separate repo to sanic or bring it back in? Is the disconnect between versioning for sanic and sanic-routing a feature?

ahopkins:
There is an ongoing conversation. It has not been decided. There might be some utility to keeping it seperate, but I suspect we will have a definitive answer by the Summer.

sanic/exceptions.py
```diff
@@ -3,26 +3,14 @@
...
-    def class_decorator(cls):
-        cls.status_code = code
-        if quiet or quiet is None and code != 500:
-            cls.quiet = True
-        _sanic_exceptions[code] = cls
-        return cls
-
-    return class_decorator
+class SanicException(Exception):
+    def __init__(self, message=None, status_code=None, quiet=None):
```
ahopkins, line 7: Can you add some type annotations here?

ahopkins (changes requested):
One more change.

miss85246:
looks interesting

sanic/exceptions.py
```diff
@@ -3,26 +3,19 @@
...
+    def __init__(
+        self,
+        message: Optional[Union[str, bytes]] = None,
+        status_code: Optional[int] = None,
+        quiet: Optional[bool] = None,
+    ) -> None:
+
+        if message is None and status_code is not None:
+            msg: bytes = STATUS_CODES.get(status_code, b"")
+            # These are stored as bytes in the STATUS_CODES dict
```
ahopkins, line 16: You can remove this comment. It is not necessary.

ahopkins (changes requested):

ahopkins (approved):

dnikolayev:
Guys, you're crazy :)
in 21.3 the decorator exists(as in all previous since even 0.6 or 0.8.3) in 21.6 - no :)
No deprecation warning in the documentation  :(

ahopkins:
> Guys, you're crazy :)
> in 21.3 the decorator exists(as in all previous since even 0.6 or 0.8.3) in 21.6 - no :)
> No deprecation warning in the documentation :(

What decorator `@add_status_code`?

dnikolayev:
Sorry, but I was using this decorator for ages :)
So, I had questions about the backward compatibility of [webargs-sanic](https://github.com/EndurantDevs/webargs-sanic). 
For instance, I know about the 0.8 or 18.12 usages still with newer and older form-data validation.
Now all is good with Sanic 0.8.3 to 21.6.2: https://app.travis-ci.com/github/EndurantDevs/webargs-sanic/builds/237030388

Thanks!

P.s. The only thing is harder to support now - changes in the testing part, but I will figure out something (sanic-testing only from 21th versions, etc.)

ahopkins:
@dnikolayev Thank you very much for bringing this up. We are really trying to narrow in on both the public and private API to get the project (hopefully) to a "stable" classifier in the next LTS. I have been trying very hard to make sure we stick to backwards compat support thru at least a couple versions for any deprecations. I apologize if this one slipped through the cracks.

PR #2083, opened 2021-03-22
review: use plural for MiddlewareMixin._future_middlewares

laggardkernel:
Convert `MiddlewareMixin._future_middleware` to `MiddlewareMixin._future_middlewares`. Using the plural make it consistent with variable names in other Mixins.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -83,7 +83,7 @@ class Sanic(BaseSanic):
         "_blueprint_order",
         "_future_routes",
         "_future_statics",
-        "_future_middleware",
+        "_future_middlewares",
         "_future_listeners",
         "_future_exceptions",
         "_future_signals",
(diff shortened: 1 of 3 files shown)


ahopkins:
Thank you @laggardkernel for your contribution. However, it is intentionally `middleware` as that word can (and does) operate as both singular and plural.


Switching to `middlewares` does not sound correct in English and would be non-intuitive like: `softwares`, `hardwares`, `silverwares`, etc.



https://english.stackexchange.com/questions/257120/middleware-vs-middlewares

ashleysommer:
I've written software in the past that uses "middlewares" and others that use "middleware" as the plural. Sanic has always used the "middleware" naming to mean both the singular and the plural. Confusing, I know, but it does sound more correct in English to use "middleware" as the plural.

PR #2085, opened 2021-03-22
Make sure that blueprints with no slash is maintained when applied

### LINKED ISSUE #2080: bug: route without a trailing slash gets routed to route with a trailing slash, when the route is the top level of blueprint (maybe also app)
Version: 21.3.1

Description:

setting two urls in a blueprint with strict slashes for "" and "/" will raise an exception "route already registered" in sanic 21.3.1

In sanic version before 21.3 this would be acceptable

Code to reproduce:

```python
from sanic import Sanic, Blueprint

app = Sanic("test")

bp = Blueprint("test", url_prefix="test")
@bp.get("", strict_slashes=True)
def _(req):
    return text("a")

@bp.get("/", strict_slashes=True)  # causes exception on sanic 21.3
def _(req):
    return text("b")

app.blueprint(bp)
app.run()
```

potential fix: https://github.com/sanic-org/sanic/pull/2079

ahopkins:
Resolves #2080
Replaces #2079

### DIFF
--- a/tests/test_routes.py
+++ b/tests/test_routes.py
@@ -12,7 +12,7 @@ from sanic_routing.exceptions import (
 )
 from sanic_testing.testing import SanicTestClient
 
-from sanic import Blueprint, Sanic
+from sanic import Blueprint, Sanic, response
 from sanic.constants import HTTP_METHODS
 from sanic.exceptions import NotFound, SanicException
 from sanic.request import Request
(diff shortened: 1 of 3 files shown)


tests/test_routes.py
```diff
@@ -12,7 +12,7 @@
 )
 from sanic_testing.testing import SanicTestClient
 
-from sanic import Blueprint, Sanic
+from sanic import Blueprint, Sanic, response
```
yunstanford, line 15: nit: redundant import ?

yunstanford (dismissed):

tests/test_routes.py
ahopkins: Thanks. Looks like VS Code was being a little over zealous on me.

ashleysommer:
Not sure if this covers the condition when BP's (`url_prefix == ""` or `url_prefix is None`) and `url == ""`.

ahopkins:
```python
bp = Blueprint("test")

@bp.get("")
def bar(req):
    return text("bar")
```

```
$ curl localhost:8888/ localhost:8888
bar
bar
```

Issue #2092, opened 2021-03-28
Flaky Tests


harshanarayana:
**Describe the bug**
There are a series of tests in our code base which seem to be behaving in a bit of a flaky manner on regular basic. This has been one of the main reason we have seen so many random Travis CI failures while running tests. 

Currently, the main one seem to be isolated to `py39-no-ext`

While enabling GitHub Actions for the repo using #2050 I noticed that there are a few more flaky tests and the behavior is a little more prominent and visible when you extend your test matrix to include `macos-latest` or `windows-latest`

It would be nice if we can audit these tests and fix some of them or mark them aptly as `@pytest.mark.xfail` with Proper conditional selector to make sure we don't run into random PR patchest failures on regular basis

**Code snippet**
NA

**Expected behavior**
Lesser Flaky Tests

**Environment (please complete the following information):**
NA

**Additional context**
Check the PR commit history for #2050. You will notice a series of failures of tests and most of those are flaky test failures that work on retires or shows up randomly.

ahopkins:
It seems to always get stuck (if it will get stuck) on this:

```
$ tox -e py39-no-ext -- tests/test_signal_handlers.py
```

But, I cannot recreate this locally.

harshanarayana:
@ahopkins  Yeah. I haven't had any success either. Tried on my Personal mac as well as the Ubuntu machines we use at work. Can't recreate at will but the pipeline seem to get stuck in this loop way more than I would like it to.

PR #2097, opened 2021-04-04
Allow case-insensitive HTTP Upgrade header

ENT8R:
As far as I understand [Section 11.2 of RFC 6455](https://tools.ietf.org/html/rfc6455#section-11.2) ("The WebSocket Protocol"), the HTTP Upgrade header should actually be `WebSocket`. At least Firefox uses the header like that, which leads to a wrong detection of the underlying protocol in Sanic (e.g. timeouts after a minute because Sanic thinks the request is a normal HTTP request). This is why I propose to make the check for the Upgrade header case-insensitive in order to correctly handle the incoming request.


ashleysommer (commented):
The scheme check: `headers` is not defined, you should use `self.headers`.

ashleysommer:
Looks great. I think it might've been done this way (with just lowercase "websockets" because the `Headers` Class is a CIMultidict, meaning we don't normally need to worry about case. But if course that only applies to the key names, not the values. So when checking for specific values, we do need to convert case.

Thanks for this fix. I'm surprised we havent had reports of broken websockets in Sanic v21.3 due to this.

ashleysommer (dismissed):

ashleysommer:
Ah, type checking has picked up a valid error case.

`headers.get('upgrade')` could return None, and you can't do `.lower()` on None.

I know it's a pain and an extra step, but you probably need to put a `and headers.get('upgrade', None) is not None` in there too.

ahopkins:
I agree. I am surprised as well since this has certainly been tested and out in usage already. Which kind of makes me curious as to why we haven't seen this already. Ideally I'd also like to have a test coverage for it, but that might be hard to do with our current test client.

ENT8R:
Thank you for the fast response and quick review.
At least in my case, the code continued to work, but I was seeing lots of timeouts and websocket reconnections every 60 seconds. The behavior was definitely not wanted but didn't seem to cause other effects besides the reconnections. Also, if I wouldn't have checked the logs, I would have probably never seen this bug...

harshanarayana:
Interesting. I did a bit of digging. 

1. [Mozilla Dev](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Upgrade)
2. [RFC](https://tools.ietf.org/html/rfc6455#section-1.2)

In the `RFC` the sample always shows 

```bash
        GET /chat HTTP/1.1
        Host: server.example.com
        Upgrade: websocket
        Connection: Upgrade
        Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==
        Origin: http://example.com
        Sec-WebSocket-Protocol: chat, superchat
        Sec-WebSocket-Version: 13
```
So does the documentation of the Mozilla Dev page. 

Running a quick inspect on Websocket using Firefox indicates the same. 

Screenshot 2021-04-04 at 9 11 52 PM

sanic/http.py
```diff
@@ -218,7 +218,7 @@ async def http1_request_header(self):
             raise InvalidUsage("Bad Request")
 
         headers_instance = Header(headers)
-        self.upgrade_websocket = headers_instance.get("upgrade") == "websocket"
+        self.upgrade_websocket = "upgrade" in headers_instance and headers_instance.get("upgrade", None) is not None and headers_instance.get("upgrade").lower() == "websocket"
```
harshanarayana, line 221: If we really want to do this check the easier might be 

```suggestion
        self.upgrade_websocket =  headers_instance.get("upgrade", "").lower() == "websocket"
```

harshanarayana (changes requested):

sanic/http.py
ahopkins: I like this better

ashleysommer:
I don't know whats going on with those travis tests. It should be passing now. I'll try to restart them.

ashleysommer:
Nope, it won't let me restart it. (Error: "Oops, this Job cannot be Restarted.")

I think we can just merge this regardless.

On a different note, for @sanic-org/core-devs 
I'm wondering if we should try to avoid using `headers.get('key')` in our codebase.
In the CIMultidict that holds our headers, each key can have multiple values. Thats why the `.getone()` and `.getall()` methods exist. From the Docs, the `.get()` method simply gets the first value, and has implicit `None` as default. Doing `.get('key')` is the same as `.getone('key', None)`. I wonder if we should be more explicit with the intention in the code and change all occurances of `.get('key')` to `.getone('key', None)`.

harshanarayana:
> I don't know whats going on with those travis tests. It should be passing now. I'll try to restart them.

Those tests I've been really flaky lately. Especially the `py3{8,9}-no-ext` and like you noticed, Travis restart tests aren't working anymore. 😥

harshanarayana (approved):
Approving this change. But I do like the idea of using `getone` as suggested by @ashleysommer.. May be it can be a total cleanup work across the board using a different issue item?

ahopkins (changes requested):
Looks like there is no new line at the end of the files. It should be failing linting. Can you run `make pretty` to make sure that it complies with `black`. You can also run `tox -e lint` to make sure that everything is formatted properly.

ahopkins:
+1 the explicit usage of `getone` and `getall` as a separate PR across the board.

ENT8R:
> Looks like there is no new line at the end of the files

Are you sure that this is the actual cause of the failing tests? As far as I see, there is generally no new line at the end of files in this project. I feel like this is more related to #2092 because it's only `py39-no-ext` which fails to build... I ran `make pretty` and the only complaint was related to the line length. Let's see what Travis thinks about it 🙃

ENT8R:
Looks like Travis has no complaints this time 🎉

harshanarayana (approved):

ahopkins (approved):

harshanarayana:
> +1 the explicit usage of `getone` and `getall` as a separate PR across the board.

Added a tracker item for this

ahopkins:
> > +1 the explicit usage of `getone` and `getall` as a separate PR across the board.
> 
> Added a tracker item for this

#2098

PR #2099, opened 2021-04-05
fixing static request handler logging exception on 404

fredlllll:


Tronic:
I gather this was throwing and logging 500 errors for all file opening problems? 404 seems like a good improvement but perhaps 403 for PermissionError should also be added, that being a quite common issue with files (there are a number of other OSErrors that could be risen too but that seldom occur).

ahopkins:
@Tronic this stemmed from a conversation on the Discord channel where @fredlllll was annoyed at full stack traces showing up in the logs on 404s for files that do not exist.

Even adding:

```python
@app.exception(FileNotFound)
async def file_not_found(request, exception):
    return text(f"No file: {request.path}", status=404)
```

The exception would still log because of `error_logger.exception`. This way, the `FileNotFound` will log out if the exception is not handled. Or, something like the above could be used to catch it.

Tronic:
That's what I thought. I've been annoyed by the same problem since 404s may frequently occur even though there is nothing wrong with the app (this was already fixed for routing-related 404 errors last year if I recall correctly).

ahopkins (approved):

PR #2104, opened 2021-04-07
Explicit usage of CIMultiDict getters

### LINKED ISSUE #2098: Cleanup Usage of CIMultidict
While reviewing #2097, @ashleysommer pointed out how we have been using `header` which is a `CIMultidict` type. We have been using `header.get` instead of explicitly using either `getone` or `getall`

In order to keep the original PR clean, we are opening a new tracker item to make sure we can cleanup the usage of this across the board in `sanic` codebase.

1. Replace usage of `get` with a more suitable appropriate replacement of `getone` or `getall`

ENT8R:
Fixes #2098
- Changed all occurrences of `get(key)` to `getone(key, None)`
- Changed `self.headers = headers` to `self.headers = Header(headers)` in [`request.py#L128`](https://github.com/sanic-org/sanic/blob/1a352ddf559f62d40b6fbb34608fffd493db0ade/sanic/request.py#L128) (I don't know if this change is wanted, but it seemed logical to me as the type hint suggests that this parameter should be a Header class)

The tests are currently failing in the following snippet because multidict [expects the first parameter (i.e. the key) to be a string](https://github.com/aio-libs/multidict/blob/ea74a0d0287bc64020cc1e407253598fd956071f/multidict/_multidict_py.py#L193-L199). When running the tests, this value (`self.app.config.REQUEST_ID_HEADER`) is no string but instead an instance of `unittest.mock.Mock` which multidict doesn't know how to handle...
https://github.com/sanic-org/sanic/blob/16a320f268115d5ace586b76468e899c0f8f6bfd/sanic/request.py#L265-L268

### DIFF
--- a/sanic/mixins/routes.py
+++ b/sanic/mixins/routes.py
@@ -666,7 +666,7 @@ class RouteMixin:
                     "%a, %d %b %Y %H:%M:%S GMT", gmtime(stats.st_mtime)
                 )
                 if (
-                    request.headers.getone("If-Modified-Since", None)
+                    request.headers.getone("if-modified-since", None)
                     == modified_since
                 ):
                     return HTTPResponse(status=304)
--- a/sanic/mixins/routes.py
+++ b/sanic/mixins/routes.py
@@ -665,7 +665,10 @@ class RouteMixin:
                 modified_since = strftime(
                     "%a, %d %b %Y %H:%M:%S GMT", gmtime(stats.st_mtime)
                 )
-                if request.headers.get("If-Modified-Since") == modified_since:
+                if (
+                    request.headers.getone("If-Modified-Since", None)
+                    == modified_since
+                ):
                     return HTTPResponse(status=304)
                 headers["Last-Modified"] = modified_since
             _range = None
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -673,7 +673,9 @@ class Sanic(BaseSanic):
         try:
             # Fetch handler from router
             route, handler, kwargs = self.router.get(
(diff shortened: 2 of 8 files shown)


sanic/mixins/routes.py
```diff
@@ -665,7 +665,10 @@ async def _static_request_handler(
                 modified_since = strftime(
                     "%a, %d %b %Y %H:%M:%S GMT", gmtime(stats.st_mtime)
                 )
-                if request.headers.get("If-Modified-Since") == modified_since:
+                if (
+                    request.headers.getone("If-Modified-Since", None)
```
ahopkins, line 669: While you are at it, maybe we should be consistent. Sometimes when we access a header we ask for it in title-case. Sometimes lower case. I think it would be nice to just change all these to lower case strings.

ahopkins (changes requested):

ahopkins:
> The tests are currently failing in the following snippet because multidict expects the first parameter (i.e. the key) to be a string. When running the tests, this value (self.app.config.REQUEST_ID_HEADER) is no string but instead an instance of unittest.mock.Mock which multidict doesn't know how to handle...

Can you fix this test to use a string then for the `REQUEST_ID_HEADER`?

ENT8R:
BTW: There are still some usages of `get()` instead of `getone()` in the tests. Do you want them to be changed too?

ahopkins:
Not too concerned with those since it is a closed environment.

ENT8R:
Alright, then I will leave them the way they are currently

ahopkins (approved):

PR #2107, opened 2021-04-09
Fix type hinting for load_env

jdraymon:
Type hint for load_env shows as a boolean, but it's documented and used as a `Union[bool, str]`. Update hinting to match

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -43,7 +43,7 @@ from sanic.asgi import ASGIApp
 from sanic.base import BaseSanic
 from sanic.blueprint_group import BlueprintGroup
 from sanic.blueprints import Blueprint
-from sanic.config import BASE_LOGO, Config
+from sanic.config import BASE_LOGO, SANIC_PREFIX, Config
 from sanic.exceptions import (
     InvalidUsage,
     SanicException,
@@ -125,7 +125,8 @@ class Sanic(BaseSanic):
         router: Optional[Router] = None,
         signal_router: Optional[SignalRouter] = None,
         error_handler: Optional[ErrorHandler] = None,
-        load_env: bool = True,
+        load_env: Union[bool, str] = True,
+        env_prefix: Optional[str] = SANIC_PREFIX,
         request_class: Optional[Type[Request]] = None,
         strict_slashes: bool = False,
         log_config: Optional[Dict[str, Any]] = None,
@@ -150,7 +151,7 @@ class Sanic(BaseSanic):
         self._test_manager = None
         self.asgi = False
         self.blueprints: Dict[str, Blueprint] = {}
-        self.config = Config(load_env=load_env)
+        self.config = Config(load_env=load_env, env_prefix=env_prefix)
         self.configure_logging = configure_logging
         self.ctx = SimpleNamespace()
         self.debug = None
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -1,7 +1,8 @@
 from inspect import isclass
 from os import environ
 from pathlib import Path
-from typing import Any, Union
+from typing import Any, Dict, Optional, Union
+from warnings import warn
 
 from .utils import load_module_from_file_location, str_to_bool
 
@@ -41,7 +42,13 @@ DEFAULT_CONFIG = {
 
 
 class Config(dict):
-    def __init__(self, defaults=None, load_env=True, keep_alive=None):
+    def __init__(
+        self,
+        defaults: Dict[str, Union[str, bool, int, float, None]] = None,
+        load_env: Optional[Union[bool, str]] = True,
+        env_prefix: Optional[str] = SANIC_PREFIX,
+        keep_alive: Optional[int] = None,
+    ):
         defaults = defaults or {}
         super().__init__({**DEFAULT_CONFIG, **defaults})
 
@@ -50,9 +57,20 @@ class Config(dict):
         if keep_alive is not None:
             self.KEEP_ALIVE = keep_alive
 
-        if load_env:
-            prefix = SANIC_PREFIX if load_env is True else load_env
-            self.load_environment_vars(prefix=prefix)
+        if env_prefix != SANIC_PREFIX:
+            if env_prefix:
+                self.load_environment_vars(env_prefix)
+        elif load_env is not True:
+            if load_env:
+                self.load_environment_vars(prefix=load_env)
+            warn(
+                "Use of load_env is deprecated and will be removed in "
+                "21.12. Modify the configuration prefix by passing "
+                "env_prefix instead.",
+                DeprecationWarning,
+            )
+        else:
+            self.load_environment_vars(SANIC_PREFIX)
 
     def __getattr__(self, attr):
         try:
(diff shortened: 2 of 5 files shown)


jdraymon:
Note - this was done on a whim in the github editor. I'll come back in a bit and do a proper check later with an actual repo clone

ahopkins (changes requested):
This is correct, since we do accept either a `bool` or a `str`. To be honest, I would rather this get fixed properly and _only_ be `str` with `SANIC_PREFIX` as its default.

```python
class Config(dict):
    def __init__(
        self,
        defaults: Dict[str. Union[str, bool, int, float, None]] = None,
        load_env: Optional[str] = SANIC_PREFIX,
        keep_alive: Optional[int] = None
    ):
```

Then, the `load_env` in `Sanic` should match.

jdraymon:
I can do that... I wasn't sure about introducing a breaking change as a non core contributor... Is there anywhere this needs to be discussed?

jdraymon:
@ahopkins it looks as though the reason this accepts a bool is that you disable loading from the environment entirely if the value is `False` - How would we expect this to work now? I see a few options:

* Pass `None` for `load_env` to disable config loading
  * Not as semantically clear as passing `False` to  `load_env`
  * Maybe explicitly change the name to `env_prefix`?
* Add a separate, `env_prefix` parameter, keeping `load_env` with its boolean meaning
  * Adds clarity by demuxing the option
  * Bloats app parameter list
* Always load from environment, using `SANIC_PREFIX` if `None` is passed through
  * Removes currently available behaviour
  
  Thoughts?

ahopkins:
I like the suggestion of `env_prefix` as it evokes consistency with `url_prefix`. Let's do a proper depreciation for `load_env`. Raise a UserWarning if it is not `True`.

The warning should state that the feature will be removed in 21.12.

sanic/app.py
```diff
@@ -481,9 +475,7 @@ def url_for(self, view_name: str, **kwargs):
 
         route = self.router.find_route_by_view_name(view_name, **kw)
         if not route:
-            raise URLBuildError(
-                f"Endpoint with name `{view_name}` was not found"
-            )
+            raise URLBuildError(f"Endpoint with name `{view_name}` was not found")
```
jdraymon, line 478: Looks like my editor didn't read in the project's black/flake configuration properly. Will fix

jdraymon:
Sorry for all the FPs - looks happy now

sanic/app.py
```diff
@@ -125,7 +125,8 @@ def __init__(
         router: Optional[Router] = None,
         signal_router: Optional[SignalRouter] = None,
         error_handler: Optional[ErrorHandler] = None,
-        load_env: bool = True,
+        load_env: Union[bool, str] = True,
+        env_prefix: str = SANIC_PREFIX,
```
ahopkins, line 129: Looks good. I'll take a closer look at the tests when I'm at my computer.

Quick question about this though, what about making this `Optional`? To be able to pass `None` to skip environment loading. I'm honestly not sure the use case of when you would have env variables that you would explicitly not want, but we have that functionality now, so we would be removing it.

tests/test_config.py
```diff
@@ -80,6 +80,11 @@ def test_dont_load_env():
     del environ["SANIC_TEST_ANSWER"]
 
 
+def test_load_env_deprecation():
+    with pytest.warns(DeprecationWarning, match=r"21\.12"):
+        _ = Sanic(name=__name__, load_env=False)
```
ahopkins, line 85: I would `pytest.mark.parameterize` this to test with `False` and also a string: `FOO_`

sanic/config.py
```diff
@@ -50,9 +57,17 @@ def __init__(self, defaults=None, load_env=True, keep_alive=None):
...
+        if load_env is not True:
+            if load_env:
+                self.load_environment_vars(prefix=load_env)
+            warn(
+                "Use of load_env is deprecated and will be removed in "
+                "21.12. Modify the configuration prefix by passing "
+                "env_prefix instead.",
+                DeprecationWarning,
+            )
+        else:
```
ahopkins, line 69: I think this should be:

```python
elif env_prefix:
```

ahopkins: And then add one more test with:

```python
Sanic(..., env_prefix=None)
```

where the variable is not loaded. You probably need to add a catch for this in the above line if they are still using `load_env` though.

OR, if `env_prefix != SANIC_PREFIX` then skip the whole `load_env` part altogether.

ahopkins (changes requested):

sanic/app.py
jdraymon: > Looks good. I'll take a closer look at the tests when I'm at my computer.
> 
> Quick question about this though, what about making this `Optional`? To be able to pass `None` to skip environment loading. I'm honestly not sure the use case of when you would have env variables that you would explicitly not want, but we have that functionality now, so we would be removing it.

Seems ok. In my own code i've been trying to avoid using `None` as a meaningful value, but the semantics here are probably clear enough where it makes sense and makes for less of a breaking change

ahopkins: Point taken. But, it is better than having two values.

jdraymon: Explicitly `None` or just false-y?

ahopkins: ooof.... this is getting more complicated than I was originally thinking. 

My thought had been falsey. But, what if you want to import all env, and therefore an empty string?

The more I think about it  though, we probably should not support that. Explicit env variables are better than just taking everything.

ahopkins: Yeah, I agree. If you _really_ want everything, then I think that's a case for a roll your own solution to loop thru `os.environ`.

jdraymon:
side note - is it objectionable to add things to the `.gitignore`?

sanic/config.py
jdraymon: :heavy_check_mark:

tests/test_config.py
jdraymon: :heavy_check_mark:

ahopkins:
> side note - is it objectionable to add things to the `.gitignore`?

Personal identifying information: yes

Some missing patterns: no

jdraymon:
Test failure seems unrelated - is that test environment generally flaky?

sjsadowski:
It's just the one macOS test, and macOS has been pretty consistently inconsistent with py38 - not sure why. I'd say we're probably safe but I'm trying to find time to deep dive on why the macOS py38 tests are having issues.

jdraymon:
My first commit went through just fine - I can drop the `.gitignore` commit if you would like to play it safer

ahopkins:
No worries. Leave it as is.

sjsadowski:
I think the .gitignore is fine, it's some weird interaction (mostly) with macOS/py38/gunicorn though there was the one timeout for large request size. I generally think you're fine.

ahopkins (approved):

ahopkins:
Looks good. The GH actions are not yet set as required to pass. We're still experimenting with them. Travis is still the go to for now.

Thanks for the work on this PR.

jdraymon:
How are the docs updated? Should I PR that as well or is there a translation process?

ahopkins:
Actually, we should have added something in the PR here for the API docs.

For these docs: https://sanicframework.org/, don't worry about that now. We will add that later.

PR #2110, opened 2021-04-10
Fix request.args.pop removes parameters inconsistently

sjsadowski:
Per the issue in subject, should fix request.args.pop inconsistency.

When merged closes #2106

Issue #2106, opened 2021-04-09
request.args.pop removes parameters inconsistently


artcg:
Environment : 20.12.3, presumably other environments too

Code to reproduce:

send this request to this app
`GET http://localhost:8000/?hello=a&world=4`
```python
from sanic import Sanic


app = Sanic('test')

@app.get('/')
def _(request):
    a = request.args.pop('hello')
    b = request.args.pop('world')
    c = request.args.pop('world')
    d = request.args.pop('world')   << crashes here

app.run()
```

Expected:
crashes at line second call of .pop('world'), or not at all

artcg:
I have no ideas about this since the RequestParameters seems to be a really straightforward wrapper around the `dict`, would be interested if anyone can reproduce it ...

artcg:
Started to look at this a bit more it seems that the problem is not with the implementation of RequestParameters but rather with this somewhere

`    args = property(get_args)`

vs 18.12 where it still works as expected

```
    @property
    def args(self):
        if self.parsed_args is None:
            if self.query_string:
                self.parsed_args = RequestParameters(
                    parse_qs(self.query_string)
                )
            else:
                self.parsed_args = RequestParameters()
        return self.parsed_args
```

artcg:
Ah I worked it out!

Because parsed_args is a defaultdict, once the request.args is exhausted, it will evaluate as False again, and then a fresh parsing of the query will occur.

sjsadowski:
Reopening until merge is complete

PR #2112, opened 2021-04-12
(20.12 LTS backport) fix request.args.pop removing parameters inconsistently 

artcg:


ahopkins:
See #2106 and #2110

ashleysommer:
Looks good.
Is there any other recent fixes from 21.3 or master that we should push back into 20.12 to make a new LTS release?

ahopkins:
None. Since this has been out there for at least 2 years, my thought was to roll LTS updates with the next release.

ashleysommer:
Submitter confirmed that https://github.com/sanic-org/sanic/issues/2113 affects 20.12 too, so when a fix for that comes we should roll that into LTS too.

ahopkins (approved):

Issue #1901, opened 2020-07-30
IPV6 address wrapped in brackets


ahopkins:
@Tronic why are we wrapping IPV6 addresses with square brackets here?

```python
class ConnInfo:
    ...
    def __init__(self, transport, unix=None):
        ...
        # IPv4 (ip, port) or IPv6 (ip, port, flowinfo, scopeid)
        ...
        if isinstance(addr, tuple):
            self.client = addr[0] if len(addr) == 2 else f"[{addr[0]}]"
            self.client_port = addr[1]
```
[source](https://github.com/huge-success/sanic/blob/0072fd1573d43c64a4a2b9b89b4e4f887bc07a70/sanic/server.py#L81)

Tronic:
Because in URLs and in `Forwarded` headers they are wrapped.

sjsadowski:
Also it's required as part of RFC 3986 when used in conjunction with other parts such as user:pass and :port as both use colon designators and could look like hex refs.

ahopkins:
Well, then someone ought to tell Postgres.

```
update some_table set ip_addr='[::ffff:172.20.0.1]';                                                                                           
invalid input syntax for type inet: "[::ffff:172.20.0.1]"
LINE 1: update some_table set ip_addr='[::ffff:172.20...
                                                     ^
```

I think maybe we should make it somewhat easier to get at the raw value than: `request.conn_info.sockname[0]`.

sjsadowski:
The brackets are only valid as part of the address when there's a prefix or suffix as part of a URI... IE stephen:password@[::ff] or [::ff]:8080

The address itself should never contain the brackets. Think of it like this, using an f-stringINET_ADDR 

`f'https://{prefix}[{ipv6_addr}]:{port}'`

Technically postgres is correct because you shouldn't be storing the the entire uri as an inet type, you'd do it in a char type field., same as you wouldn't store 172.20.0.1:443 in an inet column, as postgres would reject that.


Hope this helps.

ahopkins:
Right, that makes sense. but, given that Sanic is setting these properties, it looks like Sanic is wrong, and to get the actual value from the tuple yourself is not intuitive. Either, we should add another top level property to request object, or only do the wrapping when needed. I'm particularly not even sure that Sanic should do it at all except when building an address maybe. Grabbing request.ip for DB insertion seems common enough pattern that it shouldn't appear as a gotcha.

Tronic:
[...]

PR #2114, opened 2021-04-14
Added new client_ip accessor 

sjsadowski:
Added new client_ip accessor for ConnInfo class, updated request to use client_ip instead of client to be more representative of what will be returned (actual ipv6 ip instead of bracket wrapped ip)


sjsadowski:
This will close #1901 when merged

ahopkins:
Nice. Some tests when you have a chance? 😎

sjsadowski:
@ahopkins that's why it's still a draft, need to make sure test_client will do what I want.

ahopkins:
😆 totally didn't see that. sorry. I was on my phone.

Might need to mock it. IIRC, I had some problems in the past. But, that might have been ASGI client.

Tronic (approved):
LGTM. Leaving discussion of other necessary work on bracket handling and client IPs elsewhere (I don't believe that socket level client IP is very useful; usually you want true remote IP instead).

sjsadowski:
Test client doesn't support ipv6, looks like. Going to have to come at the tests from a different direction... probably mocking like @ahopkins suggested.

ahopkins:
> Test client doesn't support ipv6, looks like. Going to have to come at the tests from a different direction... probably mocking like @ahopkins suggested.

I'll spend some time looking at the test client tonight. It should be able to.

ahopkins (approved):

PR #2117, opened 2021-04-17
Implement 0.6 routing and some cleanup

### LINKED ISSUE #2089: put request returns error 405
with sanic version 20.12 the below code worked perfect when using a put statement following the first route definition.
(for completeness i added all the relevant routes)

```python3

@app_he.route("/api/<user_id>/lights/<id_num:string>/state", methods=["PUT"])
async def hue_api_put_light(request, user_id, id_num):
    """light change state request"""
    ...
    return resp

@app_he.route("/api/<user_id>/lights/<id_num:string>", methods=["GET"])
async def hue_api_individual_light(request, user_id, id_num):
    """individual device state, probably called to make sure command went through"""
    ...
    return response.json(my_a.hue_light(id_num))

```

Installing the latest version of sanic gives the follow error when using a put statement:

```
2021/03/25 12:15:41 !!Sanic except error http://192.168.15.31/api/h632JhAtwDezSwGk3VKEeFn4snXZUj1541885352/lights/bed%20light/state, Method PUT not allowed for URL /api/h632JhAtwDezSwGk3VKEeFn4snXZUj1541885352/lights/bed%20light/state
stack=/home/pi/lucy/lucy_app/base/logger.py@115,/home/pi/lucy/lucy_app/base/mail_print.py@266,/home/pi/lucy/lucy_app/base/pi_utils.py@534,/home/pi/lucy/lucy_app/base/pi_utils.py@427,/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/handlers.py@94,/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/app.py@624,/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/app.py@748,/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/http.py@125,/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/server.py@198,/home/pi/.pyenv/versions/3.9.1/lib/python3.9/asyncio/runners.py@44,/home/pi/lucy/lucy_app/base/tasker.py@278,/home/pi/lucy/main.py@10
  File "/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic_routing/route.py", line 84, in get_handler
    return self.handlers[raw_path][method][idx]
IndexError: list index out of range

  File "/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/router.py", line 33, in _get
    return self.resolve(
  File "/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic_routing/router.py", line 84, in resolve
    handler = route.get_handler(raw_path, method, handler_idx)
  File "/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic_routing/route.py", line 86, in get_handler
    raise self.router.method_handler_exception(
sanic_routing.exceptions.NoMethod: Method 'PUT' not found on <Route: name=Hue_emul.hue_api_user path=api/<user_id>>

  File "/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/app.py", line 679, in handle_request
    route, handler, kwargs = self.router.get(
  File "/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/router.py", line 61, in get
    return self._get(path, method, host)
  File "/home/pi/.pyenv/versions/3.9.1/lib/python3.9/site-packages/sanic/router.py", line 41, in _get
    raise MethodNotSupported(
sanic.exceptions.MethodNotSupported: Method PUT not allowed for URL /api/h632JhAtwDezSwGk3VKEeFn4snXZUj1541885352/lights/bed%20light/state

```

when reverting to the older sanic version, with the below statement, everything works again as it should.
install "sanic>=20.12,<20.13" 

Is there an issue in the new routing script?

Kind Regards,
Rudy

### LINKED ISSUE #2090: 404 returned from endpoint when using regex in routes while registering more than 3 routes in some cases
I have some code in which a route used a regular expression to catch any text in the URL after a certain point. After the 21.3 update this stopped working and kept returning a 404 (not found) error. After trying many things I noticed that if I changed the regex to string (i.e., doing `<foo:string>`) the route suddenly worked again (interestingly using `<foo:path>` also doesn't work). Also, if I limit the number of routes I register to 3 or less the route works with the original regex expression. Below is some code that reproduces this.

**Update:**
It seems the issue stems from the route being missing from the routing tree that is generated. I am not sure how the tree is generated, but it seems that something about the route `/foo/bar/<path_param:.*>` causes it to not be added.

```python
from sanic import Sanic

app = Sanic(__name__)

@app.route('/foo/bar/<path_param:.*>', methods=["GET", "POST"])
async def foo_bar(request, path_param: str):
    return json({"path_param": path_param})

@app.route('/foo/lorep/<long_text:string>', methods=["GET", "POST"])
async def foo_lorep(request, long_text: str):
    return json({"path_param": long_text})

@app.route('/another/foo/bar/<path_param:string>', methods=["GET", "POST"])
async def another_foo_bar(request, path_param: str):
    return json({"path_param": path_param})

@app.route('/another/foo/bar/lorep/<path_param:string>', methods=["GET", "POST"])
async def another_foo_bar_lorep(request, path_param: str):
    return json({"path_param": path_param})

@app.route('/some/route1/<some_text:string>', methods=["GET", "POST"])
async def some_route1(request, some_text: str):
    return json({"path_param": some_text})

@app.route('/some/route2/<some_text:string>', methods=["GET", "POST"])
async def some_route2(request, some_text: str):
    return json({"path_param": some_text})

@app.route('/special/foo/bar/<path_param:string>', methods=["GET", "POST"])
async def special_foo_bar(request, path_param: str):
    return json({"path_param": path_param})

@app.get('/long/sub/route/param_a/<param_a:string>/param_b/<param_b:string>')
@app.post('/long/sub/route/')
async def long_sub_route(request, param_a: str, param_b: str):
    return json({"path_param": [param_a, param_b]})


if __name__ == "__main__":
    app.run(host="0.0.0.0", port=1234, debug=True)
```

Once the code above is running then navigating to `http://0.0.0.0:1234/foo/bar/123` should result in a `404`, but if you remove 5 routes (leaving only 3) the route will work **with** the regex. Alternatively, in the first route if `<input_text:.*>` is changed to `<input_text:string>` it will also work but this is not the ideal solution as I would like to be able to use regex in the routes.

**Environment:**
 - OS: Arch
 - Version: 
   - Sanic: 21.3.2
   - Sanic-Routing: 0.5.0

### LINKED ISSUE #2101: ASGI routes with body not extracted from the request
[See source](https://github.com/sanic-org/sanic/blob/824f41d6e06006076725d7be0140ebb4381440ff/sanic/asgi.py#L169)

This bug only applies to ASGI when sending.

When sending a body to a request on ASGI, the body is not making it through. I have not yet tested if this also applies with streaming requests.

The problem is here:

```python
    async def read(self) -> Optional[bytes]:
        """
        Read and stream the body in chunks from an incoming ASGI message.
        """
        message = await self.transport.receive()
        if not message.get("more_body", False):
            self.request_body = False
            return None         # <<<< we are not returning the body
        return message.get("body", b"")
```

If a small message comes through that has not hit this `read()` more than once (again, I am assuming that a streaming request might actually work), the message will look like this:

```python
message = await self.transport.receive()
# message={'type': 'http.request', 'body': b'{"foo": "bar"}', 'more_body': False}
```

Therefore, we need to return the `body` even if `more_body` is `False` (and also properly set `self.request_body`.

More testing is needed

### LINKED ISSUE #2103: Request body is not automatically loaded
I had to manually call `req.receive_body()` to have `req.body` and others filled in, in a *non-streaming* handler. Isn't `streaming=False` still the default? Let me know if I missed something obvious changed in 21.3, otherwise I'll debug more tomorrow.

### LINKED ISSUE #2116: Access logs not properly displaying body size
In certain circumstances, the access logger is displaying the wrong body size.

```python
@app.post("/")
async def post(request):
    print(request.json)
```
```
$ curl localhost:9999 -d '{"foo": "bar"}
```
```
{'foo': 'bar'}
[2021-04-18 01:43:33 +0300] - (sanic.access)[INFO][127.0.0.1:52172]: POST http://localhost:9999/  200 -1
```

In this example, it is being reported as `-1`

ahopkins:
Implements release from https://github.com/sanic-org/sanic-routing/pull/24

Fixes #2116 
Fixes #2103 
Fixes #2101
Fixes #2090


vltr (approved):
All good to go :sunglasses:

ahopkins:
Nice. Going to merge, then cherry pick the squash commit into new branch for 21.3.3

PR #2119, opened 2021-04-17
Alternatate classes on instantiation for Config and Sanic.ctx

### LINKED ISSUE #2115: Allow an alternate configuration class or object to be passed to application objects
It is currently difficult to extend the `Config` class, and have an `Sanic` instance actually use that configuration class throughout its entire lifecycle. This is because the `Sanic` class's `__init__` method is hard-coded to use `sanic.config.Config`. Anyone wishing to use a different class must do one of:
  - Patch and replace `sanic.config.Config`.
  - Re-implement `Sanic.__init__` in a sub-class, duplicating most of the base implementation.
  - Assign a different value to `Sanic.config`.

The first solution is inelegant, and hard to do correctly, the second involves redundant code duplication, and the third means that users are only able to introduce custom behavior post-init.

A simple fix for this is to modify `Sanic.__init__` to allow passing in a custom configuration class. This is already done for the class used to represent requests, and users can also pass in custom router and error handler instances.

ahopkins:
Resolves #2115


sanic/app.py
```diff
@@ -145,15 +147,23 @@ def __init__(
         if configure_logging:
             logging.config.dictConfig(log_config or LOGGING_CONFIG_DEFAULTS)
 
+        if config and load_env or env_prefix:
+            raise SanicException(
+                "When instantiating Sanic with config, you cannot also pass "
+                "load_env or env_prefix"
+            )
+
```
Tronic, line 155: I believe that the condition should be `config and (load_env or env_prefix)`

Tronic (changes requested):
Otherwise looks good.

Tronic (dismissed):

Tronic (approved):

PR #2127, opened 2021-04-26
Make sure ASGI ws subprotocols is a list

### LINKED ISSUE #2070: Sanic WebSockets not working in ASGI mode with daphne as a server
Sanic WebSockets not working in ASGI mode with daphne as a server

Backend code:
```
from sanic import Sanic

app = Sanic("bug")


@app.websocket("/ws/")
async def handler(request, ws):
    await ws.accept()
    while True:
        data = "Hello"
        print("Sending:", data)
        await ws.send(data)
        data = await ws.recv()
        print("Received: ", data)

```
Ran with `daphne -e tcp:80:interface=0.0.0.0 bug:app`

Frontend code:
```
socket = new WebSocket("ws://localhost:80/ws/", []);

socket.onopen = function(e) {
  console.log("Sending to server");
  socket.send("My name is John");
};

socket.onmessage = function(event) {
  console.log(`[message] Data received from server: ${event.data}`);
};

socket.onclose = function(event) {
  if (event.wasClean) {
    console.log(`[close] Connection closed cleanly, code=${event.code} reason=${event.reason}`);
  } else {
    // e.g. server process killed or network down
    // event.code is usually 1006 in this case
    console.log('[close] Connection died');
  }
};

socket.onerror = function(error) {
  console.log(`[error] ${error.message}`);
};
ƒ (error) {
  console.log(`[error] ${error.message}`);
}
```

Expected behavior is for the sockets to connect and exchange messages.

Observed behavior is that the backend crashes with `AttributeError: 'WebSocketProtocol' object has no attribute 'handshake_deferred'`. 

Removing the `await ws.accept()` line causes backend to crash with `ValueError: Socket has not been accepted, so cannot send over it`.

Switching the order of send and recv prevents the crash, but no messages are received nor sent.

In all cases frontend disconects with message `[close] Connection died`.

OS: Win10
sanic==20.12.2
daphne==3.0.1

ahopkins:
Fixes #2070 

See #2070 for further discussion


Tronic (dismissed):
Didn't test with Daphne but code looks good.

ahopkins:
I want to double check a few things on the ASGI spec. Like I said in the issue, I was about to get everything working on all three servers. Daphne is a little more troublesome because it doesn't respond to lifespan. Not sure if that is something about our implementation or the server.

ahopkins:
@ashleysommer Is this relevant given your PR? Will you make it for 21.6?

Tronic (approved):

PR #2129, opened 2021-05-04
unpin uvloop

graingert:
https://github.com/MagicStack/uvloop/pull/401 now sets python_requires ~and so as long as you ignore the broken versions of uvloop==0.15.0 and 0.15.1~ so you can still support py3.6 here

0.15.0 and 0.15.1 have been yanked so this pinning can go away completely!


ahopkins (approved):
Happy to approve, but for several reasons we are not going to go back to supporting Python 3.6 for v21. But, this is good for v20.12

setup.py
```diff
@@ -81,7 +81,7 @@ def open_local(paths, mode="r", encoding="utf8"):
     '; sys_platform != "win32" ' 'and implementation_name == "cpython"'
 )
 ujson = "ujson>=1.35" + env_dependency
-uvloop = "uvloop>=0.5.3,<0.15.0" + env_dependency
+uvloop = "uvloop>=0.5.3,!=0.15.0,!=0.15.1" + env_dependency
```
graingert, line 84:
```suggestion
uvloop = "uvloop>=0.5.3" + env_dependency
```

graingert: This pinning goes away completely with https://github.com/MagicStack/uvloop/issues/411

graingert:
This isn't stale

ahopkins (approved):

PR #2133, opened 2021-05-12
v2 AST router

### LINKED ISSUE #2122: Handler would be modify after wrapped by `app.route`

Handler would be modify after wrapped by `app.route`

```python
from sanic import Sanic

app = Sanic("test")


@app.get("/")
def get(request):
    """
    here is my doc
    """
    return text("123")

print(get.__doc__)
print(get.__name__)
```
Original function has been changed into tuple.

```sh
Built-in immutable sequence.

If no argument is given, the constructor returns an empty tuple.
If iterable is specified the tuple is initialized from iterable's items.

If the argument is a tuple, the return value is the same object.
  File "/Users/zinklu/code/WebProject/SanicDebug/app.py", line 22, in <module>
    print(get.__name__)
AttributeError: 'tuple' object has no attribute '__name__'
```

`app.route` won't change original handler.

```python
from sanic import Sanic

app = Sanic("test")


@app.get("/")
def get(request):
    """
    here is my doc
    """
    return text("123")

print(get.__doc__)
print(get.__name__)
```

```sh

    here is my doc
    
get
```

**Additional context**

ahopkins:
Related to https://github.com/sanic-org/sanic-routing/pull/37

Fixes #2122


Tronic (approved):

Tronic:
Hmmh, I am a bit quick on the approve button. Anyway, the signals code seems OK. You'll have to look at those tests.

Tronic (approved):

Issue #2087, opened 2021-03-24
Event registration


ahopkins:
When defining a signal, an event is created:

```python
@app.signal("do.something.start")
async def signal_handler():
    ...
```

But, what if you want to dispatch that signal, and then wait for another event:

```python
@app.post("/trigger")
async def trigger(request):
    await app.dispatch("do.something.start")
    await app.event("do.something.complete")
    return text("Done.")
```

Currently, this would fail because we never defined `do.something.complete`. And, I think this is correct. I do not think that `app.event` should start waiting on something that will never happen.

However, we might still want to dispatch `do.something.complete` without needing to have a signal handler. This usage would really facilitate intra-application messaging.

```python
@app.signal("do.something.start")
async def signal_handler():
    await asyncio.sleep(2)
    await app.dispatch("do.something.complete")
```

We need some sort of an API to register a signal with no handler. I think the solution is to allow `add_signal` to have `handler=None`.

```python
app.add_signal(None, "do.something.complete")
```

The easiest way to achieve that would likely be a change something like this:

```python
    def add_signal(
        self,
        handler: Optional[Callable[..., None]],
        event: str,
        condition: Dict[str, Any] = None,
    ):
        if not handler:
            handler = lambda: ...
        self.signal(event=event, condition=condition)(handler)
        return handler
```

kc-correa:
Hello. I'd like to give this a shot. Any specific place I should put this?

ahopkins:
Awesome! 🎉 

Take a look in `./mixins/signals.py`

PR #2135, opened 2021-05-15
Reworked event registration in signals file.

kc-correa:
Reworked the event registration in signals.py for issue #2087. I put it in quotes because I need to work on the lambda. What would the lambda need to do?


ahopkins:
> What would the lambda need to do?

Nothing. You can leave it as is.

```python
handler=lambda: ...
```

ahopkins:
Close in favor of #2140

PR #2137, opened 2021-05-19
Version prefix

ahopkins:
Well I cannot quite remember where this came up recently. But somewhere there was a point made that when you add a version, it _always_ will be at the beginning of your path:

```python
# /v1/my/path
app.route("/my/path", version=1)
```

This makes it difficult if you want to prefix _before_ the version: `/api/v1`.

---

This PR adds a new `version_prefix` keyword arg to:
- `app.route` and `bp.route` decorators (and all the convenience decorators also)
- `Blueprint` instantiation
- `Blueprint.group` constructor
- `BlueprintGroup` instantiation
- `app.blueprint` registration

The default value for all of these is: `/v`. Now, if you wanted some path to come before the version:

```python
Blueprint("Example", version=1.1, url_prefix="/example", version_prefix="/api/v")

# /api/v1.1/example/my/path
bp.route("/my/path")
```

### DIFF
--- a/sanic/blueprint_group.py
+++ b/sanic/blueprint_group.py
@@ -58,13 +58,20 @@ class BlueprintGroup(MutableSequence):
         app.blueprint(bpg)
     """
 
-    __slots__ = ("_blueprints", "_url_prefix", "_version", "_strict_slashes")
+    __slots__ = (
+        "_blueprints",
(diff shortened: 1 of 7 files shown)


Tronic (approved):
LGTM, useful feature.

PR #2140, opened 2021-05-20
Event autoregister

### LINKED ISSUE #2087: Event registration

ahopkins:
Resolves #2087
Replaces #2135 

- Adds a new config: `EVENT_AUTOREGISTER` (which defaults to `False`
- When `True`, it will allow applications to await events that have not previously been registered


sanic/mixins/signals.py
```diff
@@ -60,10 +60,16 @@ def decorator(handler: SignalHandler):
...
     def add_signal(
         self,
-        handler,
+        handler: Optional[Callable[..., Any]],
         event: str,
         condition: Dict[str, Any] = None,
     ):
+        if not handler:
+
+            async def noop():
```
ashleysommer, line 69: The nested function definition here might re-introduce app pickling issues, but might be fine to leave it.
(Not sure if users use multi-workers on Windows anymore, thats where the app pickling issues were popping up).

ashleysommer (approved):
Looks good. Autoregister looks like a feature I would like to use in one of my apps.

sanic/mixins/signals.py
ahopkins: It would be helpful if our Windows tests were running right now :unamused:

PR #2141, opened 2021-05-23
Signal integration

ahopkins:
This is the beginning of the second stage of #1630. More particularly, it introduces a new module to Sanic internals: `sanic.touchup`.

The idea with this module is that there will be predefined comment/markers that will modify executable functions at startup.

---

For example, the code to run request middleware looks like this:

```python
            for middleware in applicable_middleware:
                # ODE: http.middleware.before {"request": request}
                response = middleware(request)
                if isawaitable(response):
                    response = await response
                # ODE: http.middleware.after {"request": request, "response": response}  # noqa
                if response:
                    return response
```

At startup, there is a predefined interface that will lookup whether or not the `http.middleware.before` or `http.middleware.after` signals have been implemented. If not, then nothing happens. If yes, then the touchup module will inject signal dispatchers in those locations.

This should allow Sanic developers to make good use of signals to inject functionality, but not to slow down the req/resp cycle needlessly.

---

## TODO:

- Conditional block scheme
    ```python
    # CB: start some_expression
    do_something_only_if_startup_condition_exists()
    # CB: end
    ```     
- unit tests

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -368,11 +368,13 @@ class Sanic(BaseSanic):
         *,
         condition: Optional[Dict[str, str]] = None,
         context: Optional[Dict[str, Any]] = None,
+        inline: bool = False,
     ) -> Coroutine[Any, Any, Awaitable[Any]]:
         return self.signal_router.dispatch(
             event,
             context=context,
             condition=condition,
+            inline=inline,
         )
 
     def event(self, event: str, timeout: Optional[Union[int, float]] = None):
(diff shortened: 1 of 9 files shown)


ahopkins:

https://community.sanicframework.org/t/signals-integrations/824/4

ahopkins:
Scapping this idea (for now) and going to use AST for signals. I still want to explore something like this for conditional execution though.

PR #2146, opened 2021-05-30
Require stricter object names

ahopkins:
I think we need some tighter controls around naming `Sanic()` and `Blueprint()`. The main reason is that these names are used when generating route names: `<app>.<bp>.<route>`. It is a bit awkward when a route is called: `My Application.Something blueprint.foobar_handler`.


ashleysommer (approved):
Looks great! This is something I've thought about in the past, and I'm glad to see it implemented.

PR #2155, opened 2021-06-02
Allow 8192 header max to be breached

### LINKED ISSUE #2153: Maximum request header size capped at 8k
**Describe the bug**
Sending a request with a large header to a Sanic server results in 413 (Payload Too Large) response. 

This happens regardless of the `request_max_size` setting.

If I undserstand the implementation correctly, the maximum HTTP header size is hardcoded to be 8k:

https://github.com/sanic-org/sanic/blob/main/sanic/http.py#L172

**Expected behavior**
I think maximum header size should be configurable and values > 8k should be allowed.

 - OS: CentOS 7
 - Version 21.3.4

ahopkins:
Resolves #2153 

This moves the header size cap to a constant value on `Http` to make it easier to monkeypatch for more advanced use cases.


Tronic:
Why not use a proper config variable? Perhaps initialised to a more local variable on server start if accessing per request would otherwise affect performance (in that case I think the `min()` expression could also be done on server start).

ahopkins:
> Why not use a proper config variable? Perhaps initialised to a more local variable on server start if accessing per request would otherwise affect performance (in that case I think the `min()` expression could also be done on server start).

Two things:

1. The `min` probably only should be done once. I agree, and was going to make the change, but I think there was some talk somewhere about also allowing per-request overrides. So, I didn't want to go down that route just yet.
2. Why not proper config? Because I intentionally do not want people to mess with it. If it is just a constant on a class you really need to go out of your way. Maybe this is me being too cynical.
3. We probably could also add a config value. This was my initial thought: config value that cannot exceed the ceiling, and make the ceiling somewhat possible to monkeypatch. If we decide to run this all at startup, I think this might be the approach.

Tronic:
Okay, sounds reasonable. Apparently it also is very rare to have to increase the limit since we haven't received more complaints in these few months.

Tronic (approved):

ahopkins:
@Tronic new config value: `REQUEST_MAX_HEADER_SIZE`

Let me know your thought on this. We might even consider raising the ceiling to give them some more space to play with in the config value.

Tronic (approved):

hansraj1999:
which version tag this was fixed in?

ahopkins:
image

v21.6

PR #2158, opened 2021-06-09
New websockets

### LINKED ISSUE #2000: RFC: Internal websocket handling
Sanic uses the `websockets` package to handle the websocket protocol. It has [been discussed](https://github.com/sanic-org/sanic/issues/1908#issuecomment-694172043) that we could remove this dependency and implement them ourselves. More info to come.

ashleysommer:
Resolves #2000

@aaugustin @ahopkins
I've spent another couple of days on it, and its now passing all of the tests, including some other demo applications I've written to stress it out.

It certainly needs cleaning up, and could probably get some efficiency improvements in places too, but otherwise I'm pretty happy with it.

This branch also incorporates another change I've written, that splits the existing `server.HttpProtocol` out into a base `SanicProtocol` class, that contains just all of the connection-handling and transport-layer stuff, and `HttpProtocol` that is a subclass of `SanicProtocol` and adds the HTTP-specific handlers on top of that. I'd like the new `WebSocketProtocol` to subclass `SanicProtocol` but for now it still subclasses `HttpProtocol`.


sanic/server.py
```diff
@@ -99,21 +99,118 @@ def __init__(self, transport: TransportProtocol, unix=None):
...
+        self.app: Sanic = app
+        self.signal = signal or Signal()
+        self.transport: Optional[Transport] = None
+        self.connections = connections if connections is not None else set()
+        self.conn_info: Optional[ConnInfo] = None
+        self._can_write = asyncio.Event()
+        self._can_write.set()
+        self._unix = unix
+        self._time = 0.0  # type: float
+        self._task = None  # type: Optional[asyncio.Task]
```
ahopkins, line 140:
```suggestion
        self._time: float = 0.0
        self._task: Optional[asyncio.Task] = None
```

ashleysommer: Ignore the SanicProtocol changes (and other changes in `server.py`) for now, they'll come in a different PR, and I'll add typing and formatting to the variables at that stage.

sanic/server.py
```diff
@@ -99,21 +99,118 @@ def __init__(self, transport: TransportProtocol, unix=None):
...
+        self._data_received = asyncio.Event()
+
+    @property
+    def ctx(self):
+        if self.conn_info is not None:
+            return self.conn_info.ctx
+        else:
+            return self.ctx
```
ahopkins, line 148: RecursionError

sanic/websocket.py
```diff
@@ -40,11 +768,12 @@ def __init__(
         **kwargs
     ):
         super().__init__(*args, **kwargs)
-        self.websocket = None
+        self.websocket = None # type: Union[None, WebsocketImplProtocol]
```
ahopkins, line 771:
```suggestion
        self.websocket: Union[None, WebsocketImplProtocol] = None
```

sanic/websocket.py
```diff
@@ -40,11 +768,12 @@ def __init__(
...
         # self.app = None
         self.websocket_timeout = websocket_timeout
         self.websocket_max_size = websocket_max_size
-        self.websocket_max_queue = websocket_max_queue
+        if websocket_max_queue is not None and int(websocket_max_queue) > 0:
+            error_logger.warning(DeprecationWarning("websocket_max_queue is no longer used. No websocket message queueing is implemented."))
```
ahopkins, line 776: We should standardize on how we want to raise these.

Also, maybe add a code comment to remind use to remove this after 21.12.

aaugustin:
Looking at this from the perspective of websockets, I'm planning to review how you handle connection termination, normal or abnormal, because this is the area where I'm least confident that websockets gives you a nice API. Fortunately you're only doing the server-side and it's [more straightforward](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.1) than the client side.

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+            message: Data = joiner.join(self.chunks)  # type: ignore
+
+            assert not self.message_fetched.is_set()
+            self.message_fetched.set()
+            self.chunks = []
+            assert self.chunks_queue is None
+
+            return message
+
+    async def get_iter(self) -> AsyncIterator[Data]:
```
aaugustin, line 134: Pretty cool — I never implemented this [feature request](https://github.com/aaugustin/websockets/issues/479) in websockets itself!

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+                     break
+                 # Allow an eventloop iteration for the
+                 # next message to pass into the Assembler
+                 await asyncio.sleep(0)
+        finally:
+            self.can_pause = True
+            self.recv_lock.release()
+        return messages
+
+    async def recv_streaming(self) -> AsyncIterator[Data]:
```
aaugustin, line 573: Pretty cool — I never implemented this [feature request](https://github.com/aaugustin/websockets/issues/479) in websockets itself!

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+        """
+        Process a list of incoming events.
+        """
+        # Wrapped in a mutex lock, to prevent other incoming events
+        # from processing at the same time
+        async with self.process_event_mutex:
+            for event in events:
+                if event.opcode == OP_PONG:
+                    await self.process_pong(event)
+                else:
```
aaugustin, line 323: I would use `elif event.opcode in DATA_OPCODES` here, where `DATA_OPCODES` comes from `websockets.frames`. I should make it a public API.

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+                    except asyncio.TimeoutError:
+                        error_logger.warning("Websocket timed out waiting for pong")
+                        self.fail_connection(1011)
+                        break
+        except asyncio.CancelledError:
+            raise
+        except ConnectionClosed:
+            pass
+        except BaseException:
+            error_logger.warning("Unexpected exception in keepalive ping task")
```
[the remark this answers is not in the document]
aaugustin: Also should we really catch `BaseException`? `Exception` seems enough.

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+            # In Python Version 3.7: pause_reading is idempotent
+            # i.e. it can be called when the transport is already paused or closed.
+            self.io_proto.transport.pause_reading()
+
+            # Keeping fail_connection() synchronous guarantees it can't
+            # get stuck and simplifies the implementation of the callers.
+            # Not draining the write buffer is acceptable in this context.
+
+            # clear the send buffer
+            _ = self.connection.data_to_send()
```
aaugustin, line 404: This works because `data_to_send()` always aligns with frame boundaries and you can interject a close frame anywhere. (Mentioning it because it wasn't immediately obvious to me that this was legal.)

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+                error_logger.warning("Timeout waiting for TCP connection to close. Aborting")
+                if self.io_proto.transport is not None:
+                    self.io_proto.transport.abort()
+
+    def abort_pings(self) -> None:
+        """
+        Raise ConnectionClosed in pending keepalive pings.
+        They'll never receive a pong once the connection is closed.
+        """
+        assert self.connection.state is CLOSED
```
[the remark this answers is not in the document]
aaugustin: This is copy-pasted from websockets, where I use asserts to encode my beliefs about the state of the program. Sometimes I get it wrong e.g. https://github.com/aaugustin/websockets/issues/984.

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+
+            # Catch a common mistake -- passing a dict to send().
+
+            elif isinstance(message, Mapping):
+                raise TypeError("data is a dict-like object")
+
+            # Fragmented message -- regular iterator.
+
+            elif isinstance(message, Iterable):
+                # TODO use an incremental encoder maybe?
```
aaugustin, line 637: To be honest you don't really need to support fragmentation.

websockets had a bug where using compression + fragmentation crashed. Compression is enabled by default and it took year before someone reported it. So I'm pretty sure no one uses fragmentation.

You could ship without that feature and see if someone wants it!

sanic/websocket.py
```diff
@@ -70,73 +799,94 @@ def connection_lost(self, exc):
...
+        # If we've upgraded to websocket, shut it down
+        if self.websocket is not None:
+            if self.websocket.connection.state in (CLOSING, CLOSED):
+                return True
+            else:
+                return self.websocket.fail_connection(1001)
+        else:
+            return super().close_if_idle()
 
+    async def websocket_handshake(self, request, subprotocols=Optional[Sequence[str]]):
```
aaugustin, line 833: I see that you copy-pasted and modified `ServerConnection.accept` instead of wrapping it.

I thought it would be manageable to call it and convert its return value into a `SanicException` when the response code isn't 101. Does websockets make this approach difficult? Can I help?

sanic/websocket.py
ahopkins: Is there a possibility of `event.opcode` _not_ being in `DATA_OPCODES`? 

If no, why the `elif`? Wouldn't we want everything that is not a `PONG` to fall thru to this condition?

If yes, then how do we handle that case?

sanic/websocket.py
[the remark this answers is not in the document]
ahopkins: I would rather see:

```python
if self.connection.state is not CLOSED:
    raise SomethingGotMessedUp()
```

sanic/websocket.py
aaugustin: `DATA_OPCODES = OP_CONT, OP_TEXT, OP_BINARY`

=> you want to extract frames from these

`CTRL_OPCODES = OP_CLOSE, OP_PING, OP_PONG`

=> here you only care about OP_PONG because you provide an abstraction over ping / pong with a Future that completes when receiving a pong; websockets deals with OP_PING and OP_CLOSE

sanic/server.py
```diff
@@ -99,21 +99,118 @@ def __init__(self, transport: TransportProtocol, unix=None):
...
+        # Cause a call to connection_lost where further cleanup occurs
+        if self.transport:
+            self.transport.close()
+            self.transport = None
+
+    # asyncio.Protocol API Callbacks #
+    # ------------------------------ #
+    def connection_made(self, transport):
+        try:
+            # TODO: Benchmark to find suitable write buffer limits
```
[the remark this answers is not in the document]
Tronic: Generally it is best to keep buffers the minimal size that still allows for (near) max throughput. Larger buffers don't only waste memory but also increase latencies. Agreed, benchmarking is in order, and I did some on the streaming branch but not for websockets.

sanic/server.py
```diff
@@ -99,21 +99,118 @@ def __init__(self, transport: TransportProtocol, unix=None):
...
+            if not data:
+                return self.close()
+
+            if self._data_received:
+                self._data_received.set()
+        except BaseException:
+            error_logger.exception("protocol.data_received")
+
+
+class HttpProtocol(SanicProtocol):
```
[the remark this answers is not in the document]
Tronic: FWIW, that was kind of the purpose of separating much of server.py into http.py. But now that websockets are implemented on different level, this is getting complicated.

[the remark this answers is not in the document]
ahopkins: Absolutely, it was a good move. One that we can hopefully replicate with something like this:

```
./sanic
├── protocols
│    ├── interfaces
│    │    ├── http1.py (contains the current `Http` class)
│    │    └── http3.py (maybe 21.9???)
│    ├── base.py  (contains `SanicProtocol`)
│    ├── http.py  (contains `HttpProtocol`)
│    └── websocket.py  (contains `WebSocketProtocol`)
├── websockets
│    ├── connection.py  (contains `WebSocketConnection`)
│    ├── frame.py  (contains `WebsocketFrameAssembler`)
│    └── impl.py  (contains `WebsocketImplProtocol`)
└── server.py
```

sanic/server.py
ashleysommer: Yep, thanks. That is a diff left over from a different change I made, not sure if I need it.

sanic/server.py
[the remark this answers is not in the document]
ashleysommer: That value is just cut+pasted from the old `HttpProtocol` `connection_made()` method, I didn't add that comment or choose those values.

sanic/server.py
```diff
@@ -99,21 +99,118 @@ def __init__(self, transport: TransportProtocol, unix=None):
...
+        except Exception:
+            error_logger.exception("protocol.connect_made")
+
+    def connection_lost(self, exc):
+        try:
+            self.connections.discard(self)
+            self.resume_writing()
+            if self._task:
+                self._task.cancel()
+        except BaseException:
```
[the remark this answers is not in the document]
ashleysommer: I'll change it back to `Exception`. I had changed it to `BaseException` during my debugging in an attempt to catch a different error, but it can go back to `Exception`.

sanic/server.py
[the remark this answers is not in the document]
ashleysommer: Yep, looks good. I can implement that as part of the aforementioned upcoming `SanicProtocol` PR.

sanic/websocket.py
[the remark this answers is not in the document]
ashleysommer: All of the `assert` statements were leftover from copy+pasting large chunks of code from the sample `wesockets.Connection` implementation. I left them in for debugging at this stage, but I intend to either remove them or rewrite them into `if: raise` blocks.

sanic/websocket.py
ashleysommer: It was in your sample implementation, so I've put it into our implementation. I've tested it, seems to work fine, and I'm sure our users will take advantage of it for streaming in large messages.

sanic/websocket.py
ashleysommer: Thanks. I was meaning to ask about that. That answers my question. I'll change it to `elif event.opcode in DATA_OPCODES`.

sanic/websocket.py
ashleysommer: Yeah, I've looked at other deprecation warnings in Sanic, and there's not really a standardized way of doing it. Perhaps some more discussion is needed.

sanic/websocket.py
ashleysommer: Yeah, I tried it a few ways, and I concluded that this is fine. On `fail_connection` we no longer care about may or not need to be send from the send buffer, but we also need to clear it completely in order to get the Close frame, then send that.

sanic/websocket.py
ashleysommer: Ok. I'll remove that bit.

sanic/websocket.py
aaugustin: Me:

Dory

sanic/websocket.py
ashleysommer: I used my own websocket_handshake instead of `ServerConnection.accept` for several reasons:
1) `websockets.accept()` handles errors by returning `self.reject()` that creates custom `websockets` `http.Response` objects, that are not quite compatible with Sanic's response handler. Sanic works better with `raise InvalidUsage()`-style exception throwing (and utilizing sanic's built-in error formatting, logging, and error Response generation).
2) `websockets.accept()` builds a `Headers()` structure that is not quite compatible with Sanic's response handler.
3) I wanted to write the `101 Switching Protocols` success-response back to the SanicProtocol's `send()` method directly, without having to go via the `websockets` `http.Response` object.

Having said all that, after thinking about it talking it through, I see that it probably wouldn't be too hard to refactor this method to use `websockets.accept()` instead of this custom logic.

aaugustin: That makes sense to me; up to you to pick your favorite strategy.

Going through websockets' abstractions will create a little bit of overhead but opening handshakes aren't supposed to be the bottleneck of websockets (unless the user misses the persistent aspect of WebSocket connections). So perhaps the maintainability gains are worth the API overhead.

sanic/websocket.py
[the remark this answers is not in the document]
ashleysommer: Thats right. I'd changed it to `BaseException` for debugging something else, I can change it back to `Exception`.

sanic/websocket.py
ahopkins: image

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+
+    def abort_pings(self) -> None:
+        """
+        Raise ConnectionClosed in pending keepalive pings.
+        They'll never receive a pong once the connection is closed.
+        """
+        assert self.connection.state is CLOSED
+
+        for ping in self.pings.values():
+            ping.set_exception(ConnectionClosedError(1006, ""))
```
aaugustin, line 469: After https://github.com/aaugustin/websockets/pull/998 you should change this to `ConnectionClosedError(None, None)`. It's a backwards-incompatible change affecting users who create `ConnectionClosed` exceptions, which doesn't happen often.

ahopkins: @aaugustin On a separate note... is this a recent thing that `ConnectionClosedError` subclasses `Exception`? I feel like it had been `CancelledError`.

aaugustin: Well, my memory could fail me 😎 but as far as I remember the websockets exception hierarchy as always rooted at `class WebSocketException(Exception): ...`.

sanic/websocket.py
```diff
@@ -7,23 +10,748 @@
...
+from sanic.log import error_logger, logger
 
+import asyncio
 
-__all__ = ["ConnectionClosed", "WebSocketProtocol", "WebSocketConnection"]
 
 ASIMessage = MutableMapping[str, Any]
+UTF8Decoder = codecs.getincrementaldecoder("utf-8")
+
+class WebsocketFrameAssembler:
```
aaugustin, line 38: FYI there's a strong chance I end up adding a version of this to websockets, when I write an asyncio implementation on top of the Sans I/O layer. I'll document it so you can reuse it!

Tronic: Which reminds me that Sanic has been using non-incremental (= buggy) UTF-8 decoding in some example code in docs and also in tests. Hopefully this got cleared when someone rewrote the manual a few months ago but checking to make sure wouldn't be a bad idea. Tests don't hit the bug because they only use ASCII strings.

ahopkins:
@ashleysommer How is this coming? Are we going to be able to get this done this week to sneak it into the release?

aaugustin:
I'm still tuning a few things in websockets, especially around connection termination and error handling, so it may be safest to aim for the following Sanic release.

aaugustin:
Also this branch currently depends on the main branch of websockets, so I need to make a release before you can merge it.

ahopkins:
Good to know. Will plan accordingly

aaugustin:
@ashleysommer I'm working on an integration guide here: https://github.com/aaugustin/websockets/pull/999

This weekend I reworked the implementation to ensure `data_to_send` returns the EOF marker (`b""`) at the right time — it didn't. This shouldn't affect you.

Please note the addition of the `close_expected` method, which is designed to help with dangling connections. Since you're doing the server side, it's especially helpful in the following circumstances:

* you send a close frame
* due to network unreliability, you never hear about the client again
* you have a dangling TCP connection

In this case, you'd rather time out after 10 seconds than 2 hours (which I believe is the TCP timeout if you don't do anything) :-)

ahopkins:
@ashleysommer Any update? Are we targeting next release?

ahopkins:
@ashleysommer update?

ahopkins:
@aaugustin I spoke with @ashleysommer and he said he was hoping you'd have some of your changes released soon. Is this the case? We are hoping to target a release next month and I'd love to get this in since there are a few websocket issues to address with current implementation.

ahopkins:
> by the end of August

That would be amazing :heart_eyes:

sanic/websocket.py
ashleysommer: @aaugustin 
Just clarifying here. Do you mean use `ConnectionClosedError(1006, None)`?
The first argument to ConnectionClosedError must be an int, it cannot be None.

aaugustin:
Well, I missed my "end of August" deadline for the next websockets release, but hopeful to get it done this week-end.

ashleysommer:
Finished my big refactoring, splitting out the new websockets code into modules as described by @ahopkins above a few months ago.
All the functionality is in, moving onto testing phase. There are still some excess `assert` statements hanging around that can be removed once we've thoroughly tested everything.
Similarly, there might be some `asyncio.Lock` mutexes that can be removed after everything is proved out.

ahopkins:
Why would you want to do that, what is the concern that you are trying to overcome? That seems like a huge breaking change, so even if we wanted to, it would need to wait until next year.

ashleysommer:
Yes, I understand it is a big change, and it would have to be implemented after the 21.12 release.
Though it wouldn't necessarily be a breaking change, unless some users rely on websocket requests having request-middleware applied, which I'm not sure is something people want to need to do.
My reason for suggesting it is the current way websockets requests (pre-handshake/upgrade) are handled in Sanic, feels messy and I think could be made shorter and cleaner.

ahopkins:
I get that it may not be ideal, but I would be more concerned with no longer allowing access to middleware, and to routing these differently. In my mind, that would make websockets feel more "hacked" on since it would be fundamentally different than other routes from the very beginning. If there is a performance concern, then that might be worth considering. But, if the issue is code cleanliness, I would rather have spaghetti-soup without breaking changes and reduced functionality than a more seamless integration. Let's keep the discussion open and revisit after the LTS.

sanic/mixins/routes.py
```diff
@@ -121,8 +121,11 @@ def decorator(handler):
                         "Expected either string or Iterable of host strings, "
                         "not %s" % host
                     )
-
-            if isinstance(subprotocols, (list, tuple, set)):
+            if isinstance(subprotocols, list):
+                # Ordered subprotocols, maintain order
+                subprotocols = tuple(subprotocols)
+            if isinstance(subprotocols, set):
```
ahopkins, line 127: Can we make this `elif` so lists will not trigger it?

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,113 @@
...
+        websocket_max_size=None,
+        websocket_max_queue=None,
+        websocket_read_limit=2 ** 16,
+        websocket_write_limit=2 ** 16,
+        websocket_ping_interval=20,
+        websocket_ping_timeout=20,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self.websocket = None  # type: Union[None, WebsocketImplProtocol]
```
ahopkins, line 34: Convert to regular annotation

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,113 @@
...
+        # self.app = None
```
ahopkins, line 35: Can remove this if unneeded.

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,113 @@
...
+                subprotocols = list(subprotocols)
+            ws_conn = ServerConnection(max_size=self.websocket_max_size, subprotocols=subprotocols,
+                                       state=OPEN, logger=error_logger)
+            resp: "http11.Response" = ws_conn.accept(request)
+        except Exception as exc:
+            msg = (
+                    "Failed to open a WebSocket connection.\n"
+                    "See server log for more information.\n"
+                )
+            raise SanicException(msg, status_code=500)
```
ahopkins, line 98: How about make this a `ServerError`?

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,113 @@
...
+        if 100 <= resp.status_code <= 299:
+            rbytes = b"".join([b"HTTP/1.1 ", b'%d' % resp.status_code, b" ", resp.reason_phrase.encode("utf-8"), b"\r\n"])
```
ahopkins, line 100: If you join on `b" "` you can remove the space after the protocol declaration, and the item after the `status_code`.

Or, just format the whole thing as an f-string and convert to byte string. It might be easier to read that way too.

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,113 @@
...
+            for k, v in resp.headers.items():
+                rbytes += k.encode("utf-8") + b": " + v.encode("utf-8") + b"\r\n"
+            if resp.body:
+                rbytes += b"\r\n" + resp.body + b"\r\n"
+            rbytes += b"\r\n"
+            await super().send(rbytes)
```
ahopkins, line 106: Now that I am looking at this whole block, I would use regular strings up until this point and just encode once after it is constructed.

sanic/mixins/routes.py
ashleysommer: Yes. I think this is a typo, should be `elif`.

Tronic:
> Ideally I'd like something like this:
> 
> ```
> TCPServer -> HTTPProtocol -> Handshake/Upgrade -> WebsocketProtocol -> WebsocketRouter -> Endpoint handler
> ```
> 
> But that is a big change

I already implemented it like that in one of my testing branches. Makes perfect sense to detect upgrade right after a request is received (http.py?), prior to entering any routing, middleware or handlers. I suppose the rest of the pipeline needs some changes so that websocket requests are properly rejected if no websocket handler is found. Middleware not being run is of course a breaking change.

sanic/server/protocols/websocket_protocol.py
Tronic: Yes, much faster to encode once all is finished. Also I prefer not saying "utf-8" in the argument because that goes without saying (UTF-8 is the default).

It is also faster to combine different stages and to use f-strings, so lines 103-105 would be

```python
if resp.body is not None:
    rbytes += f"\r\n{resp.body}\r\n\r\n"
else:
    rbytes += "\r\n"
```

Tronic: And also faster to

```python
rbytes += "".join(f"{k}: {v}/r/n" for k, v in resp.headers.items())
```

I think this code already exists for HTTP response formatting in Sanic, perhaps reuse the existing function?

ahopkins:
> Middleware not being run is of course a breaking change.

Perhaps this can be mitigated next Winter. One of the startup optimizations I have been playing with is the ability to preprocess middleware and combine them with the route handler into a single function.

ahopkins:
Leaving the middleware issue aside for now, the other concern/problemto addresd is the pipeline  divergence it introduces with route handling. If we do not go thru the regular request handler, and we implement routing in two places we are creating a potential for inconsistency.

Thinking more on it, maybe the route handler could be turned into a common interface so that `HTTPHandler` and `WebSocketHandler` could be two implementations of it. It also might be a good direction to start cleaning up the app instance.

I am thinking something like this:

```
TCPServer -> HTTPProtocol -> HTTPHandler
```

```
TCPServer -> HTTPProtocol -> Handshake/Upgrade -> WebsocketProtocol -> WebSocketHandler
```

```
BaseHandler -> Router -> Request Middleware -> [Subclass impl] -> Endpoint handler
```

sjsadowski:
Heyo, I'm late to the party on this one. I personally would love to have some sort of flagged option to handle the upgrade request earlier in the chain, but I also think there's valid reasons for handling it after the router is executed, such as only allowing the ws upgrade on a specific route. In that case, it really couldn't be handled before the router because the upgrade request wouldn't be valid for all routes.

Just my thoughts. Going back to my hidey-hole now.

Tronic:
I guess this really ought to be part of routing, like HTTP methods, so that only ws handlers get matched in the first place (but this probably needs bigger changes).

ahopkins:
I'm not sure I understand the need for this though (besides "cleanliness"). What is the problem with doing the upgrade after the router identifies it as a websocket route?

aaugustin:
FYI I made a last-minute API change in `ServerConnection.reject` API but it doesn't appear in the diff of this PR so it doesn't affect you. I also made API changes in `ClientConnection` that don't affect you.

aaugustin:
There you go:

- https://pypi.org/project/websockets/10.0/
- https://websockets.readthedocs.io/en/stable/howto/sansio.html
- https://websockets.readthedocs.io/en/stable/reference/server.html#sans-i-o

Please let me know if you have any questions are if I can help!

aaugustin:
One thing I'd like to do is ensuring sanic only depends on public APIs. I can submit a later PR for this once you have the initial implementation done. I will also consider making more APIs in websockets public if that makes sense.

ahopkins:
@aaugustin As always, thank you so much for your support in getting this done. Much appreciation for your dedication to the greater open source community.

ashleysommer:
@aaugustin 
> One thing I'd like to do is ensuring sanic only depends on public APIs

I thought I was using all public APIs now. Are you able to point out which surfaces we're using that are not Public API?

aaugustin:
For example:

- `CLOSED`, `OPEN`, etc
    - at this point, only the `State` enum is a public API
    - its values are still available at the module level because the original implementation was just module-level variables and I couldn't be bothered to refactor everything
- there's the same situation with the `Opcode` enum and the `OP_*` variables
- `prepare_ctrl`
    - it's a trivial helper
    - either I make it a public API or we copy-paste it in Sanic

These are just details; nothing really interesting. I don't think it changes anything to the code you're currently writing.

I think the easiest would be for me to propose a pull request (with appropriate changes synchronized to websockets' docs) as soon as you merge this pull request. I can also audit the current diff and make suggestions if you'd rather avoid a follow-up commit.

ahopkins:
I say let's get this merged and done. I am hoping to ship this in two weeks to meet our deadline. We can come back and add a cleanup PR. If that is before the release, great. If not, then we can patch release it.

ashleysommer:
Yep. I agree. I'm going to push a final commit tomorrow to this PR, that will include that latest round of change suggestions, I'll remove some assert statements, and write a couple more basic tests.

ashleysommer:
Another round of work on the new websockets impl
* Added websocket_timeout support (matching previous/legacy support)
* Lots more comments
* Incorporated suggested changes from previous round of review
* Changed RuntimeError usage to ServerError
* Changed SanicException usage to ServerError
* Removed some redundant asserts
* Change remaining asserts to ServerErrors
* Fixed some timeout handling issues
* Fixed websocket.close() handling, and made it more robust
* Made auto_close task smarter and more error-resilient
* Made fail_connection routine smarter and more error-resilient

ashleysommer:
Further new websockets impl fixes
* Update compatibility with Websockets v10
* Track server connection state in a more precise way
* Try to handle the shutdown process more gracefully
* Add a new end_connection() helper, to use as an alterative to close() or fail_connection()
* Kill the auto-close task and keepalive-timeout task when sanic is shutdown
* Deprecate WEBSOCKET_READ_LIMIT and WEBSOCKET_WRITE_LIMIT configs, they are not used in this implementation.

I'm finally happy with the state of the implementation, and I've given it lots of testing.

However I haven't written any new tests, I don't have much more free time to give to this feature.

ashleysommer:
I've done some basic benchmarks to test this new websockets implementation vs the old (existing) one.

```
New impl, packet size 1k, client Chromium
 3,129,344 bytes in 30 seconds = 104,311.5
 3,956,736 bytes in 30 seconds = 131,891.2
 6,812,672 bytes in 30 seconds = 227,089.1
   742,400 bytes in 30 seconds =  24,746.7

New impl, packet size 1k, client Firefox
 3,099,648 bytes in 30 seconds = 103,321.6
 7,199,744 bytes in 30 seconds = 239,991.5
   584,704 bytes in 30 seconds =  19,490.1
 1,179,648 bytes in 30 seconds =  39,321.6

Old impl, packet size 1k, client Chromium
 1,346,560 bytes in 30 seconds =  44,885.3
   875,520 bytes in 30 seconds =  29,184.0
 2,617,344 bytes in 30 seconds =  87,244.8
 1,193,984 bytes in 30 seconds =  39,799.5

Old impl, packet size 1k, client Firefox
 6,332,416 bytes in 30 seconds = 211,080.5
   427,008 bytes in 30 seconds =  14,233.6
 1,091,584 bytes in 30 seconds =  36,386.1
   277,504 bytes in 30 seconds =   9,250.1

New impl, packet size 4k, client Chromium
26,910,720 bytes in 30 seconds = 897,024.0
26,869,760 bytes in 30 seconds = 895,658.7
26,836,992 bytes in 30 seconds = 894,566.4
26,865,664 bytes in 30 seconds = 895,522.1

New impl, packet size 4k, client Firefox
 3,227,648 bytes in 30 seconds = 107,588.2
12,132,352 bytes in 30 seconds = 404,411.7
28,725,248 bytes in 30 seconds = 957,508.3
28,598,272 bytes in 30 seconds = 953,275.7 

Old impl, packet size 4k, client Chromium
17,334,272 bytes in 30 seconds = 577,809.0
26,718,208 bytes in 30 seconds = 890,607.0
27,373,568 bytes in 30 seconds = 912,452.3
27,172,864 bytes in 30 seconds = 905,762.1

Old impl, packet size 4k, client Firefox
12,779,520 bytes in 30 seconds = 425,984.0
 1,363,968 bytes in 30 seconds =  45,465.6
28,602,368 bytes in 30 seconds = 953,412.3
28,803,072 bytes in 30 seconds = 960,102.4

New impl, packet size 16k, client Chromium
106,184,704 bytes in 30 seconds =3,539,490

Old impl, packet size 16k, clinet Chromium
105,349,120 bytes in 30 seconds =3,511,637
```

Turns out its very hard to benchmark websockets. The browser (both firefox and chromium) will try very hard to slow down the rate of transfer, to keep the browser responsive and prevent the page from crashing (and I guess to reduce CPU usage).
I found around halfway through my testing that the speeds I get are highly dependent on whether the browser window is in the foreground or background. If the window is in the background, the transfer rate falls to 0.5 to 0.1 of the max transfer speed.

Anyway, it seems the new impl is roughly the same performance of the old impl, and I'm very happy about that.

ashleysommer:
There's still some flake8 issues, looks like the black configuration is not quite compatible with the flak8 rules, in regard to line length.
I'll try to work it out.

ahopkins:
Great info on the benchmarks.

examples/websocket_bench.py
```diff
@@ -0,0 +1,70 @@
...
+from sanic import Sanic
+from sanic.response import file
+from sanic.log import error_logger
+import time
+error_logger.setLevel(logging.INFO)
+app = Sanic(__name__)
+
+@app.route('/')
+async def index(request):
+    return await file('websocket_bench.html')
```
ahopkins, line 12: We should use `app.static` if we want this to be an example implementation.

sanic/config.py
```diff
@@ -35,12 +35,9 @@
     "REQUEST_MAX_SIZE": 100000000,  # 100 megabytes
     "REQUEST_TIMEOUT": 60,  # 60 seconds
     "RESPONSE_TIMEOUT": 60,  # 60 seconds
-    "WEBSOCKET_MAX_QUEUE": 32,
     "WEBSOCKET_MAX_SIZE": 2 ** 20,  # 1 megabyte
     "WEBSOCKET_PING_INTERVAL": 20,
     "WEBSOCKET_PING_TIMEOUT": 20,
-    "WEBSOCKET_READ_LIMIT": 2 ** 16,
-    "WEBSOCKET_WRITE_LIMIT": 2 ** 16,
 }
```
ahopkins, line 41: The removal of these values is a breaking change that we will need to call out in release notes.

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,165 @@
...
+        if (
+            websocket_write_limit is not None
+            and int(websocket_write_limit) > 0
+        ):
+            error_logger.warning(
+                DeprecationWarning(
+                    "websocket_write_limit is no longer used. "
+                    "No websocket rate limiting is implemented."
+                )
+            )
```
ahopkins, line 64: Can you add an internal comment here that acts as a reminder to remove these notices after the LTS? We can remove them in v22.3

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,165 @@
...
+from ..websockets.impl import WebsocketImplProtocol
+
+
+if TYPE_CHECKING:
+    from websockets import http11
+
+
+class WebSocketProtocol(HttpProtocol):
+
+    websocket: Union[None, WebsocketImplProtocol]
```
ahopkins, line 19:
```python
Optional[WebsocketImplProtocol]
```

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,165 @@
...
+    websocket_timeout: float
+    websocket_max_size = Union[None, int]
+    websocket_ping_interval = Union[None, float]
+    websocket_ping_timeout = Union[None, float]
```
ahopkins, line 23: Use `Optional`

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,165 @@
...
+        websocket_read_limit: Optional[int] = None,  # read_limit is deprecated
+        websocket_write_limit: Optional[int] = None,  # write_limit deprecated
+        websocket_ping_interval: Optional[float] = 20.0,
+        websocket_ping_timeout: Optional[float] = 20.0,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self.websocket = None
+        self.websocket_timeout = websocket_timeout
+        self.websocket_max_size = websocket_max_size
```
ahopkins, line 40: We have some inconsistencies in typing here. `websocket_timeout` for example. A `None` value is valid in the instantiation, but not for the instance.

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,165 @@
...
+                DeprecationWarning(
+                    "websocket_write_limit is no longer used. "
+                    "No websocket rate limiting is implemented."
+                )
+            )
+        self.websocket_ping_interval = websocket_ping_interval
+        self.websocket_ping_timeout = websocket_ping_timeout
+
+    def connection_lost(self, exc):
+        if self.websocket is not None:
```
ahopkins, line 69: Is there some other valid falsey value for `self.websocket`? If not, then we should use `if not self.websocket`

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,165 @@
...
+            self.websocket.connection_lost(exc)
+        super().connection_lost(exc)
+
+    def data_received(self, data):
+        if self.websocket is not None:
```
ahopkins, line 74: Same.

Actually, this (and `eof_received`, `close`, and other methods following this pattern) probably could use an affirmative pattern instead: `if self.websocket` since I am presuming that there is only one valid type for that attribute.

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,165 @@
...
+                # subprotocols can be a set or frozenset,
+                # but ServerConnection needs a list
+                subprotocols = list(subprotocols)
+            ws_conn = ServerConnection(
+                max_size=self.websocket_max_size,
+                subprotocols=subprotocols,
+                state=OPEN,
+                logger=error_logger,
+            )
+            resp: "http11.Response" = ws_conn.accept(request)
```
ahopkins, line 126: Why the string type? We are 3.7+

```
from __future__ import annotations
```

sanic/server/protocols/websocket_protocol.py
```diff
@@ -0,0 +1,165 @@
...
+        except Exception:
+            msg = (
+                "Failed to open a WebSocket connection.\n"
+                "See server log for more information.\n"
```
ahopkins, line 130: Is the `\n` at the end of the message intentional?

sanic/server/websockets/connection.py
```diff
@@ -0,0 +1,82 @@
...
+
+class WebSocketConnection:
+    """
+    This is for ASGI Connections.
+    It provides an interface similar to WebsocketProtocol, but
+    sends/receives over an ASGI connection.
+    """
+
+    # TODO
+    # - Implement ping/pong
```
ahopkins, line 24: What's the story here?

sanic/server/websockets/connection.py
```diff
@@ -0,0 +1,82 @@
...
+        send: Callable[[ASIMessage], Awaitable[None]],
+        receive: Callable[[], Awaitable[ASIMessage]],
+        subprotocols: Optional[List[str]] = None,
+    ) -> None:
+        self._send = send
+        self._receive = receive
+        self._subprotocols = subprotocols or []
+
+    async def send(self, data: Union[str, bytes], *args, **kwargs) -> None:
+        message: Dict[str, Union[str, bytes]] = {"type": "websocket.send"}
```
ahopkins, line 37:
```
AnyStr
```

sanic/server/websockets/frame.py
```diff
@@ -0,0 +1,281 @@
...
+    def __init__(self, protocol) -> None:
+
+        self.protocol = protocol
+
+        self.read_mutex = asyncio.Lock()
+        self.write_mutex = asyncio.Lock()
+
+        self.completed_queue = asyncio.Queue(
+            maxsize=1
+        )  # type: asyncio.Queue[Data]
```
ahopkins, line 62: Type should not be set in comment

sanic/server/websockets/impl.py
```diff
@@ -0,0 +1,776 @@
...
+        return True
+
+    async def connection_made(self, io_proto: SanicProtocol, loop=None):
+        if loop is None:
+            try:
+                loop = getattr(io_proto, "loop")
+            except AttributeError:
+                loop = asyncio.get_event_loop()
+        self.loop = loop
+        self.io_proto = io_proto  # this will be a WebSocketProtocol
```
ahopkins, line 123: This comment is redundant

sanic/server/websockets/impl.py
```diff
@@ -0,0 +1,776 @@
...
+                    try:
+                        await asyncio.wait_for(ping_waiter, self.ping_timeout)
+                    except asyncio.TimeoutError:
+                        error_logger.warning(
+                            "Websocket timed out waiting for pong"
+                        )
+                        self.fail_connection(1011)
+                        break
+        except asyncio.CancelledError:
+            raise
```
ahopkins, line 222: Does this bubble up to the handler? Can the developer catch this?

sanic/server/websockets/impl.py
```diff
@@ -0,0 +1,776 @@
...
+        except ConnectionClosed:
+            pass
+        except Exception:
+            error_logger.warning("Unexpected exception in keepalive ping task")
```
ahopkins, line 226: We should probably log out what the exception is instead of swallowing it.

sanic/server/websockets/impl.py
```diff
@@ -0,0 +1,776 @@
...
+                    except asyncio.TimeoutError:
+                        error_logger.warning(
+                            "Websocket timed out waiting for pong"
+                        )
+                        self.fail_connection(1011)
+                        break
+        except asyncio.CancelledError:
+            raise
+        except ConnectionClosed:
+            pass
```
ahopkins, line 224: Maybe a debug message here?

sanic/server/websockets/impl.py
```diff
@@ -0,0 +1,776 @@
...
+                self.connection.send_text(message.encode("utf-8"))
+                await self.send_data(self.connection.data_to_send())
+
+            elif isinstance(message, (bytes, bytearray, memoryview)):
+                self.connection.send_binary(message)
+                await self.send_data(self.connection.data_to_send())
+
+            elif isinstance(message, Mapping):
+                # Catch a common mistake -- passing a dict to send().
+                raise TypeError("data is a dict-like object")
```
ahopkins, line 628: :tada:

ahopkins (changes requested):
@ashleysommer Nice work on this. It looks like this is a fantastic upgrade to a constant pain point for support.

sanic/server/websockets/impl.py
ashleysommer: This was implemented as suggested by @aaugustin

sanic/server/websockets/impl.py
ashleysommer: Yep. That was a reminder for me during dev. It can be removed.

sanic/server/websockets/impl.py
ashleysommer: Good idea.

sanic/server/websockets/impl.py
ashleysommer: This exception will be seen if the websocket connection closes or was closed immediately before `await self.ping()` above, but before the keepalive task is cancelled (ie, CancelledError was thrown yet), it requires bad timing, but will happen relatively often. Swallowing the error and continuing to exit the task is the right thing to do, but maybe a debug message will be helpful here as you suggested.

sanic/server/websockets/impl.py
ashleysommer: I'll double check. This might really need to be a debug log and continue, like the ConnectionClosed exception below.

examples/websocket_bench.py
ashleysommer: I didn't mean to commit `websocket_bench.py` and `websocket_bench.html`, they were for my testing. We can leave them if you want, or I will remove them.
Note, this code is based on the [original `examples/websocket.py` file, which uses the same `file()` response](https://github.com/sanic-org/sanic/blob/404c5f9f9eb54f06d1f30250bc2cc840c18a0eab/examples/websocket.py#L9). Probably both need to be updated to use `app.static()`.

sanic/server/websockets/connection.py
ashleysommer: This isn't my code, it is the ASGI-websocket wrapper/adapter class. It was moved from [here](https://github.com/sanic-org/sanic/blob/404c5f9f9eb54f06d1f30250bc2cc840c18a0eab/sanic/websocket.py#L146). I don't know if it makes sense for Sanic to do the ping/pong handling for an ASGI websocket, or perhaps the ASGI server handles that itself.

sanic/server/protocols/websocket_protocol.py
ashleysommer: I did that so I didn't have to import the http11 module at runtime. Its import is protected by `if TYPE_CHECKING` above, so it's not available for type annotations at runtime. Maybe I'm misunderstanding how Py 3.7 annotations work.

sanic/server/protocols/websocket_protocol.py
ashleysommer: Good point. I thought there might be some mypy errors pop up around there. I'll hunt them all down.

sanic/server/protocols/websocket_protocol.py
ashleysommer: Yep. Can do.

sanic/server/protocols/websocket_protocol.py
ashleysommer: I tend to err on the side of explicit when checking for `is not None`. But you're right, there is no other falsey value for this, and `if not self.websocket` will work. There are lots of other places in this code where I've used `is not None` where `not var` would suffice.

sanic/server/protocols/websocket_protocol.py
ashleysommer: Switching things around to affirmative pattern will probably make some chunks of code clearer too.

sanic/server/protocols/websocket_protocol.py
ashleysommer: This was copied as-is from some Websockets example implementation, that included the `\n` chars. I'll remove them.

examples/websocket_bench.py
ahopkins: It is nice to have them, but I would rather have them in a different repo since they are not really intended to be a "how to use Sanic". Personally, I would remove.

sanic/config.py
ashleysommer: This is a feature that may be able to be implemented into the sans-i/o implementation in future revisions, if there is demand for it. It was out of scope for this initial version.

aaugustin:
You could also benchmark sanic from a client written in websockets. You can open one or many connections from the client script. Here's a benchmarking script you can take inspiration from: https://github.com/aaugustin/websockets/blob/main/experiments/compression/client.py

It's benchmarking compression, not throughput, but you get the basic structure for connecting one or several clients to a server and sending traffic. The server can either drop the traffic or echo it, as you like.

To stress the server, you can also run multiple copies of the client script. For example, a 6 core machine, I'd try starting 4 clients for 1 server. Then the server will max out 1 CPU core while clients run on 4 CPU cores. This is somewhat valid as long as sanic isn't 4x faster than websockets — else you end up benchmarking the clients instead of the server :-)

ashleysommer:
Some changes including suggestions from code reviews:

- Integrate suggestions from code reviews
- Use Optional[T] instead of union[T,None]
- Fix mypy type logic errors
- change "is not None" to truthy checks where appropriate
- change "is None" to falsy checks were appropriate
- Add more debug logging when debug mode is on
- Change to using sanic.logger for debug logging rather than error_logger.

ashleysommer:
Looks like the remaining Flake8 errors are due to line lengths on long strings (mostly in error and debug messages), but Black says those are fine.

ahopkins:
> Looks like the remaining Flake8 errors are due to line lengths on long strings (mostly in error and debug messages), but Black says those are fine.

Black will not know how to break the lines up in the middle of a string. You will need to do that part yourself as elsewhere in the repo:

https://github.com/sanic-org/sanic/blob/1f06902ce12a5946741cbb71620825837746b9f1/sanic/app.py#L1011

ahopkins (dismissed):

ashleysommer:
Final changes:
- Fix long line lengths of debug messages
- Add some new debug messages when websocket IO is paused and unpaused for flow control
- Fix websocket example to use app.static()

ashleysommer:
Finally.. all green lights

ashleysommer:
@sanic-org/core-devs  Anyone else want to do a review on this?

ahopkins (approved):

aaugustin:
🎉

PR #2160, opened 2021-06-13
Signals Integration

ahopkins:
Taking another stab at integrating #1630 into the core of the project. This replaces #2141.

The biggest difference is that there is no comment based parsing. The basic startup is like this:

- Certain Sanic classes have a `TouchUpMeta` class that registers methods to be altered
- At startup, we loop through the registered methods, and apply one of the registered schemes for altering methods
- So far, the only "scheme" is `OptionalDispatchEvent` (ODE) that will use `ast` module to look through the method for a `app.dispatch` call.
- Once found, the ODE will determine if the signal that it is calling has been registered
- If yes, do nothing; If no, remove that `app.dispatch` call

Signals implemented:
- [x] `http.routing.before`
- [x] `http.routing.after`
- [x] `http.lifecycle.begin`
- [x] `http.lifecycle.read_head`
- [x] `http.lifecycle.request`
- [x] `http.lifecycle.handle`
- [x] `http.lifecycle.read_body`
- [x] `http.lifecycle.exception`
- [x] `http.lifecycle.response`
- [x] `http.lifecycle.send`
- [x] `http.lifecycle.complete`
- [x] `http.middleware.before` (condition: `attach_to=request/response`)
- [x] `http.middleware.after` (condition: `attach_to=request/response`)
~`server.main.start`~ _no support in 21.6_
~`server.main.stop`~ _no support in 21.6_
- [x] `server.init.before`
- [x] `server.init.after`
- [x] `server.shutdown.before`
- [x] `server.shutdown.after`

---

**Breaking changes**

- when using `create_server(..., return_asyncio_server=True)`, need to run `server.startup()`


ahopkins:
Test suite is 100% passing. Requires changes to testing lib though, so before merging and releasing https://github.com/sanic-org/sanic-testing/pull/22 I want to make sure we have a go ahead on this.

ahopkins:

https://community.sanicframework.org/t/log-a-request-end/873/2

sanic/signals.py
```diff
@@ -10,13 +10,39 @@
...
+        "http.lifecycle.read_head",
+        "http.lifecycle.request",
+        "http.lifecycle.response",
+        "http.routing.after",
+        "http.routing.before",
+        "http.lifecycle.send",
+        "http.middleware.after",
+        "http.middleware.before",
+    ),
+}
```
vltr, line 41: future thought: this could be turned into a tree, like:

```python
RESERVED_NAMESPACES = {
    "server": {
        "init": (
            "before",
            "after",
        ),
        "shutdown": (
            "before",
            "after",
        )
    },
    ...
}
```

just for the sake of brevity? I'm just not sure my example is even the best approach ... I mean, that's why the "future thought", in case I'm over complicating this in my head :sweat_smile:

sanic/signals.py
```diff
@@ -98,14 +146,23 @@ async def dispatch(
...
+            event,
+            context=context,
+            condition=condition,
+            fail_not_found=fail_not_found and inline,
+            reverse=reverse,
         )
+        logger.debug(f"Dispatching signal: {event}")
+
+        if inline:
+            return await dispatch
```
vltr, line 163: that's clever! :sunglasses:

sanic/signals.py
```diff
@@ -98,14 +146,23 @@ async def dispatch(
...
+
+        task = asyncio.get_running_loop().create_task(dispatch)
         await asyncio.sleep(0)
```
vltr, line 166: this is necessary? :grimacing:

sanic/signals.py
vltr, line 41: future thought: this could be a tree, perhaps? I mean, even I'm not sure if that's a good idea, hence the "future thought" ...

sanic/signals.py
vltr, line 163: oh I like that! :sunglasses:

sanic/signals.py
vltr, line 166: any particular reason for this `asyncio.sleep` call?

sanic/touchup/schemes/base.py
```diff
@@ -0,0 +1,24 @@
...
+    @abstractmethod
+    def run(self, method, module_globals) -> None:
+        ...
+
+    def __init_subclass__(cls):
+        BaseScheme._registry.add(cls)
+
+    @property
+    def trigger(self):
+        return f"# {self.ident}: "
```
vltr, line 21: ehhh ... what's the purpose of this method?

vltr (commented):
overall, I think this is awesome! I even had a small low-level code feeling :sweat_smile:

sanic/app.py
```diff
@@ -1118,25 +1172,67 @@ async def _run_request_middleware(self, request, request_name=None):
...
+                )
+
                 if response:
                     return response
         return None
 
     async def _run_response_middleware(
         self, request, response, request_name=None
-    ):
+    ):  # no cov
```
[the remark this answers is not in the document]
ahopkins: Yup.... it is a serious downside in the AST method. Coverage.py has no way of tracking whether tests covered them, so they get marked as missed. :unamused: This is one of the biggest downsides, but I am not sure how else to go about it.

sanic/app.py
```diff
@@ -772,7 +831,7 @@ async def handle_request(self, request: Request):
                     )
 
                 # Run response handler
-                response = handler(request, **kwargs)
+                response = handler(request, **request.match_info)
```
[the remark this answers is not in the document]
ahopkins: What do you mean? Basically I wanted to copy over the contents so the originals are not disturbed, but we allow for `request.match_info` to be modifiable (something I have been wanting for years).

sanic/app.py
```diff
@@ -1253,20 +1341,43 @@ def _build_endpoint_name(self, *parts):
...
 
     @classmethod
     def _cancel_websocket_tasks(cls, app, loop):
         for task in app.websocket_tasks:
             task.cancel()
 
+    @staticmethod
+    async def dispatch_delayed_tasks(app, loop):
+        for name in app._delayed_tasks:
+            await app.dispatch(name, context={"app": app, "loop": loop})
```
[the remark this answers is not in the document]
ahopkins: Yeah, that's a good call out.

sanic/signals.py
ahopkins: It was originally, and then I realized I was over-complicating it. I think this simple structure (1) does what we need, and (2) is super simple to whoever is looking at the code.

sanic/signals.py
ahopkins: Yeah, in playing around with it, I have found this to be super handy.

sanic/signals.py
ahopkins: I am not in love with it, and open to other solutions. I should probably add a comment to explain. The reason it is there is to yield back to the loop one time which is useful in avoiding a deadlock when waiting on `await app.event`.

sanic/touchup/schemes/base.py
ahopkins: :thinking: I'm not sure. Did I add that? :laughing: 

It looks to me like it might be a relic of development and can be deleted. I will confirm.

sanic/app.py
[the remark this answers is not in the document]
vltr: I mean, now the `handler` could use only the `request` argument ... I know it would break an undefined number of apps, so that would not be a good change. Even though it makes sense to be the way it is ... NM, I'm just thinking out loud I guess :sweat_smile:

sanic/app.py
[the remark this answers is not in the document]
vltr: I think I stumbled upon something like that in the past ... if I ever find it again, I'll let you know

sanic/app.py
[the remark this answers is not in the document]
vltr: :+1:

sanic/signals.py
vltr: yeah, then +1 to my attempt of over-complicate things :laughing:

sanic/signals.py
vltr: oh yeah, let's use some Python internals in our favor too :muscle:

sanic/signals.py
vltr: well, have you bumped into any deadlocks without it?

sanic/touchup/schemes/base.py
vltr: well, I leave breadcrumbs everywhere in my code (specially development stuff), so ... :shrug:

sanic/app.py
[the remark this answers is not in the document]
ahopkins: I spent a lot of time trying to find a way around this.

sanic/signals.py
ahopkins: Yes.

ahopkins: I think the case in the docs will deadlock without it:

```python
async def wait_for_event(app):
    while True:
        print("> waiting")
        await app.event("foo.bar.baz")
        print("> event found\n")

@app.after_server_start
async def after_server_start(app, loop):
    app.add_task(wait_for_event(app))

```

vltr (approved):
all good, awesome work! :sunglasses:

PR #2162, opened 2021-06-14
Smarter auto fallback

ahopkins:
This relates back to the error formatting from #1937.

It adds:
- `error_format` on `@app.route` and convenience methods
- _smarter_ "auto" check

At startup, every route will now have an `error_format` assigned to it. This can be configured at the route level (perhaps we can think about adding it at Blueprint or BlueprintGroup).

```python
@app.route(..., error_format="text")
```

If it is not set, then it will use the value of `FALLBACK_ERROR_FORMAT`, which can be set to "text", "json", "html", or "auto". **It will _now_ default to "auto".**

---

What does `auto` do now? Previously it would try and look at the incoming headers and make a guess by looking at the request content-type and request body (for json).

Now, it at startup it will evaluate the returns of the route handlers and look for `text()`, `json()`, `html()`, etc. It is still not perfect obviously. And, I think we still might want to add a check incoming `Accept` headers to make sure the selection makes sense.

If `auto` does not come up with a guess it will fallback `HTMLRenderer`, which can be reconfigured if you use your own base `Config` (see #2119).

---

TODO:

- [x] Make sure that `FALLBACK_ERROR_FORMAT` is a proper value
- [x] Tests
- [x] Check incoming headers (pushed to #2171)


Tronic:
> What does auto do now? Previously it would try and look at the incoming headers and make a guess by looking at the request content-type and request body (for json).

Should be using the `accept` header instead:
* If it includes (starts with, even) `text/html`: HTML
* If it includes `application/json`: JSON
* Else (it is probably `*/*` the default of curl, wget and fetch):
  - check `content-type` for `application/json`: JSON (Note 1!)
  - otherwise return TEXT.

Note 1: Request body and content-type detection has the drawback that requests with no body "randomly" return different format error responses, so I might even leave out this detection entirely but certainly wouldn't be looking at `request.json`. Tough luck for those JSON users who do not correctly specify `accept`.

Tronic:
Other relevant cases for the above: img, script, video, audio etc. tags, where that logic would return a text response. What is returned doesn't matter much because browsers won't show the response even in developer tools; maybe an empty response would be more appropriate.

ahopkins:
> Other relevant cases for the above: img, script, video, audio etc. tags, where that logic would return a text response. What is returned doesn't matter much because browsers won't show the response even in developer tools; maybe an empty response would be more appropriate.

Not sure that I agree that it should be an empty response; you could be accessing the resource directly in a browser.

Tronic:
> > Other relevant cases for the above: img, script, video, audio etc. tags, where that logic would return a text response. What is returned doesn't matter much because browsers won't show the response even in developer tools; maybe an empty response would be more appropriate.
> 
> Not sure that I agree that it should be an empty response; you could be accessing the resource directly in a browser.

I believe in that case the accept header lists text/html. It only shows image and other formats if the request is from an HTML element.

ahopkins:
OK. I'll check that.

ahopkins:
This is roughly the logic:

- Use explicit `error_format`, if available
- Use likely intended response type, if available (this will cover the MAJORITY of use cases)
- Use the fallback
- If the fallback is `auto`:
    - Look to see if `text/html` is first in Accept header, use HTMLRenderer
    - Look to see if `application/json` is anywhere in Accept or if incoming request has `content-type: application/json`, use JSONRenderer
    - Check if completely missing Accept header, use TextRenderer (Likely coming from a raw socket)
    - If still nothing matches...
        - Check if its a poorly formatted JSON request, and use JSONRenderer (covers backwards compat)
        - Fallback to TextRenderer if curl
- In all cases, if there is an Accept header, make sure it is allowed
- If not, loop thru the Accept and stop at the first match (we may need to add some more potential matches here, but that can be in another iteration)
- If still nothing, use base renderer (which is HTMLRenderer for compat. Change this to TextRenderer in v22.3)

sanic/errorpages.py
```diff
@@ -398,34 +401,89 @@ def exception_response(
...
+                    # package like requests or httpx
+                    try:
+                        # Give them the benefit of the doubt if they did:
+                        # $ curl localhost:8000 -d '{"foo": "bar"}'
+                        # And provide them with JSONRenderer
+                        renderer = (
+                            JSONRenderer  # type: ignore
+                            if request.json
+                            else default
+                        )
```
Tronic, line 464: I would leave these out. Custom handling for curl can be confusing when debugging why it doesn't work in browser or elsewhere.

ahopkins: Good point

Tronic: Okay, I guess this gets simplified at 22.3. Maybe emit warnings for the cases that would be changed then or in case for the json body detection, could be changed maybe a year later? (reading body to determine type is problematic and just evil).

sanic/errorpages.py
```diff
@@ -340,41 +340,138 @@ def escape(text):
...
                 try:
-                    renderer = JSONRenderer if request.json else HTMLRenderer
-                except InvalidUsage:
+                    render_format = request.route.ctx.error_format
+                except AttributeError:
+                    ...
+
+            content_type, *_ = request.headers.getone(
+                "content-type", ""
+            ).split(";")
```
Tronic, line 402: Suggesting `.split(";", 1)[0]` instead of unpacking assignment.

tests/test_errorpages.py
```diff
@@ -74,13 +80,194 @@ def test_auto_fallback_with_content_type(app):
     app.config.FALLBACK_ERROR_FORMAT = "auto"
 
     _, response = app.test_client.get(
-        "/error", headers={"content-type": "application/json"}
+        "/error", headers={"content-type": "application/json", "accpet": "*/*"}
```
Tronic, line 83: I assume that the typo in accept is unintentional.

tests/test_errorpages.py
```diff
@@ -74,13 +80,194 @@ def test_auto_fallback_with_content_type(app):
...
     )
     assert response.status == 500
     assert response.content_type == "application/json"
 
     _, response = app.test_client.get(
-        "/error", headers={"content-type": "text/plain"}
+        "/error", headers={"content-type": "foo/bar", "accpet": "*/*"}
```
Tronic, line 89: Ditto

vltr (approved):
it seems all good to me! :sunglasses:

PR #2167, opened 2021-06-16
Add reloading on addtional directories

### LINKED ISSUE #2143: Auto reload resource files in the a Sanic app
When a Sanic app is in production, I use different configs for development, staging, and production. I use 3 YAML files to implement that. However, they are not tracked by auto_reload in the development environment. I would be happy to be able to see Sanic can restart when I change the config.

It can be a helper API to register resources for auto_reload. importlib provides a semantic method [`importlib.resources.read_text`](https://docs.python.org/3/library/importlib.html#importlib.resources.read_text). It does no more than read(<file>, 'r'). It might be nice to hook with readers from importlib.

**Additional context**
Example app structure:
```
awesomeapp
| - config
   | - development.yaml
   | - staging.yaml
   | - production.yaml
| app.py
| __init__.py
```

Potential API signature:
`sanic.resources.read_text(package, resource, encoding='utf-8', errors='strict')`

ahopkins:
Resolves #2143 

This allows the auto-reloader to watch additional directories beyond just the Python modules for changes. Implemented in two ways:

**Via app.run**
```python
app.run(..., reload_dir="./some/path")
# or
app.run(..., reload_dir=["./some/path", "./some/other/path"])
```

**Via cli**
```
sanic --reload-dir=./some/path
sanic -D ./some/path -D ./some/other/path
```

### DIFF
--- a/sanic/reloader_helpers.py
+++ b/sanic/reloader_helpers.py
@@ -3,6 +3,7 @@ import signal
 import subprocess
 import sys
 
+from pathlib import Path
 from time import sleep
 
 from sanic.config import BASE_LOGO
@@ -59,6 +60,20 @@ def restart_with_reloader():
     )
 
 
+def _check_file(filename, mtimes):
+    need_reload = False
+
+    mtime = os.stat(filename).st_mtime
+    old_time = mtimes.get(filename)
+    if old_time is None:
+        mtimes[filename] = mtime
+    elif mtime > old_time:
+        mtimes[filename] = mtime
+        need_reload = True
+
+    return need_reload
+
+
 def watchdog(sleep_interval, app):
     """Watch project files, restart worker process if a change happened.
 
@@ -85,18 +100,22 @@ def watchdog(sleep_interval, app):
         while True:
             need_reload = False
 
-            for filename in _iter_module_files():
-                try:
-                    mtime = os.stat(filename).st_mtime
-                except OSError:
-                    continue
+            for collection in (_iter_module_files(), *app.reload_dirs):
+                if isinstance(collection, Path):
+                    collection = collection.glob("**/*")
 
-                old_time = mtimes.get(filename)
-                if old_time is None:
-                    mtimes[filename] = mtime
-                elif mtime > old_time:
-                    mtimes[filename] = mtime
-                    need_reload = True
+                for filename in collection:
+                    try:
+                        check = _check_file(filename, mtimes)
+                    except OSError:
+                        continue
+
+                    if check:
+                        need_reload = True
+                        break
+
+                if need_reload:
+                    break
 
             if need_reload:
                 worker_process.terminate()
--- a/sanic/reloader_helpers.py
+++ b/sanic/reloader_helpers.py
@@ -36,11 +36,6 @@ def _iter_module_files():
                 yield filename
 
 
-def _iter_dirs(dirs):
-    for filename in (d.glob("**/*") for d in dirs):
-        yield filename
-
-
 def _get_args_for_reloading():
     """Returns the executable."""
     main_module = sys.modules["__main__"]
--- a/sanic/reloader_helpers.py
+++ b/sanic/reloader_helpers.py
@@ -1,9 +1,9 @@
+import itertools
 import os
 import signal
 import subprocess
 import sys
 
-from pathlib import Path
 from time import sleep
 
 from sanic.config import BASE_LOGO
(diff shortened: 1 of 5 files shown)


Tronic:
I would use name other than `include_dir`, which connotes header paths and doesn't implicate that they are used with auto-reloader only. Maybe `reloader_paths` or something alike? (can these be files too, not just folders?)

I suppose the use case for this is to handle changes in templates, config files etc?

Tronic:
Unrelated to this PR, but worth of notice is the use of livereload on client side. Dunno if there are any Sanic extensions that could already do that, and also connect with the server reloads.

ahopkins:
> I would use name other than `include_dir`

This is a good call out. I was having trouble coming up with a good name. I looked at `uvicorn`, and they use `--reload-dir`, which makes sense because their reload is `--reload`. And then... I just wrote it off. But, looking back on it again during the daytime hours with some more caffeine in my system, and `--reload-dir` still works for our use case too.

> can these be files too, not just folders?

Not right now. I really just wanted to sneak the functionality in a fairly low touch way. The `reloader_helper.py` probably needs some more work. Under the hood it is just looping on the dir + glob.
 
> I suppose the use case for this is to handle changes in templates, config files etc?

Yeah. The other place that I have personally run into it is with maintaining openapi spec files.

> Unrelated to this PR, but worth of notice is the use of livereload on client side. Dunno if there are any Sanic extensions that could already do that, and also connect with the server reloads.

Agreed. It is probably the realm for extension. I built this that I use when working on slide presentations:

https://github.com/ahopkins/pyconil2021-liberate-your-api/blob/main/presentation/server.py

ahopkins:
Also added `--reload` as an alias to `--auto-reload`

Tronic:
Great! A could also change the short argument to `-R` to go with `-r` for reload, and perhaps change the argument name in help to DIR or PATH so this would fit in one line:
> ```
>   -D RELOAD_DIR, --reload-dir RELOAD_DIR
>                                  Extra directories to watch and reload on changes
> ```

> ```
>   -r, --reload, --auto-reload    Watch source directory for file changes and reload on changes
>   -R PATH, --reload-dir PATH     Extra directories to watch and reload on changes
> ```

Tronic:
Also unrelated to this PR, but since you are working on CLI...

Single-flag toggles for production mode (benchmark mode) and development mode would be useful:
- Production: no access logs, workers=os.cpu_count()
- Development: debug, reloader

ahopkins:
```
  -d, --debug
  -r, --reload, --auto-reload    Watch source directory for file changes and reload on changes
  -R PATH, --reload-dir PATH     Extra directories to watch and reload on changes
```

sanic/reloader_helpers.py
```diff
@@ -85,18 +100,22 @@ def interrupt_self(*args):
...
+
+                for filename in collection:
+                    try:
+                        check = _check_file(filename, mtimes)
+                    except OSError:
+                        continue
+
+                    if check:
+                        need_reload = True
+                        break
```
Tronic, line 115: Shouldn't this be inside the try block?

Also the two level iteration is a bit convoluted - I'm sure you can find a cleaner way to do it. Flatten the iterators, itertools.chain or something?

Tronic: IIRC, one needs to iterate over all the files, updating timestamps, and not return early even when differences are found. This is to avoid multiple reloads when many files are changed, so please pay attention to preserving that feature (fixed in 2019 probably).

I did not properly review this yet but overall it seems OK, especially if you double check the reloader loops and the code within. Possibly no changes required, I just cannot wrap my head around it right now.

ahopkins: > Flatten the iterators, itertools.chain or something?

Yeah... that also would give us the ability to add files individually. I was trying to sneak this change in without being too invasive and then leave another opportunity to overhaul how the whole module works with better iteration.

> Shouldn't this be inside the try block?

Will check :eyes:

> IIRC, one needs to iterate over all the files, updating timestamps, and not return early even when differences are found. This is to avoid multiple reloads when many files are changed, so please pay attention to preserving that feature (fixed in 2019 probably).

Good catch :+1:

ahopkins: > Shouldn't this be inside the try block?

You mean because of scoping? Variables are not scoped inside `try`

```
>>> try:
...  foo = 1
... except Exception:
...  ...
... 
>>> print(foo)
1
```

ahopkins: I'm not sure if this functionally is any different. The point is we still need multiple iterators since we need to generate new globs on every run.

```python
            for filename in itertools.chain(
                _iter_module_files(),
                *(d.glob("**/*") for d in app.reload_dirs),
            ):
```

Tronic: > > Shouldn't this be inside the try block?
> You mean because of scoping? Variables are not scoped inside `try`

Nevermind. The `continue` in the exception handler takes care that this code is not reached when `check` is not set, which was my concern.

Tronic (approved):

PR #2170, opened 2021-06-20
CBV alternate attach; CompositionView deprecate

### LINKED ISSUE #2069: deprecate CompositionView ?
Currently sanic offers a class called `CompositionView`

I really am struggling to find any utility in this class, since 

```python
from sanic.views import CompositionView

def get_handler(request):
    return text("I am a get method")

view = CompositionView()
view.add(["GET"], get_handler)
view.add(["POST", "PUT"], lambda request: text("I am a post/put method"))

# Use the new view to handle requests to the base URL
app.add_route(view, "/")
```

Seems much more confusing to me than

```python
def get_handler(request):
    return text("I am a get method")

app.route("/", methods=["GET"])(get_handler)
app.route("/", methods=["POST", "PUT"])(lambda request: text("I am a post/put method"))
```

Can anyone offer a compelling use case for CompositionView?

If not, I would suggest to deprecate it 

https://github.com/sanic-org/sanic/blob/master/sanic/views.py

ahopkins:
Resolves #2069 

`CompositionView` is deprecated and will be removed in 21.12

Adds two new methods for attaching CBVs.

```python
class DummyView(HTTPMethodView):
    def get(self, request):
        return text("I am get method")

DummyView.attach(Sanic.get_app(), "/")
```

and

```python
class DummyView(HTTPMethodView, attach=app, uri="/"):
    def get(self, request):
        return text("I am get method")
```

### DIFF
--- a/sanic/views.py
+++ b/sanic/views.py
@@ -1,6 +1,14 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Iterable,
+    List,
+    Optional,
+    Union,
+)
 from warnings import warn
 
 from sanic.constants import HTTP_METHODS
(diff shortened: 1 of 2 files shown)


Tronic (approved):
LGTM, didn't test (never used the class API anyway).

PR #2184, opened 2021-07-07
Copy Blueprints Implementation

ChihweiLHBird:
Related issue: https://github.com/sanic-org/sanic/issues/2163


ChihweiLHBird:
@ahopkins The parameters list of the `copy` method breaks the codeclimate rule: _"Function copy has 5 arguments (exceeds 4 allowed)."_

Do you think key words parameters will be a better choice? Or can we ignore the codeclimate issue?

ahopkins:
Ignore that. I need to configure the rules. The defaults might work for an app, but are too restrictive for a framework.

ChihweiLHBird:
> Ignore that. I need to configure the rules. The defaults might work for an app, but are too restrictive for a framework.

Thank you. I think this PR is roughly done, how can I get some feedbacks from you and other members of the community?

ahopkins:
Sure, I'll take a look. Mark it as ready for review.

@sanic-org/framework  thoughts?

ChihweiLHBird:
> Sure, I'll take a look. Mark it as ready for review.
> 
> @sanic-org/framework thoughts?

One doubt I got is that in the copy method, should we use the default values of the parameters of the method in the new blueprint if the parameters was not provided by caller? Or we don't change anything in the new blueprint and let it use the old values copied from the old blue print.

Tronic:
> Another way to do this, by using key words parameters, we can allow user to change any attribute (methods and variables) in the new blueprint. Something like:

It is better to explicitly list the parameters. This gives auto completion and error checking in IDEs, not possible if a function takes `**kwargs`.

ahopkins:
There are four solutions that I can see:

### Pull from `kwargs`
**Pro**: Explicit knowledge that the dev intentionally passed a value
**Con**: No IDE auto-completion

### Keyword with `None` as default value
**Pro**: Easy to check that a value was intentionally set
**Con**: Does not allow passing `None` as a value

### Keyword with carrying over existing defaults
**Pro**: Easy to check that a value is NOT default and set to that
**Con**: What if you are returning a value back to its default?

### Keyword with sentinel as default
**Pro**: Allows for both `None` and orig defaults to be passed
**Con**: Difficult to type annotate

---

Therefore, I suggest a two step approach:

```python
# ./helpers.py
class Default:
    ...


_default = Default()
```

```python
class Blueprint:
    def copy(
        self,
        name: str,
        *,
        url_prefix: Optional[Union[str, Default]] = _default,
        version: Optional[Union[int, str, float, Default]] = _default,
        strict_slashes: Optional[Union[bool, Default]] = _default,
        version_prefix: Optional[Union[str, Default]] = _default,
    ):
        ...
```

ChihweiLHBird:
> There are four solutions that I can see:
> 
> ### Pull from `kwargs`
> **Pro**: Explicit knowledge that the dev intentionally passed a value
> **Con**: No IDE auto-completion
> 
> ### Keyword with `None` as default value
> **Pro**: Easy to check that a value was intentionally set
> **Con**: Does not allow passing `None` as a value
> 
> ### Keyword with carrying over existing defaults
> **Pro**: Easy to check that a value is NOT default and set to that
> **Con**: What if you are returning a value back to its default?
> 
> ### Keyword with sentinel as default
> **Pro**: Allows for both `None` and orig defaults to be passed
> **Con**: Difficult to type annotate
> 
> ---
> 
> Therefore, I suggest a two step approach:
> 
> ```python
> # ./helpers.py
> class Default:
>     ...
> 
> 
> _default = Default()
> ```
> 
> ```python
> class Blueprint:
>     def copy(
>         self,
>         name: str,
>         *,
>         url_prefix: Optional[Union[str, Default]] = _default,
>         version: Optional[Union[int, str, float, Default]] = _default,
>         strict_slashes: Optional[Union[bool, Default]] = _default,
>         version_prefix: Optional[Union[str, Default]] = _default,
>     ):
>         ...
> ```

Hi Adam, thanks for the suggestion. Do you think the default class can be used globally in the framework? We can achieve similar functionality in other functions as well with it, I think. 

By the way, I don't quite understand what's the Con of the third solution you mentioned. Can you explain what is the situation of returning a value back to its default? Thank you so much.

Tronic:
Derive the sentinel type from `NoneType`, then plain `Optional` should work for annotation. Seems a bit hackish though, is it a frequent need to set some value to None?

ChihweiLHBird:
> Derive the sentinel type from `NoneType`, then plain `Optional` should work for annotation. Seems a bit hackish though, is it a frequent need to set some value to None?

I think generally, we are not able to use `NoneType` as a class.

Tronic:
> I think generally, we are not able to use `NoneType` as a class.

Today I learned something new. Looks like None and a few other built in values are not proper objects (in contrast to integers and floats which are).

```
TypeError: type 'NoneType' is not an acceptable base type
```

ahopkins:
```python
>>> class D(type(None)):
...  ...
... 
TypeError: type 'NoneType' is not an acceptable base type
```

Hardcoded protection: https://github.com/python/cpython/blob/main/Objects/typeobject.c#L2207

---

Off the top of my head, I do not have another use case for `_default`. Typically `object()` is the standard sentinel, but typing it is hard, broken, or awkward. Hence the `Default` class.

ChihweiLHBird:
I think the natural way to do copy is keeping everything same as the old object unless the user request a change. So I would suggest we to think about a way that allows users to pass or apply these below to the attributes of the new object:
1. `None` as a value.
2. The default value of an attribute when the object is created via its constructor with a minimal parameter input.

The second can be achieved by the `Default` class and object @ahopkins mentioned above. But `None` value still conflicts with the attributes from the old objects, meaning we cannot use `None` as the default value of the parameters of the `copy` method, otherwise we don't know whether or not the `None` is from the user.

I guess we must take a trade off when implementing this feature, 😂

ahopkins:
> I think the natural way to do copy is keeping everything same as the old object unless the user request a change. So I would suggest we to think about a way that allows users to pass or apply these below to the attributes of the new object:
> 
> 1. `None` as a value.
> 
> 2. The default value of an attribute when the object is created via its constructor with a minimal parameter input.
> 
> 
> 
> The second can be achieved by the `Default` class and object @ahopkins mentioned above. But `None` value still conflicts with the attributes from the old objects, meaning we cannot use `None` as the default value of the parameters of the `copy` method, otherwise we don't know whether or not the `None` is from the user.
> 
> 
> 
> I guess we must take a trade off when implementing this feature, 😂

The default values of all copy args should be `_default` for that reason.

ChihweiLHBird:
@ahopkins So, `_default` means the value of the attributes of the old object?

ahopkins:
> @ahopkins So, `_default` means the value of the attributes of the old object?

In this case it means "do nothing"

```python
if version != _default:
    # set version on new blueprint
```

ChihweiLHBird:
Thank you @ahopkins @Tronic 

Can you take a look and see how is it now?

sanic/blueprints.py
```diff
@@ -148,17 +149,33 @@ def signal(self, event: str, *args, **kwargs):
...
+        version_prefix: str = _default,
+        strict_slashes: Optional[bool] = _default,
     ):
+        """
+        Copy a blueprint instance with some optional parameters to 
+        override the values of attributes in the old instance.
+
+        :param name: unique name of the blueprint
+        :param url_prefix: URL to be prefixed before all route URLs
+        :param host: IP Address of FQDN for the sanic server to use.
```
ahopkins, line 163: Can remove this

sanic/blueprints.py
```diff
@@ -148,17 +149,33 @@ def signal(self, event: str, *args, **kwargs):
...
+        :param version: Blueprint Version
+        :param version_prefix: the prefix of the version number shown in the URL.
+        :param strict_slashes: Enforce the API urls are requested with a
+            training */*
```
ahopkins, line 167: typo: "trailing"

This is technically only part of `strict_slashes`. It enforces the path matching will meet the existence or non-existence of the trailing slash as defined.

sanic/blueprints.py
```diff
@@ -148,17 +149,33 @@ def signal(self, event: str, *args, **kwargs):
...
+        """
         new_bp = deepcopy(self)
```
ahopkins, line 169: I am not sure `deepcopy` alone will do the job. Once a blueprint is registered to an app, you might have problems with this. Perhaps the solution might just be to say that once it is registered there can be no copying. In which case we would need to make sure that is the case and raise an exception up front. (check bp._apps).

ahopkins (changes requested):
Thanks for putting this together. I think it will be a nice add. Make sure to add some unit tests.

sanic/blueprints.py
ChihweiLHBird: Thank you @ahopkins.
Let me try to find the reason and potential solution for this problem.

ChihweiLHBird: @ahopkins I found a workaround to remove all registered info in the the new bp instance (see code below for details). That way we can deepcopy the instance after it is being registered with some Sanic app. What do you think?
```
temp_attrs = {
    "_apps": self._apps,
    "routes": self.routes,
    "websocket_routes": self.websocket_routes,
   "middlewares": self.middlewares,
    "exceptions": self.exceptions,
    "listeners": self.listeners,
}
self._apps = set()
self.routes = []
self.websocket_routes = []
self.middlewares = []
self.exceptions = []
self.listeners = {}

new_bp = deepcopy(self)
        
#...... more code here ......

for key, value in temp_attrs.items():
    setattr(self, key, value)
```

ahopkins: That works for me. It might be nice to make it a little less verbose:

```python
set_aside = (
    ("_apps", set),
    ("routes", list),
    ...
)

temp_attrs = {}

for attr, factory in set_aside:
    temp_attrs[attr] = getattr(self, attr)
    setattr(self, attr, factory())

...
```

ahopkins: The more I think... I like this idea.

```python
def __init__(...):
    # set variables from args, remove the rest of the code that is now in reset
    self.reset()

def copy(self, ..., with_ctx=False):
    new_bp = deepcopy(self)
    new_bp.reset()

    options = OVERRIDE_VALUES_HERE

    for app in self._apps:
        new_bp.register(app, options)

def reset(self):
    # remove all of these from __init__
    self._apps: Set[Sanic] = set()
    self.exceptions: List[RouteHandler] = []
    self.listeners: Dict[str, List[ListenerType]] = {}
    self.middlewares: List[MiddlewareType] = []
    self.routes: List[Route] = []
    self.statics: List[RouteHandler] = []
    self.websocket_routes: List[Route] = []
```

ChihweiLHBird: @ahopkins I think `reset` is a good idea, however, we will still have the problem with `deepcopy` after the old bp being registered... We cannot directly reset old bp before copying it because we don't want to change anything in the old bp, so, we probably still need something like the temporarily reset of registration info mentioned above. We can have an option to allow users to keep the registration info from the old bp by doing something like:

```
if keep_old_registerations:
    for app in temp_attrs["_apps"]:
        app.blueprint(new_bp, **options)
```

ChihweiLHBird: By the way, I didn't understand what's meaning by `with_ctx` and `self.statics`.

I think `self.statics` is related to registration. Should we reset it during copying?

ahopkins: > we probably still need something like the temporarily reset of registration info mentioned above. We can backup the attrs, call `reset`, and then restore attrs after copying. We can also have an option to allow users to keep the registration info from the old bp by doing something like:

Perhaps we are overthinking that. It is only a handful of fields:

```python
def copy(self, name, ..., with_ctx=False, with_registration=True):
    options = {
        "host": host,
        "strict_slashes": strict_slashes,
        "url_prefix": url_prefix,
        "version": version,
        "version_prefix": version_prefix,
    }

    new_bp = Blueprint(name)
    for attr, value in options.items():
        value = getattr(self, attr) if value is _default else value
        setattr(attr, new_bp, value)

    if with_registration:
        for app in self._apps:
            new_bp.register(app, options)

    if not with_ctx:
        new_bp = SimpleNamespace()
```



> I think sometimes users don't want to keep registration from the old bp, so this could be an optional feature.

I suppose. Although, it is probably not often that someone is building more than one app in a given impl. More often than not, there is only one application to even register to. So, I would default it to `True`.



> I didn't understand what's meaning by `with_ctx` and `self.statics`

`with_ctx` meaning that the `blueprint.ctx` should be copied. Default to `False`

`self.statics` are static file handlers: `bp.static(...)`

ChihweiLHBird: @ahopkins I updated the branch, I consider we solved this issue. Can you review it and see if we can resolve this? Thanks! Please let me know if you have any suggestion.

ChihweiLHBird:
Not sure why linter failed...

sanic/blueprints.py
```diff
@@ -177,6 +188,21 @@ def copy(
         if not isinstance(version_prefix, Default):
             new_bp.version_prefix = version_prefix
 
+        for key, value in attrs_backup.items():
+            setattr(self, key, value)
+
+        if with_registration:
+            if len(new_bp._future_statics) > 0 and len(self._apps) > 0:
+                raise Exception(
```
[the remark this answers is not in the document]
ChihweiLHBird: I checked the `SanicException` and I found they are related to http errors. So I was not sure about whether to use it or its subclasses. Can I make a subclass of `SanicException` for this kind of problem? Or maybe just use `SanicException` directly?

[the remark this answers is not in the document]
ChihweiLHBird: Relaced with `SanicException`.

sanic/blueprints.py
```diff
@@ -177,6 +188,21 @@ def copy(
         if not isinstance(version_prefix, Default):
             new_bp.version_prefix = version_prefix
 
+        for key, value in attrs_backup.items():
+            setattr(self, key, value)
+
+        if with_registration:
+            if len(new_bp._future_statics) > 0 and len(self._apps) > 0:
```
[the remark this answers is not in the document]
ChihweiLHBird: Thank you for your suggestion, this make sense.

sanic/blueprints.py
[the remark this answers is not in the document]
ahopkins: FYI, you only get the http code logic if there is a status code along with it.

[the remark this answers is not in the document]
ChihweiLHBird: Thank you!

sanic/blueprints.py
```diff
@@ -77,14 +79,9 @@ def __init__(
...
 
-        self._apps: Set[Sanic] = set()
         self.ctx = SimpleNamespace()
-        self.exceptions: List[RouteHandler] = []
         self.host = host
-        self.listeners: Dict[str, List[ListenerType]] = {}
-        self.middlewares: List[MiddlewareType] = []
-        self.routes: List[Route] = []
-        self.statics: List[RouteHandler] = []
+        self.reset()
```
ahopkins, line 84: Why is `reset()` in the middle of setting host, strict_slashes, etc?

tests/test_blueprint_copy.py
```diff
@@ -0,0 +1,40 @@
...
+    assert id(bp1) != id(bp2)
+
+    app.blueprint(bp1)
+    app.blueprint(bp2)
+    bp3 = bp1.copy(name="test_bp3", version=3, with_registration=True)
+    bp4 = bp1.copy(name="test_bp4", version=4, with_ctx=True)
+    bp5 = bp1.copy(name="test_bp5", version=5, with_registration=False)
+    assert id(bp2) != id(bp3)
+    assert id(bp3) != id(bp4)
+    assert id(bp4) != id(bp5)
```
ahopkins, line 24: It is nice that we have bp[1-5], but we should expand these tests to include a wider variety of properties.

ahopkins (changes requested):

sanic/blueprints.py
```diff
@@ -93,6 +90,14 @@ def __init__(
         )
         self.version = version
         self.version_prefix = version_prefix
+
+    def reset(self):
```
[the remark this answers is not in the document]
ChihweiLHBird: Yeah, fixed.

sanic/blueprints.py
```diff
@@ -144,6 +140,81 @@ def signal(self, event: str, *args, **kwargs):
...
+            if new_bp._future_statics:
+                raise SanicException(
+                    "Static routes registered with the old blueprint instance,"
+                    " cannot be registered again."
+                )
+            for app in self._apps:
+                app.blueprint(new_bp)
+
+        if not with_ctx:
+            new_bp.ctx = SimpleNamespace()
```
ahopkins, line 214: Should `ctx` be one of the items that is backed up and reset since we do not know what it will contain and whether we can copy it?

ahopkins (commented):
Okay... now that I closed some of the conversations so it is easier to follow, this LGTM. Just a couple more comments and we should be good to wrap this up. Thanks for all the work on this one!! :rocket:

sanic/blueprints.py
ChihweiLHBird: Fixed

sanic/blueprints.py
ChihweiLHBird: ctx is copyable. The variables are not copyable because they use `Immutable` defined in sanic_routing. 

Basically, `_future_xxxxxx` (e.g. `_future_middleware`) is copyable and `xxxxxx` (e.g. `middlewares`) is not copyable.

```python

class Immutable(dict):
    def __setitem__(self, *args):
        print(self)
        raise TypeError("Cannot change immutable dict")

    def __delitem__(self, *args):
        raise TypeError("Cannot change immutable dict")
```

tests/test_blueprint_copy.py
ChihweiLHBird: Test case has been updated. Considered this as solved. Can you take a look again? Thanks!

sanic/blueprints.py
```diff
@@ -144,6 +139,81 @@ def signal(self, event: str, *args, **kwargs):
...
+        :param name: unique name of the blueprint
+        :param url_prefix: URL to be prefixed before all route URLs
+        :param version: Blueprint Version
+        :param version_prefix: the prefix of the version number shown in the
+            URL.
+        :param strict_slashes: Enforce the API urls are requested with a
+            trailing */*
+        :param with_registration: whether register new blueprint instance with
+            sanic apps that were registered with the old instance or not.
+        :param with_ctx: whether `ctx` will be copied or not.
```
ahopkins, line 174:
```suggestion
        :param with_ctx: whether ``ctx`` will be copied or not.
```

Uses RTD

ahopkins (approved):

ahopkins:
```
  >           raise FinalizationError("Cannot finalize router more than once.")
```

On this line:

```
           bp6 = bp1.copy(
              name="test_bp6",
              version=6,
              with_registration=True,
  >           version_prefix="/version",
          )
  
  tests/test_blueprint_copy.py:64: 
```

---

Easy solve:

before the `b6` copy, add:

```python
app.router.reset()
```

ahopkins (approved):

PR #2188, opened 2021-07-08
Fix the handling of the end of a chunked request.

Tronic:
Fixes a bug introduced in the streaming branch where a quick but incorrect fix was merged in #2183.

### DIFF
--- a/sanic/http.py
+++ b/sanic/http.py
@@ -486,19 +486,22 @@ class Http:
                 self.keep_alive = False
                 raise InvalidUsage("Bad chunked encoding")
 
-            del buf[: pos + 2]
-
             if size <= 0:
                 self.request_body = None
-                # Because we are leaving one CRLF in the buffer, we manually
-                # reset the buffer here
-                self.recv_buffer = bytearray()
 
                 if size < 0:
                     self.keep_alive = False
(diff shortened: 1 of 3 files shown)


Tronic:
If the following packet was received, the streaming branch left `\r\nGET...` in buffer after the POST request was handled. The earlier attempt at removing that CRLF would also end up removing the GET request (in this case all of it). It is crucial that only the CRLF is removed, nothing more.

```
POST / HTTP/1.1\r\n
transfer-encoding: chunked\r\n
\r\n
0\r\n
\r\n
GET / HTTP/1.1\r\n
\r\n
```

Secondly, there is no guarantee that the CRLF to be removed has been received yet. Consider the following:
```
POST / HTTP/1.1\r\n
transfer-encoding: chunked\r\n
\r\n
0\r\n
(pause sending here, so that the buffer flushing code runs)
\r\n
GET / HTTP/1.1\r\n
\r\n
```

Once again there is a CRLF left in buffer causing problems for the next request. Thus, we need to wait until two CRLF after the zero are received before we can return from request handling, and be sure to only remove those but to keep everything else that there might be in the buffer in tact.

ahopkins (approved):

Issue #2189, opened 2021-07-09
Pending tasks on exit


DirtyHairy:
**Describe the bug**
Sanic does not wait for the `connection_task()` tasks to terminate after cleaning up non-idle connections after `GRACEFUL_SHUTDOWN_TIMEOUT`. This can cause dangling tasks on exit:

```
Task was destroyed but it is pending!
task: <Task pending name='Task-3' coro=<HttpProtocol.connection_task() done, defined at /Users/pestix/.virtualenvs/server-25sdMkop/lib/python3.9/site-packages/sanic/server.py:188> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x108263fa0>()]>>
```

**Code snippet**
I can reproduce this reliably by running a simple server with a SSL context and doing the same dance as in #2186 in Chrome. The surplus connection opened by Chrome will be terminated after `GRACEFUL_SHUTDOWN_TIMEOUT` expires, and I get the above warning.

**Expected behavior**
Clean exit.

**Environment (please complete the following information):**
 - OS: MacOS
 - Version: 21.6.0

**Additional context**
A possible workaround on the consumer side is

```
    @app.listener("after_server_stop")
    async def afterServerStop(app, loop):
        await asyncio.gather(*[task for task in asyncio.all_tasks() if task != asyncio.current_task()])
```

However, this exposes #2190 😛

ahopkins:
Resolved in #2231

PR #2191, opened 2021-07-10
Avoid dangling sockets by using abort() instead of close()

### LINKED ISSUE #2190: Transport may remain open after `close()`
**Describe the bug**
After applying the workaround in #2189 Sanic hangs on shutdown if it tries to force close SSL connection after `GRACEFUL_SHUTDOWN_TIMEOUT` expires. Digging into the code shows that the transport does not close after `close()` has been called but stays open for a prolonged time (I guess it will timeout eventually), and `connection_lost` is not called.

There are other reports of the same issue with asyncio SSL transports on the web, i.e. this [stack overflow thread](https://stackoverflow.com/questions/65168635/python-asyncio-ssl-transport-not-closing). I don't think this is really a bug on Sanic's end, but more of a pitfall with the transport.

Changing `HttpProtocol::close` to call `abort()` instead of `close()` fixes the issue. As this is a non-graceful shutdown of the connection anyway I think the change would be acceptable, but of course something more elaborate (like waiting another timeout before aborting the transports) would be possible, too.

I can prepare a PR if you like.

**Code snippet**
This can be reproduced the same way as #2189 --- launch a simple server with SSL, connect using Chrome and terminate. If

```
    @app.listener("after_server_stop")
    async def afterServerStop(app, loop):
        await asyncio.gather(*[task for task in asyncio.all_tasks() if task != asyncio.current_task()])
```

is added to the server it will hang on exit. Debugging shows that `connection_lost` is never called.

**Expected behavior**
The transport closes.

 - OS: MacOS
 - Version: 21.6.0

DirtyHairy:
Fixes #2190

### DIFF
--- a/sanic/server.py
+++ b/sanic/server.py
@@ -286,7 +286,7 @@ class HttpProtocol(asyncio.Protocol):
         """
         # Cause a call to connection_lost where further cleanup occurs
         if self.transport:
-            self.transport.close()
+            self.transport.abort()
             self.transport = None
 
     # -------------------------------------------- #


Tronic (approved):
Seems acceptable, considering that this is force closing only after graceful shutdown timeout.

ahopkins:
I am not sure this is _always_  the desirable outcome.

`HttpProtocol.close()` is called in the following cases:

1. At the end of `connection_task`
1. When `close_if_idle()` is called (in graceful server shutdown `serve()` before the timeout)
1. At the completion of attempting a graceful shutdown
1. When `data_received` is called with falsey value

I think it might be okay to (1) add some sort of a timeout as you suggested in #2190, and/or (2) add an `abort=True` flag for the cases where we know it is safe.

It seems more like we need to look deeper into #2189 to see why that task is being terminated and we are not waiting for it to complete.

I think 1 and 3 are probably safe for `abort()`.

Perhaps another alternative is to run the `close_if_idle()` loop once again after the `GRACEFUL_SHUTDOWN_TIMEOUT`, before attempting an `abort()`. Or, just moving `close_if_idle()` to after timeout in general. Presumably, this would provide time for them to close on their own.

ahopkins (changes requested):
`pause`

ahopkins:
@DirtyHairy Are you looking further into this?

DirtyHairy:
@ahopkins Not at the moment, sorry. I don't know the code base well enough for a quick fix other than this PR (your criticism of which is valid), and I don't have time to dig into this deeper, in particular since I switched my current project to aiohttp and don't have another project using Sanic atm.

ahopkins:
Replaced by #2231

PR #2193, opened 2021-07-12
Change dumps to AnyStr

ahopkins:
A dumps method that returns a bytes string (like orjson.dumps) is valid. So, the type annotation needs to be `AnyStr`.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -738,11 +738,12 @@ class Sanic(BaseSanic):
             request.route = route
 
             if (
-                request.stream.request_body  # type: ignore
+                request.stream
+                and request.stream.request_body
                 and not route.ctx.ignore_body
             ):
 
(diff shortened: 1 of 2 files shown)


sanic/app.py
```diff
@@ -737,15 +738,14 @@ async def handle_request(self, request: Request):
             request.route = route
 
             if (
-                request.stream.request_body  # type: ignore
+                request.stream
+                and request.stream.request_body
                 and not route.ctx.ignore_body
             ):
```
Tronic, line 744: This change seems unrelated, accidentally included in this PR?

ahopkins: `mypy` was complaining. We could use a type ignore, but I think this should be a light enough change.

Tronic (approved):

PR #2194, opened 2021-07-12
HTTP tests

ahopkins:
Introduces a new "client" for running some raw HTTP request testing.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -10,3 +10,15 @@ exclude_patterns:
   - "examples/"
   - "hack/"
   - "scripts/"
+  - "tests/"
+checks:
+  argument-count:
+    enabled: false
+  file-lines:
+    config:
+      threshold: 1000
+  method-count:
+    config:
+      threshold: 40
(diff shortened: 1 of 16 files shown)


Tronic:
I would suggest using `str` rather than `bytes` if you are about to perform any "cleanup", which is definitely unexpected to occur when bytes are passed. This cleanup puts some restrictions on what can be tested (no initial whitespace or mixed newlines permitted). Maybe take str which go through cleanup and also bytes (actually raw) to allow testing for such special cases?

Instead of lstrip, `.removeprefix("\n")` is more appropriate for removing the initial newline.

ahopkins:
removeprefix is python 3.9+.

ahopkins:
@Tronic New version only runs `clean` if you pass it a `str`. This way we have the convenience of simpler formatting:

```
    client.send(
        """
        POST /upload HTTP/1.1
        transfer-encoding: chunked

        """
    )
```
Or, just passing `bytes` to have complete control.

Tronic (approved):

PR #2196, opened 2021-07-12
style(app,blueprints): add some type hints

SaidBySolo:
I'm using pylance's strict mode.

This PR makes my linter a little less complaining.


sanic/app.py
```diff
@@ -411,7 +411,11 @@ def enable_websocket(self, enable=True):
 
         self.websocket_enabled = enable
 
-    def blueprint(self, blueprint, **options):
+    def blueprint(
+        self,
+        blueprint: Union[Blueprint, list[Blueprint], tuple[Blueprint], BlueprintGroup],
```
ahopkins: We should use `List` and `Tuple` since we also need to support 3.7

sanic/app.py
```diff
@@ -869,7 +873,7 @@ def run(
         *,
         debug: bool = False,
         auto_reload: Optional[bool] = None,
-        ssl: Union[dict, SSLContext, None] = None,
+        ssl: Union[dict[str,str], SSLContext, None] = None,
```
ahopkins, line 877:
```suggestion
        ssl: Union[Dict[str, str], SSLContext, None] = None,
```

sanic/app.py
```diff
@@ -999,7 +1003,7 @@ async def create_server(
         port: Optional[int] = None,
         *,
         debug: bool = False,
-        ssl: Union[dict, SSLContext, None] = None,
+        ssl: Union[dict[str,str], SSLContext, None] = None,
```
ahopkins, line 1007: Same

sanic/blueprints.py
```diff
@@ -146,10 +146,10 @@ def signal(self, event: str, *args, **kwargs):
 
     @staticmethod
     def group(
-        *blueprints,
-        url_prefix="",
-        version=None,
-        strict_slashes=None,
+        *blueprints: Union[Blueprint, BlueprintGroup],
+        url_prefix: Optional[str] = "",
```
ahopkins, line 150: I am not sure we do accept `None` as a value. I think this should only be `str`.

ahopkins (changes requested):

sanic/app.py
```diff
@@ -411,7 +412,11 @@ def enable_websocket(self, enable=True):
 
         self.websocket_enabled = enable
 
-    def blueprint(self, blueprint, **options):
+    def blueprint(
+        self,
+        blueprint: Union[Blueprint, List[Blueprint], Tuple[Blueprint], BlueprintGroup],
```
ahopkins, line 417: Instead of `List` and `Tuple`, maybe just `Sequence`

sanic/blueprints.py
SaidBySolo: Referenced here. Which one would be more accurate?

https://github.com/sanic-org/sanic/blob/ac784759d56428bcb714eab59caeb92f58391b96/sanic/blueprints.py#L72

ahopkins: Hmm... maybe an inconsistency? Let me jump into and see if there is a reason it is `""` for a default there and not `None`.

ahopkins: Looks like an inconsistency. The signature in `BlueprintGroup` is correct. Looking at the implementation in `Sanic.blueprint`, a `None` value is safe.

So, let's keep it as `Optional`, but also update the default value to `None`.

ahopkins (approved):
Thanks for putting this together. :clap:

sanic/app.py
SaidBySolo, line 417: Maybe i have to specify it separately as a variable?

ahopkins: Run `make pretty`

SaidBySolo: Done ! :D

ahopkins (approved):

PR #2200, opened 2021-07-18
Accept header parsing

### LINKED ISSUE #2171: RFC: Accept header parsing
The `Accept` header in RFC 7231 § 5.3.2 has a very structured format for specifying media types that the client _can_ accept, and also would prefer to accept.

[Source](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2)

    A more elaborate example is

             Accept: text/plain; q=0.5, text/html,
                 text/x-dvi; q=0.8, text/x-c

       Verbally, this would be interpreted as "text/html and text/x-c are
       the equally preferred media types, but if they do not exist, then
       send the text/x-dvi representation, and if that does not exist, send
       the text/plain representation".

       Media ranges can be overridden by more specific media ranges or
       specific media types.  If more than one media range applies to a
       given type, the most specific reference has precedence.  For example,

             Accept: text/*, text/plain, text/plain;format=flowed, */*

       have the following precedence:

       1.  text/plain;format=flowed
       2.  text/plain
       3.  text/*
       4.  */*

#2162 adds some context based error formatting based upon the initial #1937. Where both of those PRs are lacking is the ability to confirm that the media being sent is warranted and acceptable.

The proposal is to add a new accessor: `Request.accept` that provides a simplified list of acceptable types in preference order. Where there is a tie for equal precedence, Sanic should prefer the first to be listed in the raw headers.

This change will allow for the error response to make sure that it is sending an appropriate format, and also provides the developer an easy way to select one of multiple formats for responding. The last potential benefit is that Sanic could offer a "strict mode" (defaul=False) that would raise an error if it attempts to respond with a type that is not acceptable.

ahopkins:
Resolves #2171 

This adds a new `request.accept` property that will breakdown the `Accept` header and order the values according to their relative weights.

Originally I was not going to add any sort of wildcard matching logic. But, it was really very simple to add it.

```python
# Accept: text/*
assert "text/html" in request.accept
```


sanic/request.py
```diff
@@ -296,6 +298,11 @@ def load_json(self, loads=json_loads):
 
         return self.parsed_json
 
+    @property
+    def accept(self) -> List[Accept]:
+        accept_header = self.headers.getone("accept", "")
+        return parse_accept(accept_header)
+
```
Tronic, line 305: Should this be cached? And could we have a decorator for caching instead of doing it manually like in most other `request` accessors? (does that affect performance?)

sanic/headers.py
```diff
@@ -30,6 +33,121 @@
...
+    def __eq__(self, other):
+        other_is_wildcard = (
+            other.is_wildcard
+            if isinstance(other, MediaType)
+            else self.check_if_wildcard(other)
+        )
+        other_value = other.value if isinstance(other, MediaType) else other
+        return (
+            self.value == other_value or self.is_wildcard or other_is_wildcard
+        )
```
Tronic, line 62: Suggest `if ... return True` on wildcard checks to avoid useless processing.

Tronic:
The problem with wildcard matching is that way too often `*/*` is included as the sole option or after all the others. Because of this, e.g. an exception formatter cannot use `"text/html" in request.accept` for any meaningful results. Could we have an alternative API that does no wildcard matching, or that at least ignores `*/*`?

ahopkins:
That's a good idea. I uncovered a bug in the wildcard matching anyway. I'll make it toggleable by flag, default false.

sanic/request.py
ahopkins: That's a good idea.

ahopkins: The reason I did not implement is that I did not want to waste resources adding extra values to the request object that would not be used in most scenarios. Maybe however the solution is to have a single `_cache` object that can contain these values.

sanic/headers.py
ahopkins: As soon as the first condition is met, it will exit. If `self.value == other_value` is `True`, Python will not proceed to `self.is_wildcard` etc.

```python
>>> foo = 123
>>> bar = 123
>>> if foo == bar or foo.something_that_does_not_exist:
...     print("hi")
... 
hi
```

ahopkins:
> That's a good idea. I uncovered a bug in the wildcard matching anyway. I'll make it toggleable by flag, default false.

Default is `allow_type_wildcard=True` and `allow_subtype_wildcard=True`. But, can be set here:

```python
assert not Accept.parse("foo/bar").match("*/*", allow_type_wildcard=False)
```

Tronic (dismissed):
Approving, although the code is quite scary and there is a lot more of it than what I'd like for such a small feature (e.g. the comparison operators).

PR #2208, opened 2021-08-01
Blueprint specific exception handlers

### LINKED ISSUE #2202: Cannot define a blueprint-specific exception handler
**Describe the bug**
The documentation at https://sanicframework.org/en/guide/best-practices/blueprints.html#exceptions says that:

> Just like other exception handling, you can define blueprint specific handlers.

However, exception handlers defined this way don't seem to be blueprint-specific at all. Instead, they handle exceptions in other blueprints as well.

**Code snippet**
```python
#!/usr/bin/env python3

from sanic import Sanic, Blueprint, response

class Error(Exception):
    pass

handled = Blueprint("handled")
@handled.exception(Error)
def handle_error(req, e):
    return response.text("handled {}".format(e))

b = Blueprint("b")
@b.route("/")
def e(request):
    raise Error("error in e")

app = Sanic(__name__)
app.blueprint(handled)
app.blueprint(b)
app.run()
```

**Expected behavior**
A request to http://localhost:8000 should generate 500, because `Error` should not be handled.
Instead, the request generates 200, `Error` is handled, even though the handler is registered in `handled`, while the endpoint is in `b`.

 - OS: ubuntu 19.10
 - Version: Sanic 21.6.0; Routing 0.7.0

**Additional context**
The order of blueprint registration does not seem to change anything. Not registering blueprint `handled` obviously fixes the issue.

Is it a bug or am I missing something here?

ahopkins:
Resolves #2202


ashleysommer (approved):
Wow, I've looked at implementing blueprint-specific exceptions like this in the past, but I've always thought it wasn't really possible without huge refactoring.
Well done.

PR #2209, opened 2021-08-02
Add convenience methods to BP groups

ahopkins:
Adds the `on_request` and `on_response` convenience decorators to BP Groups.


tests/test_blueprint_group.py
```diff
@@ -116,6 +116,16 @@ def blueprint_group_1_middleware_not_called(request):
         global MIDDLEWARE_INVOKE_COUNTER
         MIDDLEWARE_INVOKE_COUNTER["request"] += 1
 
+    @blueprint_group_1.on_request
+    def blueprint_group_1_convenience_1(request):
+        global MIDDLEWARE_INVOKE_COUNTER
+        MIDDLEWARE_INVOKE_COUNTER["request"] += 1
```
Tronic, line 122: I don't think you need `global` in these because the variable is not being assigned to but rather used via `__setitem__` operator.

Tronic (approved):

PR #2215, opened 2021-08-06
Try to Fix Windows GitHub Workflow PIP Setup

ChihweiLHBird:


ahopkins:
What's the change? What is failing?

ahopkins:
Thanks so much for the fix! We should really move that repo to SCO. I am closing this for now since it is not needed.

ChihweiLHBird:
> Thanks so much for the fix! We should really move that repo to SCO. I am closing this for now since it is not needed.

Thank you, Adam. I am not sure what's SCO, can you tell something about it? 😂

PR #2217, opened 2021-08-08
Style: add type hints

ombe1229:
Add some type hints for strict mode

### DIFF
--- a/sanic/mixins/listeners.py
+++ b/sanic/mixins/listeners.py
@@ -6,9 +6,6 @@ from sanic.models.futures import FutureListener
 from sanic.models.handler_types import ListenerType
 
 
-Listener = Union[ListenerType, str]
-
-
 class ListenerEvent(str, Enum):
     def _generate_next_value_(name: str, *args) -> str:  # type: ignore
         return name.lower()
@@ -30,7 +27,7 @@ class ListenerMixin:
 
     def listener(
         self,
-        listener_or_event: Listener,
+        listener_or_event: Union[ListenerType, str],
         event_or_none: Optional[str] = None,
         apply: bool = True,
     ):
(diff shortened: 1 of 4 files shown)


ahopkins:
Can you take a look at `./sanic/models/handler_types.py`? I think a lot of what you are trying to implement may be there already.

sanic/mixins/routes.py
```diff
@@ -168,7 +169,7 @@ def decorator(handler):
 
     def add_route(
         self,
-        handler: Union[Coroutine[Any, Any, HTTPResponse], HTTPMethodView],
+        handler: MiddlewareResponse,
```
ahopkins, line 172: Why `MiddlewareResponse`?

sanic/views.py
```diff
@@ -86,7 +86,7 @@ def dispatch_request(self, request, *args, **kwargs):
         return handler(request, *args, **kwargs)
 
     @classmethod
-    def as_view(cls, *class_args, **class_kwargs):
+    def as_view(cls, *class_args: Any, **class_kwargs: Any) -> Any:
```
ahopkins, line 89: The return value should be a handler.

sanic/mixins/routes.py
ombe1229: I missed it. I'll fix that

sanic/mixins/listeners.py
```diff
@@ -1,9 +1,12 @@
 from enum import Enum, auto
 from functools import partial
-from typing import Any, Callable, Coroutine, List, Optional, Union
+from sanic.models.handler_types import SignalHandler
+from typing import Any, Callable, Coroutine, List, Optional, TypeVar, Union
 
 from sanic.models.futures import FutureListener
 
+L = TypeVar("L", bound=SignalHandler)
```
ahopkins, line 8: Why `TypeVar` here? I think you meant `ListenerType`. Also, `L` is not a very descriptive variable name.

ahopkins:
I think if we are going to be taking this step, there are probably some more opportunities here. Do you mind if I push a commit to your branch?

ombe1229:
Yes, please. I think there's something I didn't understand exactly.

ahopkins (approved):

PR #2224, opened 2021-08-29
Explicit static directive for serving file or dir

### LINKED ISSUE #2222: Explicit static directive for serving file or dir
See #2132

```python
app.static('/', './resources/web', as="dir")
app.static('/', './resources/web/example.html', as="file")
```

With an explicit kwarg, we would not need to run `path.isfile`. This is not really new behavior. If Sanic 20.12 is not functioning this way, I think that is more a problem on the earlier version than with the current.

---

A bit of behind the scenes info...

Let's look at what these two are doing:

```python
app.static("/", "./resources/web")
# Sanic checks this, sees that it is not a file and then converts to:
# "/<__file_uri__:path>"
# meaning that it will look for anything that matches using the `path` param type

app.static('/', './resources/web/example.html')
# If this is indeed a file, Sanic knows that you meant for an explicit path "/"
# There is no ambiguity, and the path does not need to be altered
# If the file does not exist, then like the first example, it tries to convert it to a `path` type
# But, since you already have a path expansion on that base path, there is now ambiguity
# and Sanic raises RouteExists
```

__Originally posted by @ahopkins in https://github.com/sanic-org/sanic/issues/2132#issuecomment-839477709__

anbuhckr:
Closes #2222

### DIFF
--- a/sanic/blueprint_group.py
+++ b/sanic/blueprint_group.py
@@ -197,6 +197,27 @@ class BlueprintGroup(MutableSequence):
         """
         self._blueprints.append(value)
 
+    def exception(self, *exceptions, **kwargs):
+        """
+        A decorator that can be used to implement a global exception handler
+        for all the Blueprints that belong to this Blueprint Group.
+
+        In case of nested Blueprint Groups, the same handler is applied
+        across each of the Blueprints recursively.
+
+        :param args: List of Python exceptions to be caught by the handler
+        :param kwargs: Additional optional arguments to be passed to the
+            exception handler
+        :return a decorated method to handle global exceptions for any
+            blueprint registered under this group.
+        """
+
+        def register_exception_handler_for_blueprints(fn):
+            for blueprint in self.blueprints:
+                blueprint.exception(*exceptions, **kwargs)(fn)
+
+        return register_exception_handler_for_blueprints
+
     def insert(self, index: int, item: Blueprint) -> None:
         """
         The Abstract class `MutableSequence` leverages this insert method to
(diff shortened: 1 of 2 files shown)


sanic/mixins/routes.py
```diff
@@ -831,13 +832,17 @@ def _register_static(
...
+        name = static.name        
         # If we're not trying to match a file directly,
-        # serve from the folder
-        if not path.isfile(file_or_directory):
+        # serve from the folder        
+        if not static.src and not path.isfile(file_or_directory):
+            uri += "/<__file_uri__:path>"
+        elif static.src == "dir": 
             uri += "/<__file_uri__:path>"
+        elif static.src != "file":
```
ahopkins, line 844: What if `src` is `None` and it is a file (so the first check is not triggered?)

ahopkins: Are you sure about that? We need to add some unit testing. Add some unit tests for all the possible combinations to make sure that the code behaves as expected. Feel free to use an existing test as a template.

sanic/mixins/routes.py
```diff
@@ -592,6 +592,7 @@ def static(
         strict_slashes=None,
         content_type=None,
         apply=True,
+        src=None,
```
ahopkins, line 595: How about `resource_type`?

anbuhckr: okay

.gitignore
```diff
@@ -21,4 +21,5 @@ dist/*
 pip-wheel-metadata/
 .pytest_cache/*
 .venv/*
+.test-sanic/*
```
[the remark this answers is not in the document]
anbuhckr: first time using venv, thats my venv folder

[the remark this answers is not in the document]
ahopkins: Can you make it use `.venv` so we do not need to add unique config?

PR #2229, opened 2021-08-31
Split HttpProtocol parts into base SanicProtocol and HTTPProtocol subclass

ashleysommer:
This is the first change needed before new Websockets changes can land.
This is a big enough and self-contained enough change to be its own PR.


ahopkins (approved):
Looks straightforward.

PR #2230, opened 2021-08-31
Move server.py into its own module

ahopkins:
Following @ashleysommer #2229 where he split the protocol into a base, this PR takes the entire `server.py` and breaks it into a module with smaller components.

There should be no functionality changes. To keep things backwards compat, a bunch of the objects are exposed still as `sanic.server.___`. Not all. I didn't think things like `trigger_events` and some of the other internal functions needed to be exposed. But, if anyone thinks otherwise, LMK.


ashleysommer (approved):
Awesome. Much cleaner.

Tronic (approved):
This refactoring was badly needed.

ahopkins:
Next up is app.py :dizzy_face:

PR #2231, opened 2021-09-01
Call abort() on sockets after close() to prevent dangling sockets

### LINKED ISSUE #2190: Transport may remain open after `close()`

ashleysommer:
Fixes #2190
Replaces #2191 

This adds some help helper functionality to the base SanicProtocol
Added `SanicProtocol.abort()` method to force-close a connection to prevent dangling sockets.
Added optional timeout arg to `SanicProtocol.close()` which waits a given amount of time after calling `close()` to call `abort()` if the connection is still open.
Defaults to the same timeout period as `GRACEFUL_SHUTDOWN_TIMEOUT` (for convenience).

Also change one usage of `.close()` to `.abort()` outright, on server shutdown when `GRACEFUL_SHUTDOWN_TIMEOUT` has already passed.

### DIFF
--- a/sanic/server/protocols/base_protocol.py
+++ b/sanic/server/protocols/base_protocol.py
@@ -81,13 +81,24 @@ class SanicProtocol(asyncio.Protocol):
         self._data_received.clear()
         await self._data_received.wait()
 
-    def close(self):
+    def close(self, timeout: Optional[float] = None):
+        """
+        Attempt close the connection.
+        """
+        # Cause a call to connection_lost where further cleanup occurs
+        if self.transport:
+            self.transport.close()
+            if timeout is None:
+                timeout = self.app.config.GRACEFUL_SHUTDOWN_TIMEOUT
+            self.loop.call_later(timeout, self.abort)
+
+    def abort(self):
         """
         Force close the connection.
         """
         # Cause a call to connection_lost where further cleanup occurs
         if self.transport:
-            self.transport.close()
+            self.transport.abort()
             self.transport = None
 
     # asyncio.Protocol API Callbacks #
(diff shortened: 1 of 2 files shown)


ashleysommer:
Test failures are caused by loop closing in the test suite, while the _abort_task Task is pending.
Even thought _abort_task is cancelled, it doesn't kill the task until next cycle of the event loop, and the loop is closed before that. So the _abort_task is never really cancelled.

ashleysommer:
I force-pushed a much simpler implementation, that should solve the problem, but does not leave a dangling task when loop is destroyed.

ahopkins (approved):
Nice work. TY

PR #2237, opened 2021-09-11
Add new CLI options

### LINKED ISSUE #2169: Production and development modes (default workers)
While it is possible to specify these individual features by hand, I would rather see a simple mode flag to get sensible defaults:
- Production: no access logs, workers=os.cpu_count()
- Development: debug, access logs, auto-reloader

Sanic server by default runs with a single worker process and doesn't enable features that would be useful in development. Many people accidentally use these default settings for benchmarking and production but the performance is poor due to lack of multiprocessing and due to logging if `app.run` was used. The debug mode cannot be enabled by default because running it in production is a potential security risk, so making the production mode the new default seems a better option.

Notably the reloader setting by default already follows the debug flag. I believe this should done with the access log setting too (currently `app.run(debug=True)` and `app.run(debug=False)` both enable access logs, while `sanic` CLI disables logs irregardless of the `--debug` flag). Alike, the number of workers could depend on the debug flag, keeping the single worker default in debug mode, but using the number of CPUs by default.

By these changes we get **production mode by default**, and adding `debug=True` or `sanic --debug` would accomplish full switch of all relevant settings into development mode.

This is potentially a breaking change with application servers that run multiple Sanic instances on different ports with the intention that each Sanic instance stays a single process. If this is a problem, specifying workers=1 in these installs is a simple and backwards-compatible fix.

Relevant `app.run` parts shown with changes:
```python
    def run(
        debug: bool = False,
        auto_reload: Optional[bool] = None,  # already follows debug flag if not set
        workers: Optional[int] = None,   # instead of: workers: int = 1
        access_log: Optional[bool] = None,  # was already this but doesn't work as expected
    ) -> None:
        # New logic
        if workers is None:
            workers = 1 if debug else os.cpu_count()
        if access_log is None:
            access_log = debug
```

I'd like to hear your thoughts before implementing this. This can probably be done better, or perhaps I overlooked something.

ahopkins:
Implements and closes #2169 

[As mentioned here](https://github.com/sanic-org/sanic/issues/2169#issuecomment-917472800), this adds some new flags the the CLI:

- `--mode` Can be either `debug` or `prod`
- PROD mode is new, it bumps logging up to `WARNING`
- DEBUG mode will deprecate auto-reload
- `--dev` is a shortcut to DEBUG mode + auto-reload
- `--fast` is a shortcut to max out the available workers

### DIFF
--- a/sanic/__main__.py
+++ b/sanic/__main__.py
@@ -226,9 +226,7 @@ def main():
             try:
                 args.workers = len(os.sched_getaffinity(0))
             except AttributeError:
-                error_logger.warning(
-                    "Ignoring '--fast' since it is not supported on this OS."
-                )
+                args.workers = os.cpu_count()
 
         kwargs = {
             "host": args.host,
--- a/sanic/__main__.py
+++ b/sanic/__main__.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import sys
 
@@ -11,7 +12,7 @@ from sanic_routing import __version__ as __routing_version__  # type: ignore
 from sanic import __version__
 from sanic.app import Sanic
 from sanic.config import BASE_LOGO
-from sanic.log import error_logger
+from sanic.log import error_logger, logger
 from sanic.simple import create_simple_server
 
 
@@ -87,15 +88,53 @@ def main():
     parser.add_bool_arguments(
         "--access-logs", dest="access_log", help="display access logs"
     )
-    parser.add_argument(
+    group_worker = parser.add_mutually_exclusive_group()
+    group_worker.add_argument(
         "-w",
         "--workers",
         dest="workers",
         type=int,
         default=1,
-        help="number of worker processes [default 1]\n ",
+        help="Number of worker processes [default 1]",
     )
-    parser.add_argument("-d", "--debug", dest="debug", action="store_true")
+    group_worker.add_argument(
+        "--fast",
+        action="store_true",
+        help=(
+            "Run with maximum allowable processes.\nThis should be equal "
+            "to the number of CPU processes available.\n "
+        ),
+    )
+    group_mode = parser.add_mutually_exclusive_group()
+    group_mode.add_argument(
+        "--mode",
+        help="Set the mode",
+        nargs="?",
+        choices=("debug", "prod"),
+        default="default",
+    )
+    group_mode.add_argument(
+        "-d",
+        "--debug",
+        dest="mode",
+        const="debug",
+        action="store_const",
+        help="Shortcut to set to DEBUG mode",
+    )
+    group_mode.add_argument(
+        "--prod",
+        dest="mode",
+        const="prod",
+        action="store_const",
+        help="Shortcut to set to PROD mode",
+    )
+    group_mode.add_argument(
+        "--dev",
+        dest="mode",
+        const="dev",
+        action="store_const",
+        help="Shortcut to set to DEBUG mode, and enables auto-reload\n ",
+    )
     parser.add_argument(
         "-r",
         "--reload",
@@ -116,10 +155,36 @@ def main():
         help=(
             "Path to your Sanic app. Example: path.to.server:app\n"
             "If running a Simple Server, path to directory to serve. "
-            "Example: ./\n"
+            "Example: ./"
         ),
     )
     args = parser.parse_args()
+
+    if "-d" in sys.argv:
+        error_logger.error(
+            DeprecationWarning(
+                "The -d flag has been deprecated and will be removed in v22.3"
+            )
+        )
+
+    if args.mode == "dev":
+        args.mode = "debug"
+        args.auto_reload = True
+    elif (
+        args.mode == "default"
+        and not args.access_log
+        and "--no-access-logs" not in sys.argv
+    ):
+        args.access_log = True
+        error_logger.error(
+            DeprecationWarning(
+                "Default access logging has been deprecated. Beginning v22.3 "
+                "you must either expressly use '--access-logs', or run in "
+                "DEBUG mode to display them."
+            )
+        )
+    elif args.mode == "prod":
+        args.access_log = False
 
     try:
         module_path = os.path.abspath(os.getcwd())
@@ -157,12 +222,20 @@ def main():
         else:
             ssl = None
 
+        if args.fast:
+            try:
+                args.workers = len(os.sched_getaffinity(0))
+            except AttributeError:
+                error_logger.warning(
+                    "Ignoring '--fast' since it is not supported on this OS."
+                )
+
         kwargs = {
             "host": args.host,
             "port": args.port,
             "unix": args.unix,
             "workers": args.workers,
-            "debug": args.debug,
+            "debug": args.mode == "debug",
             "access_log": args.access_log,
             "ssl": ssl,
         }
@@ -179,13 +252,16 @@ def main():
                     "changes, consider using --debug or --auto-reload."
                 )
 
+        if app.configure_logging and args.mode == "prod":
+            logger.setLevel(logging.WARNING)
+
         app.run(**kwargs)
     except ImportError as e:
         if module_name.startswith(e.name):
             error_logger.error(
                 f"No module named {e.name} found.\n"
                 "  Example File: project/sanic_server.py -> app\n"
-                "  Example Module: project.sanic_server.app"
+                "  Example Module: project.sanic_server:app"
             )
         else:
             raise e
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1014,6 +1014,14 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
             )
 
         if auto_reload or auto_reload is None and debug:
+            if auto_reload is None and debug:
+                error_logger.warning(
+                    DeprecationWarning(
+                        "Enabling auto-reload via debug mode is deprecated "
+                        "and will be removed in v22.3. Consider using "
+                        "'sanic server:app --dev' instead."
+                    )
+                )
             self.auto_reload = True
             if os.environ.get("SANIC_SERVER_RUNNING") != "true":
                 return reloader_helpers.watchdog(1.0, self)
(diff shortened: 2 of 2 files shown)


sanic/__main__.py
```diff
@@ -157,12 +222,20 @@ def main():
...
             ssl = None
 
+        if args.fast:
+            try:
+                args.workers = len(os.sched_getaffinity(0))
+            except AttributeError:
+                error_logger.warning(
+                    "Ignoring '--fast' since it is not supported on this OS."
+                )
+
```
Tronic, line 232: Does this exclude MacOS from using `--fast`? I cannot test on Mac right now but I recall someone saying that this is specific to Linux.

sanic/__main__.py
```diff
@@ -87,15 +88,53 @@ def main():
...
         default=1,
-        help="number of worker processes [default 1]\n ",
+        help="Number of worker processes [default 1]",
+    )
+    group_worker.add_argument(
+        "--fast",
+        action="store_true",
+        help=(
+            "Run with maximum allowable processes.\nThis should be equal "
+            "to the number of CPU processes available.\n "
```
Tronic, line 105: This message is hard to understand. Maybe something along the lines of `use all available CPUs`?

Tronic: It is worth a note that this does not (until 22.3) disable access logs, or change logging level (not sure if the latter would affect benchmarks). I suppose this is fine if at 22.3 the default mode is steered towards or even merged with the production mode.

Tronic (changes requested):

sanic/__main__.py
ahopkins: My mac (or rather, my wife's mac) is being repaired now so I cannot test. The docs just say "UNIX".

> These functions control how a process is allocated CPU time by the operating system. They are only available on some Unix platforms. For more detailed information, consult your Unix manpages.

sjsadowski: I'll pull and test - I have both intel and arm (m1) macs available to work with.

Tronic:
An afterthought, perhaps we should utimately have `--debug` for dev without reload, and `--dev` with reload as you suggested, while the default would be production mode so no need for separate switch for that, and still have `--fast` to use all CPUs. In particular, this would be avoid the parametric `--mode` arguments altogether, and reduce the total number of modes which I still find confusing.

The drawback is that the default mode would then have to be a genuine production mode, probably no motd or anything extra on the logs, which then could be confusing for some newcomers.

Tronic:
Most daemons print a single-line "motd" on logs when starting, which is actually informative, perhaps that could be kept in default/dev mode to avoid some of the confusion? Print the address and port being listened to and that it is in production mode using N CPUs (processes), but not the full ASCII art logo?

Tronic:
Related but not to this PR: it would be cool if someone redid the logo in xterm256-color or ANSI color and Unicode graphemes. A lot of Unix software is now taking advantage of these, and most terminals seem to support the extended colors and a fairly large selection of Unicode graphemes, even colored emoji and all that now. Although granted, they might display wrong on `less` while viewing logs via journalctl or something.

Which brings us to one option, only print the motd when a (suitable) TTY is detected, not if the output is redirected (to syslog). Then it would not need to depend on CLI options at all.

Tronic:
Also one point, since all CPUs nowadays have hyperthreading, my benchmarks show nearly identical performance no matter if you use the number of physical or logical CPUs, although the double number of workers (i.e. logical CPU count) seems to be a tad faster. This gives some headroom for some workers being stuck on I/O and not consuming a CPU, and I think both Intel and AMD can take advantage of instruction reordering via HT to better avoid pipeline stalls, giving a modest increase in total throughput if but HT cores of a physical CPU are fully loaded (contrary to HT when it was first introduced many years ago, when it would actually slow throughput a a bit when fully loaded, although it still improved scheduling latency, which was particularly helpful for interactive use on Windows which at the time had a very bad kernel scheduler that would otherwise cause noticeable stalls on desktop use).

Tronic:
Yet another concern, I would prefer to keep `app.run` arguments at least to some degree compatible with the CLI. In particular do the CPU count detection there instead within the CLI frontend (might already be addressed, I didn't look at the code).

sanic/__main__.py
sjsadowski: So on MacOS on an m1, the tests pretty much result in fails across the board. @ahopkins it may make sense to work with you to get a runner set up on my m1 mac mini that can run the tox tests and we're at least tracking the rate of failure, that way if someone gets a bee in their bonnet to fixup things for MacOS they have a starting point.

sjsadowski: @Tronic - the answer is yes:
```
$ uname -a && python
Darwin goldfinger.local 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:20 PDT 2021; root:xnu-7195.141.6~3/RELEASE_ARM64_T8101 arm6
Python 3.9.7 (default, Sep  3 2021, 04:31:11) 
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>> import os
>>> len(os.sched_getaffinity(0))
AttributeError: module 'os' has no attribute 'sched_getaffinity'
```

Tronic: So I suggest adding a fallback to "normal" cpu count if this function is not available. OS detection or special handling of Windows might not be necessary, because there is a limit to one worker elsewhere (app.py IIRC).

ahopkins: Adding `cpu_count()` as the fallback.

sanic/__main__.py
ahopkins: I threw together a quick little test to check the performance of a simple loop with different logging.

```python
import logging

import timy

LOOPS = 10_000
RANGE = 1_000
LOG = logging.getLogger(__name__)


@timy.timer(loops=LOOPS)
def normal():
    LOG.setLevel(logging.WARNING)
    c = 0
    for i in range(RANGE):
        c += 1


@timy.timer(loops=LOOPS)
def with_logging_warning():
    LOG.setLevel(logging.WARNING)
    c = 0
    for i in range(RANGE):
        c += 1
        LOG.debug("something")


@timy.timer(loops=LOOPS)
def with_logging_debug():
    LOG.setLevel(logging.DEBUG)
    c = 0
    for i in range(RANGE):
        c += 1
        LOG.debug("something")


normal()
with_logging_warning()
with_logging_debug()
```

The results on my machine:

```
Timy executed (normal) for 10000 times in 0.338087 seconds
Timy best time was 0.000031 seconds
Timy executed (with_logging_warning) for 10000 times in 2.469493 seconds
Timy best time was 0.000234 seconds
Timy executed (with_logging_debug) for 10000 times in 77.279574 seconds
Timy best time was 0.007269 seconds
```

Conclusion: changing the logging level has a huge impact on performance. But, even the existence of the not impacted log call is not trivial either. If we are steering towards trying to impress benchmarks, then limiting or removing unused `log` calls completely should make a difference.

ahopkins:
> An afterthought, perhaps we should utimately have `--debug` for dev without reload, and `--dev` with reload as you suggested, while the default would be production mode so no need for separate switch for that, and still have `--fast` to use all CPUs. In particular, this would be avoid the parametric `--mode` arguments altogether, and reduce the total number of modes which I still find confusing.

I am not opposed to this, with the caveat that we slowly introduce the change so no breaks between now and the end of the year.

> probably no motd or anything extra on the logs

:thinking: MOTD may be okay still. I am thinking specifically of celery which still has a startup message even when you are running in prod. I suppose maybe we can add an arg or config to suppress it if someone really does not want to add it. I honestly am moving more towards the opinion that we should make it more clear what is being run at startup, which can then provide people with an easy copy/paste for support.

> Most daemons print a single-line "motd" on logs when starting, which is actually informative, perhaps that could be kept in _all_ modes to avoid some of the confusion?

Okay... so I think we are in agreement here. This is what I was thinking.

> Related but not to this PR: it would be cool if someone redid the logo in xterm256-color or ANSI color and Unicode graphemes.

This is a nice point. I have gone back and forth in my head a lot for a while about whether it would or would not be a good idea to add ANSI colors to logging. This might be a good way to do it. Regarding the logo in particular, in what modes do you think it is appropriate to display it? If it is not so big and in the way, maybe we add it to MOTD if in TTY. No logo if logs are being redirected, and if we have a `--no-motd` option, then obviously no logo there as well.

> Yet another concern, I would prefer to keep `app.run` arguments at least to some degree compatible with the CLI. In particular do the CPU count detection there instead within the CLI frontend (might already be addressed, I didn't look at the code).

I agree in principle. But, if you are doing `app.run` yourself, then likely you have some specific reason that you want to control some other aspects manually. In this case, the logic that comes with `--fast` is sort of irrelevant. So, yes it would be confusing if `--debug` is not the same as `app.run(debug=True)`, but I don't think all of the CLI functionality needs to be available.

Tronic:
> > Yet another concern, I would prefer to keep `app.run` arguments at least to some degree compatible with the CLI. In particular do the CPU count detection there instead within the CLI frontend (might already be addressed, I didn't look at the code).
> 
> I agree in principle. But, if you are doing `app.run` yourself, then likely you have some specific reason that you want to control some other aspects manually. In this case, the logic that comes with `--fast` is sort of irrelevant. So, yes it would be confusing if `--debug` is not the same as `app.run(debug=True)`, but I don't think all of the CLI functionality needs to be available.

Could use `workers=True` for automatic selection, and just have `--fast` enable that flag for now.

As it turned out, implementing good logic for finding the CPU count is non-trivial, thus it is better if app developers can use the existing logic even with `app.run`.

ahopkins:
Okay, so I have gone back to the drawing board a little on this:

1. `app.mode` is either `"prod"` or `"debug"`. It follows with `app.debug` for compat. There is no `--mode`, and no `--prod`. You are in prod mode out of the box unless you ask for debug.
2. If you are running in a TTY, then there is a message telling you about how you could turn on debug mode.
3. A more robust MOTD will probably need to wait until the next release, but something like this:
```
┌───────────────────────┐
│    version 21.9.0     │
├───────────────────────┤
│        host: 0.0.0.0  │
│        port: 8000     │
│        mode: prod     │
│     workers: 12       │
│ auto-reload: disabled │
└───────────────────────┘
```

ahopkins:
Revised proposal:

| Flag    | Mode  | Tracebacks | Logging | Access logs | Reload | Max workers |
|---------|-------|------------|---------|-------------|--------|-------------|
| --debug | DEBUG | yes        | DEBUG   | yes         | ^1     |             |
|         | PROD  | no         | INFO ^2 | ^3          |        |             |
| --dev   | DEBUG | yes        | DEBUG   | yes         | yes    |             |
| --fast  |       |            |         |             |        | yes         |



- ^1 `--debug` to deprecate auto-reloading and remove in 22.3
- ^2 After 22.3 this moves to WARNING
- ^3 After 22.3: no

ahopkins:
Moved to #2295

PR #2244, opened 2021-09-25
Static DIR and FILE resource types

### LINKED ISSUE #2132: `RouteExists` thrown when registering a directory and a missing file to the same route.
**Describe the bug**
When registering a route to a file and a directory, if the file doesn't exist a `sanic_routing.exceptions.RouteExists` error is raised. This may be misleading and unintended.

**Code snippet**

`example.html` does not exist in this example. The directory  `./resources/web` does.

```python
app.static('/', './resources/web')
app.static('/', './resources/web/example.html')
```

**Expected behavior**
Currently, if a non existent file is registered to a route, no error is thrown on startup and any attempt to retrieve registered resource raises a `FileNotFoundError: [Errno 2] No such file or directory` and a `404 Not Found` error is displayed in the browser. Instead of this happening, as mentioned before a `sanic_routing.exceptions.RouteExists` is raised on startup which may be misleading and unintended.

 - OS: Ubuntu 20.04.2 LTS

ahopkins:
Replaces #2224 
Closes #2132


ahopkins:
@Tronic I meant to add unit tests to #2224, and then pushed to the wrong branch. That is what I get for adding tests at 1am.

Tronic (approved):
Cannot comment on the feature but the changes seem OK.

ahopkins:
Closes #2132 and allows for a more explicit use of `static()`.

Tronic:
Without looking much into this, my intuition is that the distinction between files and folders could be determined by trailing slash (i.e. no trailing slash causes autodetection as before and with trailing slash forces it to be a directory). The separate type argument seems odd because I have never seen anyone else require such argument in code handling files and folders.

ahopkins:
> Without looking much into this, my intuition is that the distinction between files and folders could be determined by trailing slash (i.e. no trailing slash causes autodetection as before and with trailing slash forces it to be a directory). The separate type argument seems odd because I have never seen anyone else require such argument in code handling files and folders.

That seems to me like a recipe for confusion.

```python
app.static(..., "/path/to/is_maybe_file_or_dir")
app.static(..., "/path/to/is_def_dir/")
app.static(..., "/path/to/still_ambiguous.txt")
```

This still lacks an expressive ability to definitively tell Sanic how to treat two out of three resources.

PR #2246, opened 2021-09-27
Account for BP with exception handler but no routes

### LINKED ISSUE #2242: Blueprint exceptions are sometimes handled by other blueprint's exception handlers
**Describe the bug**
Exceptions thrown from within one blueprint are sometimes handled by the exception handler of another, unrelated blueprint instead of by its own. I have not had time to check whether this happens all the time or only if specific conditions are met, but I have attached a code snippet with which the issue can be reproduced.

**Code snippet**
```py
from sanic import Blueprint, HTTPResponse, Request, Sanic
from sanic.exceptions import SanicException


class BaseException(SanicException):
    code: int
    description: str
    status_code: int
    headers: dict = {}


class ExceptionA(BaseException):
    error: str


class ExceptionB(BaseException):
    code = 0
    status_code = 400


class ExceptionBA(ExceptionB, ExceptionA):
    error = "foo"
    description = "Bar!"


app = Sanic("my_app")
bp1 = Blueprint("auth", url_prefix="/auth")
bp2 = Blueprint("token")


@bp1.route("/network", version=1)
async def error_bp1(_: Request):
    raise ExceptionBA()


@bp2.route("/token")
async def hello_bp2(_: Request):
    return text("P3rry7hePl4typu5")


bpg1 = Blueprint.group(bp1, version_prefix="/api/v")
bpg2 = Blueprint.group(bp2, url_prefix="/api/oauth2")


for bp in bpg1.blueprints:
    @bp.exception(BaseException)
    async def _(_, ex: BaseException) -> HTTPResponse:
        return text("BPG1_BaseException")

    @bp.exception(Exception)
    async def _(request: Request, ex: Exception) -> HTTPResponse:
        return text("BPG1_Exception")


for bp in bpg2.blueprints:
    @bp.exception(ExceptionA)
    async def _(_, ex: ExceptionA):
        return text("BPG2_ExceptionA")

    @bp.exception(Exception)
    async def _(request: Request, ex: Exception):
        return text("BPG2_Exception")


bpg_all = Blueprint.group(bpg1, bpg2)
app.blueprint(bpg_all)


if __name__ == "__main__":
    app.run(debug=True)
```

**Expected behavior**
When accessing `/api/v1/auth/network`, an exception is raised, which is captured by the `bpg1` exception handler; rendering a final response which displays the string `BPG1_BaseException`.

**Actual behavior**
When accessing `/api/v1/auth/network`, an exception is raised, which is (somehow) captured by the `bpg2` exception handler; rendering a final response which displays the string `BPG2_ExceptionA`.

 - OS: Windows 10 Pro; 21H1; Compilation 19043.1165.
 - Versions: Sanic 21.6.2; Routing 0.7.1

ahopkins:
See [discussion](https://github.com/sanic-org/sanic/issues/2202#issuecomment-927693978). If a BP has an exception handler and no routes assigned, it could inadvertently attach the handler to the global scope. This would be an unintended and difficult to debug consequence.

Fixes #2242


komar007 (approved):
Works as expected

ahopkins:
Thanks @komar007 for checking :muscle:

sanic/blueprints.py
```diff
@@ -337,10 +337,11 @@ def register(self, app, options):
                 middleware.append(app._apply_middleware(future, route_names))
 
         # Exceptions
-        for future in self._future_exceptions:
-            exception_handlers.append(
-                app._apply_exception_handler(future, route_names)
-            )
+        if route_names:
```
prryplatypus, line 340: Just being a bit pedantic here, but you should be able to remove this `if` statement only and it should still work (since the code is now indented inside another `if route_names` check :)

ahopkins: You're right. I was thinking it would be cleaner this way since each primitive we are applying is separate.

ashleysommer (approved):
Looks good.

prryplatypus (approved):

PR #2248, opened 2021-09-30
Some fixes to the new Websockets impl

ashleysommer:
Some fixes based on more testing I was doing today

- Will throw WebsocketClosed exception instead of ServerException now when attempting to read or write to closed websocket, this makes it easier to catch
- Keepalive-ping task no longer emits a warning when WebsocketClosed exception is encountered.
- The various ws.recv() methods now have the ability to raise CancelledError into your websocket handler
- Fix a niche close-socket negotiation bug
- Fix bug where http protocol thought the websocket never sent any response.
- Allow data to still send in some cases after websocket enters CLOSING state.
- Fix some badly formatted and badly placed comments


sanic/http.py
```diff
@@ -187,12 +187,12 @@ async def http1(self):
                 if self.response:
                     self.response.stream = None
 
-            self.init_for_request()
-
             # Exit and disconnect if no more requests can be taken
             if self.stage is not Stage.IDLE or not self.keep_alive:
                 break
 
+            self.init_for_request()
```
ahopkins, line 194: @Tronic Do you have any issues with moving this?

Tronic: I gather this should only alter the state in which this returns to protocol, but I don't see any reason for this change or against it. Maybe @ashleysommer could elaborate. Neither `.stage` nor `.keep_alive` is set by `init_for_request`.

ahopkins: The point is that the new websockets implementation raises a silent exception when it is the other way around.

Tronic: Ok then.

ahopkins (approved):

sanic/http.py
ashleysommer: Just to clarify.
One of the things that `init_for_request()` does is it resets `self.upgrade_websocket` to None.
A websocket request will always `break` in the next if statement, because its not KeepAlive.
But the new check in http_protocol for `and not self._http.upgrade_websocket` needs `_http.upgrade_websocket` to still be populated.

PR #2260, opened 2021-10-03
Allow early Blueprint registrations to still apply later added objects

### LINKED ISSUE #2255: log a warning or error when adding something to blueprint after adding it to application
when i add a blueprint to the sanic app, and then add routes to it, those routes dont work. i have to add the blueprint to the app after adding routes. but there is no indication that im doing something wrong if i do it the wrong way

log a warning or error message telling the programmer that they added routes to a blueprint after adding it to the app, and that those routes will not work. or straight up throw an error

ahopkins:
Resolves #2255 

The purpose of this PR is to allow blueprints to be registered at anytime. The following example (which previously was not allowed) should now operate as expected.

```python
from sanic import Blueprint, Sanic, text

app = Sanic("app")
bp = Blueprint("bp")

app.blueprint(bp)


@bp.route("/", methods=["GET", "post"])
async def handler(request):
    return text("...")
```

The difference is the order in which you register the Blueprint. With this PR, you can run `app.blueprint(bp)` at anytime: either before or after handlers are added to the Blueprint.

This change will behave slightly more obviously as it would otherwise be expected. That is to say that any item added to a Blueprint will be applied to the application instance(s) that the blueprint is registered on.


ahopkins:
This one turned out to be a lot more difficult than originally anticipated. I think this is mainly attributable to the fact that we are storing the `Future*` objects as tuples. If they were proper objects, then we might be able to make some of the checking a little easier. This would of course be breaking, so it may be best to leave it for another day.

vltr (approved):
this looks good! some interesting Python code applied to the solution as well :sunglasses:

PR #2262, opened 2021-10-03
Add ability to log all exceptions

prryplatypus:


sanic/__main__.py
```diff
@@ -96,6 +96,12 @@ def main():
         help="number of worker processes [default 1]\n ",
     )
     parser.add_argument("-d", "--debug", dest="debug", action="store_true")
+    parser.add_argument(
+        "-ne",
```
ahopkins, line 100: I don't think this needs a shorthand.

prryplatypus:
Would you like me to add the tests in `test_exceptions_handler.py` or in `test_logging.py`?

prryplatypus:
Would have to change the version number in the deprecation message, but everything else should be good to go! Let me know which version number that would be, or feel free to change it yourselves 🤘

prryplatypus:
After a hundred thousand iterations, I'm finally happy enough with this for it to be reviewed and merged!

PR #2264, opened 2021-10-03
Optional uvloop use

### LINKED ISSUE #2223: SANIC_NO_UVLOOP=true doesn't seem to work
**Describe the bug**
Installing sanic with a SANIC_NO_UVLOOP=true in my environment still results in uvloop being installed

 - OS: macOS Big Sur
 - Version 11.5.2

**Additional context**
Python 3.9.6

Screen Shot 2021-08-22 at 9 15 49 am

prryplatypus:
Resolves #2223.


prryplatypus:
I've made it `NO_UVLOOP` for consistency, but it's not super user friendly when we do `if not self.config.NO_UVLOOP:`. I'd suggest changing it to `USE_UVLOOP` (both in the install and in config), but make it default to `True` instead.

ahopkins:
That's a breaking change. Not idea but I think keeping it negative might be the way to go.

prryplatypus:
So we're not willing to do this? 🙈 
```py
import os

USE_UVLOOP = os.getenv('USE_UVLOOP', True)
no_uvloop = os.getenv('NO_UVLOOP', None)

if no_uvloop is not None:
    # Raise DeprecationWarning
    USE_UVLOOP = no_uvloop
```

ahopkins:
Why do they need to be connected? The `SANIC_NO_UVLOOP` variable is meant to be an install time switch. Something that itself needs to be fixed (if possible). But, if it is set, then theoretically, you should not have uvloop installed, and it should not matter.

I think we can just go forward with `Config.USE_UVLOOP` and default to that. If there is a conflict (``USE_UVLOOP == True` and `NO_UVLOOP == True`), then this is a situation that we could raise a warning. But, this should only matter if uvloop is installed. If it is not installed, and you have `USE_UVLOOP == True`, then that is another scenario we need to raise a startup warning.

prryplatypus:
Alright, that sounds good to me!

prryplatypus:
I uhhh... may need some help with these tests. I'm not too used to writing tests, and I think this is not one of the easiest things to write them for either 😅. Could anyone lend me a hand, please?

prryplatypus:
I'll probably wait for #2295 to be merged before completing this one. There's some things in that PR that would make this change easier/cleaner.

sanic/app.py
```diff
@@ -1146,7 +1146,7 @@ async def create_server(
             ssl=ssl,
             sock=sock,
             unix=unix,
-            loop=get_event_loop(),
+            use_existing_loop=True,
```
ahopkins: This is a breaking change on a public API. We need to keep the same outward facing API on this.

ahopkins: We can think about changing it to something like this. But ultimately, if someone does:

```
create_server(..., loop=loop)
```

We still need to support that. Meaning it would have to override `use_existing_loop`.

ahopkins: Unless there is a compelling reason, I would lean towards tabelling this change until v22.

prryplatypus: The `create_server` function didn't/doesn't have a `loop` parameter, and the `_helper` function isn't a public API. Am I missing something here?

ahopkins: image

Sorry, I did not expand the view to see that this was not in the signature

prryplatypus:
Hey y'all, sorry for the radio silence lately! There was something with this change I wasn't too sure about and wanted to test before actually trying to merge it into the repo, and I really didn't have time to try it out... until today!

My main concern was whether changing the event loop policy within an asynchronous function (e.g.: `create_server`) was useful at all, or whether the loop itself was not going to change because it was already running. After testing it out today, I believe my suspicions are correct, and if an event loop is already running, `asyncio.get_event_loop()` will still return the same loop even after changing the event loop policy to use uvloop until the running loop is no longer running:
```py
import asyncio
import uvloop

async def test():
        loop_type = type(asyncio.get_event_loop())
        new_loop_type = type(asyncio.new_event_loop())
        print(f"In function, before change: {loop_type}")
        print(f"In function, before change (new): {new_loop_type}")

        loop_type = type(asyncio.get_event_loop())
        new_loop_type = type(asyncio.new_event_loop())
        print(f"In function, after change: {loop_type}")
        print(f"In function, after change (new): {new_loop_type}")


new_loop = asyncio.new_event_loop()
print(f"Out of function, before change: {type(loop)}")
print(f"Out of function, before change (new): {type(new_loop)}")

loop.run_until_complete(test())

new_loop = asyncio.new_event_loop()
print(f"Out of function, after change: {type(loop)}")
print(f"Out of function, after change (new): {type(new_loop)}")
```
Results

Therefore, I would like to suggest **NOT** changing the event loop policy within async methods such as `create_server` (as these use an already-running loop for the worker that we can't change), and instead only change the loop policy within non-async methods such as `run` (where each worker uses a new event loop created via `asyncio.new_event_loop()` which would be affected by changing the policy).

Please let me know your thoughts!

prryplatypus:
Ok, just missing tests for logging and should be good to go!
Might also move the `use_uvloop` method out of `__init__.py`.

ahopkins:
Agreed, we should not be changing the loop inside of a coroutine.

ahopkins (commented):
Looks good. When you are ready, LMK and I will pull locally. But, this looks like the right direction.

prryplatypus:
> Looks good. When you are ready, LMK and I will pull locally. But, this looks like the right direction.

I think I am ready if you want to pull locally!

Wondering if in `create_server` maybe we should check if the config value is not default and if it isn't, check if the running loop matches the config value (e.g: not from `uvloop` is `USE_UVLOOP` is `False`, or viceversa) and raise a warning if not. For now I haven't because -up until now- it's just been failing silently.

prryplatypus:
Things worth noting:

- If running more than one app, do we want to raise a warning if their configs are different? Is there any way we can globally know the state of the loop policy?
- When running the app with `create_server`, do we maybe want to check if the config loop policy matches the currently running event loop and raise a warning only if it doesn't match? For example, if `config.USE_UVLOOP` is explicitly set to `True` but `asyncio.get_event_loop()` doesn't return a `uvloop.Loop` instance.
- When running Sanic with Gunicorn (`worker.py`), uvloop is always configured to be used. Do we want to add a warning about this or just put it in the docs?

ahopkins (approved):
This looks great :tada:

Nice work on this PR :muscle:

prryplatypus:
> Hmm... Yes, but if one is set to _default then I suppose it does not matter. The issue would be if one is True and one False. Correct?
Correct! Or if one is _default and the other one if False and the _default one is started first (as it will then override the event loop config).

> `create_server` is a pretty DIY. If anything, I would say that we could do a check in the `AsyncioServer.startup` to see if it is not `_default`, in which case we throw a warning. But no need to check the types.
We already do this inside the `create_server` function, so I guess that works then!

> Docs
💯

sanic/app.py
```diff
@@ -1274,6 +1281,13 @@ async def create_server(
             run_async=return_asyncio_server,
         )
 
+        if self.config.USE_UVLOOP is not _default:
+            error_logger.warning(
+                "You are trying to configure uvloop, but this is only "
+                "supported when using the run(...) method. Sanic will now "
+                "continue to run using the existing event loop."
+            )
```
ahopkins, line 1289: This is a little confusing. It is possible to use `create_server` and `uvloop`. We are inside a coroutine, so it is possible that they already applied the policy. Maybe we should instead mention that the config value is ineffecive.

ahopkins (changes requested):

sanic/app.py
prryplatypus: I’ll try to reword this! I did mention “existing event loop” on purpose, but I agree it’s not totally clear.

prryplatypus:
This is as far as I can get. I've tried adding a test for the asgi warning, but have not succeeded at it. Any help is welcome. That should be the last thing left to do!

sanic/app.py
```diff
@@ -220,6 +223,12 @@ def __init__(
         self.go_fast = self.run
 
         if register is not None:
+            warn(
+                "The register argument is deprecated and will stop working "
+                "in vXX.X. After vXX.X all apps will be added to the Sanic "
+                "app registry.",
+                DeprecationWarning,
+            )
```
ahopkins, line 231: Maybe not the right PR for this. But, I suppose we can sneak it in here instead. v22.6

prryplatypus: I can submit a separate PR if you wish?

ahopkins: belongs here https://github.com/sanic-org/sanic/pull/2306

ahopkins (approved):

ahopkins (approved):

PR #2268, opened 2021-10-06
Make HTTP connections start in IDLE stage, avoiding delays and error messages

### LINKED ISSUE #2249: KEEP_ALIVE_TIMEOUT causing the changes in server (mode debug) to be delayed?
**Describe the bug**
When running Sanic with the following options:

`app.run(auto_reload=True, debug=True)`

If I make any changes to my code and refresh, the new code is not taken into consideration. I need to wait a few seconds, and I'll see the following message in the console:

> [DEBUG] KeepAlive Timeout. Closing connection.

Reading the [documentation](https://sanicframework.org/en/guide/deployment/configuration.html#keep-alive-timeout), settings KEEP_ALIVE to False in dev mode would solve my issue, but it doesn't change anything. The error still is displayed after about 8 seconds.

What can I do to have my new code refreshed immediately when running Sanic?

**Expected behavior**
New code executed on refresh, not the previous one.

 - OS: Fedora 24
 - Version: 21.6.2
 - Python: 3.8.10

### LINKED ISSUE #2289: Exception when stopping the server.
**Describe the bug**
After using the browser like Chrome to access the API, and then stop the server, it will block and wait for a moment, and an exception will occur. 100% reproduction.
```shell
[2021-10-25 15:21:56 +0800] [1879] [INFO] Goin' Fast @ http://127.0.0.1:8000
[2021-10-25 15:21:56 +0800] [1879] [INFO] Starting worker [1879]
[2021-10-25 15:22:00 +0800] - (sanic.access)[INFO][127.0.0.1:51319]: GET http://127.0.0.1:8000/  200 13
[2021-10-25 15:22:00 +0800] - (sanic.access)[INFO][127.0.0.1:51319]: GET http://127.0.0.1:8000/favicon.ico  404 673
^C[2021-10-25 15:22:03 +0800] [1879] [INFO] Stopping worker [1879]
[2021-10-25 15:22:19 +0800] [1879] [ERROR] protocol.connection_task uncaught
  File "/Users/zeb/miniconda3/lib/python3.9/site-packages/sanic/http.py", line 138, in http1
    await self.http1_request_header()
  File "http1_request_header", line 18, in http1_request_header
    ServerError,
  File "/Users/zeb/miniconda3/lib/python3.9/site-packages/sanic/server/protocols/base_protocol.py", line 82, in receive_more
    await self._data_received.wait()
  File "/Users/zeb/miniconda3/lib/python3.9/asyncio/locks.py", line 226, in wait
    await fut
asyncio.exceptions.CancelledError

  File "connection_task", line 15, in connection_task
    from sanic.http import Http, Stage
  File "/Users/zeb/miniconda3/lib/python3.9/site-packages/sanic/http.py", line 153, in http1
    f"Request: {self.request.method} {self.request.url} "
AttributeError: 'NoneType' object has no attribute 'method'
[2021-10-25 15:22:19 +0800] [1879] [INFO] Server Stopped
```

**Code snippet**
```python
from sanic import Sanic

@app.get("/")
async def hello_world(request):
	return text("Hello, world.")

app.run()
```

**Expected behavior**
The server should exit gracefully.

 - OS: MacOS 10.15.7
 - Version: sanic-21.9.1

Tronic:
Change new connections to start in IDLE stage, and switch to REQUEST stage only once any bytes are received from client. This makes new connections without any request obey keepalive timeout rather than request timeout like they did before. New connections where no request was ever done also no longer print access log entries or crash with exception trying to do so. Such connections are now immediately disconnected during graceful shutdown, avoiding server shutdown and restart delay when it actually is idle. This was affecting at least some Chrome users.

Keep-alive timeout is the primary control for disconnecting clients that are taking too long to make a request. Request timeout only governs the receiving of the request header, starting of its first packet, and thus has very little significance for normal clients where this is always instantaneous.

Fixes #2249
Fixes #2289


Tronic:
Tests require a lot of work to make compatible with this, and possibly this needs to be delayed until 22.3 if the changes are deemed too much breaking. Also some other approaches might be taken instead, if anyone comes up with better solutions. Because of all this, I have marked this PR a draft.

ashleysommer:
For clarity, can you confirm: this bug is that the very first request  on a connection has its state set to `REQUEST` _before_ the user client sends data, where in that case it should be IDLE. If I'm following the loop correctly, I believe subsequent requests should each still be set to IDLE correctly without this fix, is that right?

sanic/http.py
```diff
@@ -380,7 +374,7 @@ def head_response_ignored(self, data: bytes, end_stream: bool) -> None:
         HEAD response: body data silently ignored.
         """
         if end_stream:
-            self.response_func = None
+            self.q = None
```
ahopkins, line 377: What is this?

ashleysommer: Looks like a typo or change that wasnt supposed to be committed

Tronic:
> For clarity, can you confirm: this bug is that the very first request on a connection has its state set to `REQUEST` _before_ the user client sends data, where in that case it should be `IDLE`. If I'm following the loop correctly, I believe subsequent requests should each still be set to IDLE correctly without this fix, is that right?

The old code set stage IDLE in constructor, but entered REQUEST stage pretty much immediately after, once the http1 function was entered. After a request was processed, if keep-alive was still active, it would set stage IDLE and wait for more data at the **end** of the main while loop.

The new code moves this waiting to the beginning of the loop, so that it also applies to the first request, and thus the connection stays initially IDLE until there is something in the receive buffer.

ashleysommer (approved):

ashleysommer:
Another user in Discord has mentioned encountering this issue. I'd like to see this merged.

ahopkins:
> Another user in Discord has mentioned encountering this issue. I'd like to see this merged.

I plan to merge and release it (probably tomorrow). I am pulling it now because IIRC there was one change I wanted to make re: KEEP_ALIVE.

PR #2270, opened 2021-10-15
Vhost support using multiple TLS certificates

Tronic:
Usage:

```
app.run(ssl=['/etc/letsencrypt/live/example.com/', '/etc/letsencrypt/live/mysite.example/'])
```

A connecting client provides TLS SNI which we then use to match against all loaded certificates, using the first one that matches. If no certificates match, the TLS connection gets disconnected prior to any request being made.

This PR also changes default SSLContext creation with more secure settings, so that A and A+ grades are possible on SSL Test.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -61,10 +61,6 @@ def test_server_run(appname):
             "--tls=certs/localhost/",
             "--tls-strict-host",
         ),
-        (
-            # No certs, all connections will get rejected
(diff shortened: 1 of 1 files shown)


Tronic:
TODO 2:
[ ] Tests
[ ] Documentation

Anything else still overlooked?

ahopkins:
> Can we somehow suppress this warning?

Yes.

PR #2279, opened 2021-10-19
style: add some type hints

SaidBySolo:
Since the listener type is a generic type in the listener part, this has been slightly modified, and type hints have been added to the arguments of various functions.

Easier to use sanic!


Tronic:
Thanks. Looks good and passes the typing tests, so it must be fairly good. However, would it be possible to use concrete types instead of some of those you have marked `Any`?

PR #2302, opened 2021-11-07
Add route context

### LINKED ISSUE #2300: Labeling / tagging routes
Some of my routes requires authentication and some of them don't (such as home page). I'm currently using a middleware and hard-code URLs that requires authentication, however it would be better to have a property on routes when registering them.

```python
@bp.on_request
async def before_request(request : Request):
    if request.route and request.route.uri in [
        "/authorize",
        "/profile/<user:str>"
    ]:
        request.ctx.token = decrypt_token(request.cookies.get("access_token"))
        # If user is not signed in, redirect user to OAuth.
        if (request.route.uri != "/authorize") and (not request.ctx.token):
            return redirect(oauth_url("ID", "/oauth_callback", request.path))
    # No need to get token for other endpoints becase other pages
    # doesn't require auth and decrypt_token() takes time.

@bp.get("/authorize")
async def authorize(request : Request):
    ...

@bp.get("/profile/<user:str>")
async def profile(request : Request, user : str):
    ...

@bp.get("/")
async def home(request : Request):
    ...
```

It can be much easier (and may look cleaner) if we have `tags` or `labels` parameter when registering the routes. (It can accept a list of enums or strings or anything). So we can access these tags in middlewares and other places. I know I can also use a decorator and inject tags to routes, but I prefer having an in-built property instead of monkey-patching the routes. 

```python
class Tags(Enum):
    BASE_AUTH = "base_auth"
    REQUIRES_AUTH = "requires_auth"

@bp.on_request
async def before_request(request : Request):
    if Tags.REQUIRES_AUTH in request.route.tags:
        request.ctx.token = decrypt_token(request.cookies.get("access_token"))
        # If user is not signed in, redirect user to OAuth.
        if (Tags.BASE_AUTH in request.route.tags) and (not request.ctx.token):
            return redirect(oauth_url("ID", "/oauth_callback", request.path))
    # No need to get token for other endpoints becase other pages
    # doesn't require auth and decrypt_token() takes time.

@bp.get("/authorize", tags = [Tags.REQUIRES_AUTH, Tags.BASE_AUTH])
async def authorize(request : Request):
    ...

@bp.get("/profile/<user:str>", tags = [Tags.REQUIRES_AUTH])
async def profile(request : Request, user : str):
    ...

@bp.get("/")
async def home(request : Request):
    ...
```

**Additional context**
N/A

ahopkins:
This PR implements the idea discussed in #2300. It is an initial PR meant to explore the concept and idea of passing value to the route context using the suggestion from @Tronic to use a `ctx_` prefix in the route kwargs.

```python
@app.route("/", ctx_foo=123)
```

Closes #2300


sanic/mixins/routes.py
```diff
@@ -65,6 +74,7 @@ def route(
         static: bool = False,
         version_prefix: str = "/v",
         error_format: Optional[str] = None,
+        **kwargs,
```
Tronic, line 77: Perhaps also name this `ctx_kwargs` if no other kwargs are accepted. IDEs show the name while typing in code.

sanic/mixins/routes.py
```diff
@@ -957,3 +990,18 @@ def visit_Return(self, node: Return) -> Any:
...
+            if key.startswith("ctx_")
+        }
+        restricted = [
+            key for key in ctx_kwargs.keys() if key in RESTRICTED_ROUTE_CONTEXT
+        ]
+        if restricted:
+            raise AttributeError(
+                f"Cannot use restricted route context: {restricted}."
+            )
+        return HashableDict(ctx_kwargs)
```
Tronic, line 1007: Should raise an exception (NameError?) if any kwargs without that prefix are found.

ahopkins: Agreed. I was planning on doing something like that so we didn't accept just anything.

vltr (approved):
:+1:

PR #2304, opened 2021-11-07
Add named tasks

### LINKED ISSUE #2267: app.stop() does not run tasks to completion
Stopping the app instantly drops all existing tasks, meaning that try-finally sections, context manager exits etc. do not get executed. This is bad. Tasks should get cancelled and then run to completion prior to stopping the loop.

I suppose one should use `loop.run_until_completion()` if doing such manual management of asyncio loops. The modern way of course is to remove the explicit loop passing and handling entirely, like asyncio have done in the last few Python versions, but that with Sanic would be a huge undertaking. (deprecation and removal of loop arguments on public APIs would be a start though)

### LINKED ISSUE #2301: Task naming
The `app.add_task` API does not allow support for passing a `name` variable to `loop.create_task`.

ahopkins:
This is a fairly simple PR that adds something that adds naming `asyncio.Task`. If you create a named task, it will also allow you to add it to a registry (if requested) for simple callback after.

```python
app.add_task(dummy, name="dummy_task")
task = app.get_task("dummy_task")
```

This PR adds the following new attributes to application instances:

- `app.get_task`
- `app.cancel_task`
- `app.purge_tasks`
- `app.shutdown_tasks`
- `app.tasks`

In addition, on `app.add_task` it has a return value of the new task IF you are inside of the running loop.

### TODO:
- [x] Simple API to cleanup the registry from done tasks
- [x] Some more tests around task registration
- [x] Potentially an easy callable to cancel tasks: `app.cancel_task("foo")`
- [x] Potentially also providing the task as a return value on `add_task` when created inside application (and not in global scope where it is added to delayed execution)
- [x] Solve #2267 

Closes #2301 
Closes #2267


prryplatypus (approved):

ahopkins:

https://community.sanicframework.org/t/add-task-that-contains-an-infinite-loop-will-cause-an-error-when-stopping-the-server/948/4

PR #2305, opened 2021-11-07
Fix examples to work as expected

ahopkins:
This started out as an attempt to fix the `run_async` examples. But, there were a number of other outdated problems with the examples that needed some updates.


examples/run_async_advanced.py
```diff
@@ -28,20 +43,20 @@ async def test(request):
 loop = asyncio.get_event_loop()
 serv_task = asyncio.ensure_future(serv_coro, loop=loop)
 signal(SIGINT, lambda s, f: loop.stop())
-server: AsyncioServer = loop.run_until_complete(serv_task)  # type: ignore
-server.startup()
+server: AsyncioServer = loop.run_until_complete(serv_task)
+loop.run_until_complete(server.startup())
```
Tronic, line 47: a) Does this actually work? I recall having trouble getting something like this started (for hosting http and https at the same time on tls docs page).
b) Any chance this could be done without loop arguments?

examples/try_everything.py
```diff
@@ -52,53 +55,69 @@ async def test_file_stream(request):
 
 @app.exception(ServerError)
 async def test(request, exception):
-    return response.json({"exception": "{}".format(exception), "status": exception.status_code},
-                         status=exception.status_code)
+    return response.json(
+        {"exception": "{}".format(exception), "status": exception.status_code},
```
Tronic, line 59: Could use `dict(exception=str(exception), status=exception.status_code)`, or at least avoiding the `format`

examples/unix_socket.py
```diff
@@ -10,14 +11,15 @@
...
-            os.unlink(server_address)
+        os.unlink(server_address)
     except OSError:
-            if os.path.exists(server_address):
-                        raise
+        if os.path.exists(server_address):
+            raise
     sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
     sock.bind(server_address)
     app.run(sock=sock)
```
Tronic, line 25: Suggest removing this example, or using the built-in UNIX socket support instead.

Tronic (approved):

examples/run_async_advanced.py
ahopkins: 1. Yes, this works. I plan to post the "proper" way to run nested Sanic in the docs later this week.
2. Without the loop? Hmm... I suppose it could wrap it all into a single coroutine. The problem is that `startup` now requires the existence of the loop to be running.

examples/try_everything.py
ahopkins: I did not even look at this.

I think this is probably the easiest to read:
```python
{"exception": str(exception), "status": exception.status_code},
```

vltr: I agree on that.

examples/unix_socket.py
vltr: @Tronic you mean by using the `bind_unix_socket` method (from Sanic) or `loop.create_unix_server`?

Tronic: `app.run(unix=....)` unless there is some specific reason not to?

ahopkins: TBH, these examples are best served in the documentation. This PR was mainly to get the examples that were broken running again. Going forward I do not intend to continue adding to this section, but for the people that know this exists and may return back to it, we should make sure they are operational. I do not think we need to make it a showcase.

vltr: @Tronic thanks. there are N ways to get an asyncio server running with UNIX sockets, we just need to stick to one I guess ... after testing which one is the best, of course :sweat_smile:

examples/amending_request_object.py
```diff
@@ -1,30 +1,29 @@
...
+    request.ctx.num = randint(0, 100)
 
 
-@app.get('/pop')
+@app.get("/pop")
 def pop_handler(request):
-    # Pop key from request object
-    num = request.pop('num')
-    return text(num)
+    return text(request.ctx.num)
```
prryplatypus, line 17: Might want to rename this route since it doesn't `pop` the value anymore?

examples/amending_request_object.py
```diff
@@ -1,30 +1,29 @@
...
 
 
-@app.get('/key_exist')
+@app.get("/key_exist")
```
prryplatypus, line 20: `/attr_exist` may be better now?

examples/amending_request_object.py
```diff
@@ -1,30 +1,29 @@
...
 def key_exist_handler(request):
     # Check the key is exist or not
-    if 'num' in request:
-        return text('num exist in request')
+    if hasattr(request.ctx, "num"):
+        return text("num exist in request")
 
-    return text('num does not exist in reqeust')
+    return text("num does not exist in reqeust")
```
prryplatypus, line 26:
```suggestion
    return text("num does not exist in request")
```

examples/limit_concurrency.py
```diff
@@ -1,36 +1,40 @@
...
 app = Sanic(__name__)
 
 sem = None
 
 
-@app.listener('before_server_start')
-def init(sanic, loop):
+@app.before_server_start
+def init(sanic, _):
     global sem
```
prryplatypus, line 16: Could put this in `app.ctx` instead of using the `global` keyword.

prryplatypus (approved):
`examples/blueprint_middlware_execution_order.py` has a small typo in the filename too, in case you wanna rename ;)

PR #2306, opened 2021-11-07
LTS v21.12 Deprecations

ahopkins:
There are a number of things we need to remove in the next release:

- [x] `abort`
- [x] `CompositionView`
- [x] `StreamingHTTPResponse`
- [x] `Sanic` and `Blueprint` to use `__slots__`
- [x] `Config.load_env`

Most of these are straightforward. The open question is re: `StreamingHTTPResponse`. More particularly, the two convenience methods we had for instantiating it: `response.stream` and `response.file_stream`.

While the release notes and warning have made it clear that `StreamingHTTPResponse` would be removed, we also said that the convenience methods would stick around for some time. Therefore: **we must provide backwards compat in v21.12**.

I have accomplished this with a new `ResponseStream` object which is a fairly lightweight and is meant to be a compat layer. What we really need to do is to decide the fate of `stream` and `file_stream`.

I propose the following:

1. Deprecate `stream` and mark it for removal in v22.6.
2. Continue with support for `file_stream` in the form proposed in this PR.

There really is no longer a utility in `stream`. Even if we do not mark it for removal, I would advocate for removing it from the docs and leaving it as undocumented in the User Guide since it is no longer the preferred method for handling these. On the other hand, `file_stream` does have utility separate from `StreamingHTTPResonse`, so I think we should carry that one forward. 

Other thoughts? Opinions?


Tronic:
I am happy to see old streaming removed in all forms, even if it takes some time. The `return file_stream(...)` thing is quite useful, so we should support that based on the new streaming system. I cannot quite recall what the current state of this is.

Also, there actually might not be much use for supporting `file` and `file_stream` return values separately, they ought both be streaming when necessary, e.g. based on file size.

ahopkins:
> Also, there actually might not be much use for supporting `file` and `file_stream` return values separately, they ought both be streaming when necessary, e.g. based on file size.

@Tronic another point worth mentioning is the discussion that started in Discord re: sync v async file reads. While `async` like what is currently in Sanic is non-blocking, it is also much slower. Combining `file` and `file_stream` does make sense, but we should also consider what method to use for the reading since smaller files will actually be better served with a little bit of blocking using regular `open`.

Tronic:
> Combining `file` and `file_stream` does make sense, but we should also consider what method to use for the reading since smaller files will actually be better served with a little bit of blocking using regular `open`.

Strictly **all** disk I/O is best avoided when the server is running. Unfortunately even the simplest disk access can take 10+ seconds when SSDs are doing their reorganization, when btrfs is doing its, or when mechanical drives are spinning up. Throughput needs to be optimized elsewhere, e.g. reading all small static files (if known not to change on disk, or if there is a background thread that monitors for changes) to RAM and serving from there (probably needs to be done by the application or at the very least an extension, as it is hard to see how Sanic core could offer that).

tests/test_app.py
```diff
@@ -2,6 +2,7 @@
 import logging
 import re
 
+from email import message
```
prryplatypus, line 5: I think this one might be my fault from the register test haha. VSCode likes auto-importing things 😓 
```suggestion
```

tests/test_deprecation.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+from sanic.log import deprecation
+
+
+def test_deprecation():
+    message = r"\[DEPRECATION v9\.9\] hello"
```
prryplatypus, line 7: Hello 🙃 👋

tests/test_exceptions.py
```diff
@@ -262,7 +252,7 @@ def tempest(_):
 
 
 def test_exception_in_ws_logged(caplog):
-    app = Sanic(__file__)
+    app = Sanic(__name__)
```
prryplatypus, line 255: Seems to be the only app name you haven't changed to a fixed string. Missed it or intentional?

tests/test_exceptions_handler.py
```diff
@@ -77,6 +77,7 @@ def handler_exception_with_ctx(request, exception):
 
     @exception_handler_app.exception(ServerError)
     def handler_exception(request, exception):
+        print(exception)
```
prryplatypus, line 80:
```suggestion
```

tests/test_unix_socket.py
```diff
@@ -21,6 +21,9 @@
 pytestmark = pytest.mark.skipif(os.name != "posix", reason="UNIX only")
 SOCKPATH = "/tmp/sanictest.sock"
 SOCKPATH2 = "/tmp/sanictest2.sock"
+httpx_version = tuple(
+    map(int, httpx.__version__.strip(ascii_lowercase).split("."))
+)
```
prryplatypus, line 26: Seems to be defined a few lines before already
```suggestion
```

tests/test_versioning.py
```diff
@@ -19,7 +19,7 @@ def test_route(app, handler):
 
 
 def test_bp(app, handler):
-    bp = Blueprint(__file__, version=1)
+    bp = Blueprint(__name__, version=1)
```
prryplatypus, line 22: Do we not want fixed names in these tests like in all others + examples? There's a few more like this in this specific file, but have just marked the first one.

prryplatypus (approved):

tests/test_exceptions.py
ahopkins: Should be fine here. The problem was the examples are actually used in some tests (which should really not be the case). So, their `__name__` includes a dot because of the import path.

examples/redirect_example.py
```diff
@@ -1,18 +1,18 @@
...
 
-    
-@app.route('/')
+app = Sanic("Example")
+
+
+@app.route("/")
 def handle_request(request):
-    return response.redirect('/redirect')
+    return response.redirect("/redirect")
```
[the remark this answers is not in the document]
ahopkins: Another PR

ahopkins:

https://community.sanicframework.org/t/v22-6-cant-not-use-sanic-auto-register-true/1035/2

PR #2310, opened 2021-11-16
More consistent config setting with post-FALLBACK_ERROR_FORMAT apply

### LINKED ISSUE #2276: ``FALLBACK_ERROR_FORMAT`` cannot be set even when the ``update_config`` function is called because the ``ErrorHandler`` object receives the default value of the config and is initialized.
**Describe the bug**

https://github.com/sanic-org/sanic/blob/3262878ebd41aa2230ef15d4475bbcf223b2356b/sanic/app.py#L139-L184

Referring to the code above, since the config is set when the Sanic object is initialized, you cannot pass a fallback value to the ErrorHandler even if you set ``config.update_config`` or ``app.config.FALLBACK_ERROR_FORMAT = "json"``.

To solve this, you can pass the config argument when initializing the Sanic app, but this doesn't seem to be intended.

**Code snippet**
```py
app = Sanic(__name__)

@app.main_process_start
async def start(app, loop):
```

**Expected behavior**
Even if you use ``config.update_config`` or ``app.config.FALLBACK_ERROR_FORMAT = "json"``, the error handler should normally receive this value.

 - OS: [Windows 10]
 - Version [21.9.1]

**Additional context**

I wanted to open a PR on this issue, but I'm not sure where to touch.

ahopkins:
**INTENDED FOR v21.9.2 RELEASE**

## Background re: `Config`

The `Config` object in Sanic is a heavily modified `dict`. There already were instances where we wanted to do some checks _post_ modification of a config value. **HOWEVER**, we were not consistently applying this post-apply.

It would trigger here:

```python
app.config.FOO = 123
```

But not here:

```python
app.config["FOO"] = 123
```

Or here:

```python
app.config.update({"FOO": 123})
```

## Issue re: `FALLBACK_ERROR_FORMAT`

The `ErrorHandler` took a value for fallback _at the time `Sanic()` was initialized_. This meant that applying a value after initializing your app would not always work as might be expected. In fact, this value was almost _always_ being ignored.

Furthermore, routes were taking the `app.config.FALLBACK_ERROR_FORMAT` when they were being applied. This is the reason that setting it in the global scope was working.

## What this PR does

This PR is meant to handle #2276. While investigating this issue, I uncovered the inconsistency with the `Config` setter, so the scope of the PR grew. But, I think this is enough of a "gotcha" that it too should be included in an update to the current Sanic release.

Here are all of the changes

### Change `Config` setters

All of the methods for setting values on the `Config` option are now consistent. No matter which method used, we will apply the post-set logic.

```python
app.config.FOO = 1
app.config["FOO"] = 2
app.config.update({"FOO": 3})
app.config.update([("FOO", 4)])
app.config.update(FOO=5)
app.config.update_config({"FOO": 6})  # everything supported here ultimately uses `config.update(...)`
```

### Add a new post-set to apply `FALLBACK_ERROR_FORMAT` to the error handler

Previously, we were already checking the value of `FALLBACK_ERROR_FORMAT` to make sure that it would be valid. Now, when you set this value in any scope it will also reach into the `error_handler` attached to the application instance and update its `fallback` value. 

There is a little bit of a break here though. To accomplish this, we needed to bring the `app` instance into the `Config` object. Ultimately, I think this makes sense and for most users out there they will never notice a difference. However, if you are using a custom `Config` object, you will not have the `app` set. Therefore, this post-apply `FALLBACK_ERROR_FORMAT` will not work. We will have to provide a notice about this.

### Default `route.ctx.error_handler` to `None` if it is not explicitly set

We will not longer _always_ add an `error_handler` to the route. Instead we will do it only when there is an explicit value from the route definition. This way any values or changes to the `config` or the `error_handler` will still be impactful at render-time.

Closes #2276

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -190,14 +190,12 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         self._state: ApplicationState = ApplicationState(app=self)
         self.blueprints: Dict[str, Blueprint] = {}
         self.config: Config = config or Config(
-            load_env=load_env, env_prefix=env_prefix
+            load_env=load_env, env_prefix=env_prefix, app=self,
         )
         self.configure_logging: bool = configure_logging
         self.ctx: Any = ctx or SimpleNamespace()
         self.debug = False
-        self.error_handler: ErrorHandler = error_handler or ErrorHandler(
-            fallback=self.config.FALLBACK_ERROR_FORMAT,
-        )
+        self.error_handler: ErrorHandler = error_handler or ErrorHandler()
         self.listeners: Dict[str, List[ListenerType[Any]]] = defaultdict(list)
         self.named_request_middleware: Dict[str, Deque[MiddlewareType]] = {}
         self.named_response_middleware: Dict[str, Deque[MiddlewareType]] = {}
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -190,7 +190,9 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         self._state: ApplicationState = ApplicationState(app=self)
         self.blueprints: Dict[str, Blueprint] = {}
         self.config: Config = config or Config(
-            load_env=load_env, env_prefix=env_prefix, app=self,
+            load_env=load_env,
+            env_prefix=env_prefix,
+            app=self,
         )
         self.configure_logging: bool = configure_logging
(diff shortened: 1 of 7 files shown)


ashleysommer (approved):
I hope nobody considers this a breaking change, eg if they were relying on the other methods of setting a config variable to _not_ trigger the post-set validation.

PR #2320, opened 2021-11-21
Remove app instance from Config for error handler setting

ahopkins:
This stems from a conversation with @prryplatypus and @ashleysommer re: the change to get the ErrorHandler functioning as expected in v21.9. This unfortunately introduced `app` as a property of the Config object.

This PR removes that, reverting back to the previous methodology, but still retaining the as expected ability to apply changes to the `Config` post-instantiation and carry through to the route handler that was missing in earlier v21.

TODO
- [x] How to handle post-init change to `Config` where `ErrorHandler` is not default
- [x] Value checking when setting `fallback`


ashleysommer (approved):
Thanks for making this change

sanic/handlers.py
```diff
@@ -27,24 +37,92 @@ class ErrorHandler:
...
+                ),
+            )
+
+        if config is None:
+            error_logger.warning(
+                DeprecationWarning(
+                    "Starting in v22.3, config will be a required argument "
+                    "for ErrorHandler.finalize()."
+                ),
+            )
```
prryplatypus, line 118: This shouldn't be needed. As soon as we remove `fallback` we should be able to deprecate the `finalize` method.
```suggestion
```

sanic/handlers.py
```diff
@@ -27,24 +37,92 @@ class ErrorHandler:
...
+            ),
+        )
+
+    @classmethod
+    def _get_fallback_value(cls, error_handler: ErrorHandler, config: Config):
+        if error_handler._fallback is not _default:
+            if config._FALLBACK_ERROR_FORMAT is _default:
+                return error_handler.fallback
+
+            error_logger.warning(
```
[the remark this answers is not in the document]
ahopkins: Not set in stone on this, so I can be convinced otherwise. My thought is that doing it only once will likely mean it is overlooked.

sanic/handlers.py
ahopkins: We'll get there when we get there.

prryplatypus (approved):
LGTM except the `DescriptorMeta` in the `Config` class, because I don't understand what it does 😅

ahopkins:
It stores an internal reference to descriptors so that when setting a value we determine if we need to execute the setter or fallback to the dict update.

PR #2321, opened 2021-11-24
Refactor environment variable hydration logic.

Varriount:
- Refactor environment variable hydration logic to be less nested.  
  This allows possible future extension of the hydration logic.
- Fix a spelling mistake in `load_environment_vars`'s docstring.

### DIFF
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -174,11 +174,11 @@ class Config(dict):
 
     def load_environment_vars(self, prefix=SANIC_PREFIX):
         """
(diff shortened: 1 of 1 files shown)


prryplatypus (approved):

ahopkins (commented):
Thanks for the PR. It seems the code change is mainly stylistic, unless as you say, there were a method to extend that. I suppose that's the question, should we add an api for extension the same way you can extend path parameter types?

ahopkins (approved):

PR #2322, opened 2021-11-26
Fix Type Hint

ChihweiLHBird:
Union is not necessary here? My IDE pops an error for it.

`Union requires two or more type arguments. Pylance`

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -101,7 +101,7 @@ class BaseHTTPResponse:
 
     async def send(
         self,
-        data: Optional[Union[AnyStr]] = None,
+        data: Optional[AnyStr] = None,
         end_stream: Optional[bool] = None,
     ) -> None:
         """


ahopkins (approved):

prryplatypus (approved):

ChihweiLHBird:
The CI doesn't work?

prryplatypus:
It should. Can you try pushing another empty commit?

ChihweiLHBird:
> It should, unless #2311 broke something. Can you try pushing another empty commit?

Pushed, and seems still not working...

prryplatypus:
@ahopkins is it possible #2311 might've borked tests? 😳

ChihweiLHBird:
> @ahopkins is it possible #2311 might've borked tests? 😳
> 
> P.D.: Wonder if they're not re-running because they're not seeing any changes in the `sanic` or `tests` directories 🤔

I think the reason is that #2311 made the CI only run when these two types of events (`review_requested` or `ready_for_review`) are triggered. Pushing a new commit is not considered as the event defined above.

ChihweiLHBird:
I am requesting @prryplatypus's review now, and you can see the CI start running.

ahopkins:
Yeah, clearly some more work is needed on CI. I really was just trying to get it to stop spamming my notifications with test failures whenever there was a push to a DRAFT PR branch. It was getting very annoying.

I just implemented what I saw someone else suggest, but it is not really sufficient, so we need to add something more.

In the meanwhile, feel free to undo those changes, and it will get it working again and we can go back to the drawing board on that.

Anyone interested in looking into this? :pleading_face:

ChihweiLHBird:
> Yeah, clearly some more work is needed on CI. I really was just trying to get it to stop spamming my notifications with test failures whenever there was a push to a DRAFT PR branch. It was getting very annoying.
> 
> I just implemented what I saw someone else suggest, but it is not really sufficient, so we need to add something more.
> 
> In the meanwhile, feel free to undo those changes, and it will get it working again and we can go back to the drawing board on that.
> 
> Anyone interested in looking into this? 🥺

I think current version is okay but we would need document its behavior (if we decided to keep it), lol. But if we want the CI only runs for non-draft PR and runs for every new commit in the non-draft PR, I can try some to do it.

https://github.com/sanic-org/sanic/pull/2324

What do you think? @ahopkins

Issue #2219, opened 2021-08-17
Nicer error handling with mishandled responses


ahopkins:
```python
@app.get("/")
async def handler(request):
    response1 = await request.respond(headers={"one": "one"})
    response2 = await request.respond(headers={"two": "two"})

    await response1.send("One")
    await response2.send("Two")

    await response1.eof()
    await response2.send("???")
    await response2.eof()
```

```
$ curl localhost:9999 -i
HTTP/1.1 200 OK
one: one
transfer-encoding: chunked
connection: keep-alive
content-type: None

OneTwo
```

```
[2021-08-17 15:14:06 +0300] [368537] [ERROR] Exception occurred while handling uri: 'http://localhost:9999/'
Traceback (most recent call last):
  File "handle_request", line 83, in handle_request
    """
  File "/tmp/p.py", line 21, in handler
    await response2.send("???")
  File "/home/adam/Projects/Sanic/sanic/sanic/response.py", line 122, in send
    await self.stream.send(data, end_stream=end_stream)
TypeError: 'NoneType' object is not callable
```

Two things here:
1. When we create the second response, there is no indication to the user that this is somewhat useless. The headers here are ignored, although you can continue sending from it. As seen in example.
2. When `send` is called after the `eof`, there should be a nicer error than `NoneType` that does not explain what happened to the stream.

ChihweiLHBird:
Should has been fixed in https://github.com/sanic-org/sanic/pull/2327

ChihweiLHBird:
```python
@app.get("/case1")
async def handler(request):
    response1 = await request.respond(headers={"one": "one"})
    response2 = await request.respond(headers={"two": "two"})

    await response1.send("One")
    await response2.send("Two")

    await response1.eof()
    await response2.send("???")
    await response2.eof()

@app.get("/case2")
async def handler(request):
    response1 = await request.respond(headers={"one": "one"})

    await response1.send("One")

    await response1.eof()
    await response1.send("Two")


app.run()
```

Log:

```
[2021-12-09 15:33:54 -0600] [1606] [ERROR] Exception occurred while handling uri: 'http://localhost:8000/case1'
Traceback (most recent call last):
  File "handle_request", line 83, in handle_request
    from sanic.models.handler_types import Sanic as SanicVar
  File "test.py", line 87, in handler
    response2 = await request.respond(headers={"two": "two"})
  File "/home/chihwei/sanic/sanic/request.py", line 194, in respond
[...]

Issue #2197, opened 2021-07-14
Streaming response middleware executes twice


ahopkins:
**Describe the bug**
When responding with the response object from a streaming handler, the response middleware is executed twice.

**Code snippet**
Compare:
```python
@app.get("/")
async def handler(request: Request):
    logger.info("handler")
    resp = await request.respond()
    logger.info("sending")
    await resp.send(b"Now I'm free, free-falling")
    logger.info("sent")
    await resp.send(b"Now I'm free, free-falling")
    logger.info("sent")
    await resp.eof()
    logger.info("done")
    return resp


@app.on_request
async def req(_):
    logger.info("request")


@app.on_response
async def res(*_):
    logger.info("response")
```
With:
```python
@app.get("/")
async def handler(request: Request):
    logger.info("handler")
    resp = await request.respond()
    logger.info("sending")
    await resp.send(b"Now I'm free, free-falling")
    logger.info("sent")
    await resp.send(b"Now I'm free, free-falling")
    logger.info("sent")
    await resp.eof()
    logger.info("done")
    # return resp


@app.on_request
async def req(_):
    logger.info("request")


@app.on_response
async def res(*_):
    logger.info("response")
```

```
[2021-07-14 09:23:36 +0300] [1772072] [INFO] request
[2021-07-14 09:23:36 +0300] [1772072] [INFO] handler
[2021-07-14 09:23:36 +0300] [1772072] [INFO] response
[2021-07-14 09:23:36 +0300] [1772072] [INFO] sending
[2021-07-14 09:23:36 +0300] - (sanic.access)[INFO][127.0.0.1:48476]: GET http://localhost:9999/  200 26
[2021-07-14 09:23:36 +0300] [1772072] [INFO] sent
[2021-07-14 09:23:36 +0300] [1772072] [INFO] sent
[2021-07-14 09:23:36 +0300] [1772072] [INFO] done
[2021-07-14 09:23:36 +0300] [1772072] [INFO] response
---
[2021-07-14 09:24:04 +0300] [1772197] [INFO] request
[2021-07-14 09:24:04 +0300] [1772197] [INFO] handler
[2021-07-14 09:24:04 +0300] [1772197] [INFO] response
[2021-07-14 09:24:04 +0300] [1772197] [INFO] sending
[2021-07-14 09:24:04 +0300] - (sanic.access)[INFO][127.0.0.1:48484]: GET http://localhost:9999/  200 26
[...]

PR #2323, opened 2021-11-26
Nicer error handling after eof, warning for second response being created

ChihweiLHBird:
Trying to solve issue: https://github.com/sanic-org/sanic/issues/2219

Some doubts I still have:
1. Whether it's okay to put a `response` reference in `request` instance. I recently learned from the memory leak issue; we should not create more reference if not necessary. I am considering replace it with a bool flag. 
2. Should we only allow one response instance for each request? If so, we can consider keep and cache the response reference and return it if `request.respond` method is called more than 1 time.
3. I believe that `self.response_func is None and self.stage == Stage.IDLE` implies the request has been responded and the response stream is ended. Am I correct on this?

Please let me know your suggestions, thanks! @sanic-org/core-devs

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -61,13 +61,12 @@ from sanic.compat import OS_IS_WINDOWS, enable_windows_color_support
 from sanic.config import SANIC_PREFIX, Config
 from sanic.exceptions import (
     InvalidUsage,
-    ResponseException,
     SanicException,
     ServerError,
-    ShouldNotHandleException,
     URLBuildError,
 )
 from sanic.handlers import ErrorHandler
+from sanic.http import Http, Stage
 from sanic.log import LOGGING_CONFIG_DEFAULTS, Colors, error_logger, logger
 from sanic.mixins.listeners import ListenerEvent
 from sanic.models.futures import (
@@ -737,17 +736,15 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
             context={"request": request, "exception": exception},
         )
 
-        if isinstance(exception, ShouldNotHandleException):
-            error_logger.exception(exception)
-            return
-        try:
-            request.reset_response()
-        except ResponseException as _:
+        if isinstance(request.stream, Http) and request.stream.stage in (
+            Stage.RESPONSE,
+            Stage.IDLE,
+        ):
             error_logger.exception(exception)
             logger.error(
-                "Server error page was not sent to the client for the "
-                "exception above because a previous response has been "
-                "sent at least partially."
+                "The error page was not sent to the client for the "
+                "exception raised above because a previous response has "
+                "already been sent at least partially."
             )
             return
 
@@ -1290,8 +1287,15 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         return None
 
     async def _run_response_middleware(
-        self, request, response, request_name=None
+        self,
+        request: Request,
+        response: BaseHTTPResponse,
+        request_name: Optional[str] = None,
     ):  # no cov
+        if response.middlewares_ran:
+            return response
+        else:
+            response.middlewares_ran = True
         named_middleware = self.named_response_middleware.get(
             request_name, deque()
         )
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -61,6 +61,7 @@ from sanic.compat import OS_IS_WINDOWS, enable_windows_color_support
 from sanic.config import SANIC_PREFIX, Config
 from sanic.exceptions import (
     InvalidUsage,
+    ResponseException,
     SanicException,
     ServerError,
     ShouldNotHandleException,
@@ -739,8 +740,16 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         if isinstance(exception, ShouldNotHandleException):
             error_logger.exception(exception)
             return
-
-        request.reset_response()
+        try:
+            request.reset_response()
+        except ResponseException as _:
+            error_logger.exception(exception)
+            logger.error(
+                "Server error page was not sent to the client for the "
+                "exception above because a previous response has been "
+                "sent at least partially."
+            )
+            return
 
         # -------------------------------------------- #
         # Request Middleware
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -742,8 +742,8 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         ):
             error_logger.exception(exception)
             logger.error(
-                f"The error response won't be sent to the client for the "
-                "{exception} because a previous response has "
+                "The error response won't be sent to the client for the "
+                f"{exception} because a previous response has "
                 "already been sent at least partially."
             )
             return
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1289,10 +1289,6 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
     async def _run_response_middleware(
         self, request, response, request_name=None
     ):  # no cov
-        if response.middlewares_executed:
-            return response
-        else:
-            response.middlewares_executed = True
         named_middleware = self.named_response_middleware.get(
             request_name, deque()
         )
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -63,6 +63,7 @@ from sanic.exceptions import (
     InvalidUsage,
     SanicException,
     ServerError,
+    ShouldNotHandleException,
     URLBuildError,
 )
 from sanic.handlers import ErrorHandler
@@ -734,6 +735,12 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
             inline=True,
             context={"request": request, "exception": exception},
         )
+
+        if isinstance(exception, ShouldNotHandleException):
+            error_logger.exception(exception)
+            return
+
+        request.reset_response()
 
         # -------------------------------------------- #
         # Request Middleware
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -742,8 +742,8 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         ):
             error_logger.exception(exception)
             logger.error(
-                "The error response won't be sent to the client for the exception: "
-                f'"{exception}" because a previous response has '
+                "The error response won't be sent to the client for the "
+                f'exception: "{exception}" because a previous response has '
                 "already been sent at least partially."
             )
             return
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -743,7 +743,7 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
             error_logger.exception(exception)
             logger.error(
                 "The error response won't be sent to the client for the exception: "
-                f"\"{exception}\" because a previous response has "
+                f'"{exception}" because a previous response has '
                 "already been sent at least partially."
             )
             return
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1287,10 +1287,7 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         return None
 
     async def _run_response_middleware(
-        self,
-        request: Request,
-        response: BaseHTTPResponse,
-        request_name: Optional[str] = None,
+        self, request, response, request_name=None
     ):  # no cov
         if response.middlewares_executed:
             return response
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -742,8 +742,8 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         ):
             error_logger.exception(exception)
             logger.error(
-                "The error page was not sent to the client for the "
-                "exception raised above because a previous response has "
+                f"The error response won't be sent to the client for the "
+                "{exception} because a previous response has "
                 "already been sent at least partially."
             )
             return
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -742,8 +742,8 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         ):
             error_logger.exception(exception)
             logger.error(
-                "The error response won't be sent to the client for the "
-                f"{exception} because a previous response has "
+                "The error response won't be sent to the client for the exception: "
+                f"\"{exception}\" because a previous response has "
                 "already been sent at least partially."
             )
             return
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1292,10 +1292,10 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         response: BaseHTTPResponse,
         request_name: Optional[str] = None,
     ):  # no cov
-        if response.middlewares_ran:
+        if response.middlewares_executed:
             return response
         else:
-            response.middlewares_ran = True
+            response.middlewares_executed = True
         named_middleware = self.named_response_middleware.get(
             request_name, deque()
         )
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -32,7 +32,7 @@ from httptools import parse_url  # type: ignore
 
 from sanic.compat import CancelledErrors, Header
 from sanic.constants import DEFAULT_HTTP_CONTENT_TYPE
-from sanic.exceptions import InvalidUsage
+from sanic.exceptions import InvalidUsage, SanicException
 from sanic.headers import (
     AcceptContainer,
     Options,
@@ -175,9 +175,9 @@ class Request:
         content_type: Optional[str] = None,
     ):
         if self.response:
-            logger.warning(
-                "Another response instance was created before, please consider "
-                "re-use it rather than creating a new one."
+            raise SanicException(
+                "Another response instance was created, "
+                "creating the second response is not allowed for this request."
             )
         # This logic of determining which response to use is subject to change
         if response is None:
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -18,7 +18,6 @@ from sanic_routing.route import Route  # type: ignore
 if TYPE_CHECKING:
     from sanic.server import ConnInfo
     from sanic.app import Sanic
-    from sanic.http import Http
 
 import email.utils
 import uuid
@@ -32,7 +31,7 @@ from httptools import parse_url  # type: ignore
 
 from sanic.compat import CancelledErrors, Header
 from sanic.constants import DEFAULT_HTTP_CONTENT_TYPE
-from sanic.exceptions import InvalidUsage, SanicException
+from sanic.exceptions import InvalidUsage, ResponseException, SanicException
 from sanic.headers import (
     AcceptContainer,
     Options,
@@ -42,6 +41,7 @@ from sanic.headers import (
     parse_host,
     parse_xforwarded,
 )
+from sanic.http import Http, Stage
 from sanic.log import error_logger, logger
 from sanic.models.protocol_types import TransportProtocol
 from sanic.response import BaseHTTPResponse, HTTPResponse
@@ -109,7 +109,6 @@ class Request:
         "stream",
         "transport",
         "version",
-        "response",
     )
 
     def __init__(
@@ -150,7 +149,6 @@ class Request:
         self.parsed_not_grouped_args: DefaultDict[
             Tuple[bool, bool, str, str], List[Tuple[str, str]]
         ] = defaultdict(list)
-        self.response: Optional[BaseHTTPResponse] = None
         self.request_middleware_started = False
         self._cookies: Optional[Dict[str, str]] = None
         self._match_info: Dict[str, Any] = {}
@@ -166,6 +164,15 @@ class Request:
     def generate_id(*_):
         return uuid.uuid4()
 
+    def reset_response(self):
+        if isinstance(self.stream, Http):
+            if self.stream.stage in (Stage.RESPONSE, Stage.IDLE):
+                raise ResponseException(
+                    "Response can't be reset because it was already sent."
+                )
+            elif self.stream.response:
+                self.stream.response = None
+
     async def respond(
         self,
         response: Optional[BaseHTTPResponse] = None,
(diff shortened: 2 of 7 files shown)


Tronic:
1-2. Trying to make another response should probably just fail, especially if the headers of the previous one have already been sent. Perhaps this can be handled if the original one was never actually sent anywhere.

3. I believe you are right. IDLE means that no request is being served but that the pipeline is still alive. I am not exactly sure why that also needs to test for `response_func` separately but certainly there was a reason for that when I wrote it but the stage meanings have also changed since. I think that during `Stage.RESPONSE` you can check `response_func` to see whether a response has already been initiated.

ChihweiLHBird:
> 3 - I believe you are right. IDLE means that no request is being served but that the pipeline is still alive. I am not exactly sure why that also needs to test for `response_func` separately but certainly there was a reason for that when I wrote it but the stage meanings have also changed since. I think that during `Stage.RESPONSE` you can check `response_func` to find which state the response is in (not started, something sent, already EOF).

Currently, the exception raised for sending some data after `end_stream` being set to `True` is `send` method is NonType, and the `send` method is a property that returns `response_func`. I am not near my laptop right now and above is from my memory. That's why I thought testing whether `response_func` is None or not.

ChihweiLHBird:
> 1-2. Trying to make another response should probably just fail, especially if the headers of the previous one have already been sent. Perhaps this can be handled if the original one was never actually sent anywhere.

Like raising an exception when the user is trying to create second response and the first response sent some data?

ChihweiLHBird:
Found a tricky problem that in some circumstances, if an exception raised during the making response, the exception handler would try to generate another response which can lead to the exception again. Have no idea how to make progress so far. Probably need to look into the exception handling mechanism, which is complicated.

ChihweiLHBird:
I would like to propose the change included in this PR for issues: https://github.com/sanic-org/sanic/issues/2219 and https://github.com/sanic-org/sanic/issues/2197

Basically, it prevents second response object being created from `request.respond`. ResponseException is not always handled by usual exception handler that returns a default 500 page because a part of the response stream might have been sent. To be honest, these changes is not perfect in my mind because it seems added code snippets that makes the modified modules unclear as before, but it is the best I can do so far. Can you guys let me know if these changes are ok? Maybe I got some of them or all of them wrong? Thank you so much for guiding me, I really appreciate it.

ahopkins:
> I would like to propose the change included in this PR for issues: #2219 and #2197

Agreed. Those can be tackled in this PR.


> 1-2. Trying to make another response should probably just fail, especially if the headers of the previous one have already been sent. Perhaps this can be handled if the original one was never actually sent anywhere.

I am not sure what you mean by fail. I agree that we should not allow multiple response instances to be generated. As the original use case from the issue shows, weird behaviors arise. However, I am not sure that raising an exception is appropriate either. It sounds like it might an area ripe for race conditions as the sending of the response might be in different states. The exception might be thrown inconsistently during the response.

Therefore, I am more in favor of the idea that it either fails silently to the logs (probably also difficult to debug), or just return the same response (as mentioned, what would the memory impact look like?) Clearly any attempt to change the headers after the first byte was sent should fail. Again, should it sack the whole response or just do it to the logs? I am more inclined to allow the response to complete.

Another though popped into my head. I am not sure this is a road we want to go down, but at startup we already examine route handlers to see what kind of responses they might send. Perhaps we piggyback off that to inspect if there might be the case for multiple responses. If we throw an exception, I would rather do it at startup.

Tronic:
Where could we end up in this situation? If a handler attempts to start another response or returns something after using `request.respond`, there is a clear logic error in the program. Other common case supposedly is when an exception is risen and Sanic then tries to send its error response. The latter is a case where Sanic exception handling should probably detect this and not even try to send an error response, but only log it.

Now, in either of these situations, what to do with the response in progress?

I believe that in any case, if the handler has not signaled EOF yet, it is wrong for Sanic to complete the response (even if it can, depending on content-length header). It is better to disconnect the TCP connection, which will cause an error in browser that is not pretty but it is better than giving "successful" 200 responses with incomplete data.

On the other hand, if the handler has signaled EOF, from client perspective there is nothing wrong, and Sanic can just silently log the error and even keep the keep-alive alive. A bit of risky business there in case we have any logic errors in Sanic, so it might still be safer to disable keep-alive at this point. No error on client but it will have to make another connection.

ahopkins:
Something like this is what I had in mind:

```diff
diff --git a/sanic/mixins/routes.py b/sanic/mixins/routes.py
index a543a05..667ad64 100644
--- a/sanic/mixins/routes.py
+++ b/sanic/mixins/routes.py
@@ -20,6 +20,7 @@ from sanic.exceptions import (
     FileNotFound,
     HeaderNotFound,
     InvalidUsage,
+    SanicException,
 )
 from sanic.handlers import ContentRangeHandler
 from sanic.log import error_logger
@@ -934,6 +935,7 @@ class RouteMixin:

     def _get_response_types(self, node):
         types = set()
+        responds = 0

         class HttpResponseVisitor(NodeVisitor):
             def visit_Return(self, node: Return) -> Any:
@@ -954,6 +956,14 @@ class RouteMixin:
                 except AttributeError:
                     ...

-        HttpResponseVisitor().visit(node)
+            def visit_Call(self, node) -> Any:
+                nonlocal responds
+
+                responds += node.func.attr == "respond"

+        HttpResponseVisitor().visit(node)
+        if responds > 1:
+            raise SanicException(
+                "Multiple streaming responses identified in handler"
+            )
         return types
```

Tronic:
@ahopkins Is that Visitor pattern? I am too tired to understand that code now.

ahopkins:
> @ahopkins Is that Visitor pattern? I am too tired to understand that code now.

Yeah. It's a super simple method to look for callables named "respond". Just meant as an illustration, not a complete solution.

ChihweiLHBird:
@ahopkins Do you think we should still handler the second response situation beside the code inspection for the user specified request handler? One reason I can think of is that second responding not only happens in the request handler, but also in Sanic's internal code, for example exception handler. Consider the code below:

```python
@app.get("/")
    resp1: HTTPResponse = await request.respond()
    await resp1.send(b"Now I'm free, free-falling")
    raise SanicException("Some text here.")
```

`resp1` was created as a response, but since a Sanic exception is raised, `handle_exception` will be invoked to handle it and try to create another response for the Sanic's 500 error page. Some weird behaviours would happen.

ChihweiLHBird:
Currently, I plan to make the following changes as we discussed:
1. Try to implement code inspection for request handler.
2. When `request.respond` is being called second time, return the same response instance. 
a. If different `headers`, `status`, and `content_type` are provided, then `response` instance's corresponding attributes would be overridden if it was never sent (partially or fully). 
b. If `response` has already been sent while different parameters mentioned above are provided, log an error and return the same response instance

Does it sound good for you guys? @Tronic @ahopkins Thanks for feedbacks!

Tronic:
> 2. When `request.respond` is being called second time, return the same response instance.
>    a. If different `headers`, `status`, and `content_type` are provided, then `response` instance's corresponding attributes would be overridden if it was never sent (partially or fully).
>    b. If `response` has already been sent while different parameters mentioned above are provided, log an error and return the same response instance

I believe there should be **no effort to do any of this**. 2a may lead to problematic responses that accidentally combine unrelated headers, status or content types when the latter response assumes to have the default values. 2b should raise an exception to terminate the code that attempts another response (which will certainly lead to invalid results).

In 2b, as a practical example:
1. The handler makes a 200-response and starts sending data (using chunked encoding)
2. The client receives and saves a file as it is being sent
3. The handler has some logic error such that it for whatever reason starts sending another unrelated response
4. The client appends the other message to the file it saved thinking that everything was OK, while it actually got corrupt data!
5. The handler sees no error and writes to database that the request was successfully completed

A very similar but maybe more common flow follows if step 3 is an exception (e.g. cannot read more data from database) and the exception is handled by Sanic, and then in step 4 the client silently appends the error message to the file it downloads, but there is no step 5 because the exception caused the handler to terminate.

The only case of another response that **could be accepted** is when no response packets have been sent to the client yet, and in that case this might be best implemented in **http.py**:

```python
    def respond(self, response: BaseHTTPResponse) -> BaseHTTPResponse:
        """
        Initiate new streaming response.

        Nothing is sent until the first send() call on the returned object, and
        calling this function multiple times will just alter the response to be
        given.
        """
        if self.stage is not Stage.HANDLER:
            self.stage = Stage.FAILED
            raise RuntimeError("Response already started")

        # Disconnect any earlier but unused response object  (this part is new code)
        if self.response is not None:
            self.response.stream = None

        # Connect and return the response
        self.response, response.stream = response, self
        return response
```

The ASGI mode does not use http.py and that backend might be sending the headers earlier anyway (I don't recall, @ahopkins might), but in any case would then need separate handling of this to safely support replacing unused response objects with new ones.

```python
        # This logic of determining which response to use is subject to change
        if response is None:
            response = (self.stream and self.stream.response) or HTTPResponse(
```

The above part in **request.py** is written by me and seems to be a part of the problem. I now believe it is wrong to use an existing response object (this would silently ignore any status, headers and content type parameters). Also, this function is another potential place to implement what I suggested for http.py above, while also covering ASGI.

> 3. Make `Sanic.handle_exception` method detect whether the response has been sent (partially or fully), and only log the error without making second 500 response if previous response was sent. This is somehow already implemented in this PR, but need some improvements, I think.

This is a good thing, and is what I proposed in the previous message.

Tronic:
I would very much like to see this PR included in 21.12 LTS. Looks like it fixes some rather nasty and dangerous bugs.

ahopkins:
I am for getting this one in. We have time, but it is December now, so that means we need to finish this in the next week or so.

I agree that the checks on the subsequent respond calls are unnecessary and too complicated. I am okay with raising an error there, but does bring into question the existing response.

```python
resp1 = await request.respond(...)
await resp.send(...)

resp2 = await request.respond(...)
```
Raising an exception here would be weird. Headers have been sent. The client is already expecting a complete response. This is why I was suggesting to return the same instance and have some sort of logging only failure.

I suppose it is really the same as this:

```python
resp = await request.respond()
await resp.send(...)

raise Exception(...)
```

So, ultimately I agree that is probably the correct approach. Plus it should be simpler to implement.

@ChihweiLHBird As for Item 1 (code inspection), if we can sneak that in great. But do not focus on that. It is much more important to get the rest of this right and in 21.12. If the inspection needs to happen in 22.3, then that is fine with me.

ahopkins:
Correction, now that I am looking at this and playing again I remember that this fails anyway:

```python
resp1 = await request.respond(...)
await resp.send(...)

resp2 = await request.respond(...)
```

The use case that is the problem is instantiating the response twice before anything is sent which seems to work okay client side but fails server side.

In that case, it really should be fine to throw the exception the second time `respond` is called.

ChihweiLHBird:
Oh, just realized that `stream.respond` already handled and raised a `RuntimeError` when being called after a response has at least been partially sent....

ChihweiLHBird:
Currently, Sanic assumes exception handler specified by user would return a response and Sanic would make a response if the handler doesn't make one. In my experiment, if `Sanic.handle_exception` method is skipped for the exception that involves a partially sent response, then the exception handler specified by the user wouldn't work. Does this sound like a problem or this is out wanted behavior?

ahopkins:
> Does this sound like a problem or this is out wanted behavior?

Exception handlers should have a response. The only use case when handlers should not have a response is when they are streaming, and I am not sure that we should even be concerned with streaming from an exception handler.

sanic/app.py
```diff
@@ -735,6 +736,18 @@ async def handle_exception(
...
         )
 
+        if isinstance(request.stream, Http) and request.stream.stage in (
+            Stage.RESPONSE,
+            Stage.IDLE,
+        ):
+            error_logger.exception(exception)
+            logger.error(
+                "The error page was not sent to the client for the "
+                "exception raised above because a previous response has "
```
ahopkins, line 746: The loggers may not be going to the same location, so we should not say "above"

```python
f"exception raised ({exception}) because ...
```

sanic/app.py
```diff
@@ -1276,6 +1289,10 @@ async def _run_request_middleware(
     async def _run_response_middleware(
         self, request, response, request_name=None
     ):  # no cov
+        if response.middlewares_executed:
+            return response
+        else:
+            response.middlewares_executed = True
```
ahopkins, line 1295: This seems like a bit of a hack to me. Why not try to prevent them from being called multiple times in the first place?

sanic/request.py
```diff
@@ -172,6 +172,10 @@ async def respond(
         headers: Optional[Union[Header, Dict[str, str]]] = None,
         content_type: Optional[str] = None,
     ):
+        if isinstance(self.stream, Http) and self.stream.stage is Stage.IDLE:
+            raise SanicException(
+                "Another response was sent previously."
+            )
```
ahopkins, line 178: What is `self.stream` otherwise? `None`?

I would either check for `is None` or:

```python
try:
    if self.stream.stage is Stage.IDLE:
        raise SanicException(...)
except AttributeError:
    ...
```

sanic/response.py
```diff
@@ -112,8 +114,16 @@ async def send(
...
             end_stream = True
-        if end_stream and not data and self.stream.send is None:
-            return
+        if self.stream.send is None:
+            if end_stream and not data:
+                return
+            elif self.stream.stage == Stage.IDLE:
+                raise SanicException(
+                    "Response stream was ended, no more response data is"
+                    " allow ed to be sent."
```
ahopkins, line 123: Extra space in `allow ed`.
Also, add the trailing space to the first line and remove from before `allow`.

sanic/response.py
```diff
@@ -112,8 +114,16 @@ async def send(
...
+                )
+            else:
+                raise SanicException("Send response function isn't available")
```
ahopkins, line 126: Should we add the `Stage` to this?

tests/test_middleware.py
```diff
@@ -297,3 +297,28 @@ async def handler(request):
...
+    @app.on_request
+    def request(_):
+        nonlocal request_middleware_run_count
+        request_middleware_run_count += 1
+
+    @app.get("/")
+    async def handler(request):
+        resp1 = await request.respond()
+        await resp1.eof()
+        return text("")
```
ahopkins, line 320: Food for thought, maybe we should never allow a return on a `respond` handler.

sanic/request.py
ChihweiLHBird: https://github.com/sanic-org/sanic-testing/issues/27

In some test cases, the stream is not `Http` 😂

sanic/app.py
ChihweiLHBird: Let me thinking about this

sanic/app.py
ChihweiLHBird: fixed, the new one looks like:

```python
logger.error(
    f"The error response won't be sent to the client for the "
     {exception} because a previous response has "
    "already been sent at least partially."
)
```

sanic/response.py
ChihweiLHBird: I am not sure. I couldn't think about any other situation that the function is None, but I added a else just in case it falls to this.

tests/test_middleware.py
ChihweiLHBird: Got it, will rewrite the tests

Tronic:
> Furthermore, if there is a partially sent response, then it does not make sense that the handler would be able to take over. It couldn't send a valid HTTP response at that point.

Suppose a custom handler was used to do something server side (like log errors somewhere), it might be expected that they still get executed, and Sanic could then just never send the response that they return. But then, if implementing that is not absolutely trivial, I wouldn't bother (given how unlikely such situation is - need an exception being raised during a streaming response, and then a custom handler that attempts to do something about it server side).

sanic/app.py
ChihweiLHBird: Now it is being checked in `request.respond`, where middleware is executed. Do you think it is better now?

ChihweiLHBird:
> > Furthermore, if there is a partially sent response, then it does not make sense that the handler would be able to take over. It couldn't send a valid HTTP response at that point.
> 
> Suppose a custom handler was used to do something server side (like log errors somewhere), it might be expected that they still get executed, and Sanic could then just never send the response that they return. But then, if implementing that is not absolutely trivial, I wouldn't bother (given how unlikely such situation is - need an exception being raised during a streaming response, and then a custom handler that attempts to do something about it server side).

Exactly, that's what I concerned about; some exception handlers not necessarily have to return a response but might have some business logic for the exception, like sending a report to an email address or log it somewhere else. And I am wondering whether it is worth to still execute the handler without sending its response.

ahopkins:
> Exactly, that's what I concerned about; some exception handlers not necessarily have to return a response but might have some business logic for the exception, like sending a report to an email address or log it somewhere else. And I am wondering whether it is worth to still execute the handler without sending its response.

I am not sure that this have ever been an express supported use case for exception handlers. This is precisely the sort of thing that `@app.signal("http.lifecycle.exception")` is meant to cover.

sanic/response.py
ahopkins: I agree that it is the correct approach to have a failsafe, but I was just thinking that if that happens maybe we should include a little bit of information about whatever unexpected state was reached.

sanic/request.py
ahopkins: Got it. That brings up the earlier point. I think the more proper way to handle this then would be to add stage to that as discussed elsewhere.

sanic/app.py
ahopkins: will look.

ahopkins:
I spent a little time hacking on the code inspection. Again, this is not necessarily meant to be included here, but it seems as good a place as any to keep it for now.

<details>
<summary>Source code for respond inspector</summary>

```python
class Visitor(ast.NodeVisitor):
    def __init__(self, tree, assignments) -> None:
        super().__init__()
        self.respond_refs = 0
        self.tree = tree
        self.assignments = assignments

    @classmethod
    def load(cls, handler):
        src = dedent(getsource(handler))
        tree = ast.parse(src)
        node = tree.body[0]
        idx = 1 if node.args.args[0].arg == "self" else 0
        assignments = [(node.args.args[idx].arg, "respond")]
        return cls(tree, assignments)

    def count_refs(self):
        self.visit(self.tree)
        return self.respond_refs

    def visit_Assign(self, node: ast.Assign) -> Any:
        for name, attr in self.assignments:
            if (
                isinstance(node.value, ast.Attribute)
                and node.value.value.id == name
                and node.value.attr == attr
            ) or (
                isinstance(node.value, ast.Name)
                and not attr
                and node.value.id == name
            ):
                for target in node.targets:
                    self.assignments.append((target.id, None))

    def visit_Call(self, node: ast.Call) -> Any:
        for name, attr in self.assignments:
            if (
                attr
                and isinstance(node.func, ast.Attribute)
                and node.func.value.id == name
                and node.func.attr == attr
            ):
                self.respond_refs += 1
            elif (
                not attr
                and isinstance(node.func, ast.Name)
                and node.func.id == name
            ):
                self.respond_refs += 1
```
</details>

Current limitations:
- Only works on CBV if it is a `staticmethod` or the instance ref is `self`
- Does not take scoping into account, therefore two `respond` calls that are in different paths could still trigger a count

What it does handle:
```python
@app.get("/")
async def handler(request):
    await request.respond()
    foo = request.respond
    bar = foo
    await bar()
    await foo()

class HandlerView(HTTPMethodView):
    async def get(self, request: Request):
        await request.respond()
        await request.respond()

    @staticmethod
    async def post(req: Request):
        await req.respond()
        await req.respond()
```

---

One potential option might just be to use some iteration of this and instead of throwing an exception, just raising a warning:

```
Route handler "handler" may contain more than one response object.
```

sanic/request.py
```diff
@@ -179,20 +179,34 @@ async def respond(
...
-            response = await self.app._run_response_middleware(
-                self, response, request_name=self.name
-            )
-        except CancelledErrors:
-            raise
-        except Exception:
-            error_logger.exception(
-                "Exception occurred in one of response middleware handlers"
-            )
+        if not re_use_res:
```
ahopkins, line 199: I like this better. Let's just use a better variable name: `should_reuse_response`

sanic/app.py
ChihweiLHBird: Thanks, I think I can consider this as resolved. 😊

Tronic:
@ChihweiLHBird Please check for `Stage.HANDLER` in all code that needs to check whether a response can be sent, to stay consistent with http.py. This is the only stage where you can, no need to check for IDLE or RESPONSE anywhere.

Tronic:
One thing I mentioned earlier, that is not implemented, is that responses should never be reused or combined. Could you fix that as well?

And possibly we could delay the running of response middlewares to the first send? This would entirely avoid the problem with them being executed twice, or being executed for responses that never get sent.

Tronic:
asgi.py lines 166-190 show no Stages and no handling of multiple responses. Namely, calling respond() again does not check for a response already started and does not disconnect any previous response, just connects with the new one blindly. The send() function then uses the most recent response and clears it. So, then one can initiate another response and corrupt the stream (which hopefully then raises an exception in the ASGI backend).

This needs to be fixed. I suppose that adding a stage variable on ASGI would be a rather simple way to do it.

ahopkins:
> And possibly we could delay the running of response middlewares to the first send?

I am assuming you mean right before it? As in the first `.send()` will run the middleware before hitting the transport? That would be a breaking change. Not sure if it matters.

Tronic:
> I am assuming you mean right before it? As in the first `.send()` will run the middleware before hitting the transport? That would be a breaking change. Not sure if it matters.

Yes, but I am not very convinced that this is a good idea. I gather that it is very rare for multiple responses to be created, and that an extra execution of the middleware then (for the response never sent) is not a big deal. Also, the way that it is done now potentially allows middlewares to pass information back to handler (not that I could immediately think of any use for that either). But maybe most importantly, it might be a bit confusing to developers if these middlewares are triggered on `send()`, rather than on `respond()`

tests/test_middleware.py
ChihweiLHBird: Considered as resolved.

ahopkins:
Agreed.

ChihweiLHBird:
> This needs to be fixed. I suppose that adding a stage variable on ASGI would be a rather simple way to do it.

Do you expect that to be done in this PR? I think this PR became complicated and we might want a separate PR for it, lol.

sanic/request.py
ChihweiLHBird: Can you remind me what's the proper way? I don't remember I saw it somewhere else...

sanic/response.py
ChihweiLHBird: What info can be included do you think? I couldn't think about more than that.

ChihweiLHBird:
> I spent a little time hacking on the code inspection. Again, this is not necessarily meant to be included here, but it seems as good a place as any to keep it for now.
> 
> Source code for respond inspector
> Current limitations:
> 
> * Only works on CBV if it is a `staticmethod` or the instance ref is `self`
> * Does not take scoping into account, therefore two `respond` calls that are in different paths could still trigger a count
> 
> What it does handle:
> 
> ```python
> @app.get("/")
> async def handler(request):
>     await request.respond()
>     foo = request.respond
>     bar = foo
>     await bar()
>     await foo()
> 
> class HandlerView(HTTPMethodView):
>     async def get(self, request: Request):
>         await request.respond()
>         await request.respond()
> 
>     @staticmethod
>     async def post(req: Request):
>         await req.respond()
>         await req.respond()
> ```
> 
> One potential option might just be to use some iteration of this and instead of throwing an exception, just raising a warning:
> 
> ```
> Route handler "handler" may contain more than one response object.
> ```

Yeah, let's open another PR later for it later.

ChihweiLHBird:
Too many stuffs going on, let's continue here to summarize them: https://github.com/sanic-org/sanic/pull/2327

PR #2327, opened 2021-12-02
Prevent sending multiple or mixed responses on a single request

ChihweiLHBird:
Previous PR thread was getting long. This is a new implementation includes the logic that prevents second `request.respond` call. 

Let's summarize what we expect to be done in this PR and work toward that. Also, the future works planned. Sounds good? @Tronic @ahopkins 

I think most comments were resolved here, but I may miss some either not sure about the implementation or forget.

Previous PR: https://github.com/sanic-org/sanic/pull/2323


sanic/request.py
```diff
@@ -164,6 +166,20 @@ def __repr__(self):
...
+    def reset_response(self):
+        try:
+            if (
+                self.stream is not None
+                and self.stream.stage is not Stage.HANDLER
+            ):
+                raise ResponseException(
+                    "Cannot reset response because previous response was sent."
+                )
+            self.stream.response = None
```
Tronic, line 178: Should also clear the link from response to stream, i.e. `self.stream.response.stream = None`. Always disconnect them both ways to avoid misuse and to allow faster freeing of memory.

sanic/request.py
```diff
@@ -172,13 +188,19 @@ async def respond(
...
+                raise ResponseException("Second respond call is not allowed.")
+        except AttributeError:
+            pass
         # This logic of determining which response to use is subject to change
         if response is None:
             response = (self.stream and self.stream.response) or HTTPResponse(
                 status=status,
                 headers=headers,
                 content_type=content_type,
             )
```
Tronic, line 202: Still need to change this not to test for `self.stream and self.stream.response` (thanks to the test you added right before this, it is always `None`).

sanic/request.py
```diff
@@ -104,6 +104,7 @@ class Request:
         "parsed_json",
         "parsed_forwarded",
         "raw_url",
+        "responded",
```
Tronic, line 107: Looks like this variable is unnecessary, as you can simply check `self.response` for the same thing.

sanic/response.py
```diff
@@ -112,8 +113,16 @@ async def send(
...
+        if self.stream.send is None:
+            if end_stream and not data:
+                return
+            elif self.stream.stage == Stage.IDLE:
+                raise ResponseException(
+                    "Response stream was ended, no more response data is "
+                    "allowed to be sent."
+                )
+            else:
+                raise SanicException("Send response function isn't available")
```
Tronic, line 125: Minor style issues. 'elif' and 'else' are unnecessary because each case returns/raises. I suggest two regular ifs and nothing on the final part (when does the final one occur, anyway? -- if it doesn't, just give the stream has ended exception without any test). Also `is` comparison for Stage enum.

sanic/request.py
ChihweiLHBird: Fixed, thanks!

sanic/request.py
ChihweiLHBird: Fixed.

sanic/response.py
ChihweiLHBird: Thanks, fixed.

ChihweiLHBird: Actually, if the last raising removed, type checks will fail because None is not callable. So, I would propose make a message string (e4b53a01c2f633fbfea1a1d5b580fc457bf0cc1b), wondering whether it looks okay.

sanic/app.py
```diff
@@ -735,6 +736,18 @@ async def handle_exception(
             context={"request": request, "exception": exception},
         )
 
+        if (
+            request.stream is not None
+            and request.stream.stage is not Stage.HANDLER
+        ):
+            error_logger.exception(exception)
```
ahopkins, line 743: We should explicitly add `exc_info=True`.

sanic/app.py
```diff
@@ -735,6 +736,18 @@ async def handle_exception(
...
+            logger.error(
+                "The error response won't be sent to the client for the "
+                f'exception: "{exception}" because a previous response has '
+                "already been sent at least partially."
```
ahopkins, line 747: > The error response will not be sent to the client for the following exception:___. A previous response has at least partially been sent.

sanic/exceptions.py
```diff
@@ -266,3 +266,7 @@ def abort(status_code: int, message: Optional[Union[str, bytes]] = None):
     )
 
     raise SanicException(message=message, status_code=status_code)
+
+
+class ResponseException(SanicException):
```
ahopkins, line 271: Do we need this? Why not `ServerError`?

ahopkins (commented):
Thanks for all the work on this 🤘

sanic/app.py
ChihweiLHBird: Why we would need to explicitly add this? I think `error_logger.exception` has that as the default behavior?

sanic/app.py
ChihweiLHBird: Fixed, thanks!

sanic/exceptions.py
ChihweiLHBird: Originally, I thought we need a new type of `Exception` that we would need to ignore in `handle_exception`, but now `handle_exception` can detect the stage, so we don't need it anymore. 

Fixed.

sanic/app.py
ChihweiLHBird: Added.

ChihweiLHBird:
> Thanks for all the work on this 🤘

Thank you and @Tronic for helping me to understand everything and get them correct. 😊

sanic/asgi.py
```diff
@@ -195,6 +204,8 @@ async def __call__(self) -> None:
         Handle the incoming request.
         """
         try:
+            self.stage = Stage.HANDLER
             await self.sanic_app.handle_request(self.request)
         except Exception as e:
             await self.sanic_app.handle_exception(self.request, e)
+            self.stage = Stage.FAILED
```
Tronic, line 211: This should probably not enter `Stage.FAILED`. Just having an exception is not a failure, but an incomplete/unfinished response is. If either the handler or `handle_exception` managed to send a complete response, the stream should go to `Stage.IDLE` and continue normally.

sanic/asgi.py
```diff
@@ -164,10 +169,14 @@ async def __aiter__(self):
                 yield data
 
     def respond(self, response):
+        if self.stage is not Stage.HANDLER:
+            self.stage = Stage.FAILED
+            raise RuntimeError("Response already started")
         response.stream, self.response = self, response
         return response
```
Tronic, line 176: Disconnect an earlier response here as well, to stay compatible with http.py.

ChihweiLHBird:
> I suggest replacing the whole block with
> 
> ```python
>             if end_stream and not data:
>                 return
>             raise ResponseException(
>                 "Response stream was ended, no more response data is allowed to be sent."
>             )
> ```
> 
> Because I cannot find out any other situation where we would end up here, justifying the "response function is not available" error.

Agreed and added, thanks.

sanic/asgi.py
ChihweiLHBird: Fixed

Tronic:
It has been a huge undertaking, but looks like this one is getting real close to being approved and merged. Great work :)

sanic/asgi.py
ChihweiLHBird: I added a finally block and set stage to idle, do you think that works?

sanic/asgi.py
```diff
@@ -213,4 +213,5 @@ async def __call__(self) -> None:
             await self.sanic_app.handle_request(self.request)
         except Exception as e:
             await self.sanic_app.handle_exception(self.request, e)
-            self.stage = Stage.FAILED
+        finally:
+            self.stage = Stage.IDLE
```
Tronic, line 217: This might not be correct either. What if it goes to FAILED state during handler or exception? Then this sets it back to IDLE.

However, I cannot have a proper look at ASGI to see how it behaves, but I gather we should go FAILED only when a response is left without EOF. But no idea of how to actually handle that with ASGI servers. Needs some more investigation, in any case. Also prefer to keep this as close to http.py as possible.

sanic/app.py
ahopkins: I think you're right, but I want to check. I was (and still am) on my phone. Please confirm.

sanic/asgi.py
ChihweiLHBird: Would the lifecycle of the object (`Http` or `ASGI`, how do you guys call this type of object? Connection or something else?) be ended after `Stage.IDLE` or `stage.FAILED`? Can it accept further in-coming messages from the client? If it can, that means setting it to `Stage.IDLE` can make it accept more messages? 

By the way, I only see the stage being set to failed here in Sanic repo:
image

http.py, line 575
```python
    def respond(self, response: BaseHTTPResponse) -> BaseHTTPResponse:
        """
        Initiate new streaming response.

        Nothing is sent until the first send() call on the returned object, and
        calling this function multiple times will just alter the response to be
        given.
        """
        if self.stage is not Stage.HANDLER:
            self.stage = Stage.FAILED
            raise RuntimeError("Response already started")
```

ChihweiLHBird: What about this?
```python
finally:
    if self.stage == Stage.RESPONSE:
        self.stage = Stage.IDLE
```

ChihweiLHBird: I think maybe we don't need that in `__call__` at all, because it will be done in the send function, same in `http.py`

Tronic: I have a gut feeling that using try-finally is never good for this sort of operation. Can you set it IDLE where the final ASGI message (with EOF) is sent, and not in this function? Following the same flow as http.py is a good idea, and doing it in send() seems reasonable.

The meaning of `FAILED` is that nothing else can be done on the connection anymore, meaning that it needs to be disconnected. Not quite sure how that translates to ASGI, so cannot help you with that, but if possible, we should somehow tell the server that we are unable to make a valid response.

ChihweiLHBird: > I have a gut feeling that using try-finally is never good for this sort of operation. Can you set it IDLE where the final ASGI message (with EOF) is sent, and not in this function? Following the same flow as http.py is a good idea, and doing it in send() seems reasonable.

Thanks, I think it has been done. ✅

> if possible, we should somehow tell the server that we are unable to make a valid response.

Yeah, I can definitely start looking to it. Since the current version worked somehow well on this, I think we can introduce this functionality in the future PR? 😂

Tronic (approved):
I believe all notes have been addressed, but suggest additional review by someone else, considering the complexity of this.

sanic/asgi.py
ChihweiLHBird: Resolved below in another note.

ChihweiLHBird:
> I believe all notes have been addressed, but suggest additional review by someone else, considering the complexity of this.

Thanks @Tronic.

@ahopkins May I have you to review this? Would you suggest any other changes to this PR? 

And is there anyone else that can help to review this?

Thank you!

ahopkins:
> I believe all notes have been addressed, but suggest additional review by someone else, considering the complexity of this.

I am taking a look at this now.

ahopkins:
## Overview

I wanted to test the following items and what impact would be with multiple responses with different `send()` patterns.

1. Varying the location of a `send` with multiple `respond`
2. Varying the response with different headers
3. Returning another response object

In the end, item 2 had no impact. I just set the headers on each `respond`, and the results were no different from item 1, so I did not put the results below. Further, creating a `HTTPResponse` object normally seems to have no impact at all. It does not generate an error, and it does not impact the response.

## Thoughts

While I generally see no problems with the code, and I think ___ did a great job on this PR, I am not sure that it is complete and/or correct. First, it seems we are no longer requiring `eof()` (or, the empty `send` with `end_stream=True`). It does not seem like that was introduced in this PR, and I am not sure where that made it in. While it certainly does seem like a nice convenience to be able to skip that step, I feel that if we require the explicit finalization of the stream it could be much cleaner. This is not something we need to tackler here, just bringing up the observation.

Second, if you open a `stream` and then instantiate a `HTTPResponse` object thru the traditional means (`json`, `text`, etc), there is no exception and no logs. This is a regression and we should bring it back. Which leads to the main point I think we need to discuss.

Third, we need to think thru the main breaking change here re: exception handling. Mainly I am talking about this pattern:

```python
class CustomException(SanicException):
    message = "Hello"
    quiet = True
    status = 418


@app.exception(CustomException)
async def custom(*_):
    print("Something bad happened")


@app.get("/")
    response = await request.respond()
    await response.send("...")
    raise CustomException
```

In this scenario, we no longer execute the exception handler. While this is correct--and I think (as I discussed previously) the proper way to deal with this is to use the signal instead--it nonetheless is a breaking change to exception handling in two ways: the handler is never executed, and the exception is always logged ignoring `quiet`.

We are reporting now that `[ERROR] The error response will not be sent to the client for the following exception:"Hello". A previous response has at least partially been sent.`. This makes sense, and so I am less concerned about ignoring the `quiet` by logging out the exception. We are explicit here. So that seems fine.

However, there is an inconsistency. With no error, we allow the route handler to carry on just fine even though we have another HTTPResponse (see test 6). We do not log the issue or raise an exception :-1: (my second point above), and the fact that there is a second HTTPResponse does not impact the ability for the route handler to continue to send body, output logs, etc. I feel like the above situation should operate the same way as test 6.

In both of those cases (raised exception as seen above and Test 6), we should log out the existence of an ignored response object. That means that I think we _should_ execute the exception handler and just do nothing with that response object the same as in Tests 4-6. We should then also throw a `DeprecationWarning` and say something to the effect that we will no longer execute that exception handler beginning in v22.6.

In this way we do not breaking anyone's existing exception handlers, and we also are treating all HTTPResponses consistently.

## Results

### Test 1 - Matching responses, send() between respond()
|        | Error logged       | Status Code   |
|--------|--------------------|---------------|
| PR     | :heavy_check_mark: | 200 :ok_hand: |
| `main` | :heavy_check_mark: | 200 :ok_hand: | Invalid HTTP, missing final `\r\n` |

<details>
<summary>
Code
</summary>

```python
@app.get("/")
    response = await request.respond()
    await response.send("...")
    response = await request.respond()
```
</details>

<details>
<summary>
Raw responses
</summary>

#### PR response
```
HTTP/1.1 200 OK
transfer-encoding: chunked
connection: keep-alive
content-type: None

3
...
0



```

#### Main response
```
HTTP/1.1 200 OK
transfer-encoding: chunked
connection: keep-alive
content-type: None

3
...


```
</details>

### Test 2 - Matching responses, send() after second respond()
|        | Error logged       | Status Code   |
|--------|--------------------|---------------|
| PR     | :heavy_check_mark: | 500 :x:       |
| `main` | :x:                | 200 :ok_hand: |

<details>
<summary>
Code
</summary>

```python
@app.get("/")
    response = await request.respond()
    response = await request.respond()
    await response.send("...")
```
</details>

<details>
<summary>
Raw responses
</summary>

#### PR response
```
HTTP/1.1 500 Internal Server Error
content-length: 586
connection: keep-alive
content-type: text/plain; charset=utf-8

⚠️ 500 — Internal Server Error
==============================
Second respond call is not allowed.

ServerError: Second respond call is not allowed. while handling path /
Traceback of __main__ (most recent call last):

  ServerError: Second respond call is not allowed.
    File handle_request, line 83, in handle_request

    File /tmp/p.py, line 16, in handler
    response = await request.respond()

    File /home/adam/Projects/Sanic/sanic/sanic/request.py, line 194, in respond
    raise ServerError("Second respond call is not allowed.")

```

#### Main response
```
HTTP/1.1 200 OK
transfer-encoding: chunked
connection: keep-alive
content-type: None

3
...
0



```
</details>

### Test 3 - Matching responses, send() after each respond()
|        | Error logged       | Status Code   | Notes                                                                                |
|--------|--------------------|---------------|--------------------------------------------------------------------------------------|
| PR     | :heavy_check_mark: | 200 :ok_hand: | Only the send of the first respond makes it thru                                     |
| `main` | :heavy_check_mark: | 200 :ok_hand: | Only the send of the first respond makes it thru. Invalid HTTP, missing final `\r\n` |

<details>
<summary>
Code
</summary>

```python
@app.get("/")
    response = await request.respond()
    await response.send("abc")
    response = await request.respond()
    await response.send("123")
```
</details>

<details>
<summary>
Raw responses
</summary>

#### PR response
```
HTTP/1.1 200 OK
transfer-encoding: chunked
connection: keep-alive
content-type: None

3
abc
0



```

#### Main response
```
HTTP/1.1 200 OK
transfer-encoding: chunked
connection: keep-alive
content-type: None

3
abc


```
</details>

### Test 4 - Returned response, no send()
|        | Error logged       | Status Code   | Notes                                                              |
|--------|--------------------|---------------|--------------------------------------------------------------------|
| PR     | :x:                | 200 :ok_hand: | The returned response object seems to have been ignored completely |
| `main` | :x:                | 200 :ok_hand: | The `respond()` seems to have been ignored completely              |

<details>
<summary>
Code
</summary>

```python
@app.get("/")
    response = await request.respond(headers={"one": "one"})
    return json({"foo": "bar"}, headers={"two": "two"})
```
</details>

<details>
<summary>
Raw responses
</summary>

#### PR response
```
HTTP/1.1 200 OK
one: one
content-length: 0
connection: keep-alive
content-type: None



```

#### Main response
```
HTTP/1.1 200 OK
two: two
content-length: 13
connection: keep-alive
content-type: application/json

{"foo":"bar"}

```
</details>

### Test 5 - Returned response, send() after respond()
|        | Error logged       | Status Code   | Notes                                                                                                  |
|--------|--------------------|---------------|--------------------------------------------------------------------------------------------------------|
| PR     | :x:                | 200 :ok_hand: | The returned response object seems to have been ignored completely                                     |
| `main` | :heavy_check_mark: | 200 :ok_hand: | The returned response object seems to have been ignored completely. Invalid HTTP, missing final `\r\n` |

<details>
<summary>
Code
</summary>

```python
@app.get("/")
    response = await request.respond(headers={"one": "one"})
    await response.send("...")
    return json({"foo": "bar"}, headers={"two": "two"})
```
</details>

<details>
<summary>
Raw responses
</summary>

#### PR response
```
HTTP/1.1 200 OK
one: one
transfer-encoding: chunked
connection: keep-alive
content-type: None

3
...
0



```

#### Main response
```

```
</details>

### Test 6 - Returned response, send() after respond() and after json() response created
|        | Error logged       | Status Code   | Notes                                                                                                  |
|--------|--------------------|---------------|--------------------------------------------------------------------------------------------------------|
| PR     | :x:                | 200 :ok_hand: | The returned response object seems to have been ignored completely                                     |
| `main` | :heavy_check_mark: | 200 :ok_hand: | The returned response object seems to have been ignored completely. Invalid HTTP, missing final `\r\n` |

<details>
<summary>
Code
</summary>

```python
@app.get("/")
    response = await request.respond(headers={"one": "one"})
    await response.send("...")
    json_response = json({"foo": "bar"}, headers={"two": "two"})
    await response.send("...")
    return json_response
```
</details>

<details>
<summary>
Raw responses
</summary>

#### PR response
```
HTTP/1.1 200 OK
one: one
transfer-encoding: chunked
connection: keep-alive
content-type: None

3
...
3
...


```
</details>

tests/test_middleware.py
```diff
@@ -297,3 +297,27 @@ async def handler(request):
 
     _, response = app.test_client.get("/")
     assert response.json["foo"] == "bar"
+
+
+def test_middleware_return_response(app):
```
ahopkins, line 302: We need to add some more tests here to cover some of the new edge cases we are solving with this PR.

ChihweiLHBird:
Error and warning logs mentioned above were added, will write test cases for all these stuffs later.

ChihweiLHBird:
Hey @Tronic @ahopkins, tests were added, and one change to `app.py` between line 917-928 was added after @Tronic's review. It would need a second look. Please let me know if there is any other issue. Thank you all!

sanic/app.py
```diff
@@ -733,6 +734,47 @@ async def handle_exception(
...
+                    "by the exception handler won't be sent to the client "
+                    "because another response was at least partially sent "
+                    "The exception handler should only be used to generate "
+                    "the error page, and if you would like to perform any "
+                    "other action for an exception raised, please consider "
+                    "using a signal handler, like "
+                    '`@app.signal("http.lifecycle.exception")`\n'
+                    "Sanic signals detailed docs: "
+                    "https://sanicframework.org/en/guide/advanced/"
+                    "signals.html"
```
ahopkins, line 765:
```suggestion
            handler = self.error_handler._lookup(
                exception, request.name if request else None
            )
            if handler:
                error_logger.warning(
                    "An error occured while handling the request after at ",
                    "least some part of the response was sent to the client. ",
                    "Therefore, the response from your custom exception ",
                    f"handler {handler.__name__} will not be sent to the ",
                    "client. Beginning in v22.6, Sanic will stop executing ",
                    "custom exception handlers in this scenario. Exception ",
                    "handlers should only be used to generate the exception ",
                    "responses. If you would like to perform any other action ",
                    "on a raised exception, please consider using a signal ",
                    'handler like `@app.signal("http.lifecycle.exception")`\n',
                    "For further information, please see the docs: ",
                    "https://sanicframework.org/en/guide/advanced/",
                    "signals.html",
```

sanic/app.py
```diff
@@ -871,10 +914,18 @@ async def handle_request(self, request: Request):  # no cov
                 if isawaitable(response):
                     response = await response
 
-            if response is not None:
+            if response is not None and not request.responded:
                 response = await request.respond(response)
-            elif not hasattr(handler, "is_websocket"):
-                response = request.stream.response  # type: ignore
+            elif not hasattr(handler, "is_websocket") or request.responded:
```
ahopkins, line 919: Why in the same conditional?

sanic/response.py
```diff
@@ -112,8 +117,17 @@ async def send(
         """
         if data is None and end_stream is None:
             end_stream = True
-        if end_stream and not data and self.stream.send is None:
-            return
+        if self.stream is None:
+            raise SanicException(
+                "Stream was disconnected with this response instance."
+            )
```
ahopkins, line 123: Let's make this also `ServerError`. Also, I am not sure that the message is helpful. How is the user supposed to know which response is "this"? Can we just say "Stream is disconnected"?

ahopkins (changes requested):
Looks good. Just a couple comments about cleaning up the error messaging, then I think we are ready to merge.

sanic/app.py
```diff
@@ -733,6 +734,47 @@ async def handle_exception(
...
+                    "Sanic signals detailed docs: "
+                    "https://sanicframework.org/en/guide/advanced/"
+                    "signals.html"
+                )
+                try:
+                    response = self.error_handler.response(request, exception)
+                    if isawaitable(response):
+                        response = await response
+                except BaseException as e:
+                    logger.error("An error occurs in the exception handler.")
```
ahopkins, line 772: `occured`

sanic/app.py
```diff
@@ -871,10 +914,18 @@ async def handle_request(self, request: Request):  # no cov
...
-                response = request.stream.response  # type: ignore
+            elif not hasattr(handler, "is_websocket") or request.responded:
+                if request.stream is not None:
+                    response = request.stream.response
+                if request.responded:
+                    error_logger.error(
+                        "The response object returned by the route handler "
+                        "won't be sent to client because a previous response "
+                        "or itself was created and may have been sent the "
+                        "client. "
```
ahopkins, line 927:
```suggestion
                    error_logger.error(
                        "The response object returned by the route handler "
                        "will not be sent to client. The request has already "
                        "been responded to."
```

sanic/app.py
ChihweiLHBird: Fixed, added `occurred` as the past tense of occurs

ChihweiLHBird: https://github.com/sanic-org/sanic/pull/2327/commits/a35054715f4a95678d3f73850b623dcb2ed9006c

sanic/response.py
ChihweiLHBird: How about "No stream is connected to the response object"? This error message is raised in `Response` class, so it is related to the response, right?

sanic/app.py
ChihweiLHBird: Because I thought connect `request.stream.response` to `response` is also a wanted behavior here when the request was responded (by the response object `request.stream.response`). Would you agree that? Maybe the if else blocks can be better organized?

ChihweiLHBird: If not connect, test case 4 above may lead to an exception in one of my past experiments. I can reproduce and show that tomorrow.

sanic/response.py
ahopkins: That works.

sanic/app.py
ahopkins: It just seems that you have one conditional and then separate inside. Why not two `elif` one for `is_websocket` check, and the other for the `responded` check?

sanic/response.py
ChihweiLHBird: Thanks, fixed.

sanic/app.py
ChihweiLHBird: Is this looks better?

```python
            if response is not None:
                if not request.responded:
                    if request.stream is not None:
                        response = request.stream.response
                    error_logger.error(
                        "The response object returned by the route handler "
                        "will not be sent to client. The request has already "
                        "been responded to."
                    )
                else:
                    response = await request.respond(response)
            elif not hasattr(handler, "is_websocket"):
                response = request.stream.response  # type: ignore
```

I wanted to avoid 3 layers if statements so putted the tricky or for set `response` to `request.stream.response`, which should be done in these one of the two conditions.

tests/test_middleware.py
ChihweiLHBird: Test cases added, considered as resolved. Some new test cases failed, will try to fix.

sanic/app.py
ahopkins: Seems like you are overcomplicating it.

```python
if request.responded:
    error_logger.error(
        "The response object returned by the route handler "
        "will not be sent to client. The request has already "
        "been responded to."
    )
elif response is not None:
    response = await request.respond(response)
elif not hasattr(handler, "is_websocket"):
    response = request.stream.response  # type: ignore
```

ChihweiLHBird: That's not going to work... For example, error should not be logged if `response is None`, `response = request.stream.response` should also be done (IMO) if `request.responded`. 😂

ChihweiLHBird:
```python
if response is not None and not request.responded:
    response = await request.respond(response)
elif not hasattr(handler, "is_websocket"):
    response = request.stream.response  # type: ignore

if request.responded:
    if request.stream is not None:
        response = request.stream.response if request.stream is not None else None
    error_logger.error(
        "The response object returned by the route handler "
        "will not be sent to client. The request has already "
        "been responded to."
    )
```

This might be better?

ChihweiLHBird: No, above doesn't work, we have to check `request.responded` before the `respond` call.......

ChihweiLHBird: Please ignore my messed messages above, sorry about that. The latest push contains modified version of your code. @ahopkins 

Can you do a re-review? Thanks!

ahopkins (approved):

PR #2328, opened 2021-12-02
Add runtime checking to create_server to verify that startup has been run

ahopkins:
When using `create_server`, you need to manually call some functions. One of them being `app.startup`. If you do not, then the router is never finalized and you get a cryptic message, but this only happens on the first request. That means the server seemingly starts fine even though it is in a bad state.

```
  File "/home/user/project/.venv/lib/python3.9/site-packages/sanic/signals.py", line 73, in get
    group, param_basket = self.find_route(
TypeError: 'NoneType' object is not callable
future: <Task finished name='Task-3' coro=<HttpProtocol.connection_task() done, defined at /home/user/project/.venv/lib/python3.9/site-packages/sanic/server/protocols/http_protocol.py:92> exception=TypeError("'NoneType' object is not callable")>
```

This PR simple adds a new flag to the application state to indicate whether this function has been run.


Tronic (dismissed):

rmorshea:
@Tronic It looks like [this example](https://github.com/sanic-org/sanic/blob/f28286536211627797dc1e7ccbc09e130f4d9188/examples/run_async_advanced.py#L40-L42) needs to be updated too.

PR #2333, opened 2021-12-07
Fix type doc

ChihweiLHBird:


sanic/app.py
```diff
@@ -724,7 +724,7 @@ async def handle_exception(
         A handler that catches specific exceptions and outputs a response.
 
         :param request: The current request object
-        :type request: :class:`SanicASGITestClient`
+        :type request: Request
```
ahopkins, line 727: Can you keep the class reference so that Sphinx will create a link?

https://stackoverflow.com/a/21297186/1036843

ahopkins (changes requested):

sanic/app.py
ahopkins: You can run the following to standup the API docs locally:

```
make docs-serve
```

ChihweiLHBird: As we discussed, sphinx can auto detect the type-hints and link the doc to the class reference. Can you try it on your end and test this feature?

ahopkins (approved):

PR #2344, opened 2021-12-24
deprecation warning for not catch lowercase env var

ChihweiLHBird:
Related issue: https://github.com/sanic-org/sanic/issues/2120


sanic/config.py
```diff
@@ -222,6 +222,12 @@ def __init__(self, name) -> None:
         for key, value in environ.items():
             if not key.startswith(prefix):
                 continue
+            if not key.isupper():
+                deprecation(
+                    "Lowercase environment variables won't be "
+                    "catch for sanic config in the future version, 22.9",
```
prryplatypus, line 228: Might want to use the word `loaded into` instead of `catch for` to stay consistent with the documentation and avoid misunderstandings, as `catch` is used for exception handling in other languages.

And a small nitpick: uppercase in “Sanic”, please 😅

ChihweiLHBird: Thanks! Fixed.

prryplatypus (dismissed):

sanic/config.py
prryplatypus: Thanks! :)

sanic/config.py
```diff
@@ -222,6 +222,12 @@ def __init__(self, name) -> None:
         for key, value in environ.items():
             if not key.startswith(prefix):
                 continue
+            if not key.isupper():
+                deprecation(
+                    "Lowercase environment variables won't be "
+                    "loaded into Sanic config in the future version, 22.9",
```
ahopkins, line 228:
```suggestion
                    "Lowercase environment variables will not be "
                    "loaded into Sanic config beginning in v22.9.",
```

ahopkins (changes requested):
We should probably only raise this once per load, not for each variable. Also, can you add a unit test for coverage?

ChihweiLHBird:
@ahopkins Do you think the code climate Cognitive Complexity limit can be increased? I have no idea about how to make it below 10...

ChihweiLHBird:
Maybe we can wrap this into a function?

```python
for converter in reversed(self._converters):
    try:
        self[config_key] = converter(value)
        break
    except ValueError:
        pass
```

tests/test_config.py
```diff
@@ -120,6 +122,19 @@ def test_env_w_custom_converter():
     del environ["SANIC_TEST_ANSWER"]
 
 
+def test_env_lowercase():
+    with pytest.warns(None) as record:
+        environ["SANIC_test_answer"] = "42"
+        app = Sanic(name=__name__)
+        assert app.config.test_answer == 42
+    print(record[0].message)
```
ahopkins, line 130: You canremove this

ahopkins:
> @ahopkins Do you think the code climate Cognitive Complexity limit can be increased? I have no idea about how to make it below 10...

Don't worry about it. We are going to give up on code climate soon and try to go back to codecov.

tests/test_config.py
ChihweiLHBird: Fixed

ahopkins (approved):

PR #2346, opened 2021-12-30
credentials property added

SerGeRybakov:
Added a credentials property returning Credentials object either for NoAuth, or BasicAuth, Bearer and API tokens. 
If authorization header is empty all Credentials field are None.
Tests added

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import dataclasses
+
 from base64 import b64decode
 from typing import (
     TYPE_CHECKING,
@@ -14,6 +16,7 @@ from typing import (
 )
 
 from sanic_routing.route import Route  # type: ignore
+
 
 if TYPE_CHECKING:  # no cov
     from sanic.server import ConnInfo
@@ -52,10 +55,38 @@ try:
 except ImportError:
     from json import loads as json_loads  # type: ignore
 
-class Credentials(NamedTuple):
+
+@dataclasses.dataclass()
+class Credentials:
+    _auth_type: Optional[str]
     token: Optional[str]
-    username: Optional[str]
-    password: Optional[str]
+    _username: Optional[str] = dataclasses.field(default=None)
+    _password: Optional[str] = dataclasses.field(default=None)
+
+    def __post_init__(self):
+        if self._auth_type == "Basic":
+            self._username, self._password = (
+                b64decode(self.token.encode("utf-8")).decode().split(":")
+            )
+
+    @property
+    def username(self):
+        if not self._check_auth_type:
+            raise AttributeError("Username is available for Basic Auth only")
+        return self._username
+
+    @property
+    def password(self):
+        if not self._check_auth_type:
+            raise AttributeError("Password is available for Basic Auth only")
+        return self._password
+
+    @property
+    def _check_auth_type(self) -> bool:
+        if self._auth_type == "Basic":
+            return True
+        return False
+
 
 class RequestParameters(dict):
     """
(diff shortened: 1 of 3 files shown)


Tronic:
Cool. Although there could be some dispute whether this is better handled as an extension rather than Sanic core, I believe that such auth functionality is standardised enough and useful widely enough to warrant for inclusion in the core.

Notably this doesn't include JWT Bearer parsing or such more advanced constructs that indeed might be best left for plugins doing login and A&A in a more general sense.

Tronic:
Not sure whether HTTP Digest auth is in any use nowadays but if you handle Basic, it might be better to handle that as well.

prryplatypus:
As far as I'm aware, this PR stems after a discussion in the Discord server, where someone mentioned that Sanic already provides [a `request.token` attribute](https://sanicframework.org/en/guide/basics/headers.html#request) to parse authorization tokens in the form `Token <token>` or `Bearer <token>` (which I guess can be seen as JWT Bearer parsing, as you mention), but doesn't provide one for basic credential authentication.

prryplatypus:
Although I wonder if it'd be better to just add a method like `extract_credentials(allowed_prefixes: List[str], decode_method: Callable[..., T] = lambda x: x) -> Optional[T]:` at this point

Tronic:
> Although I wonder if it'd be better to just add a method like `extract_credentials(allowed_prefixes: List[str], decode_method: Callable[..., T] = lambda x: x) -> Optional[T]:` at this point

The few prefixes in use are well defined. What you do with the string that *follows* `Bearer` is another thing, although JWT is a pretty common format for applications to use.

ahopkins:
> Although there could be some dispute whether this is better handled as an extension rather than Sanic core, I believe that such auth functionality is standardised enough and useful widely enough to warrant for inclusion in the core.

As long as it is just parsing, and not happening until asked for (as opposed to on every request), this seems relevant enough to include here. We do similar header parsing elsewhere.

> Notably this doesn't include JWT Bearer parsing or such more advanced constructs that indeed might be best left for plugins doing login and A&A in a more general sense.

We already do this in `request.token` if by "parsing" you mean extracting. That IMO is the extent to which it is relevant as a core feature.

> Not sure whether HTTP Digest auth is in any use nowadays but if you handle Basic, it might be better to handle that as well.

That is a bit more involved. It certainly is in use as I have encountered it out in the wild. Again, at most I think we should be providing tools to easily get to the data, not necessarily handling it.

> Although I wonder if it'd be better to just add a method

> The few prefixes in use are well defined.

As I mentioned, we should probably be limiting ourselves to extraction rather than anything else. Here are the acceptable values:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes

I really think it is overkill to try and support them all. Currently, by far the usage of `Bearer` scheme is the most popular in modern authentication schemes. There also is a lot of usage of `Token`, which although not standard is widely used. As for `Basic`, it is in my experience still far less utilized on the Internet these days, but does still have use. This PR adds support in a fairly minimally intrusive way so I do not see it as a problem for inclusion.

If we start getting requests for other formats (including `Digest`), I am up for looking into handling them. But I do not see it as a worthwhile endeavor at this time.

sanic/request.py
```diff
@@ -347,6 +351,27 @@ def token(self):
...
+
+        if auth_header is not None:
+            for prefix in prefixes:
+                if prefix in auth_header:
+                    header_value = auth_header.partition(prefix)[-1].strip()
+                    if prefix == "Basic":
+                        username, password = b64decode(header_value.encode("utf-8")).decode().split(":")
+                        return Credentials(username=username, password=password, token=None)
+                    return Credentials(token=header_value, username=None, password=None)
+        return Credentials(token=auth_header, username=None, password=None)
```
ahopkins, line 373: We should follow a pattern used elsewhere where we cache the value on a `parsed_*` attribute.

ahopkins, line 373: I wonder if it makes sense to take this parsing out of the `Request` and into the model itself:

```python
credentials = Credentials.parse(auth_header)
```

This would allow it to be much more easily extended by application developers or users as discussed.

ahopkins (changes requested):

sanic/request.py
Tronic: Btw, could we switch to Python's `@cached_property` for that? New in Python 3.8... https://docs.python.org/3/library/functools.html#functools.cached_property

ahopkins: Absolutely. As soon as we are ready to drop 3.7. Regular support has expired so, we should probably bring this conversation up soon.

PR #2347, opened 2022-01-02
Sanic multi-application server

ahopkins:
_Implementation of #2348_

```python
app.prepare(port=9999, debug=True)
app.prepare(port=9998)
app.prepare(port=9997)
app2.prepare(port=8888)
app2.prepare(port=8887)

Sanic.serve()
```

---

Requires #2362

### BREAKING CHANGES :rotating_light: 

- Removes `GunicornWorker` completely without a deprecation period

### SIDE EFFECTS :cupcake: 

- Allows [named tasks](https://sanic.dev/en/guide/basics/tasks.html#named-tasks) even in Python 3.7

### NOTES :memo: 

- `auto_reload` will be enabled if `ANY` of the applications are prepared with it on
- `workers` are bound by the primary application (the first one prepared)


vltr (approved):
I think this will be awesome, a huge step forward the right direction! :sunglasses:

PR #2354, opened 2022-01-04
Add config.update support for setters

ahopkins:
This is specifically to target this style of setting when there is a descriptor pattern with setters.

```python
app.config.update({'FALLBACK_ERROR_FORMAT': 'text'})
```

### DIFF
--- a/sanic/config.py
+++ b/sanic/config.py
@@ -137,9 +137,21 @@ class Config(dict, metaclass=DescriptorMeta):
         self.update({attr: value})
 
     def update(self, *other, **kwargs) -> None:
-        other_mapping = {k: v for item in other for k, v in dict(item).items()}
-        super().update(*other, **kwargs)
-        for attr, value in {**other_mapping, **kwargs}.items():
+        kwargs.update({k: v for item in other for k, v in dict(item).items()})
+        setters: Dict[str, Any] = {
+            k: kwargs.pop(k)
+            for k in {**kwargs}.keys()
+            if k in self.__class__.__setters__
+        }
+
+        for key, value in setters.items():
+            try:
+                super().__setattr__(key, value)
+            except AttributeError:
+                ...
+
+        super().update(**kwargs)
+        for attr, value in {**setters, **kwargs}.items():
             self._post_set(attr, value)
 
     def _post_set(self, attr, value) -> None:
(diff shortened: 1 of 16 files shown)


sanic/config.py
```diff
@@ -137,9 +137,21 @@ def __setitem__(self, attr, value) -> None:
         self.update({attr: value})
 
     def update(self, *other, **kwargs) -> None:
-        other_mapping = {k: v for item in other for k, v in dict(item).items()}
-        super().update(*other, **kwargs)
-        for attr, value in {**other_mapping, **kwargs}.items():
+        kwargs.update({k: v for item in other for k, v in dict(item).items()})
+        setters: Dict[str, Any] = {
+            k: kwargs.pop(k)
+            for k in {**kwargs}.keys()
```
prryplatypus, line 143: Should work either way, I believe, but just to simplify a bit.
```suggestion
            for k in kwargs
```

tests/test_config.py
```diff
@@ -385,5 +385,14 @@ def test_config_set_methods(app, monkeypatch):
...
+    post_set.assert_has_calls(
+        calls=[
+            call("FOO", 6),
+            call("BAR", 7),
+        ]
+    )
+    post_set.reset_mock()
+
+    app.config.update_config({"FOO": 8})
+    post_set.assert_called_once_with("FOO", 8)
```
prryplatypus, line 398: Missing a test case with kwargs

prryplatypus (commented):
Do we want to clean up the `__setattr__` method, since all the logic it has is now also performed in `update`, which it calls after running the checks itself?

sanic/config.py
ahopkins: We need a copy because we are changing the values of the dict in the iteration. I believe we would get an exception otherwise.

tests/test_config.py
ahopkins: Which one?

ahopkins: You mean this?

```python
    app.config.update(FOO=5)
    post_set.assert_called_once_with("FOO", 5)
    post_set.reset_mock()
```

prryplatypus: One mixing both at once

```py
app.config.update({"FOO": 5}, BAR=6)
```

sanic/config.py
prryplatypus: It does indeed!

prryplatypus (dismissed):

PR #2355, opened 2022-01-04
Ignore name argument on Python 3.7

ahopkins:
We need to ignore the `name` argument on Python 3.7


prryplatypus (dismissed):
Do we want to add a test for this?

ashleysommer (dismissed):

ahopkins:
> Do we want to add a test for this?

Sure. If nothing else it will boost our cov %

Tronic:
I tend to bring this up every now and then, but this might also be a time to consider dropping Python 3.7 support. Even Debian is at 3.9 and other distros have already moved to 3.10. MacOS ships with 3.8 and Microsoft Store has 3.10.

ahopkins:
> I tend to bring this up every now and then, but this might also be a time to consider dropping Python 3.7 support. Even Debian is at 3.9 and other distros have already moved to 3.10. MacOS ships with 3.8 and Microsoft Store has 3.10.

We probably will by the end of the year. I see no compelling reason to do it just yet. The place where I see the biggest pain-point is with Sanic Extensions, which is already 3.8+. We do still see about 10-15% of our downloads from 3.7.

sanic/app.py
```diff
@@ -1552,10 +1552,19 @@ def _loop_add_task(
         name: Optional[str] = None,
         register: bool = True,
     ) -> Task:
-        prepped = cls._prep_task(task, app, loop)
-        task = loop.create_task(prepped, name=name)
+        if not isinstance(task, Future):
+            prepped = cls._prep_task(task, app, loop)
+            if sys.version_info == (3, 7):
```
prryplatypus, line 1557: To keep the check consistent with `get_task`:
```suggestion
            if sys.version_info < (3, 8):
```

sanic/app.py
```diff
@@ -1552,10 +1552,19 @@ def _loop_add_task(
...
+                    error_logger.warning(
+                        "Cannot set a name for a task when using Python 3.7. Your "
+                        "task will be created without a name."
+                    )
+                task = loop.create_task(prepped)
+            else:
+                task = loop.create_task(prepped, name=name)
 
-        if name and register:
+        if name and register and sys.version_info > (3, 7):
```
prryplatypus, line 1567: Again to keep the version checks consistent
```suggestion
        if name and register and sys.version_info >= (3, 8):
```

prryplatypus (dismissed):
Do we want to add a test to prevent #2356 from happening again in the future?

sanic/app.py
prryplatypus: Not sure if this check will work because of what we've just spoken on Discord (tuple len is 3)

PR #2356, opened 2022-01-05
Fixed task is created again when adding the task

SaidBySolo:
Before this PR, even if a task or a future was received, the task was created and registered once again.
When a user gives a task or future, only the task is added without creating a task or future again.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1552,16 +1552,17 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         name: Optional[str] = None,
         register: bool = True,
     ) -> Task:
-        prepped = cls._prep_task(task, app, loop)
-        if sys.version_info == (3, 7):
-            if name:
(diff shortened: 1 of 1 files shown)


ahopkins:
@SaidBySolo I changed the base of this to target another PR that is going into the LTS.

SaidBySolo:
> @SaidBySolo I changed the base of this to target another PR that is going into the LTS. Can you resolve that conflict?

sure :D

SaidBySolo:
@ahopkins Done!

ahopkins:
Thank you! :confetti_ball:

ahopkins:
Okay... so I think we have some commits in here that we do _NOT_ want to include. This is probably because you (reasonably) branched off `main`. Can we pull those out?

ahopkins:
Awesome! Thanks

ahopkins (approved):

PR #2357, opened 2022-01-05
Add credentials property to Request objects

SerGeRybakov:
`Credentials` logics refactored. 

Anytime there shall be `_auth_type` and `token` fields. Just in case of `_auth_type == 'Basic'` Credentials shall autocreate decoded (parsed) `_username` and `_password` fields, which can be reached by properties `username` and `password`. 
If `_auth_type != 'Basic'` properties `username` and `password` raise `AttributeError` with correspondent message.


ahopkins:
#2346

sanic/request.py
```diff
@@ -53,6 +56,38 @@
     from json import loads as json_loads  # type: ignore
 
 
+@dataclasses.dataclass()
```
ahopkins, line 59: In other locations in the project we use just `@dataclass`. Let's keep it consistent.

sanic/request.py
```diff
@@ -53,6 +56,38 @@
...
+class Credentials:
+    _auth_type: Optional[str]
```
ahopkins, line 61: Why is `_auth_type` private? Why `Optional`?

sanic/request.py
```diff
@@ -53,6 +56,38 @@
...
+    token: Optional[str]
```
ahopkins, line 62: Why `Optional`?

sanic/request.py
```diff
@@ -53,6 +56,38 @@
...
+        return self._username
+
+    @property
+    def password(self):
+        if not self._check_auth_type:
+            raise AttributeError("Password is available for Basic Auth only")
+        return self._password
+
+    @property
+    def _check_auth_type(self) -> bool:
```
ahopkins, line 85: A `check` attribute sounds more like a function than a property`. Also, the name does not really signify what it is actually doing.

sanic/request.py
```diff
@@ -347,6 +382,24 @@ def token(self):
...
+        """
+        prefixes = ("Basic", "Bearer", "Token")
+        auth_header = self.headers.getone("authorization", None)
+
+        if auth_header is not None:
+            for prefix in prefixes:
+                if prefix in auth_header:
+                    header_value = auth_header.partition(prefix)[-1].strip()
+                    return Credentials(_auth_type=prefix, token=header_value)
+        return Credentials(_auth_type=None, token=auth_header)
```
ahopkins, line 401: Please put this in `parse_credentials` function and follow the pattern used in other properties. Look at `forwarded` as an example.

ahopkins (changes requested):

sanic/request.py
SerGeRybakov: Optional because there can be a None. Look at the [test-cases](https://github.com/SerGeRybakov/sanic/blob/credentials/tests/test_requests.py#L455-L626) - they are all the same as for the [token property](https://github.com/SerGeRybakov/sanic/blob/credentials/tests/test_requests.py#L455-L626). I believe they must be the same for tokens as we make a combined property.

sanic/request.py
SerGeRybakov: Because I think no one is really interested in the type of auth. Everyone just need the result.

SerGeRybakov: But I can change it to public if you think it's right

ahopkins: I'd make it public. And required.

sanic/request.py
ahopkins: I am not sure I am seeing which use case you are talking about. If there is no `authorization` header, there is nothing to parse and we should not even have a `Credentials` object.

sanic/request.py
SerGeRybakov: I don't really get. It's a refactored `token` [property](https://github.com/SerGeRybakov/sanic/blob/credentials/sanic/request.py#L369-L383). What's wrong?

Or you want me to refactor token property also using DRY?

sanic/request.py
SerGeRybakov: right the first one and the last one

sanic/models/http_types.py
```diff
@@ -0,0 +1,37 @@
...
+    def password(self):
+        if not self._auth_is_basic:
+            raise AttributeError("Password is available for Basic Auth only")
+        return self._password
+
+    @property
+    def _auth_is_basic(self) -> bool:
+        if self.auth_type == "Basic":
+            return True
+        return False
```
ahopkins, line 37:
```suggestion
        return self.auth_type == "Basic"
```

sanic/models/http_types.py
```diff
@@ -0,0 +1,37 @@
...
+
+@dataclass()
+class Credentials:
+    auth_type: Optional[str]
+    token: Optional[str]
+    _username: Optional[str] = field(default=None)
+    _password: Optional[str] = field(default=None)
+
+    def __post_init__(self):
+        if self.auth_type == "Basic":
```
ahopkins, line 16:
```suggestion
        if self._auth_is_basic:
```

tests/test_requests.py
```diff
@@ -451,6 +452,180 @@ async def handler(request):
     assert request.token is None
 
 
+def test_credentials(app, capfd):
```
ahopkins, line 455: Can we break this up into multiple tests?

ahopkins (changes requested):
Looking good :muscle:

tests/test_requests.py
SerGeRybakov: done

sanic/models/http_types.py
SerGeRybakov: done

ahopkins (approved):

ahopkins (approved):

PR #2361, opened 2022-01-07
Add config option to skip Touchup step, for debugging purposes

ashleysommer:
This new option allows you to configure the App to skip the Touchup step at startup.
This is useful for debugging purposes, it means you can again set breakpoints into sanic methods like `handle_request` and have them work as expected at runtime.

This option also sets `fail_not_found` to False for built-in system signals, so the code can still dispatch signals and not get a NotFound exception.

Tests added for coverage.


sanic/app.py
```diff
@@ -443,6 +443,11 @@ def dispatch(
         inline: bool = False,
         reverse: bool = False,
     ) -> Coroutine[Any, Any, Awaitable[Any]]:
+        if self.config.get("SKIP_TOUCHUP", False):
```
ahopkins, line 446: If we are setting it as a proper config value, then we do not need to use `dict` syntax to get the value.

ahopkins, line 446: I'd rather take the value of this and cache it on the `SignalRouter` instance. Then inside the `_dispatch` method we can check for the value _inside_ the exception block. Therefore we do not need to check this value on **every** dispatch of every event.

```
# ./signals.py
async def _dispatch(...):
    try:
        ...
    except Notfound as e:
        if fail_not_found and self.allow_failures:
            raise e
        ...
```

We could set it is inside `Sanic.signalize`. The other option that probably makes more sense is to tie it into the `Config` object with `_post_set` so that we are always syncing the config value back to the `SignalRouter` since we ultimately want that to be the public facing API.

ashleysommer: Been doing some work around integrating this suggestion. An issue I'm running into: the config _post_set() method does not have access to the `app` instance, so it cannot affect the app's or get the SignalRouter instance. And the Config object may not even be attached to an app when the config property is set.

I will however look into integrating the property into the signal router

ashleysommer: Ah, I see another problem. The value of `app.config.SKIP_TOUCHUP` is likely not set at the point in time that the SignalRouter is created.

ashleysommer: We could make a setter on `app`:
```
@property
    def skip_touchup(self):
        return self.config.SKIP_TOUCHUP

    @skip_touchup.setter
    def skip_touchup(self, value: bool):
        self.signal_router.allow_fail_builtin = not value
        self.config.SKIP_TOUCHUP = value
```

ahopkins: I think we can also do it as a descriptor on the `Config` object. This is similar to what we did with `FALLBACK_ERROR_FORMAT` since we ultimately want the `Config` object to be "the source of truth".

ahopkins: > I will however look into integrating the property into the signal router. The tricky part is it should only skip throwing an exception on the built-in signals, not for any user-defined signals.

Agreed

ahopkins: Actually... as I think more and more about this, we only care about what the value is at startup time. So, it can be a regular config variable that we read inside `_startup` and pass to `signalize`. Any change to the config value later is irrelevant, and we just change the `signal_router` value at startup time.

ashleysommer:
I've implemented the changes as suggested above.
I also added a message to say that in v22.3 `--debug` switch will skip runtime optimizations, but when I merged master it pulled in the `remove-warnings` branch, which removed those v22.3 messages, so I removed my new messages too.

Still, I'd like to see `--debug` apply `app.config.SKIP_TOUCHUP=True` in Sanic v22.3

ahopkins:
> Still, I'd like to see `--debug` apply `app.config.SKIP_TOUCHUP=True` in Sanic v22.3

That seems reasonable. I would also apply that in `_startup`:

```python
if app.state.is_debug:
    app.config.SKIP_TOUCHUP = True
```

One more question though: do you think it would be better if the config were not in the negative?

```python
app.config.TOUCHUP = False
```

ahopkins:
@ashleysommer How is this looking? Can we get this merged soon?

ashleysommer:
Oh, sorry. I thought I'd finished with this. I'll check if I made those changes in my branch.

ashleysommer:
> One more question though: do you think it would be better if the config were not in the negative?
> `app.config.TOUCHUP = False`

Yeah, if we have `TOUCHUP` as a permanent configuration option that will _always_ be present in app.config, it can default to True, and I'm happy to swap the logic around.

ashleysommer:
Good to go now

ahopkins (approved):

ahopkins:
Nice :sunglasses: 

Looks like there was one place where the config swap was missed. Just pushed the fix for you.

ahopkins:
I have not tested this yet, but this might also solve some of the problems with test coverage :thinking:. Maybe this is something I will check next week after the release.

ashleysommer:
> Looks like there was one place where the config swap was missed. Just pushed the fix for you.

Doh! I missed the second test. Thanks.

ahopkins (approved):

PR #2366, opened 2022-01-11
fix: websocket dependency for websockets 9.1 security fix

sjsadowski:
Security fix in websockets dependency


ahopkins (commented):
Have you tested locally? Does it also work with 10? Maybe that should be the upper limit?

sjsadowski:
@ahopkins I wanted to minimize the change for the LTS, it probably could be 10 but I think the minimal change makes more sense in this particular scope. We do have it as 10 in the current LTS.

ahopkins:
Will get this released tonight.

ahopkins (dismissed):

ahopkins (approved):

PR #2373, opened 2022-01-14
remove error_logger on websockets

### LINKED ISSUE #2371: sanic.error - INFO - connection open
Hey there,

Why do we see:

    sanic.error - INFO - connection open

via stderr when getting new websocket connections. Shouldn't this go to stdout?

Also, is it possible to add "middleware" so we can properly log websocket connections and disconnects?  Is it possible to get a callback on websocket disconnects? 

Thanks!

jonra1993:
Fixes #2371

It also customized pyproject.toml  for easier setup with poetry. I think the email is not correct.


ahopkins (changes requested):
Looks good other than the poetry specific changes. Also, it's not helpful to have meta in two places.

ahopkins (approved):

Issue #2375, opened 2022-01-14
Wrong typing for app.listener decorator


aericson:
**Describe the bug**
Mypy accuses typing errors on the `@app.listener` decorator.

I'll try to put a PR later if I can manage to fix it 😄 

**Code snippet**
```python
from sanic import Sanic
from sanic.response import text
from asyncio import AbstractEventLoop

app = Sanic("MyHelloWorldApp")

@app.get("/")
async def hello_world(request):
    return text("Hello, world.")


@app.listener("before_server_start")
async def setup_db(app, loop: int):
    pass
```

```
mypy --pretty hello.py   
hello.py:12: error: Too few arguments
    @app.listener("before_server_start")
     ^
hello.py:12: error: Argument 1 has incompatible type "Callable[[Any, int], Coroutine[Any, Any, Any]]"; expected <nothing>
    @app.listener("before_server_start")
     ^
```

Both `app.before_server_start` decorator and `app. register_listener ` works fine.

**Expected behavior**
Should have the correct typing.

**Environment (please complete the following information):**
```
pip list
Package           Version
----------------- -------
aiofiles          0.8.0
httptools         0.3.0
multidict         5.2.0
mypy              0.931
mypy-extensions   0.4.3
pip               21.3.1
sanic             21.12.1
sanic-routing     0.7.2
setuptools        58.1.0
tomli             2.0.0
typing_extensions 4.0.1
ujson             5.1.0
uvloop            0.16.0
websockets        10.1
```
- Tested on python 3.9 and 3.10

**Additional context**

Tronic:
Partially related: the loop argument should be deprecated. I suppose the listener registration should look at how many arguments the callable expects so that listeners don't need to take the loop argument anymore but that the old ones, that do, keep working with a deprecation warning.

ahopkins:
@Tronic I agree :100: I made this exact change at one point. I cannot remember on what branch it was and it clearly was never merged. But yes, we should definitely do that.

ahopkins:
FWIW this feature was added in the last release.

sjsadowski:
Closing per @ahopkins comment

PR #2376, opened 2022-01-14
Fixing typing for ListenerMixin.listener

aericson:
### Description

This PR addresses #2375. The typing of that method is a bit complex because it has two different code paths which result in different typing of parameters and return.

I've added two overloads so that we get better type checking.

### Alternative solutions

I've also considered another way to go around this by separating `listener` into two methods which should reduce complexity but in this PR I only addressed the typing.

Here is how the alternative solution could look like:
https://gist.github.com/aericson/d36514eb45b9ec1f65b8ee51fdf5c08f


sanic/models/handler_types.py
```diff
@@ -1,11 +1,13 @@
...
 from typing import Any, Callable, Coroutine, Optional, TypeVar, Union
 
+import sanic
+
 from sanic.request import Request
 from sanic.response import BaseHTTPResponse, HTTPResponse
 
 
-Sanic = TypeVar("Sanic")
+Sanic = TypeVar("Sanic", bound="sanic.Sanic")
```
aericson, line 10: I'm not super sure about this change. Not sure we should do it, and maybe outside the scope of the PR.

ahopkins: This is okay

ahopkins:
> I've also considered another way to go around this by separating `listener` into two methods which should reduce complexity but in this PR I only addressed the typing.

Let's consider that in a separate PR.

sanic/mixins/listeners.py
```diff
@@ -59,6 +82,8 @@ def register_listener(listener, event):
             return listener
 
         if callable(listener_or_event):
+            if TYPE_CHECKING:
+                assert event_or_none is not None
```
ahopkins, line 86: Avoid `assert` in production code. There is another instance in the codebase where we have this, but we should not proliferate the issue. Use `raise` instead.

sanic/mixins/listeners.py
```diff
@@ -26,12 +26,33 @@ def __init__(self, *args, **kwargs) -> None:
     def _apply_listener(self, listener: FutureListener):
         raise NotImplementedError  # noqa
 
+    @overload
+    def listener(
+        self,
+        listener_or_event: ListenerType[Sanic],
+        event_or_none: str,
+        apply: bool = True,
```
ahopkins, line 34: Do not need a default here.

sanic/mixins/listeners.py
```diff
@@ -26,12 +26,33 @@ def __init__(self, *args, **kwargs) -> None:
...
+    ) -> ListenerType[Sanic]:
+        ...
+
+    @overload
+    def listener(
+        self,
+        listener_or_event: str,
+        event_or_none: None = None,
+        apply: bool = True,
```
ahopkins, line 43: Can remove these defaults

sanic/mixins/listeners.py
aericson: won't this make the typing invalid? I can't get the type checking to work as intended without the default. Removing them would make them required.

aericson: Oh, nice. Just found out one can use an ellipsis (`...`). Thanks for teaching me that 😄

ahopkins (approved):

ahopkins:
> ### Alternative solutions
> 
> I've also considered another way to go around this by separating `listener` into two methods which should reduce complexity but in this PR I only addressed the typing.

No rush on this, but do you think you can put this alternative version together as a PR for the June release?

aericson:
> > ### Alternative solutions
> > 
> > I've also considered another way to go around this by separating `listener` into two methods which should reduce complexity but in this PR I only addressed the typing.
> 
> No rush on this, but do you think you can put this alternative version together as a PR for the June release?

Yep, I can put some time into that. Are you thinking on us closing this PR and having a new one with that? Just wondering if I should do on top of this changes or from a clean branch.

PR #2381, opened 2022-01-17
fix(tasks): newly assigned ``None`` in registry

SaidBySolo:
## How to reproduce?
```python
from sanic import Sanic
from asyncio import sleep, create_task

app = Sanic(__name__)


async def task():
    while True:
        await sleep(10)


@app.before_server_start
async def dummy_start(app, _):
    app.add_task(create_task(task()), name="Dummy")


@app.before_server_stop
async def dummy_stop(app, _):
    # Raise RuntimeError: dictionary changed size during iteration
    app.purge_tasks()


app.run()
```
https://github.com/sanic-org/sanic/blob/b8d991420b698ca18d18c8893a219a0d0e9e55c9/sanic/app.py#L1273-L1274
If a name is not registered during create_task, the following name is obtained as ``Task-1``, so the error that occurred because it is newly assigned has been resolved.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1179,6 +1179,8 @@ class Sanic(BaseSanic, RunnerMixin, metaclass=TouchUpMeta):
                 task = loop.create_task(prepped, name=name)
 
         if name and register and sys.version_info > (3, 7):
+            if isinstance(task, Task):
+                task.set_name(name)
             app._task_registry[name] = task
 
         return task
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1179,8 +1179,6 @@ class Sanic(BaseSanic, RunnerMixin, metaclass=TouchUpMeta):
                 task = loop.create_task(prepped, name=name)
 
         if name and register and sys.version_info > (3, 7):
-            if isinstance(task, Task):
-                task.set_name(name)
             app._task_registry[name] = task
 
         return task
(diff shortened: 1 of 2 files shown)


sanic/app.py
```diff
@@ -1179,6 +1179,8 @@ def _loop_add_task(
                 task = loop.create_task(prepped, name=name)
 
         if name and register and sys.version_info > (3, 7):
+            if isinstance(task, Task):
+                task.set_name(name)
```
ahopkins, line 1183: Potentially duplicative. Perhaps we should do it in an `else` block on the first conditional? Or, set an internal flag? If someone passes a task with a name already set, we would not want to override it if their intent was to just use the name for internal Sanic ref.

SaidBySolo: When using else in the first block, it seems need to use a conditional statement that needs to check the version.
Or how about using a conditional statement in ``purge_tasks`` to first check if a name exists?

sjsadowski: @SaidBySolo Would you mind editing the title of the PR to reflect that you're setting the name of the task? either the semantic commit to fix(tasks) or in the rest of the descriptor. Your choice!

SaidBySolo: @ahopkins 
How about adding an attribute like this to the Task attribute?
```py
setattr(task, "__sanic_task_name__", name)
```

ahopkins: Why?

SaidBySolo: ``purge_task`` gets the registered task using task.get_name. 

If the task name and the name registered in sanic are different, a problem occurs.

Or there seems to be a workaround to not use ``get_name`` in purge_task.

ahopkins: I do not think we should be adding that property unless needed to avoid renaming the task. In which case we can look for it with the `get_name` as fallback. 

Iterating over the registry probably makes sense. In this case, we would not need that hack IIRC.

ahopkins:
Can you add a unit test?

SaidBySolo:
@ahopkins 

All tests passed!

I'm not sure if the test I wrote is correct

ahopkins (approved):

PR #2382, opened 2022-01-18
Downgrade warnings to backwater debug messages

ahopkins:
Most people do not care about these warnings. Let's only show them as debug messages with increased verbosity.


ashleysommer (approved):

Tronic:
Full diff coverage, perhaps?

ahopkins:
> Full diff coverage, perhaps?

Fine. :angry: I won't be so lazy...

PR #2383, opened 2022-01-18
fix: deprecation warning in ``asyncio.wait``

### LINKED ISSUE #2377: Update websockets implementation for 3.11
To make the websocket implementation 3.11 compat, we need to remove the deprecated usage of `asyncio.wait` here:  https://github.com/sanic-org/sanic/blob/4a416e177aa5037ba9436e53f531631707e87ea7/sanic/server/websockets/impl.py#L521

---

thank you both 

@ahopkins - We also get this warning in the error logs when using websockets:

    DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.  done, pending = await asyncio.wait(

Is this from the websockets package?

_Originally posted by @vgoklani in https://github.com/sanic-org/sanic/issues/2371#issuecomment-1013105803_

@vgoklani 

> Is this from the websockets package?

This is separate and should probably be opened as a new issue. It should be a simple fix:

```python
            self.recv_cancel = asyncio.Future()
            tasks = (
                self.recv_cancel,
                # NEXT LINE IS THE CHANGE NEEDED
                # TO EXPLICITLY CREATE THE TASK
                asyncio.create_task(self.assembler.get(timeout)),
            )
            done, pending = await asyncio.wait(
                tasks,
                return_when=asyncio.FIRST_COMPLETED,
            )
```

_Originally posted by @ahopkins in https://github.com/sanic-org/sanic/issues/2371#issuecomment-1013822496_

SaidBySolo:
close #2377

### DIFF
--- a/sanic/server/websockets/impl.py
+++ b/sanic/server/websockets/impl.py
@@ -520,7 +520,7 @@ class WebsocketImplProtocol:
             self.recv_cancel = asyncio.Future()
             tasks = (
                 self.recv_cancel,
-                asyncio.create_task(self.assembler.get(timeout)),
+                asyncio.ensure_future(self.assembler.get(timeout)),
             )
             done, pending = await asyncio.wait(
                 tasks,
@@ -576,7 +576,7 @@ class WebsocketImplProtocol:
             while True:
                 tasks = (
                     self.recv_cancel,
-                    asyncio.create_task(self.assembler.get(timeout=0)),
+                    asyncio.ensure_future(self.assembler.get(timeout=0)),
                 )
                 done, pending = await asyncio.wait(
                     tasks,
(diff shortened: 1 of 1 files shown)


sanic/server/websockets/impl.py
```diff
@@ -570,8 +574,12 @@ async def recv_burst(self, max_recv=256) -> Sequence[Data]:
             self.can_pause = False
             self.recv_cancel = asyncio.Future()
             while True:
+                tasks = (
```
SaidBySolo, line 577: I'm not sure if this should be inside a Loop statement

ahopkins (dismissed):

ahopkins (approved):

PR #2389, opened 2022-01-23
fix: replace distutils.strtobool() with locally implemented version

### LINKED ISSUE #2388: distutils.strtobool deprecation
**Describe the bug**
he distutils package is deprecated and slated for removal in Python 3.12. [PEP 632](https://www.python.org/dev/peps/pep-0632/)

**Code snippet**
[sanic/server/loop.py](https://github.com/sanic-org/sanic/blob/ac388d644b1e22156e228470fad8ea34932c080a/sanic/server/loop.py#L3)

**Expected behavior**
A lack of warnings about deprecation

 - OS: All
 - Version: All supported as of 2022/01/23 

**Additional context**
This is housekeeping to clean up a core python deprecation.

sjsadowski:
removes distutils.strtobool() dependency

closes #2388

### DIFF
--- a/sanic/server/loop.py
+++ b/sanic/server/loop.py
@@ -1,10 +1,23 @@
 import asyncio
 
-from distutils.util import strtobool
 from os import getenv
 
 from sanic.compat import OS_IS_WINDOWS
 from sanic.log import error_logger
+
+
+def strtobool(query: str) -> bool:
+    """
+    reimplement strtobool per PEP 632 and python 3.12 deprecation
+
+    True values are y, yes, t, true, on and 1; false values are n, no, f, false, off and 0. Raises ValueError if val is anything else.
+    """
+    if query.lower() in ["y", "yes", "t", "true", "on", "1"]:
+        return True
+    elif query.lower() in ["n", "no", "f", "false", "off", "0"]:
+        return False
+    else:
+        raise ValueError
 
 
 def try_use_uvloop() -> None:
--- a/sanic/server/loop.py
+++ b/sanic/server/loop.py
@@ -18,7 +18,7 @@ def strtobool(query: str) -> bool:
     elif query.lower() in ["n", "no", "f", "false", "off", "0"]:
         return False
     else:
-        raise ValueError
+        raise ValueError(f'String value {query} cannot be converted to bool')
 
 
 def try_use_uvloop() -> None:
--- a/sanic/server/loop.py
+++ b/sanic/server/loop.py
@@ -10,7 +10,8 @@ def strtobool(query: str) -> bool:
     """
     reimplement strtobool per PEP 632 and python 3.12 deprecation
 
-    True values are y, yes, t, true, on and 1; false values are n, no, f, false, off and 0. Raises ValueError if val is anything else.
+    True values are y, yes, t, true, on and 1; false values are n, no, f,
+    false, off and 0. Raises ValueError if val is anything else.
     """
     if query.lower() in ["y", "yes", "t", "true", "on", "1"]:
         return True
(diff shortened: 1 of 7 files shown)


ashleysommer:
Is this the only place in the code that distutils is used?

sjsadowski:
That was a good question. It's in [setup.py](https://github.com/sanic-org/sanic/blob/9d04f242983830b8f484c4c2aab7bbc2bcab2ce5/setup.py#L9) as well. I'll clean it out there too.

ashleysommer (dismissed):

sanic/server/loop.py
```diff
@@ -1,12 +1,26 @@
...
+
+    True values are y, yes, t, true, on and 1; false values are n, no, f,
+    false, off and 0. Raises ValueError if val is anything else.
+    """
+    if query.lower() in ["y", "yes", "t", "true", "on", "1"]:
+        return True
+    elif query.lower() in ["n", "no", "f", "false", "off", "0"]:
+        return False
+    else:
+        raise ValueError
```
ahopkins, line 21: Can we add a message here?

ahopkins (changes requested):
Can you add a test for catching regressions?

prryplatypus:
Seems like we already have an `str_to_bool` method here: https://github.com/sanic-org/sanic/blob/ac388d644b1e22156e228470fad8ea34932c080a/sanic/utils.py#L13-L42

Maybe we should unify all uses of `strtobool` to use this one instead?

sjsadowski:
My setup.py fu is not strong enough to figure out why tox was balking at importing str_to_bool from sanic.utils. Happy for suggestions to solve to fix the non-DRY problem of including it in setup.py itself. @sanic-org/core-devs

ahopkins:
I'm going to guess because the package has not yet been setup. non DRY may be the way to go.

sjsadowski:
@ahopkins in that case, this should be ready to go.

setup.py
```diff
@@ -61,7 +59,7 @@ def open_local(paths, mode="r", encoding="utf8"):
         "Build fast. Run fast."
     ),
     "long_description": long_description,
-    "packages": find_packages(),
+    "packages": find_packages(include=[]),
```
prryplatypus, line 62: What does this achieve vs not being there? 🤔

sanic/server/loop.py
```diff
@@ -1,12 +1,26 @@
 import asyncio
 
-from distutils.util import strtobool
 from os import getenv
 
 from sanic.compat import OS_IS_WINDOWS
 from sanic.log import error_logger
 
 
+def strtobool(query: str) -> bool:
```
ahopkins, line 9: Why do we need to do it twice inside the lib?

sanic/utils.py
```diff
@@ -11,35 +11,18 @@
...
-        "n", "no", "f", "false", "off", "disable", "disabled", "0"
-    ) returns False.
-    Else Raise ValueError."""
-
-    val = val.lower()
-    if val in {
-        "y",
-        "yes",
-        "yep",
-        "yup",
```
ahopkins, line 30: This is a breaking change. Why remove it?

sanic/server/loop.py
sjsadowski: we don't, I missed updating this to use sanic.utils

sanic/utils.py
sjsadowski: because it's not compliant with the functionality we're replacing, plus do people really use 'yep' and 'yup'? in every place we're actively using this it's only to check for 'no'

ahopkins: ... that we are using, yes. But it has been in the lib. There doesn't seem to be a compelling reason to remove this. And, if we did it would need to be done with deprecation. I'd say let's continue to support those because it already exists. The only people using "yup" would be someone that already expects it as a truthy value.

sjsadowski:
Gonna rethink the approach here. Will come back to this in a bit.

PR #2390, opened 2022-01-24
Fix "DeprecationWarning: There is no current event loop"

jmarcet:
I'm seeing the warning on Windows platforms only but it is a general deprecation of Python 3.10, to be dropped for 3.11

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -11,7 +11,6 @@ from asyncio import (
     CancelledError,
     Task,
     ensure_future,
-    get_event_loop,
     get_event_loop_policy,
     get_running_loop,
     wait_for,
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -12,6 +12,7 @@ from asyncio import (
     Task,
     ensure_future,
     get_event_loop,
+    get_event_loop_policy,
     get_running_loop,
     wait_for,
 )
@@ -254,7 +255,7 @@ class Sanic(BaseSanic, RunnerMixin, metaclass=TouchUpMeta):
                 "Loop can only be retrieved after the app has started "
                 "running. Not supported with `create_server` function"
             )
-        return get_event_loop()
+        return get_event_loop_policy().get_event_loop()
 
     # -------------------------------------------------------------------- #
     # Registration


ahopkins:
Thanks for the contribution. Can you let me know the context in which you are seeing this? I am particularly interested to know what is calling this causing the notice.

What platform are you seeing this on? OS/Python/Sanic?

jmarcet:
I am seeing this when running my iptv proxy [movistar-u7d](https://github.com/jmarcet/movistar-u7d) natively on Windows.

I am using python 3.10, installed with [scoop](https://scoop.sh/) on Windows 11 x86_64 and sanic stable, i.e., v21.12.1

For some reason I'm also getting a `[WARNING] Sanic is running in PRODUCTION mode. Consider using '--debug' or '--dev' while actively developing your application.` in the same environment. None of that happens in Linux.

ahopkins:
> For some reason I'm also getting a `[WARNING] Sanic is running in PRODUCTION mode. Consider using '--debug' or '--dev' while actively developing your application.` in the same environment. None of that happens in Linux.

This shows up on in TTY output when debug is off.

jmarcet:
> This shows up on in TTY output when debug is off.

Yeah, I checked it after writing the message. I am not seeing it in Linux in general because I usually run under docker, not in a terminal directly.

> So, if you are not calling `app.loop` manually, then something is outside of a running loop. While this may "solve" the problem, it seems more like a bandaid.

It is just `app.run`.

In those backends I only create tasks with `asyncio.create_task()` or launch external processes with `asyncio.create_task()` or  `app.add_task()`. Now that I've checked, there is a reference to `self.loop` [here](https://github.com/jmarcet/movistar-u7d/blob/625521f7442d625455d2304f4d08b2c3e24db77c/movistar_u7d.py#L532), but the warning shows up in both backends, so that cannot be it.

Furthermore, it is Windows only, I never saw it in Linux.

jmarcet:
I've just done a quick test with this simplistic sanic app:

```
#!/usr/bin/env python3

import sys

from sanic.log import logger as log


app = Sanic("test")


@app.get("/favicon.ico")
async def handle_favicon(request):
    return response.empty(200)


if __name__ == "__main__":
    try:
        app.run(
            host="127.0.0.1",
            port=9090,
            access_log=True,
            auto_reload=True,
            debug=True,
            workers=1,
        )
    except (KeyboardInterrupt, TimeoutError):
        sys.exit(1)
    except Exception as ex:
        log.critical(f"{repr(ex)}")
        sys.exit(1)
```

That's enough to trigger the warning: 

image

ahopkins (approved):

PR #2396, opened 2022-02-01
allow multidict version 6

dotlambda:
The only breaking change is the dropped support for Python 3.6, which sanic doesn't support anyway: https://github.com/aio-libs/multidict/releases/tag/v6.0.0


ahopkins (approved):

dotlambda:
ping @ahopkins

PR #2401, opened 2022-02-22
Changes to CLI

howzitcdf:
Small changes to help the developer out when their app is not starting

1. if the cli detects that a function has been passed instead of a Sanic instance it will error out and suggest that the dev uses `--factory` eg. `sanic fake.factory.run`

2. if the cli detects a path has been passed instead of a Sanic instance it will error out and suggest the that the dev uses `--simple` eg. `sanic ./fake/`

I welcome all changes and teachings! :)

### DIFF
--- a/sanic/cli/app.py
+++ b/sanic/cli/app.py
@@ -114,9 +114,10 @@ Or, a path to a directory to run as a simple HTTP server:
                 module_name, app_name = self.args.module.rsplit(delimiter, 1)
 
                 if module_name == "" and os.path.isdir(self.args.module):
-                    raise ValueError("App not found.\n"
-                    "   Please use --simple if you are passing a directory to sanic.\n"
-                    f"   eg. sanic {self.args.module} --simple"
+                    raise ValueError(
+                        "App not found.\n"
+                        "   Please use --simple if you are passing a directory to sanic.\n"
+                        f"   eg. sanic {self.args.module} --simple"
                     )
 
                 if app_name.endswith("()"):
--- a/sanic/cli/app.py
+++ b/sanic/cli/app.py
@@ -114,8 +114,8 @@ Or, a path to a directory to run as a simple HTTP server:
                 module_name, app_name = self.args.module.rsplit(delimiter, 1)
 
                 if module_name == "" and os.path.isdir(self.args.module):
-                    raise ValueError("App not found."
-                    "   Please use --simple if you are passing a directory to sanic."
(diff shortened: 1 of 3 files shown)


sanic/cli/app.py
```diff
@@ -76,7 +76,12 @@ def run(self):
             kwargs = self._build_run_kwargs()
             app.run(**kwargs)
         except ValueError:
-            error_logger.exception("Failed to run app")
+            if os.path.isdir(self.args.module):
+                error_logger.exception(
+                    "Please attach --simple to run your app from a directory."
+                )
+            else:
```
ahopkins, line 83: Why do this check here and not in the `_get_app` exception?

sanic/cli/app.py
```diff
@@ -125,9 +130,18 @@ def _get_app(self):
                 app_type_name = type(app).__name__
 
                 if not isinstance(app, Sanic):
+                    if hasattr(app, "__call__"):
```
ahopkins, line 133:
```python
if callable(app):
```

sanic/cli/app.py
howzitcal: didn't know we could, let me do that :)

sanic/cli/app.py
```diff
@@ -125,9 +130,18 @@ def _get_app(self):
...
+                        solution = f"sanic {self.args.module} --factory"
+                        raise ValueError(
+                            "Module is not a Sanic app, it is a"
+                            f"{app_type_name}\n"
+                            "  If your function returns a"
```
ahopkins, line 138: Let's not presume that it is a function.

```
If this callable returns a
```

sanic/cli/app.py
ahopkins: I would just do this check right next to the `--factory` check.

howzitcal: on it :)

sanic/cli/app.py
howzitcal: sorted, I learnt something :) thank you!

sanic/cli/app.py
howzitcal: done and dusted :)

sanic/cli/app.py
howzitcal: i changed where it raises slightly, however I think it should be fine, let me know :)

ahopkins (approved):

ahopkins (approved):

PR #2402, opened 2022-02-23
Conditionally inject CLI arguments into factory

ahopkins:
When starting with `--factory`, the CLI will attempt to inject the CLI arguments `Namespace` into the factory (if possible).


prryplatypus (approved):
LGTM. Wanna add a test?

ahopkins:
> LGTM. Wanna add a test?

I was waiting until #2401 is merged so that the tests could reuse the same fake factory module.

PR #2409, opened 2022-03-10
Add a docstring to `Request.respond()`

Bluenix2:
Without a docstring this method will not show up in the Sphinx documentation.

See [this relevant Discord conversation](https://discord.com/channels/812221182594121728/813454547585990678/951562386949677156) for information on how it was discovered.

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -196,7 +196,22 @@ class Request:
         status: int = 200,
         headers: Optional[Union[Header, Dict[str, str]]] = None,
         content_type: Optional[str] = None,
-    ):
+    ) -> BaseHTTPResponse:
+        """Respond to the request without returning.
+
+        This method can only be called once, as you can only respond once.
+        It is useful if you wish to respond to the request without returning
+        from the handler.
+
+        If `response` is not passed, a response will be created from the
+        `status`, `headers` and `content_type` keyword arguments.
+
+        :param response: response instance to send
+        :param status: status code to return in the response
+        :param headers: headers to return in the response
+        :param content_type: Content-Type of the response
+        :return: final response sent (may have changed because of middlewares)
+        """
         try:
             if self.stream is not None and self.stream.response:
                 raise ServerError("Second respond call is not allowed.")
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -196,15 +196,29 @@ class Request:
         status: int = 200,
         headers: Optional[Union[Header, Dict[str, str]]] = None,
         content_type: Optional[str] = None,
-    ) -> HTTPResponse:
+    ):
         """Respond to the request without returning.
 
         This method can only be called once, as you can only respond once.
         It is useful if you wish to respond to the request without returning
-        from the handler.
+        from the handler. In this case, you should be responsible for sending
+        data back to the client directly.
 
-        If `response` is not passed, a response will be created from the
-        `status`, `headers` and `content_type` keyword arguments.
+        .. code-block:: python
+
+            response = await request.respond(content_type="text/csv")
+            await response.send("foo,")
+            await response.send("bar")
+
+        You can control the completion of the response using ``eof()``.
+
+        .. code-block:: python
+
+            await response.eof()
+
+        If ``response` is not passed ``respond`` (like in the previous
+        example), a response object will be created from the ``status``,
+        ``headers`` and ``content_type`` keyword arguments.
 
         :param response: response instance to send
         :param status: status code to return in the response
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -196,7 +196,7 @@ class Request:
         status: int = 200,
         headers: Optional[Union[Header, Dict[str, str]]] = None,
         content_type: Optional[str] = None,
-    ):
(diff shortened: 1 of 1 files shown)


sanic/request.py
```diff
@@ -197,6 +197,21 @@ async def respond(
         headers: Optional[Union[Header, Dict[str, str]]] = None,
         content_type: Optional[str] = None,
     ):
```
SaidBySolo, line 199:
```suggestion
    ) -> HTTPResponse:
```

How about adding a type hint as well?

ahopkins (approved):

Bluenix2:
I originally did so, `BaseHTTPResponse` more specifically because it would match the parameter, but it failed type checks [as seen here](https://github.com/sanic-org/sanic/runs/5501781111?check_suite_focus=true#step:3:154).

I could not figure out how or why so I reverted that change. It did not fix the seemingly unrelated test failure though.

ahopkins:
> I originally did so, `BaseHTTPResponse` more specifically because it would match the parameter, but it failed type checks [as seen here](https://github.com/sanic-org/sanic/runs/5501781111?check_suite_focus=true#step:3:154).
> 
> I could not figure out how or why it failed so I reverted that change. It did not fix the seemingly unrelated test failure though.

I made a couple of changes. We can worry about the type annotation issue another time.

sanic/request.py
```diff
@@ -197,6 +197,35 @@ async def respond(
...
+
+        If ``response` is not passed ``respond`` (like in the previous
+        example), a response object will be created from the ``status``,
+        ``headers`` and ``content_type`` keyword arguments.
+
+        :param response: response instance to send
+        :param status: status code to return in the response
+        :param headers: headers to return in the response
+        :param content_type: Content-Type of the response
+        :return: final response sent (may have changed because of middlewares)
```
Bluenix2, line 227: Seeing the changes you made @ahopkins, is this still correct?

I thought you passed a `HTTPResponse` with a body that then would get sent. With the changes you made to the docstring that doesn't look to be the case (the user has to write the body themselves)?

The return value isn't a *sent* (past tense) response in this case because the user should write the response themselves.

ahopkins: Well... I guess the misleading part is "sent". I think my change should be clarified also.

`request.respond` _DOES_ execute the response middleware, and ultimately returns **either**:
1. the response created (either passed as an argument or created)
2. a response returned and interrupted by middleware

However, it does not actually do the sending. My example was more for how it would typically be used. But, this is also fine:

```python
@app.get("/")
    json_response = json({"foo": "bar"})
    await request.respond(json_response)


@app.on_response
async def add_header(_, response: HTTPResponse):
    response.headers["one"] = "two"
```

How about something more like this:

```
        This method can only be called once, as you can only respond once.
        It is useful typically in one of two situations: you wish to respond
        to the request without returning an object the handler, or you wish to
        respond to the client directly (for example, streaming content).

        If ``response` is not passed ``respond`` (like in the previous
        example), a response object will be created from the ``status``,
        ``headers`` and ``content_type`` keyword arguments.

        **SCENARIO #1**

        The request handler does not return an object. In this example, a
        :class:`sanic.response.HTTPResponse` object is explicitly used to set
        the content, type, and status of the response.

        .. code-block:: python

            @app.get("/")
                json_response = json({"foo": "bar"})
                await request.respond(json_response)


            @app.on_response
            async def add_header(_, response: HTTPResponse):
                response.headers["one"] = "two"

        **SCENARIO #2**

        The handler will control sending of data to the client.

        .. code-block:: python

            response = await request.respond(content_type="text/csv")
            await response.send("foo,")
            await response.send("bar")

        You can control the completion of the response using ``eof()``. But, 
        this is not necessary.

        .. code-block:: python

            await response.eof()

        The example from Scenario #1 can be rewritten:

        .. code-block:: python

            @app.get("/")
                response = await request.respond(
                    content_type="application/json"
                )
                await response.send(ujson.dumps({"foo": "bar"}))

        :param response: response instance to send
        :param status: status code to return in the response
        :param headers: headers to return in the response
        :param content_type: Content-Type of the response
        :return: final response to be sent (may have changed because of 
            middleware, and body content may differ based upon manual sending 
            of data)
```

ahopkins: These changes and some better examples probably should make it into [sanic-guide](https://github.com/sanic-org/sanic-guide).

Bluenix2: Not a fan of the bold `SCENARIO` text acting as headers, I think you could bake it into the text itself (`The other scenario...` or similar) but overall that is a big improvement.

I think that's a super clear explanation on how to use it!

ahopkins: > Not a fan of the bold `SCENARIO` text acting as headers

Feel free to incorporate and make any changes you think are appropriate. I just did not want to mess up the TOC at all while still making them visually distinctive. Not tied to the formatting or wording.

Bluenix2: I commited this in 371284fa488f89503bb7eb7f9b3926f1b0dce09e

ahopkins (approved):
Great change. I'll deal with the flaky tests 🤬

PR #2416, opened 2022-03-25
style: add msg in ``task.cancel``

### LINKED ISSUE #2325: Cancel tasks with a message where appropriate
[cancel()](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) also accepts a "msg" argument, might it be a good idea for this "message" to be added to places like https://github.com/sanic-org/sanic/blob/f7abf3db1bd4e79cd5121327359fc9021fab7ff3/sanic/server/protocols/http_protocol.py#L172 that are otherwise calling cancel() with no explanatory message?    if this is the CancelledError this user is getting, a simple message there would save everyone a lot of time.

_Originally posted by @zzzeek in https://github.com/sanic-org/sanic/issues/2296#issuecomment-983881945_

---

Where we are able to in Py3.9, we should add a message to `cancel()`.

SaidBySolo:
close #2325 

ref: https://github.com/sanic-org/sanic/blob/030987480c8d6e40d67ddfe1a4f9669b69131359/sanic/app.py#L1258-L1266


SaidBySolo:
If I test locally, everything is fine. I can't reproduce the failure in ci.

PR #2419, opened 2022-03-30
22.3 Internal version bumps

ahopkins:


tests/test_headers.py
```diff
@@ -164,11 +164,12 @@ def test_raw_headers(app):
...
-        b"Host: example.com\r\nAccept: */*\r\nAccept-Encoding: gzip, "
-        b"deflate\r\nConnection: keep-alive\r\nUser-Agent: "
-        b"Sanic-Testing\r\nFOO: bar"
-    )
+    assert b"Host: example.com" in request.raw_headers
+    assert b"Accept: */*" in request.raw_headers
+    assert b"Accept-Encoding: gzip, deflate" in request.raw_headers
+    assert b"Connection: keep-alive" in request.raw_headers
+    assert b"User-Agent: Sanic-Testing" in request.raw_headers
+    assert b"FOO: bar" in request.raw_headers
```
prryplatypus, line 172: Just to further constraint the possible matches...

```suggestion
    assert b"Host: example.com\r\n" in request.raw_headers
    assert b"Accept: */*\r\n" in request.raw_headers
    assert b"Accept-Encoding: gzip, deflate\r\n" in request.raw_headers
    assert b"Connection: keep-alive\r\n" in request.raw_headers
    assert b"User-Agent: Sanic-Testing\r\n" in request.raw_headers
    assert b"FOO: bar\r\n" in request.raw_headers
```

tests/test_routes.py
```diff
@@ -254,7 +254,7 @@ def handler2(request):
 
 
 def test_route_invalid_parameter_syntax(app):
-    with pytest.raises(ValueError):
+    with pytest.raises(InvalidUsage):
```
prryplatypus, line 257: Why does this raise an error for the client and not an error for the developer? I'd expect it to raise when declaring the route, not when accessing it. Am I missing something? 🤔

ahopkins: It raises at run time. Which we are just using the client as a quick and dirty proxy for running. Perhaps not the cleanest test.

PR #2420, opened 2022-04-01
refactor: consistent exception naming

prryplatypus:
Names exceptions consistently, so they match the corresponding HTTP code message.

- [x] Modify exceptions
- [x] Modify uses in code
- [x] Modify tests
- [x] Add tests


ahopkins (approved):
This is one of those small touches that screams of polish. Nicely done, and thanks for adding this.

PR #2421, opened 2022-04-03
Add fall back for Windows even loop fetching

ahopkins:


Tronic (dismissed):

prryplatypus (dismissed):

sjsadowski:
I don't think this needs to be held out, @ahopkins can you get this in? @prryplatypus reminded me that it's blocking for windows users.

ahopkins:
I know. I have been planning on releasing it, I just did not want to do it while I was traveling since access to my machine was less than ideal. I should be releasing this version later today.

Issue #2417, opened 2022-03-26
Expose ASGI scope on Request object


rmorshea:
**Is your feature request related to a problem? Please describe.**

I'm trying to write an application framework that works across ASGI compatible server implementations and to do this I'd like to make the ASGI scope available to users. Sanic does not make the scope publicly available though. 

**Describe the solution you'd like**
I'd like the `sanic.request.Request` object to provide a `scope` attribute which is the current `ASGIScope`

**Additional context**
Presently I'm working around this by accessing `request.app._asgi_app.transport.scope`

ahopkins:
That's a good idea, and sounds easy enough that we can probably squeeze that in for the upcoming release.

rmorshea:
You'll probably need to handle the case where the server isn't being run via ASGI, but from what I can tell it seems like the project is moving away from that.

ahopkins:
> from what I can tell it seems like the project is moving away from running without ASGI.

Just curious, what makes you think this is the case? The built-in Sanic server (TMK) is still the most common deployment method. But, I do agree that ASGI support is an important alternative.

rmorshea:
I can't recall exactly, but I may have skimmed [this thread](https://community.sanicframework.org/t/considering-asgi-support/70/3) and come away with the wrong idea.

ahopkins:
No, not necessarily. ASGI is very important and I hope will continue to be. But we are not getting rid of the internal server either. Being geared specifically for Sanic it has the benefit of being optimized for performance in ways that ASGI servers cannot be.

rmorshea:
That sounds reasonable, but then again, you could also make it a `@property` and return `None` if there's no `_asgi_app`.

I'm not very familiar with Sanic's internals though, so I can't say what the best option would be.

azimovMichael:
Yeah I can do that, but I guess the downside is that currently the Request object is decoupled completely from any ASGI internals, and this change will change that a bit

ahopkins:
@azimovMichael I would be happy for you to take this on. I see you have been assigned it already. Sorry for my delay, I am just now getting back to my regular schedule after some time off.

Instead of altering the `__init__`, we could just assign it:

```python
instance.request = request_class(
    ...
)
instance.request.stream = instance
...
instance.request._scope = scope
```

Using `@property` as suggested by @rmorshea is a nice idea so that it is less easily changed and a pattern used elsewhere on the `Request` object.

ahopkins:
[...]

PR #2432, opened 2022-04-18
Expose scope parameter in request object

azimovMichael:
From #2417

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -827,7 +827,7 @@ class Request:
         :rtype: str
         """
 
-        return self.app._asgi_app.transport.scope if self.app._asgi_app else None
+        return self.app._asgi_app.transport.scope if self.app.asgi else None
 
 
 class File(NamedTuple):
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -92,6 +92,7 @@ class Request:
         "_protocol",
         "_remote_addr",
         "_socket",
+        "_scope",
         "_match_info",
         "_name",
         "app",
@@ -168,6 +169,7 @@ class Request:
         self.stream: Optional[Http] = None
         self.route: Optional[Route] = None
         self._protocol = None
+        self._scope: Optional[ASGIScope] = None
         self.responded: bool = False
 
     def __repr__(self):
@@ -827,7 +829,7 @@ class Request:
         :rtype: str
         """
 
-        return self.app._asgi_app.transport.scope if self.app.asgi else None
+        return self._scope
 
 
 class File(NamedTuple):
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -92,7 +92,6 @@ class Request:
         "_protocol",
         "_remote_addr",
         "_socket",
-        "_scope",
         "_match_info",
         "_name",
         "app",
@@ -169,7 +168,6 @@ class Request:
         self.stream: Optional[Http] = None
         self.route: Optional[Route] = None
         self._protocol = None
-        self._scope: Optional[ASGIScope] = None
         self.responded: bool = False
 
     def __repr__(self):
@@ -829,7 +827,7 @@ class Request:
         :rtype: str
         """
 
-        return self._scope
+        return self.transport.scope if self.app.asgi else None
 
 
 class File(NamedTuple):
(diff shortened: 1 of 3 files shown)


sanic/request.py
```diff
@@ -819,6 +820,15 @@ def url_for(self, view_name: str, **kwargs) -> str:
             view_name, _external=True, _scheme=scheme, _server=netloc, **kwargs
         )
 
+    @property
+    def scope(self) -> Optional[ASGIScope]:
+        """
+        :return: the URL
```
prryplatypus, line 826:
```suggestion
        :return: the ASGI scope object if running in ASGI mode
```

sanic/request.py
```diff
@@ -819,6 +820,15 @@ def url_for(self, view_name: str, **kwargs) -> str:
...
+        :rtype: str
```
prryplatypus, line 827:
```suggestion
        :rtype: Optional[ASGIScope]
```

tests/test_request.py
```diff
@@ -191,3 +191,26 @@ def test_bad_url_parse():
...
+
+def test_request_scope_is_none_when_no_asgi():
+    app = Sanic("no_asgi")
+
+    @app.get("/")
+    async def get(request):
+        return response.empty()
+
+    request, _ = app.test_client.get("/")
+    assert request.scope == None
```
prryplatypus, line 204: Small nitpick
```suggestion
    assert request.scope is None
```

tests/test_request.py
```diff
@@ -191,3 +191,26 @@ def test_bad_url_parse():
...
+
+
+@pytest.mark.asyncio
+async def test_request_scope_is_not_none_when_runnin_in_asgi(app):
+    @app.get("/")
+    async def get(request):
+        return response.empty()
+
+    request, _ = await app.asgi_client.get("/")
+    assert request.scope != None
```
prryplatypus, line 214:
```suggestion
    assert request.scope is not None
```

prryplatypus (approved):
Thanks!

sanic/request.py
```diff
@@ -819,6 +820,15 @@ def url_for(self, view_name: str, **kwargs) -> str:
...
         )
 
+    @property
+    def scope(self) -> Optional[ASGIScope]:
+        """
+        :return: The ASGI scope of the request. If the app isn't an ASGI app, then returns None.
+        :rtype: Optional[ASGIScope]
+        """
+
+        return self.transport.scope if self.app.asgi else None
```
ahopkins, line 830: I would rather this always return `ASGIScope` without `Optional`. It will be easier for annotations and not having to add something like the following whenever you want to use `request.scope.get`, etc.

```py
if request.scope:
    ...
```

ahopkins: Instead, as mentioned in [my comment](https://github.com/sanic-org/sanic/issues/2417#issuecomment-1101342575) I would rather raise an exception if not in ASGI.

ahopkins (changes requested):

tests/test_request.py
```diff
@@ -191,3 +191,27 @@ def test_bad_url_parse():
...
+
+
+def test_request_scope_raises_exception_when_no_asgi():
+    app = Sanic("no_asgi")
+
+    @app.get("/")
+    async def get(request):
+        return request.scope
+
+    _, response = app.test_client.get("/")
```
ahopkins, line 203: You could even go back similar to what your previous test looked like and test for the exceptions:

```
request, _ = ...

with pytest.raises(NotImplementedError):
    request.scope
```

ahopkins (approved):

ahopkins (approved):

ahopkins:
@azimovMichael run `black` on the codebase. You can also just run `make pretty`.

ahopkins (approved):

azimovMichael:
@ahopkins Not sure why py38 tests are failing, they pass locally.
The typing checks on the other hand are failing because `TransportProtocol` doesn't have `scope` parameter but we actually use `MockProtocol` under-the-hood

ahopkins:
I can handle the typing issue. Don't worry about the failing tests. There are some tests that we scratch our head about because they are flaky on GH but run fine locally.

azimovMichael:
@ahopkins I added scope param to the TransportProtocol, if it's ok with you @ahopkins

PR #2434, opened 2022-04-22
Preserve blank values for form parsing (option)

### LINKED ISSUE #2427: request.form uses urllib.parse_qs which does not persist blank values by default when requests submit application/x-www-form-urlencoded data
**Describe the bug**
This seems to be an acknowledged limitation of parse_qs, but I'm raising as a bug as this introduces undesired behaviour rather than adding new features.

While `request.get_args` and and `request.get_query_args` accept the `keep_blank_values` argument in order to persist blanks, `request.form` does not provide an interface to persist blank values.

My assertion is that a framework should not remove data simply because it doesn't have a value.

Relevant anecdote: Twilio webhooks submit a POST request using `application/x-www-form-urlencoded` encoded body. The body includes all values including some potentially blank/empty values.

When the request is processed by Sanic, it is decoded using `urllib.parse_qs` and does not have a mechanism to allow blank elements to persist. This results in missing elements from the request when accessing the variables from `request.form`.

In my case, the elements from the submission are relevant to generate a signature for request validation (see https://www.twilio.com/docs/usage/security). The signature is constructed using SHA1 hash of the submitted request key/value pairs and missing pairs generates a different signature to that expected.

In this instance, I have been receiving requests which include two elements that, depending on the hook, may or may not have a value.

**Expected behavior**
All variables, irrespective of existence of a value should be passed to the application.

 - Version v22.3.0

sjsadowski:
This is a non-breaking update to form parsing based on a realistic expectation of being able to preserve blank values that are passed. As this currently works, using the default behavior of urllib.parse_qs, blank values are dropped, however this is sometimes not the preferred behavior. urllib.parse_qs has a parameter (keep_blank_values) that defaults to false, but can be set to true to ensure this is available. This fix allows for that option to be used by creating a new request property that can be updated.

closes #2427 

- New request property: keep_blank_values (default false)
if set to true, will preserve blank values in x-www-form-urlencoded forms when request.form is accessed

- New tests:
form keep blanks
form drop blanks (expected behavior)

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -427,7 +427,7 @@ class Request:
         return self.parsed_credentials
 
     @property
-    def form(self):
+    def form(self, keep_blank_values: bool = False):
         if self.parsed_form is None:
             self.parsed_form = RequestParameters()
             self.parsed_files = RequestParameters()
@@ -438,7 +438,7 @@ class Request:
             try:
                 if content_type == "application/x-www-form-urlencoded":
                     self.parsed_form = RequestParameters(
-                        parse_qs(self.body.decode("utf-8"))
+                        parse_qs(self.body.decode("utf-8"), keep_blank_values)
                     )
                 elif content_type == "multipart/form-data":
                     # TODO: Stream this instead of reading to/from memory
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -146,7 +146,7 @@ class Request:
         self.head = head
 
         # enforce existing behavior prior to issue #2427
-        self.keep_blank_values = False
+        self.keep_blank_form_values = False
 
         # Init but do not inhale
         self.body = b""
@@ -431,17 +431,12 @@ class Request:
         return self.parsed_credentials
 
     @property
-    def keep_blank_values(self) -> bool:
-        """Getter for preserving blank form values"""
+    def keep_blank_form_values(self) -> bool:
         return self._keep_blank_form_values
 
-    @keep_blank_values.setter
-    def keep_blank_values(self, keep_blank_values: bool = False):
-        """Setter for preserving blank form values
-
-        Defaults to false
-        """
-        self._keep_blank_form_values = keep_blank_values
+    @keep_blank_form_values.setter
+    def keep_blank_values(self, keep_blank_form_values: bool = False):
+        self._keep_blank_form_values = keep_blank_form_values
 
     @property
     def form(self):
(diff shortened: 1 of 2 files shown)


sanic/request.py
```diff
@@ -144,6 +145,9 @@ def __init__(
         self.transport = transport
         self.head = head
 
+        # enforce existing behavior prior to issue #2427
+        self._keep_blank_form_values = False
```
prryplatypus, line 149: Is there any specific reason to make this property private and then add a getter & setter without any logic in them?

prryplatypus (approved):

sanic/request.py
sjsadowski: Missed that in refactoring.

sanic/request.py
```diff
@@ -93,6 +93,7 @@ class Request:
         "_socket",
         "_match_info",
         "_name",
+        "_keep_blank_form_values",
```
prryplatypus, line 96:
```suggestion
        "keep_blank_form_values",
```

sanic/request.py
```diff
@@ -426,6 +430,19 @@ def credentials(self) -> Optional[Credentials]:
...
+        return self._keep_blank_form_values
+
+    @keep_blank_values.setter
+    def keep_blank_values(self, keep_blank_values: bool = False):
+        """Setter for preserving blank form values
+
+        Defaults to false
+        """
+        self._keep_blank_form_values = keep_blank_values
+
```
prryplatypus, line 445: What I meant was that the setter methods weren't really necessary, and we could just make that property public like you've now done, so I guess these methods are no longer necessary.

```suggestion
```

sjsadowski: discussed on discord

prryplatypus (approved):

sanic/request.py
```diff
@@ -144,6 +145,9 @@ def __init__(
         self.transport = transport
         self.head = head
 
+        # enforce existing behavior prior to issue #2427
+        self.keep_blank_form_values: bool = False
```
ahopkins, line 149: Should we also implement the same feature for query params?

We should think a little more on the API as well. While it absolutely should be controllable per `Request` instance, I can also see wanting to control this globally. Furthermore, if we implement for query params and form, should they have two different API values? Or, a unified API like this:

```python
class KeepBlank(Enum):
    NONE = ...
    FORM_ONLY = ...
    QUERY_ONLY = ...
    ALL = ...

keep_blank_values: bool | KeepBlank
```

ahopkins (changes requested):

ahopkins:
Should we instead be opening up the API to follow the same pattern for `args` and `query_args` using `get_*`?

prryplatypus:
I do agree that we should keep the behaviour consistent between query and form args, and -personally- I would've gone for the `get_*` approach and would've made the form attribute just be a default invocation of such method.

sjsadowski:
This is pretty cluttered. Will refactor based on discord conversation and open a new PR.

PR #2435, opened 2022-04-23
use next instead of forloop & replace set with comprehension

Mng-dev-ai:
This PR adds a slight speed improvement by using next instead of for loop and using {} instead of set() which should be faster than before

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -310,14 +310,14 @@ class Blueprint(BaseSanic):
             # Prepend the blueprint URI prefix if available
             uri = url_prefix + future.uri if url_prefix else future.uri
 
-            version_prefix = self.version_prefix
-            for prefix in (
-                future.version_prefix,
-                opt_version_prefix,
-            ):
-                if prefix and prefix != "/v":
-                    version_prefix = prefix
-                    break
+            version_prefix = next(
+                (
+                    prefix
+                    for prefix in (future.version_prefix, opt_version_prefix)
+                    if prefix and prefix != "/v"
+                ),
+                self.version_prefix,
+            )
 
             version = self._extract_value(
                 future.version, opt_version, self.version
@@ -449,16 +449,11 @@ class Blueprint(BaseSanic):
 
     @staticmethod
     def _extract_value(*values):
-        value = values[-1]
-        for v in values:
-            if v is not None:
-                value = v
-                break
-        return value
+        return next((v for v in values if v is not None), values[-1])
 
     @staticmethod
     def register_futures(
         apps: Set[Sanic], bp: Blueprint, futures: Sequence[Tuple[Any, ...]]
     ):
         for app in apps:
-            app._future_registry.update(set((bp, item) for item in futures))
+            app._future_registry.update({(bp, item) for item in futures})


sjsadowski:
Please add a description outlining both the value and purpose of this change.

Mng-dev-ai:
@sjsadowski  Added

sanic/blueprints.py
```diff
@@ -449,16 +449,11 @@ def event(self, event: str, timeout: Optional[Union[int, float]] = None):
 
     @staticmethod
     def _extract_value(*values):
-        value = values[-1]
-        for v in values:
-            if v is not None:
-                value = v
-                break
-        return value
+        return next((v for v in values if v is not None), values[-1])
```
sjsadowski, line 452: Is there a reason for using a generator instead of a list comprehension here?

sjsadowski (changes requested):

sanic/blueprints.py
Mng-dev-ai: I think generators are better to be used when it's possible as it uses less memory since it doesn't build the whole list at once

sjsadowski, line 452: @ahopkins is this to really just return the first value found that is not None? Just want to make sure I understand what you were after here. I'm okay with switching to the generator as it's a parity change, but I want to verify that's what you were really after.

Mng-dev-ai: yea, it returns the first value found that is not None

sjsadowski: @Mng-dev-ai yes, but I want to make sure that you didn't uncover a bug, @ahopkins will probably look tomorrow.

ahopkins: The logic seems fine. We probably could improve DRY a little by using `_extract_value` with the `version_prefix` and just abstract the conditional a bit since they are essentially doing the same thing, and it would keep the priority setting using a consistent methodology.

However, TBH, I am not sure how this is an improvement and what this change adds. Even running 10M iterations over it, I only had to run my benchmarks twice to get inconsistent results. I am not sure that I would agree that a generator adds any performance here.
image

We are only looping through 2 or three items, not a very large list. And, the values already exist in memory so using a generator does not reduce the footprint. Am I missing something else?

sanic/blueprints.py
```diff
@@ -449,16 +449,11 @@ def event(self, event: str, timeout: Optional[Union[int, float]] = None):
...
 
     @staticmethod
     def register_futures(
         apps: Set[Sanic], bp: Blueprint, futures: Sequence[Tuple[Any, ...]]
     ):
         for app in apps:
-            app._future_registry.update(set((bp, item) for item in futures))
+            app._future_registry.update({(bp, item) for item in futures})
```
ahopkins, line 459: What is the purpose of this change?

sanic/blueprints.py
ahopkins: I am not sure that I agree with this logic here. Arguments are passed as an existing object (a tuple). They already are existing in memory in an iterable. Where are we rebuilding the list that we are incurring a memory overhead?

Mng-dev-ai: yea, in this case here there's no memory overhead the main reason here for using generators is to use next instead of forloop

Mng-dev-ai: Didn't know that the array has only a few items so as u said the difference won't be noticed

sanic/blueprints.py
Mng-dev-ai: no need to use set() here as it's slower than just {}

ahopkins:
Sorry @Mng-dev-ai. I am closing this for now as I do not see how this improves the codebase. Feel free to convince me otherwise if you think I am wrong.

1. Using the generator instead of the forloop does not add anything.
2. Using a set comprehension instead of `set` also does not add anything in this context.

Both of these items run only at startup. While it is a stated goal to increase performance of the project, we are mainly focused on the request/response cycle and not just marginal startup costs.

PR #2438, opened 2022-04-24
feat: easier websocket interface annotation

### LINKED ISSUE #2437: Easier websocket interface annotation
Right now, to properly annotate a websocket endpoint you need to do this:

```python
from sanic.server.websockets.impl import WebsocketImplProtocol
from sanic import Request

@app.websocket("")
async def handler(request: Request, ws: WebsocketImplProtocol):
    ...
```

That is not easy or intuitive.

This would be much nicer:

```python
from sanic import Request, Websocket

@app.websocket("")
async def handler(request: Request, ws: Websocket):
    ...
```

We should just alias and put it inside `__init__.py` for convenience.

SaidBySolo:
close #2437

### DIFF
--- a/sanic/__init__.py
+++ b/sanic/__init__.py
@@ -4,6 +4,7 @@ from sanic.blueprints import Blueprint
 from sanic.constants import HTTPMethod
 from sanic.request import Request
 from sanic.response import HTTPResponse, html, json, text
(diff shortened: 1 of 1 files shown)


ahopkins (approved):

ahopkins:
Thanks @SaidBySolo

thibautd:
Thanks @SaidBySolo !

PR #2439, opened 2022-04-24
Preserve blank form values for urlencoded forms (option)

### LINKED ISSUE #2427: request.form uses urllib.parse_qs which does not persist blank values by default when requests submit application/x-www-form-urlencoded data

sjsadowski:
This is a non-breaking update to form parsing based on a realistic expectation of being able to preserve blank values that are passed. As this currently works, using the default behavior of urllib.parse_qs, blank values are dropped, however this is sometimes not the preferred behavior. urllib.parse_qs has a parameter (keep_blank_values) that defaults to false, but can be set to true to ensure this is available. This fix allows for that option to be used by creating a new request property that can be updated.

closes #2427

refactor logic from the .form property to .get_form() method which will accept the optional parameter of keep_blank_values as a boolean. By default this is False, preserving the existing behavior. If set to True, will preserve blank values in x-www-form-urlencoded forms when request.form is accessed

New tests:
form keep blanks
form drop blanks (expected behavior)

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -426,7 +426,7 @@ class Request:
                 pass
         return self.parsed_credentials
 
-    def get_form(self, keep_blank_values: bool = False):
+    def get_form(self, keep_blank_values: bool = False) -> RequestParameters:
         self.parsed_form = RequestParameters()
         self.parsed_files = RequestParameters()
         content_type = self.headers.getone(
@@ -451,6 +451,8 @@ class Request:
                 )
         except Exception:
             error_logger.exception("Failed when parsing form")
+
+        return self.parsed_form
 
     @property
     def form(self):
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -153,7 +153,7 @@ class Request:
         self.parsed_credentials: Optional[Credentials] = None
         self.parsed_json = None
         self.parsed_form: Optional[RequestParameters] = None
-        self.parsed_files = None
+        self.parsed_files: Optional[RequestParameters] = None
         self.parsed_token: Optional[str] = None
         self.parsed_args: DefaultDict[
             Tuple[bool, bool, str, str], RequestParameters
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -443,7 +443,7 @@ class Request:
                 )
             elif content_type == "multipart/form-data":
                 # TODO: Stream this instead of reading to/from memory
-                boundary = parameters["boundary"].encode("utf-8")
+                boundary = parameters["boundary"].encode("utf-8")  # type: ignore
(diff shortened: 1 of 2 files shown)


sanic/request.py
```diff
@@ -426,28 +426,36 @@ def credentials(self) -> Optional[Credentials]:
                 pass
         return self.parsed_credentials
 
+    def get_form(self, keep_blank_values: bool = False):
```
ahopkins, line 429: Doesn't have a return annotation. Looks like the function also is missing a return value.

ahopkins (changes requested):

ahopkins (approved):

sanic/request.py
```diff
@@ -426,28 +426,40 @@ def credentials(self) -> Optional[Credentials]:
...
+        )
+        content_type, parameters = parse_content_header(content_type)
+        try:
+            if content_type == "application/x-www-form-urlencoded":
+                self.parsed_form = RequestParameters(
+                    parse_qs(
+                        self.body.decode("utf-8"),
+                        keep_blank_values=keep_blank_values,
+                    )
+                )
```
prryplatypus, line 445: Should we consider caching the parsed for with the different settings like we do for query args?
https://github.com/sanic-org/sanic/blob/3a6cc7389c5166a0c9580daf5d709a0dee85a996/sanic/request.py#L561-L563

sjsadowski: I've pretty much burned my weekend on this, if we want to make additional changes to functionality someone can pick it up from here.

ahopkins: That to me sounds like its own PR and discussion.

PR #2444, opened 2022-05-01
2369 graceful shutdown of signal handlers

jonathanstaniforth:
Issue: https://github.com/sanic-org/sanic/issues/2369

This PR makes two key changes:

- It names signal handler `asyncio.Task`s "signal" so that they are distinguishable from other `asyncio.Task`s that are running in the event loop.
- It cancels running signal handler `asyncio.Task`s when sanic is stopped.

### DIFF
--- a/sanic/signals.py
+++ b/sanic/signals.py
@@ -192,7 +192,7 @@ class SignalRouter(BaseRouter):
         if inline:
             return await dispatch
 
-        task = asyncio.get_running_loop().create_task(dispatch)
+        task = asyncio.get_running_loop().create_task(dispatch, name="signal")
         await asyncio.sleep(0)
         return task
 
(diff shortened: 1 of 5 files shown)


sanic/signals.py
```diff
@@ -192,7 +192,7 @@ async def dispatch(
         if inline:
             return await dispatch
 
-        task = asyncio.get_running_loop().create_task(dispatch)
+        task = asyncio.get_running_loop().create_task(dispatch, name="signal")
```
ahopkins, line 195: Do you think we should add a more descriptive name? Also include the handler `__name__`?

ahopkins:
This does not seem to be working as intended. I am going to hold off on this until next release and try and come up with another solution. If you have other ides, LMK.

PR #2447, opened 2022-05-03
File Cache Control Headers Support

ChihweiLHBird:
Related issue: https://github.com/sanic-org/sanic/issues/2441

Would like to get some feedback about the implementation.

The behavior was made to be similar to `Werkzeug`. But I think we can only keep the auto headers option (and remove `last_modified`) since it is a fact from the file stat and user can still override the it with a `headers` dictionary passed in.


ChihweiLHBird:
Current headers looks like:
image

ChihweiLHBird:
@sanic-org/framework @ahopkins May I have some feedbacks about the draft? Thanks!

Tronic:
@ChihweiLHBird Please avoid force pushing on PRs. Use normal push so that the notification don't give an error about commits not being found.

ChihweiLHBird:
> @ChihweiLHBird Please avoid force pushing on PRs. Use normal push so that the notification don't give an error about commits not being found.
> 

Thanks. Sorry about that, and I didn't know that before. I wanted to undo the commit because the change was messed up. Won't do it in the future PRs.

sanic/response.py
```diff
@@ -149,6 +159,8 @@ async def send(
             if hasattr(data, "encode")
             else data or b""
         )
+        if self.auto_content_length and "content-length" not in self.headers:
```
ahopkins, line 162: Why are we adding this here on the base? This is something that will already be added when needed: https://github.com/sanic-org/sanic/blob/86ae5f981cbe64fd85bee34fee3989b009e90dd2/sanic/http.py#L300

sanic/response.py
```diff
@@ -317,13 +333,37 @@ async def file(
...
+        )
+    if max_age:
+        headers.setdefault("cache-control", f"max-age={max_age}")
+        headers.setdefault(
+            "expires",
+            formatdate(
+                time() + max_age,
+                usegmt=True,
+            ),
+        )
```
ahopkins, line 365: Stylistic thing: can we keep the spacing between all of these ifs consistent?

sanic/response.py
```diff
@@ -317,13 +333,37 @@ async def file(
...
     """
     headers = headers or {}
     if filename:
         headers.setdefault(
             "Content-Disposition", f'attachment; filename="{filename}"'
         )
+    if isinstance(last_modified, datetime):
+        last_modified = last_modified.timestamp()
+
+    if last_modified is None:
```
ahopkins, line 348: This could be an `elif`

sanic/response.py
```diff
@@ -317,13 +333,37 @@ async def file(
...
+        last_modified = Path(location).stat().st_mtime
+    if isinstance(max_age, Default):
+        max_age = 0  # Should change this (default) value to configable?
```
ahopkins, line 351: Yes.

sanic/response.py
```diff
@@ -149,6 +159,8 @@ async def send(
             if hasattr(data, "encode")
             else data or b""
         )
+        if self.auto_content_length and "content-length" not in self.headers:
+            self.headers["content-length"] = str(len(data))  # type: ignore
```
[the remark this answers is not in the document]
ChihweiLHBird: Solved. `data` could be `None` and now it will be checked.
Thanks!

sanic/response.py
ChihweiLHBird: Fixed

sanic/response.py
ChihweiLHBird: Oh, that's true. Removed auto content length in response.py and let Http handle it. Thanks!

sanic/response.py
ChihweiLHBird: I am sure what are the spacing... Can you tell me a little more about it? What spacings should be consistent? Current spacing is made by Black via `make pretty`.

sanic/response.py
ChihweiLHBird: Working on this.

sanic/response.py
ahopkins: Really? 
image

If black is happy with it, then I will not question it. :shrug:

ahopkins:
Marking it as a draft. Change it back when it is ready for final review :sunglasses:

sanic/response.py
ChihweiLHBird: GitHub Linter Checks workflow seems passed, and my local `tox -e lint` is an okay... 
Maybe there are some changes in the latest black version?

sanic/response.py
ChihweiLHBird: I think `conifg` is not accessible in `file` function because nether `app` nor `app.config` is being passed into the function. After exploring how werkzeug did it, I came up with a similar behavior that returning `no-cache` (the response can be stored in caches, but the response must be validated with the origin server before each reuse) when `max-age` is not set with an extra option for `no-store` (absolutely not to cache). Can you re-review these changes? Much appreciate.

PR #2448, opened 2022-05-04
Fix for running in pythonw

timmo001:
Running sanic server on windows and on `pythonw.exe` causes the app to silently crash due to `sys.stdout` being None.

NVIDIA_Share_2022-05-04_12-36-03
 

This adds a check for stdout. The rest of the server then continues to function normally.


sanic/mixins/runner.py
```diff
@@ -424,7 +424,7 @@ def _helper(
 
         self.motd(self.serve_location)
 
-        if sys.stdout.isatty() and not self.state.is_debug:
+        if sys.stdout and sys.stdout.isatty() and not self.state.is_debug:
```
ahopkins, line 427: For some reason GH is not giving me the option to create a suggestion here... :thinking: 

Nevetheless, can you change this whole block like this:

```python
        if not self.state.is_debug:
            message = (
                "Sanic is running in PRODUCTION mode. Consider using "
                "'--debug' or '--dev' while actively developing your "
                "application."
            )
            if sys.stdout and sys.stdout.isatty():
                message = "".join([Colors.YELLOW, message, Colors.END])
            error_logger.warning(message)
```

ahopkins (changes requested):

sanic/mixins/runner.py
ahopkins: At some point we can abstract some of the color stuff, but for now this at least keeps the message still populating.

ahopkins:
@timmo001 I am sorry, in my head I was looking at the original logic thinking that it was wrong. That is why I suggested the change. But, it was correct. I reverted the PR back to your original suggestion. Let's pretend I never suggested that to begin with :laughing: 

Thanks for catching this for us.

ahopkins (approved):

ahopkins (changes requested):
Here's a suggestion that _ACTUALLY_ makes sense...

There are a few more places that follow a similar pattern. Perhaps you can add the same change:

```
sanic/mixins/runner.py:427:        if sys.stdout and sys.stdout.isatty() and not self.state.is_debug:
sanic/mixins/runner.py:618:                if sys.stdout.isatty():
sanic/mixins/runner.py:659:                        if sys.stdout.isatty():
sanic/log.py:86:    if sys.stdout.isatty():
sanic/application/motd.py:39:        motd_class = MOTDTTY if sys.stdout.isatty() else MOTDBasic
sanic/application/logo.py:47:        if sys.stdout.isatty()
```

ahopkins (dismissed):

timmo001:
Update on this? I see its been approved but the checks are failing.

I've rebased, but this requires an approver to run

timmo001:
I've rebased, but this requires an approver to run

timmo001:
Fixed lint, needs approval

ahopkins (approved):

ahopkins (approved):

ahopkins (dismissed):

timmo001:
Schedule for the next release?

ahopkins:
Should be this week.

PR #2465, opened 2022-05-23
Properly catch websocket CancelledError in websocket handler in Python 3.7

ashleysommer:
Apply fix #2463 to Sanic 21.12LTS version

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -1004,10 +1004,10 @@ class Sanic(BaseSanic, metaclass=TouchUpMeta):
         cancelled = False
         try:
             await fut
-        except Exception as e:
-            self.error_handler.log(request, e)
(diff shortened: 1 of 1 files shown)


ashleysommer:
I'm guessing we don't need to worry about the linter checks failing on this?

ahopkins:
That's odd... I wonder why that is? I am assuming it must be a versioning issue. But yes, I do not care about that.

ashleysommer:
Black check failing:
```
...
would reformat sanic/config.py
...
would reformat sanic/headers.py
...
```

Maybe a newer version of black used now than in December release?
Or maybe some patches were pushed to 21.12LTS branch previously that broke formatting.

ahopkins:
I think it is a black version issue. I have run into this on some other projects.

PR #2468, opened 2022-05-25
Add trustme certs

ahopkins:
We should offer support for creation of certs using both `mkcert` and trustme`. 

I am running into [an issue](https://github.com/aiortc/aioquic/issues/295) with `trustme` certs right now though. They work okay for the HTTP/1.1 server, but fail for the `aioquic` server. More to come...


miss85246:
That's great! And I will finish the translation work next weekend.

PR #2479, opened 2022-06-10
Fix `test_cli` and `test_cookies`

ChihweiLHBird:
I was able to reproduce the test cli issues on my laptop, which is the timeout occurred when running the subprocess in `capture` function. After increasing the timeout locally, the issue never came up again during multiple tests.

Open this PR to further test the test cases.

I think this issue usually happen on smaller machine with lower performance.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -2,6 +2,7 @@ import json
 import subprocess
 
 from pathlib import Path
+from typing import List
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


ChihweiLHBird:
Test cookies issue seems a problem of httpx's async client. 
https://github.com/encode/httpx/discussions/2144

In the failed test cookie case, the headers contain setting cookie but hpptx didn't put it into `response.cookies`. I think there is not much we can do until httpx dev team fix it. If we really want to fix it right now, we can try to parse the cookies in sanic-testing after getting the response object from httpx.

ahopkins (dismissed):

ChihweiLHBird:
It's weird that fixing cli tests seems increase the chance of the failed cookies test. @ahopkins

ChihweiLHBird:
The cookies test fixed, the only (occasionally) broken test case is keep alive.

ahopkins (approved):

ahopkins:
Amazing amazing work. You rock.

PR #2482, opened 2022-06-20
Upgrade LTS to use sanic-routing 21.12 version

leotada:
Upgrade sanic-routing to a "compatible" LTS version 21.12.*


ahopkins:
That actually will be a breaking change. The LTS was released without this bump by accident, so the breaking change was not included. I intentionally did not go back to fix because it will have an impact on users. Making this change will break for anyone using "string" or "number" as a path parameter.

PR #2485, opened 2022-06-24
Add missing spaces in CLI error message

prryplatypus:


ahopkins (approved):

sanic/cli/app.py
```diff
@@ -146,9 +146,9 @@ def _get_app(self):
                     if callable(app):
                         solution = f"sanic {self.args.module} --factory"
                         raise ValueError(
-                            "Module is not a Sanic app, it is a"
+                            "Module is not a Sanic app, it is a "
                             f"{app_type_name}\n"
-                            "  If this callable returns a"
+                            "  If this callable returns a "
                             f"Sanic instance try: \n{solution}"
```
Tronic, line 152: This ought to be using a single f string literal specifically to avoid this type of problem. And the line length limitation should be long enough to fit that on a single line (or allow breaking the rule for a literal).

These are just my notes for improving code quality in general (and for reconsideration of `black` settings), not asking for any immediate changes.

PR #2490, opened 2022-06-30
Make WebsocketImplProtocol async iterable

ChihweiLHBird:
Make websocket object iterable. Will add tests later when the code has been finalized.

Update:
Manual test was okay. Currently we don't have a test client in Sanic that can send and receive WebSocket messages, so it seems impossible to have a test case for this feature at this time.


Tronic:
Still preferring a solution that only uses `__aiter__`.

ChihweiLHBird:
> ~Still preferring a solution that only uses `__aiter__`.~ Don't care which way it is done.

Just removed `__anext__`, lol.

sanic/server/websockets/impl.py
```diff
@@ -840,3 +844,10 @@ def connection_lost(self, exc):
         self.abort_pings()
         if self.connection_lost_waiter:
             self.connection_lost_waiter.set_result(None)
+
+    async def __aiter__(self):
+        while True:
+            try:
+                yield await self.recv()
+            except ConnectionClosedOK:
+                return
```
Tronic, line 853: I didn't benchmark that but I believe this can be faster if the try-except is moved outside of the loop. Someone else needs to look at whether this is the handling that we want for closed sockets.

Tronic: For cleaner code avoiding try-except for this sort of structure, consider:
```python
from contextlib import suppress

with suppress(ConnectionClosedOK, AnyOtherExceptionToIgnore, ...):
    while True:
        yield await self.recv()
```

Tronic:
Thanks for the PR. This is a very useful feature and it helps to have parity with the websockets module. Receive on a closed socket should probably follow that as well, and no-error on clean close seems like the right thing to do anyway. Will approve as is if no-one else comments on that, with the code changes suggested.

sanic/server/websockets/impl.py
ChihweiLHBird: Thank you! My first time to use `contextlib`. It looks very useful.

Tronic (approved):

ahopkins (dismissed):

ahopkins:
Can you add a test?

ChihweiLHBird:
> Can you add a test?

Our test client with WebSocket mimic client support was not released so the CI would fail in this PR. Can we do a release or point the version of sanic-testing to the repo?

ahopkins:
Ooo nice! Yeah, I'll do that.

sanic/server/websockets/impl.py
```diff
@@ -840,3 +845,8 @@ def connection_lost(self, exc):
         self.abort_pings()
         if self.connection_lost_waiter:
             self.connection_lost_waiter.set_result(None)
+
+    async def __aiter__(self):
+        with suppress(ConnectionClosedOK):
+            while True:
+                yield await self.recv()
```
ahopkins, line 852:
```python
try:
    something_bad()
except Exception:
    ...
```
This is almost always going to be faster than `suppress`.

tests/test_ws_handlers.py
```diff
@@ -0,0 +1,56 @@
...
+):
+    @app.websocket("/ws")
+    async def ws_echo_handler(request: Request, ws: Websocket):
+        while True:
+            msg = await ws.recv()
+            await ws.send(msg)
+
+    _, ws_proxy = app.test_client.websocket(
+        "/ws", mimic=simple_ws_mimic_client
+    )
```
ahopkins, line 38: :heart: I really like this new pattern.

sanic/server/websockets/impl.py
ChihweiLHBird: Okay, change it back to try except pair

Tronic: Benchmarked, yes `suppress` is slower by about 220 ns. Most definitely negligible compared to the opening of a websocket, or to even just the recv call, but still eight times the time it takes to make a simple function call with only try-except within that function and nothing else.

ahopkins (approved):

PR #2493, opened 2022-07-06
Respect error_format defined in app.config when None is given to app.route

ling7334:
If `None` is given on `@app.route`, it should respect the global setting instead of go to "auto" check directly.

### DIFF
--- a/sanic/mixins/routes.py
+++ b/sanic/mixins/routes.py
@@ -167,7 +167,7 @@ class RouteMixin(metaclass=SanicMeta):
                 # subprotocol is unordered, keep it unordered
                 subprotocols = frozenset(subprotocols)
 
-            if not error_format or error_format == "auto":
+            if error_format == "auto":
                 error_format = self._determine_error_format(handler)
 
             route = FutureRoute(


ahopkins:
Why? This would be a breaking change. Is there a use-case I'm missing?

ling7334:
when the endpoint's `error_format` is leave blank.
```python
@app.get("/")
def index(req):
    raise Exception()
```
and app.config.FALLBACK_ERROR_FORMAT is set to a specify format
```python
```
As a user, I would expect such error will format as I defined in the global setting, in this case, it's `json`, instead of `auto`.

This suppress `FALLBACK_ERROR_FORMAT` setting, and in no case `FALLBACK_ERROR_FORMAT` setting will be used.

But indeed its a breaking change.

ahopkins:
This sounds like another use case for `_default` as the default value rather than `None`.

ling7334:
A workaround is possible, catch the general exception and format as desired.

```python
def global_exception_handler(request: Request, exception: Exception) -> HTTPResponse:
    """
    Global exception handler
    """
    format = request.app.config.FALLBACK_ERROR_FORMAT
    if format in RENDERERS_BY_CONFIG:
        return RENDERERS_BY_CONFIG[format](request, exception, request.app.debug).render()
    else:
        # Default render as you desire
        return HTMLRenderer(request, exception, request.app.debug).render()
```
and register it to the app

```python
app.error_handler.add(Exception, global_exception_handler)
```


https://github.com/ling7334/sanic-scaffold/blob/74c63c4604d3f6d1b9d25f6500cf0808ba5d2485/setup.py#L150

prryplatypus:
@ling7334 are you by any chance calling `app.blueprint(...)` before updating your config?

ling7334:
> @ling7334 are you by any chance calling `app.blueprint(...)` before updating your config?

@prryplatypus Sorry for not mention the background.
I'm actually let Sanic load the config from environment.
I assume it will fetch the env vars once the Sanic is instantiate, before the blueprint was register.

prryplatypus:
All good, thanks! Was wondering if it was related at all to another issue someone mentioned in Discord :P

ahopkins:
Am I missing something?

```python
app = Sanic(__name__)
app.config.FALLBACK_ERROR_FORMAT = "text"


@app.get("/")
def index(req):
    raise Exception()
```

No matter what value I put in here it comes back as expected. Is there another use case to replicate the bad behavior?

prryplatypus:
Yeah that's what I remembered from using Sanic too. Felt odd when @ling7334 said otherwise, but maybe we're missing something 🤔

ling7334:
I'm kind of forget how it happend, closing for not reproduce.

Issue #2477, opened 2022-06-10
Sanic static directory fails when folder name ends with ".."


ekzhang:
**Describe the bug**

When serving a static file named `a../b.txt` with `app.static()`, the Sanic request handler gives a 400 status code instead of 200.

**Code snippet**

First file is called `main.py`.

```python
from sanic import Sanic

app = Sanic(name="sanic_test")

app.static('/static', './static_files')

if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8000)
```

Then, make the file in question:

```
mkdir -p static_files/a../
cat "hello world" > static_files/a../b.txt
```

And run the server with `python main.py`, followed by a GET request.

```sh-session
$ curl http://localhost:8000/static/a../b.txt
⚠️ 400 — Bad Request
====================
Invalid URL

```

**Expected behavior**

The `app.static()` handler should serve the `b.txt` file with a 200 response code.

**Environment (please complete the following information):**
 - OS: macOS and Linux
 - Version 22.3.0

ChihweiLHBird:
This has been solved in the previous PRs. Feel free to reopen if there is still an issue.

Tested and is no longer reproducible.

<img width="1042" alt="image" src="https://user-images.githubusercontent.com/43905414/200190383-9794c131-6f96-4c89-9b73-f05adce6a343.png">

PR #2495, opened 2022-07-10
Prevent directory traversion with static files

### LINKED ISSUE #2478: Sanic static handler allows parent ".." directory traversal
**Describe the bug**

The sanic static directory code checks for `../` as a substring of paths, but it also unquotes the path, which allows a malicious user to escape outside the static folder by using `..%2F`, where `%2F` is the URL-escaped version of `/`.

**Code snippet**

First, a basic server called `main.py`.

```python
from sanic import Sanic

app = Sanic(name="sanic_test")

app.static('/static', './static_files')

if __name__ == "__main__":
```

Then create a static file folder.
```
mkdir static_files
cat "hello world" > static_files/a.txt
```

Now run the server with `python3 main.py` and:

```sh-session
$ curl http://localhost:8000/static/a.txt
hello world
$ curl http://localhost:8000/static/..%2Fstatic_files/a.txt
hello world
```

This is very surprising behavior. From a security perspective it is not critical because sanic checks that the final resolved path has a prefix with the static directory, but this allows an attacker to expose information like the name of the static file folder.

Another case where this is dangerous is if you have a middleware that only allows a user to see certain subpaths like `/static/public/**` of the `/static/**` routes without authentication. Then, even without authentication, a user could visit a path like `/static/public/..%2F/private/secret_content.txt` and retrieve the contents of `/static/private/secret_content.txt`.

**Expected behavior**

Sanic should not allow parent directory traversal in static folders.

 - OS: macOS and Linux
 - Version 22.3.2

prryplatypus:
Fixes #2478 and (maybe?) #2477


sanic/mixins/routes.py
```diff
@@ -806,23 +806,21 @@ async def _static_request_handler(
...
         if __file_uri__:
+            # Strip all / that in the beginning of the URL to help prevent python
+            # from herping a derp and treating the uri as an absolute path
             unquoted_file_uri = unquote(__file_uri__).lstrip("/")
 
-            file_uri_pp = PurePath(unquoted_file_uri)
-            if any(part == ".." for part in file_uri_pp.parts):
+            segments = unquoted_file_uri.split("/")
+            if any(segment == ".." for segment in segments):
                 raise BadRequest("Invalid URL")
```
Tronic, line 818: I think we need to reject anything containing `\` as Invalid URL too. If left unchecked, Windows would treat it as path separator.

ChihweiLHBird:
Should we consider use async file APIs (`aiofiles` or `trio`) in this module?

sanic/mixins/routes.py
ChihweiLHBird: Only needed on win32 because Unix does support `\` character in the directory name.

Tronic: You could test for the OS but it would probably be faster (both in development and execution time) to just reject `if '\\' in unquoted_file_url` on all operating systems. It's not like anyone actually names their files with backslashes (or carriage returns, vertical tabs, alarm bells and various other strange characters that the UNIX standards permit).

ChihweiLHBird: We can do `any(os.sep in segment for segment in segments)`. 

`os.sep` will be `\\` on win32 and be `/` on Unix-like OSs.

ChihweiLHBird (dismissed):

sanic/mixins/routes.py
```diff
@@ -806,23 +805,23 @@ async def _static_request_handler(
...
-            file_path = path.join(
-                file_or_directory, sub("^[/]*", "", __file_uri__)
-            )
+            # Strip all / that in the beginning of the URL to help prevent
+            # python from herping a derp and treating the uri as an
+            # absolute path
+            unquoted_file_uri = unquote(__file_uri__).lstrip("/")
+
+            segments = unquoted_file_uri.split("/")
+            if ".." in segments or any(sep in segment for segment in segments):
```
ahopkins, line 818: we should benchmark this. Running two loops where only one is needed.

ChihweiLHBird: How about `if any(sep in segment or segment == ".." for segment in segments):`?

prryplatypus: That looks good to me, although my personal preference would be to place the `segment == ".."` check first, since it's quicker to check. However this is probably an unnecessary micro optimisation and either way works fine :P

ChihweiLHBird: @ahopkins Do you mean `sep in segment` was another loop or `".." in segments` was another loop?
@prryplatypus Yeah, that make sense.

ChihweiLHBird: I never had a chance to do the benchmark before. Is there any guide for it?

ChihweiLHBird (approved):

PR #2499, opened 2022-07-13
Sanic Server WorkerManager refactor

### LINKED ISSUE #1346: auto_reload support for Windows
This would be really useful for us. We don't have an option to use Linux for development unfortunately

### LINKED ISSUE #1471: Not properly handle signal to stop worker sub-process in debug mode
**Describe the bug**
In the debug mode, if you do not change your source code to trigger auto reload, the stop signals(SIGTERM, SIGINT) send to the main process will not work properly to stop the worker sub-process.  As the following code, run it and stop(Ctrl-C), the ``on_shutdown`` method will not be called.

```
# v18.12.0
from sanic import Sanic

app = Sanic(__name__)

@app.listener('after_server_stop')
def on_shutdown(app, loop):
    print('I am done!')

app.run(debug=True)
```

I think the problem maybe caused by the following code in ``reloader_helpers.py``:

```
def watchdog(sleep_interval):
    mtimes = {}
    worker_process = restart_with_reloader()
    signal.signal(
        signal.SIGTERM, lambda *args: kill_program_completly(worker_process)
    )
    signal.signal(
        signal.SIGINT, lambda *args: kill_program_completly(worker_process)
    )
```

Because of creating worker_process before register signal hander(``kill_program_completly``), the woker_process still use the default signal handler and do not kill sub-process woker.

The following snippet is a immature solution, just for reference:

```
def kill_program_completly(proc):
    if not proc:
        kill_process_children(os.getpid())
        os._exit(0)
    proc = proc.pop()
    kill_process_children(proc.pid)
    proc.terminate()
    os._exit(0)


def watchdog(sleep_interval):
    mtimes = {}
    worker_process = []
    signal.signal(
        signal.SIGTERM, lambda *args: kill_program_completly(worker_process)
    )
    signal.signal(
        signal.SIGINT, lambda *args: kill_program_completly(worker_process)
    )

    worker_process.append(restart_with_reloader())
    while True:
        for filename in _iter_module_files():
            try:
                mtime = os.stat(filename).st_mtime
            except OSError:
                continue

            old_time = mtimes.get(filename)
            if old_time is None:
                mtimes[filename] = mtime
                continue
            elif mtime > old_time:
                wp = worker_process.pop()
                kill_process_children(wp.pid)
                wp.terminate()
                worker_process.append(restart_with_reloader())
                mtimes[filename] = mtime
                break

        sleep(sleep_interval)

```

### LINKED ISSUE #2312: multiprocessing error when runs on macos big sur
**Describe the bug**

when sanic starts with multiple session, it crashes as :

```
objc[96314]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[96315]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[96314]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
objc[96315]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

**Code snippet**
    app.run(
        host="0.0.0.0",
        port=port,
        workers=sessions,
        register_sys_signals=True,
        protocol=WebSocketProtocol,
    )

**Expected behavior**
No exception, no crash

 - OS: macos big sur, apple m1
 - Version 

https://stackoverflow.com/questions/50168647/multiprocessing-causes-python-to-crash-and-gives-an-error-may-have-been-in-progr

### LINKED ISSUE #2364: RFC: Always run workers in a subprocess.
## Proposal: _Always_ run workers in a subprocess. 

Currently, when in `PRODUCTION` mode with `workers=1`, the server runs in the main process. This can lead to a discrepancy since the server(s) will be in a subprocess if:

- there are multiple workers
- `auto_reload` is enabled

We can create a more consistent experience if it will always be in a subprocess.

### LINKED ISSUE #2429: Manually triggered auto-reload
Hello fantastic community,

I am using vault and consul to dynamically configure my sanic application. I am trying to find a way to make it so everytime the vault secret is updated remotely, it kicks a reload of the sanic application. Very similar to what the autoreload feature is doing today.

I have a process running in another thread that loops through and makes sure that when credentials/secret expires, it reloads. I would like to be able to call a function to make the server reload, is that something that exists today?

Phil

### LINKED ISSUE #2467: Unix sockets are not always passed to worker processes
**Describe the bug**
If the server configuration includes the "unix" parameter to specify the path to a unix socket to open in the server, the socket is opened in the parent process but then not passed as the "sock" parameter to the worker processes.

**Code snippet**
Relevant source code, make sure to remove what is not necessary.

**Expected behavior**
Passing the unix parameter should result in the socket being opened in the parent process and then passed as an open file descriptor using the "sock" parameter to workers just like TCP/IP sockets are passed.

 - OS: Ubuntu linix
 - 22

**Additional context**
I believe it is simply a matter of a missing line here:
https://github.com/sanic-org/sanic/blob/4ee2e57ec8474cf7c9a4f018de97926d8899f7cb/sanic/server/runners.py#L237
of doing the same thing that the TCP case does
```
        server_settings["unix"] = None
        server_settings["sock"] = sock
```

### LINKED ISSUE #2494: Remove app from `_app_registry` on app.stop()
**Describe the bug**
The `_app_registry` keeps hold of any app created by Sanic, even those stopped.

**Code snippet**
Adding the following code resolve the issue:

```
    def stop(self):
        """
        This kills the Sanic
        """
        if self.state.stage is not ServerStage.STOPPED:
            self.shutdown_tasks(timeout=0)
            for task in all_tasks():
                with suppress(AttributeError):
                    if task.get_name() == "RunServer":
                        task.cancel()
            get_event_loop().stop()

            # Added part:
            del self.__class__._app_registry[self.name]
```

**Additional context**
Implementing some advanced auto reload system causes issues when fully restarting an application.

(Restarting only the server (calling `app.run()`) won't take into consideration any changes such as the environment variables.)

### LINKED ISSUE #2534: Inconsistent access logs
There is inconsistent behavior when starting from CLI and from script in how access logs perform. This is related to `--access-log` always returning a boolean and not a `None` value.

By default, access logs should only be displayed when running in DEBUG unless otherwise specified.

Ref: https://github.com/sanic-org/sanic/pull/2237#issuecomment-926205582

ahopkins:
_NOTE: When I say "Sanic" I mean Sanic Server. This PR is inapplicable to ASGI mode._

## Background

The long overdue refactor of Sanic multiprocessing...

This PR intends to redo how Sanic creates processes and sockets. As a side effect, it will also change how auto-reload works, and perhaps most importantly finally fix serving multiple workers from Windows. It will also expose a new API for manually triggering restarts, accessing the current worker process context, and passing objects to the processes like `multiprocessing.Queue` that will enable sync between single-node, multi-worker instances.

## Design concept

image

As discussed in #2364, there will _always_* be a main process when you start Sanic. The main process will create a `WorkerManger` that is responsible for the lifecycle of one or more server processes, and optionally a reload process. This pattern will also be extensible by API to allow Sanic Extension and other plugins to piggyback off the manager for example to create a health-check process.

\* *When I say __always__, I really mean "sometimes". We will also create a secondary `Sanic.single_serve()` method that creates and runs the server in one process. No auto-reload or multi-worker support. You will need to go out of your way to use this for those people that need it. OOTB, we want to make a consistent experience between DEBUG and PROD.*

Once this is complete, we can make further enhancements to `Sanic.prepare`. Currently if you prepare multiple HTTP versions and/or socket bindings, they will run in each process on the same loop. We can make this better to allow HTTP servers instances to have their own loops inside individual processes. _This is **not** a part of this PR and will likely be an addition in v22.12._

One or more sockets will be created by the Main process and passed to each worker process (as needed). Because most of this implementation will be all new, it should be pretty easy to maintain backwards compatibility by taking the existing `Sanic.serve` and moving it to `Sanic.serve_legacy`.

## New API

- `Sanic.serve_legacy` - Use the old version of `Sanic.serve`
- `Sanic.serve_single` - Run Sanic w/o multiprocessing (no auto-reload allowed)
- `app.run(..., single_process=True)`
- `app.run(..., legacy=True)`
- `app.m.restart()` - Manually restart an application
- `app.shared_ctx` - see below
- `$ sanic path.to:app --inspect` - CLI command to inspect running application
- `app.manager.manage(...)` - see below

### Shared state

For sharing _safe_ values between workers (on the same Sanic runtime), there is a new `app.shared_ctx`. This will be an object explicitly for things like `multiprocessing` sync and shared state objects (`Queue`, `Pipe`, `Value`, `Array`, etc). They should **ONLY** be added in the main process like this:

```python
@app.main_process_start
async def test(app, _):
    pipe = Pipe()
    app.shared_ctx.value = Value("i", 0, lock=False)  # OK
    app.shared_ctx.pipe = pipe                        # OK
    app.shared_ctx.pipe_side = pipe[0]                # OK
    app.shared_ctx.queue = Queue()                    # OK
    app.shared_ctx.unsafe = {}                        # raises warning
    app.shared_ctx.also_unsafe = (0,)                 # raises warning
```

### Process management

A user can hook into the Sanic worker manager to allow it to manage any additional subprocess. All it needs is a callable. If that subprocess will be blocking, then it should also handle some common signals.

```python
from signal import SIGINT, SIGTERM
from signal import signal as signal_func

def my_process(*args, **kwargs):
    run = True

    def stop(*_):
        nonlocal run
        run = False

    signal_func(SIGINT, stop)
    signal_func(SIGTERM, stop)

    while run:
        print(".")
        sleep(1)

@app.main_process_ready
async def ready(app: Sanic, _):
    app.manager.manage("MyProcess", my_process, {"foo": "bar"})
```

Sanic will now startup the process and manage its lifecycle.

### Worker state

There is an object available on the application multiplexer called "state":

```python
app.m.state
```

It is a dictionary-like object containing basic details about the current state of the *current* worker process:

```python
{
    'server': True,
    'state': 'ACKED',
    'pid': 1386076,
    'start_at': datetime.datetime(2022, 9, 14, 8, 9, 44, 651607, tzinfo=datetime.timezone.utc),
    'starts': 1
}
```

This object has some basic operations to interact with it like a dictionary:

```python
request.app.m.state["arbitrary"] = 123
request.app.m.state.update({"foo": "bar"})
```

*It should be noted that the multiplexer object is only available inside of a server worker process.*

### Inspector

There is a special worker process called the "Inspector". You must opt-in to use it:

```python
app.config.INSPECTOR = True
```

This will open a local port that is exposed to allow an outside process to get information and interact with the running application. In the below screenshot, we execute a CLI command to get the state of the current instance:

image

Other commands:

```
--inspect              Inspect the state of a running instance, human readable
--inspect-raw          Inspect the state of a running instance, JSON output
--trigger-reload       Trigger worker processes to reload
--trigger-shutdown     Trigger all processes to shutdown
```

## Breaking Changes

- Setting `app.config` values in `main_process_start` will have no impact. (_This is a sort of unintended side effect. Probably more a bug that should not have been allowed. Nonetheless, it is no longer possible and may be a breaking change for some_).
- `app.run()` or `Sanic.serve()` must be inside of a `if __name__ == "__main__"` block

## Impacted Issues

If you know of some other Issues that this touches not listed below, please LMK.

Closes #2534
Closes #2494  
Closes #2467
Closes #2429
Closes #2364
Closes #2312
Closes #1471
Closes #1346

## TODO
- [x] Windows socket binding pattern
- ~Do not pre-bind socket in legacy mode~
- [x] Unit tests
- [x] Move loop setup to global scope
- [x] Expose API for passing args to processes
- [x] Signal for reloader
- [x] Worker introspection
- [x] Move `health` module to Sanic Extensions

Technical discussion and overview: https://youtu.be/m8HCO8NK7HE


ahopkins:
FYI - If you are looking at this PR an see the `sanic.worker.health` module, I plan to pull this out and move it to Sanic Extensions as an optional add-on. If you are looking at the PR and do not see it, then I already did.

ahopkins:
See https://github.com/sanic-org/sanic-ext/pull/111 for health monitoring implementation

sanic/types/shared_ctx.py
```diff
@@ -0,0 +1,55 @@
...
+            module.startswith(prefix)
+            for prefix in ("multiprocessing", "ctypes")
+        ):
+            error_logger.warning(
+                f"{Colors.YELLOW}Unsafe object {Colors.PURPLE}{name} "
+                f"{Colors.YELLOW}with type {Colors.PURPLE}{type(value)} "
+                f"{Colors.YELLOW}was added to shared_ctx. It may not "
+                "not function as intended. Consider using the regular "
+                f"ctx. For more information, please see ____.{Colors.END}"
+            )
```
ahopkins, line 48: Not sure of a better way to do this. If anyone has some thoughts, happy to hear.

sjsadowski:
In general, I'm giving this my +1. There's a lot to chew through here, but I understand the value. I'm hoping that @sanic-org/core-devs can review, and those who are better developers and pythonistas than I can weigh in.

prryplatypus:
Would you be able to walk us through this PR some day on Discord, @ahopkins? I don’t know where to start with it :S

ahopkins:
> Would you be able to walk us through this PR some day on Discord, @ahopkins? I don’t know where to start with it :S

Before it is merged or after?

I was thinking about doing a release event with a screen share when I build the release and walk thru some new features. Need to figure out if I should do it on Discord, YT, or something else.

prryplatypus:
> Before it is merged or after?
> 
> I was thinking about doing a release event with a screen share when I build the release and walk thru some new features. Need to figure out if I should do it on Discord, YT, or something else.

I was thinking before! Kinda like peer-reviewing, so I (and hopefully others too) can add our pair of eyes to it :)

ahopkins:
> I was thinking before! Kinda like peer-reviewing, so I (and hopefully others too) can add our pair of eyes to it :)

https://youtu.be/m8HCO8NK7HE

sjsadowski:
I'm putting my chop on this. I was satisfied with the walkthrough and the Q&A, and I think with tests passing as best as can be expected, I'm ready for this to be merged.

sjsadowski (approved):

sanic/types/shared_ctx.py
sjsadowski: I think this is fine for now and don't think we need to block the merge for a possible better way.

ahopkins:
@sjsadowski Nice. I am going to merge, which will unblock a few other items I want to get done this week.

sanic/types/shared_ctx.py
ahopkins: I did play with the whitelist idea, but it really gets messy and out of control quick. Just take a look at what the `multiprocessing` module is doing and I think you will see why that seems like a bad idea.

https://github.com/python/cpython/blob/main/Lib/multiprocessing/__init__.py#L22
https://github.com/python/cpython/blob/main/Lib/multiprocessing/managers.py#L1211

cnicodeme:
Do you have a date for this release? I'm curious and eager to move it to prod :)

ahopkins:
> Do you have a date for this release? I'm curious and eager to move it to prod :)

Likely Sunday (2022-09-26).

cnicodeme:
Grea, thank you :)

PR #2503, opened 2022-07-20
Update exceptions.py

liuguangjiu:
Supports the latest version of sanIC_Restful_API


ahopkins:
This feature was deprecated in v21.6 and removed in 21.12.

https://github.com/sanic-org/sanic/pull/2077/files

https://sanic.dev/en/guide/release-notes/v21.12.html#removals

If that package needs this functionality, perhaps it should add it on its own.

PR #2504, opened 2022-07-20
Fix Test Cases: `test_http` for Py3.9+, `test_json_response_json` for ujson 5.4.0+, and `test_zero_downtime`; Test Case Type Annotations

ChihweiLHBird:
The test case was fixed.

### DIFF
--- a/tests/test_json_encoding.py
+++ b/tests/test_json_encoding.py
@@ -3,11 +3,10 @@ import sys
 from dataclasses import asdict, dataclass
 from functools import partial
 from json import dumps as sdumps
+from string import ascii_lowercase
 from typing import Dict
 
 import pytest
-
-from packaging import version
 
 
 try:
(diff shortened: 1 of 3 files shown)


tests/test_json_encoding.py
```diff
@@ -3,11 +3,16 @@
 from dataclasses import asdict, dataclass
 from functools import partial
 from json import dumps as sdumps
+from typing import Dict
 
 import pytest
 
+from packaging import version
```
ahopkins, line 10: No need to add a dependency. Just parse the version by hand. We do it in a couple of places.

ahopkins (changes requested):

tests/test_json_encoding.py
ChihweiLHBird: `packaging` is used by setuptools so I thought it is everywhere. And sure, I can manually parse it.

tests/test_json_encoding.py
```diff
@@ -75,7 +84,13 @@ def test_json_response_ujson(payload):
         json(payload)
 
 
-@pytest.mark.skipif(NO_UJSON is True, reason="ujson not installed")
+@pytest.mark.skipif(
```
ahopkins, line 87: We are going need to rethink this test. But will merge it for now.

tests/test_unix_socket.py
```diff
@@ -162,11 +166,11 @@ async def client(app, loop):
 app_multi = Sanic(name="test")
 
 
-def handler(request):
+def handler(request: Request):
     return text(f"{request.conn_info.server}")
 
 
-async def client(app, loop):
+async def client(app: Sanic, loop: AbstractEventLoop):
```
ahopkins, line 173: There really is no reason to start annotating all of these. Especially those embedded handlers.

ahopkins (approved):

PR #2506, opened 2022-07-28
Use pathlib for path resolution

ahopkins:
Piggyback off #2495 to use `pathlib`. This way we do not need to do any of the path calculations.

### DIFF
--- a/sanic/mixins/routes.py
+++ b/sanic/mixins/routes.py
@@ -26,12 +26,7 @@ from sanic.base.meta import SanicMeta
 from sanic.compat import stat_async
 from sanic.constants import DEFAULT_HTTP_CONTENT_TYPE, HTTP_METHODS
 from sanic.errorpages import RESPONSE_MAPPING
-from sanic.exceptions import (
-    BadRequest,
-    FileNotFound,
-    HeaderNotFound,
-    RangeNotSatisfiable,
-)
+from sanic.exceptions import FileNotFound, HeaderNotFound, RangeNotSatisfiable
 from sanic.handlers import ContentRangeHandler
 from sanic.log import error_logger
 from sanic.models.futures import FutureRoute, FutureStatic
@@ -808,6 +803,11 @@ class RouteMixin(metaclass=SanicMeta):
         # Merge served directory and requested file if provided
         file_path_raw = Path(unquote(file_or_directory))
         root_path = file_path = file_path_raw.resolve()
+        not_found = FileNotFound(
+            "File not found",
+            path=file_or_directory,
+            relative_url=__file_uri__,
+        )
 
         if __file_uri__:
             # Strip all / that in the beginning of the URL to help prevent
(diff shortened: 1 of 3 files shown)


seemethere (dismissed):

ahopkins:
Thanks @seemethere. Hope you've been well. :sunglasses: 

@Tronic @ChihweiLHBird @prryplatypus I changed the pattern to use `pathlib` so that we can rely upon the stdlib for all the various path resolution `Path().resolve()`. Also, I think we need `404` for all of the exceptions. A `400` could leak information.

ChihweiLHBird:
Migrate to pathlib is awesome!

sanic/mixins/routes.py
```diff
@@ -806,31 +801,39 @@ async def _static_request_handler(
...
         if __file_uri__:
             # Strip all / that in the beginning of the URL to help prevent
             # python from herping a derp and treating the uri as an
             # absolute path
             unquoted_file_uri = unquote(__file_uri__).lstrip("/")
+            file_path_raw = Path(file_or_directory, unquoted_file_uri)
+            file_path = file_path_raw.resolve()
+            if (
+                file_path < root_path and not file_path_raw.is_symlink()
+            ) or file_path_raw.match("../**/*"):
```
ChihweiLHBird, line 821: Is there any case that `file_path_raw.is_symlink()` is `False` and `file_path_raw.match("../**/*")` is `True`?

sanic/mixins/routes.py
```diff
@@ -806,31 +801,39 @@ async def _static_request_handler(
...
-                f"File not found: path={file_or_directory}, "
-                f"relative_url={__file_uri__}"
-            )
-            raise FileNotFound(
-                "File not found",
-                path=file_or_directory,
-                relative_url=__file_uri__,
-            )
+        try:
+            file_path.relative_to(root_path)
```
ChihweiLHBird, line 829: Can we use [`is_relative_to`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to) instead of catching the error on `relative_to`?

sanic/mixins/routes.py
```diff
@@ -806,31 +801,39 @@ async def _static_request_handler(
...
+        except ValueError:
+            if not file_path_raw.is_symlink():
```
ChihweiLHBird, line 831: What happens when `file_path_raw.is_symlink()` is `True`? I don't quite understand why we needed this...

sanic/mixins/routes.py
ahopkins: That's Python 3.9+

sanic/mixins/routes.py
ahopkins: If you have a symlink, once it is resolved, it could be a totally different part of the file system, so it might not be `relative_to`

sanic/mixins/routes.py
ChihweiLHBird: Oh, okay

ChihweiLHBird (approved):

PR #2508, opened 2022-07-31
Use path.parts instead of match

ahopkins:
Small change to improve this logic.


ChihweiLHBird:
Is it faster than `match`?

ChihweiLHBird (approved):

ahopkins:
Yes

PR #2513, opened 2022-07-31
Better request cancel handling

ahopkins:
Currently, there is no way to distinguish between the following:

**SCENARIO A: Something in the handler raises a `CancelledError`**
```python
@app.get("/")
    raise CancelledError
```

**SCENARIO B: A client cancels a request in the middle of handling**
```python
@app.get("/")
    for _ in range(10):
        print(".")
        await sleep(1)
    return json(None)
```

This is particularly noticeable and troublesome if there is an underlying library (like a DB driver) that decides to call `CancelledError`. It causes a 503 and there really is no way to catch this. The following will *NOT* work, which is somewhat counterintuitive.

```python
@app.exception(CancelledError)
async def exc_handler(request: Request, exc):
    ...
```

---

The idea here is to introduce a new exception: `RequestCancelled` that will bubble up through the system. It should be catchable on its own, and also `CancelledError` should also be catchable. Furthermore, it should not create a response object and attempt to push the bytes if the client connection has been lost.


ChihweiLHBird:
It looks good to me.

ahopkins:
I need to add a few more unit tests.

ChihweiLHBird (dismissed):

sanic/http/http1.py
```diff
@@ -132,15 +132,18 @@ async def http1(self):
...
                     logger.info(
                         f"Request: {self.request.method} {self.request.url} "
                         "stopped. Transport is closed."
                     )
                     return
-                e = self.exception or ServiceUnavailable("Cancelled")
+                if self.protocol.conn_info.lost:
+                    e = RequestCancelled
+                else:
+                    e = self.exception or exc
```
Tronic, line 146: Should that be RequestCancelled(...) rather than bare type? Might also prefer expression conditional:
```python
e = RequestCancelled() if self.protocol.conn_info.lost else self.exception or exc
```

tests/test_cancellederror.py
```diff
@@ -0,0 +1,21 @@
...
+    @app.get("/")
+    async def handler(request: Request):
+        raise CancelledError("STOP!!")
+
+    @app.exception(CancelledError)
+    async def handle_cancel(request: Request, exc: CancelledError):
+        return json({"message": exc.args[0]}, status=418)
+
+    _, response = app.test_client.get("/")
+    assert response.status == 418
```
Tronic, line 20: I am a Teapot?

sanic/http/http1.py
```diff
@@ -132,15 +132,18 @@ async def http1(self):
 
                 if self.stage is Stage.RESPONSE:
                     await self.response.send(end_stream=True)
-            except CancelledError:
+            except (CancelledError, RequestCancelled) as exc:
```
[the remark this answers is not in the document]
ahopkins: Yes, was just an explicit visual indicator.

sanic/http/http1.py
ahopkins: I usually try to avoid multiple branches in a single expression. But that's fine with me.

Tronic (approved):

PR #2521, opened 2022-08-07
Revise middleware handling

ahopkins:
This PR aims to:

- introduce an API to control middleware ordering
- performance boost

```python
@app.on_request(priority=99)
async def high_priority(request):
    ...
```

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -701,7 +701,10 @@ class Sanic(BaseSanic, RunnerMixin, metaclass=TouchUpMeta):
     # -------------------------------------------------------------------- #
 
     async def handle_exception(
-        self, request: Request, exception: BaseException
+        self,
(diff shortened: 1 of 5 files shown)


sanic/app.py
```diff
@@ -750,9 +754,12 @@ async def handle_exception(
         # -------------------------------------------- #
         # Request Middleware
         # -------------------------------------------- #
-        response = await self._run_request_middleware(
-            request, request_name=None
-        )
+        if (
+            run_middleware
+            and request.route
+            and request.route.extra.request_middleware
```
ahopkins, line 760: If there is no route, then we need to run the global middleware

ahopkins:
A lot more use cases to consider with some potential breaking changes. Will be safer to kick this one out until after LTS.

PR #2526, opened 2022-08-11
Validate File When Requested

ChihweiLHBird:
Return status code of 304 when the conditions are satisfied.

### DIFF
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -322,21 +322,19 @@ def html(
 async def validate_file(
     request_headers: Header, last_modified: Union[datetime, float, int]
 ):
-    try:
-
-        if_modified_since = request_headers.get(
-            "If-Modified-Since",
-            request_headers.get("if-modified-since"),
-        )
-        if_modified_since = parsedate_to_datetime(if_modified_since)
-        if not isinstance(last_modified, datetime):
-            last_modified = datetime.fromtimestamp(
-                float(last_modified), tz=timezone.utc
-            ).replace(microsecond=0)
-        if last_modified <= if_modified_since:
-            return HTTPResponse(status=304)
-    except Exception:
-        pass
+    if_modified_since = request_headers.get(
+        "If-Modified-Since",
+        request_headers.get("if-modified-since"),
+    )
+    if not if_modified_since:
+        return
+    if_modified_since = parsedate_to_datetime(if_modified_since)
+    if not isinstance(last_modified, datetime):
+        last_modified = datetime.fromtimestamp(
+            float(last_modified), tz=timezone.utc
+        ).replace(microsecond=0)
+    if last_modified <= if_modified_since:
+        return HTTPResponse(status=304)
 
 
 async def file(
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -335,7 +335,7 @@ async def validate_file(
             ).replace(microsecond=0)
         if last_modified <= if_modified_since:
             return HTTPResponse(status=304)
-    except:
+    except Exception:
         pass
 
 
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -338,7 +338,7 @@ async def file(
     location: Union[str, PurePath],
     status: int = 200,
     request_headers: Optional[Header] = None,
-    validate_when_requested: bool = True,
+    validate_when_requested: bool = False,
     mime_type: Optional[str] = None,
     headers: Optional[Dict[str, str]] = None,
     filename: Optional[str] = None,
--- a/sanic/response.py
+++ b/sanic/response.py
@@ -322,10 +322,14 @@ def html(
 async def validate_file(
     request_headers: Header, last_modified: Union[datetime, float, int]
 ):
-    if_modified_since = request_headers.get("If-Modified-Since")
-    if not if_modified_since:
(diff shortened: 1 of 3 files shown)


sanic/response.py
```diff
@@ -319,9 +319,31 @@ def html(
...
 
+async def validate_file(
+    request_headers: Header, last_modified: Union[datetime, float, int]
+):
+    try:
+        if_modified_since = parsedate_to_datetime(
+            request_headers.get(
+                "If-Modified-Since",
+                request_headers.get("if-modified-since"),
+            )
```
ahopkins, line 330: Why the redundancy?

ChihweiLHBird: Because it could be in lower case. Our dict is not case insensitive, am I right?

ahopkins: it is this: https://multidict.readthedocs.io/en/stable/multidict.html#cimultidict

ChihweiLHBird: Ah, that's nice!

sanic/response.py
```diff
@@ -319,9 +319,26 @@ def html(
...
+        ).replace(microsecond=0)
+    if last_modified <= if_modified_since:
+        return HTTPResponse(status=304)
+
+
 async def file(
     location: Union[str, PurePath],
     status: int = 200,
+    request_headers: Optional[Header] = None,
+    validate_when_requested: bool = True,
```
ahopkins, line 341: Should this default to true? It's a breaking change, no?

ChihweiLHBird: Yes, kind of? Maybe make it to be False for now? And just realized that I forgot to add a doc for the added parameters 😂. Will add them soon.

ChihweiLHBird: And it is how flask and our static files serving work. So still a reasonable behavior?

ChihweiLHBird: Changed to `False`. The default behavior should remain same as previous.

ChihweiLHBird:
I am going to improve the test case a little bit.

sanic/response.py
```diff
@@ -319,9 +319,26 @@ def html(
     )
 
 
+async def validate_file(
+    request_headers: Header, last_modified: Union[datetime, float, int]
+):
+    if_modified_since = request_headers.get("If-Modified-Since")
+    if not if_modified_since:
+        return
+    if_modified_since = parsedate_to_datetime(if_modified_since)
```
ahopkins, line 328: We should catch a potential `ValueError` here and respond appropriately.

sanic/response.py
```diff
@@ -319,9 +319,26 @@ def html(
     )
 
 
+async def validate_file(
+    request_headers: Header, last_modified: Union[datetime, float, int]
+):
+    if_modified_since = request_headers.get("If-Modified-Since")
```
ahopkins, line 325: What about `If-None-Match` support?

sanic/response.py
```diff
@@ -319,9 +319,26 @@ def html(
...
+        ).replace(microsecond=0)
+    if last_modified <= if_modified_since:
+        return HTTPResponse(status=304)
+
+
 async def file(
     location: Union[str, PurePath],
     status: int = 200,
+    request_headers: Optional[Header] = None,
+    validate_when_requested: bool = False,
```
ahopkins, line 341: I think I am going back on this. This probably should be enabled by default.

ahopkins (changes requested):

sanic/response.py
ChihweiLHBird: Thank you! Fixed.

sanic/response.py
ChihweiLHBird: Response 400? `ValueError` only for Python 3.10, and before 3.10, it was `TypeError`. I am going to try to handle both exceptions with a 400 response.

sanic/response.py
ChihweiLHBird: We don't have eTag for both static files and `file` response function at this time. So I would suggest doing this in another PR. By the way, may I try to combine the logics that handles validating files for static files and the `file` response function in order to reduce duplicated functionality of code?

sanic/response.py
ChihweiLHBird: Or maybe just ignore the invalid value and return 200?

ChihweiLHBird: Just added try except to ignore invalid If-Modified-Since header. This would be considered as solved.

sanic/response.py
```diff
@@ -319,9 +319,30 @@ def html(
...
+async def validate_file(
+    request_headers: Header, last_modified: Union[datetime, float, int]
+):
+    try:
+        if_modified_since = request_headers.getone("If-Modified-Since")
+    except KeyError:
+        return
+    try:
+        if_modified_since = parsedate_to_datetime(if_modified_since)
+    except (TypeError, ValueError):
```
ahopkins, line 331: Let's add:

```suggestion
    except (TypeError, ValueError):
		logger.warning(
			"Ignorning invalid If-Modified-Since header received: "
            "'%s'",
            if_modified_since
        )
```

ChihweiLHBird: Thanks!

ahopkins (approved):

Issue #2529, opened 2022-08-18
`tests` folder is included in wheel file


huntzhan:
**Describe the bug**

The `tests` folder is included in wheel file and hence will be installed into the `site-packages`. This could break the user's dev environment (and did break our test toolkit) since importing `tests`actually imports `tests` of sanic.

**Code snippet**

![WeChatWorkScreenshot_b78c284b-bc10-498e-bf7f-d7683fcd4896](https://user-images.githubusercontent.com/5213906/185287465-672ccbdb-aea8-4a59-8705-2d15552b77e6.png)

**Expected behavior**

Remove the `tests` folder from wheel.

Passing value to `exclude` or `include` of `find_packages` should work:
https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

**Environment (please complete the following information):**

 - OS: Linux & macOS
 - Sanic Version: 22.6.2

**Additional context**

ahopkins:
Thanks for bringing this up! Can you purpose a PR?

huntzhan:
PR purposed.

PR #2530, opened 2022-08-18
Resolve issue 2529

huntzhan:
#2529

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup_kwargs = {
         "Build fast. Run fast."
     ),
     "long_description": long_description,
-    "packages": find_packages(),
+    "packages": find_packages(exclude=("tests", "tests.*")),
     "package_data": {"sanic": ["py.typed"]},
     "platforms": "any",
     "python_requires": ">=3.7",


setup.py
```diff
@@ -61,7 +61,7 @@ def open_local(paths, mode="r", encoding="utf8"):
         "Build fast. Run fast."
     ),
     "long_description": long_description,
-    "packages": find_packages(),
```
ahopkins, line 64: None of the other dirs get added, correct?

setup.py
```diff
@@ -61,7 +61,7 @@ def open_local(paths, mode="r", encoding="utf8"):
...
+    "packages": find_packages(exclude=("tests", "tests.*")),
```
ahopkins, line 64: None of the other dirs get added, correct?

ahopkins (approved):

setup.py
huntzhan: Yeah, for now only `tests` and `sanic` are included in wheel distribution. This PR removes `tests` and leaving `sanic` untouched.

PR #2533, opened 2022-08-20
Refactor `_static_request_handler`

ChihweiLHBird:
1. Wrap logics that processes file path to a method.
2. Use built in `update` method to merge two dictionaries.
3. Fixed an issue that when `if-modified-since` in request header is larger than the file modified time, 200 is still responded. 304 might be more appropriate.


ahopkins:
Not sure that this adds much.

ChihweiLHBird:
> Not sure that this adds much.

Just added some, lol

Tronic:
Seems useful to me and it doesn't add much code, so that is also good. Header dict `update` is *much* faster than Python loop. Should keep an eye out for such loops that are better done with C builtins (perhaps there are static analyser tools that could pick up such patterns in the whole codebase?).

ChihweiLHBird:
> Seems useful to me and it doesn't add much code, so that is also good. Header dict `update` is _much_ faster than Python loop. Should keep an eye out for such loops that are better done with C builtins (perhaps there are static analyser tools that could pick up such patterns in the whole codebase?).

Thanks for the feedback. I never heard about a such tool. Let me do a search for it.

sanic/mixins/routes.py
```diff
@@ -790,24 +790,9 @@ def _generate_name(self, *objects) -> str:
...
-        file_or_directory,
-        use_modified_since,
-        use_content_range,
-        stream_large_files,
-        request,
-        content_type=None,
-        __file_uri__=None,
-    ):
-        # Merge served directory and requested file if provided
+    async def get_file_path(self, file_or_directory, __file_uri__, not_found):
```
ahopkins, line 793: The dunder seems odd here. I see it is in the handler args, but still seems out of place. Maybe out of scope for this PR to change.

ahopkins, line 793:
```suggestion
    async def _get_file_path(self, file_or_directory, __file_uri__, not_found):
```

sanic/mixins/routes.py
```diff
@@ -834,22 +819,43 @@ async def _static_request_handler(
...
+        __file_uri__=None,
+    ):
+        not_found = FileNotFound(
+            "File not found",
+            path=file_or_directory,
+            relative_url=__file_uri__,
+        )
+
+        # Merge served directory and requested file if provided
+        file_path = await self.get_file_path(
```
ahopkins, line 841:
```suggestion
        file_path = await self._get_file_path(
```

ahopkins (approved):

ahopkins (approved):

PR #2536, opened 2022-08-26
Improve and fix some type annotations

monosans:
According to [typing.Tuple](https://docs.python.org/3/library/typing.html#typing.Tuple) docs: `To specify a variable-length tuple of homogeneous type, use literal ellipsis, e.g. Tuple[int, ...]`.

All of the cases I fixed involve the use of variable-length tuples.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -458,9 +458,7 @@ class Sanic(BaseSanic, RunnerMixin, metaclass=TouchUpMeta):
 
     def blueprint(
         self,
-        blueprint: Union[
-            Blueprint, List[Blueprint], Tuple[Blueprint, ...], BlueprintGroup
-        ],
+        blueprint: Union[Blueprint, Iterable[Blueprint], BlueprintGroup],
         **options: Any,
     ):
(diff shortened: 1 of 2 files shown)


sanic/app.py
```diff
@@ -459,7 +459,7 @@ def enable_websocket(self, enable=True):
     def blueprint(
         self,
         blueprint: Union[
-            Blueprint, List[Blueprint], Tuple[Blueprint], BlueprintGroup
+            Blueprint, List[Blueprint], Tuple[Blueprint, ...], BlueprintGroup
```
ahopkins, line 462: We probably could collapse to sequence. IIRC, we just loop thru them, so a set and a frozenset would also be valid.

monosans: Sequence is not suitable for set and frozenset, so I replaced Union[list, tuple] with Iterable. This will allow the use of any object that returns a Blueprint when iterated, including set and frozenset.

seemethere (approved):

ahopkins (approved):

PR #2540, opened 2022-09-11
Add signals before and after handler execution

ahopkins:
Adds two new signals:
- `http.handler.before`
- `http.handler.after`

The purpose of these is to add a sort of middleware alternative that runs before and after the handler is called. `http.handler.before` will always run after request middleware. `http.handler.after` runs after the handler has completed. But that may or may not be after response middleware.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -896,7 +896,17 @@ class Sanic(BaseSanic, RunnerMixin, metaclass=TouchUpMeta):
                     )
 
                 # Run response handler
+                await self.dispatch(
+                    "http.handler.before",
+                    inline=True,
+                    context={"request": request},
+                )
                 response = handler(request, **request.match_info)
+                await self.dispatch(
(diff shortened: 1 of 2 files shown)


sanic/app.py
```diff
@@ -896,7 +896,17 @@ async def handle_request(self, request: Request):  # no cov
...
+                    "http.handler.before",
+                    inline=True,
+                    context={"request": request},
+                )
                 response = handler(request, **request.match_info)
+                await self.dispatch(
+                    "http.handler.after",
+                    inline=True,
+                    context={"request": request},
+                )
```
prryplatypus, line 909: Wouldn't this be better after awaiting the handler (in those cases in which it is awaitable)?

ahopkins: Yes. Good catch.

Tronic (approved):

PR #2545, opened 2022-09-14
Standardize init of exceptions

### LINKED ISSUE #2601: SanicException.message is always None

### Describe the bug

Raising SanicException with a custom message does not assign it to the message attribute. It's always empty.

### Code snippet

```
import logging

from sanic import Sanic
from sanic.exceptions import Forbidden, NotFound, SanicException

app = Sanic(__name__)


@app.route('/unauthorized')
async def unauthorized(request: Request):
    raise SanicException(message="Unauthorized!!!", status_code=401)


@app.route('/forbidden')
async def forbidden(request: Request):
    raise Forbidden(message="Forbidden!!!")


@app.route('/not-found')
async def not_found(request: Request):
    raise NotFound()


@app.exception(SanicException)
async def error_handler(req: Request, err: SanicException):
    logging.error(f"err: {err.message}", exc_info=err)
    return json({"response_code": err.status_code, "error": err.message})


if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8080, dev=True)
```

### Expected Behavior

_No response_

### How do you run Sanic?

ASGI

### Operating System

Mac

### Sanic Version

22.9.1

### Additional context

_No response_

ahopkins:
Needs some more cleanup. But this is an attempt to make raising exceptions a more consistent experience. It will be breaking however. There are a few existing exceptions that define their own `__init__` with a signature that does not match. To the extent we can preserve those orders, I will. But, I think it will be best to convert the other arguments (`status_code`, `extra`, `context`, `headers`) to keyword.

Closes #2601


ahopkins:
I am a bit torn.

On the one hand: we have exceptions that are a bit of a mess. There is not a consistent `__init__`, and properties are not guaranteed to be there.

On the other: exceptions are very fundamental. Changing them will break for users.

---

I thought about ways to solve this:

1. monkeypatch until we can deprecate
2. providing a legacy and a new module

Neither of these solutions will work, and have their own problems. Anyone else have some thoughts?

Right now I am leaning towards fixing what we can, and leaving the breaking changes (a uniform constructor) for another time.

Tronic:
I generally stand with "break it now to fix all of the future" (as opposed to keeping broken APIs for compatibility). But yes, at a bare minimum one should be able to write code that then works both with old and new Sanic versions, even if it requires changes to some old code.

If it is about kwargs' names being changed, I suppose both names could be supported on init functions as a deprecation path. Removal of positional args is harder but at least one can then always fix the code to use kwargs instead and it will work with older Sanic too. Not a big fan of either one of the two solutions that you suggest, as both would increase complexity and confusion quite a bit.

ahopkins:
> If it is about kwargs' names being changed, I suppose both names could be supported on init functions as a deprecation path. Removal of positional args is harder but at least one can then always fix the code to use kwargs instead and it will work with older Sanic too. Not a big fan of either one of the two solutions that you suggest, as both would increase complexity and confusion quite a bit.

Yup. Which is why I ruled them out.

The specific issue is this:

```python
class SanicException(Exception):
    def __init__(
        self,
        message: Optional[Union[str, bytes]] = None,
        status_code: Optional[int] = None,
        quiet: Optional[bool] = None,
        context: Optional[Dict[str, Any]] = None,
        extra: Optional[Dict[str, Any]] = None,
    ) -> None:
        ...
class MethodNotAllowed(SanicException):
    def __init__(self, message, method, allowed_methods):
        ...

class Unauthorized(SanicException):
    def __init__(self, message, status_code=None, scheme=None, **kwargs):
        ...
```

We have subclasses with incompatible signatures. I would really like to get to to keyword args (except `message`), and allow the subtypes to add their own.

```python
class SanicException(Exception):
    def __init__(
        self,
        message: Optional[Union[str, bytes]] = None,
        *,
        status_code: Optional[int] = None,
        quiet: Optional[bool] = None,
        context: Optional[Dict[str, Any]] = None,
        extra: Optional[Dict[str, Any]] = None,
    ) -> None:
        ...
class MethodNotAllowed(SanicException):
    def __init__(
        self,
        message,
        method,
        allowed_methods
        *,
        status_code: Optional[int] = None,
        quiet: Optional[bool] = None,
        context: Optional[Dict[str, Any]] = None,
        extra: Optional[Dict[str, Any]] = None,
    ):
        ...

class Unauthorized(SanicException):
    def __init__(
        self,
        message,
        status_code=None,
        scheme=None
        *,
        status_code: Optional[int] = None,
        quiet: Optional[bool] = None,
        context: Optional[Dict[str, Any]] = None,
        extra: Optional[Dict[str, Any]] = None,
    ):
        ...
```

So, yes, the issue is we have bad subclassing that messes up positional args.

sanic/exceptions.py
```diff
@@ -9,103 +10,292 @@ class RequestCancelled(CancelledError):
...
+    status_code: int
+    quiet: Optional[bool]
+    headers: Dict[str, str]
     message: str = ""
 
     def __init__(
         self,
         message: Optional[Union[str, bytes]] = None,
         status_code: Optional[int] = None,
+        *,
```
ahopkins, line 71: Yes, this is a breaking change that I think we should be making. It was a mistake that lead to a messy API allowing these to not be keyword-required.

ahopkins:
> Message taking bytes is a bit odd 
Not sure when or why that was added, but I think its been there a while.

PR #2548, opened 2022-09-17
Update runner.py

LauVenda:
Update logger.warn to logger.warning


ahopkins:
Thanks very much for this PR, but I am going to close it. #2499 does add support for muti-worker on all OSes, and therefore this line of code is now moot.

PR #2549, opened 2022-09-18
Add request counter

ahopkins:
Coming off the discussion [re: the worker manager](https://www.youtube.com/watch?v=m8HCO8NK7HE&t=1976s), I implemented the idea of adding a request counter. I thought of a few different strategies. Most important to me was that it not be forced upon users who do not opt in.

The obvious first choice was of course `config`. But, I do not want to even include a check for this on *every* request. I thought about using a touchup scheme, but we do not have one that is currently setup for general conditional checks. Maybe we should, but that is outside the scope here.

Doing it on `Request.__init__` is the easiest. So, I thought perhaps as a subclass we can get the result without any of the consequences. It does however bring the question about whether a self-contained opinionated piece of logic belongs in the main project at all, or if it should be pushed over to Sanic Extensions.

Thoughts?

### DIFF
--- a/sanic/request.py
+++ b/sanic/request.py
@@ -31,6 +31,7 @@ import uuid
 
 from collections import defaultdict
 from http.cookies import SimpleCookie
+from itertools import count
 from types import SimpleNamespace
 from urllib.parse import parse_qs, parse_qsl, unquote, urlunparse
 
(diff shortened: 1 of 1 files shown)


ahopkins:
Closed in favor of https://github.com/sanic-org/sanic-ext/pull/129

PR #2550, opened 2022-09-20
Begin middleware revamp

ahopkins:
#2521 is a bit too ambitious, and really is a bit of scope creep. This PR is meant to obtain the benefits of middleware streamlining and up front assignment without adding the complexity of a new request handling mechanism.

---

This PR aims to:

- introduce an API to control middleware ordering
- performance boost

```python
@app.on_request(priority=99)
async def high_priority(request):
    ...
```


sanic/app.py
```diff
@@ -840,7 +846,13 @@ async def handle_request(self, request: Request):  # no cov
...
         # Define `response` var here to remove warnings about
         # allocation before assignment below.
-        response = None
+        response: Optional[
+            Union[
+                BaseHTTPResponse,
+                Coroutine[Any, Any, Optional[BaseHTTPResponse]],
+            ]
+        ] = None
+        run_middleware = True
```
prryplatypus, line 855: Looks like this variable is being set to False twice (one here and one in line 900)

sanic/app.py
```diff
@@ -946,7 +960,7 @@ async def handle_request(self, request: Request):  # no cov
                 ...
                 await response.send(end_stream=True)
             elif isinstance(response, ResponseStream):
-                resp = await response(request)
+                resp = await response(request)  # type: ignore
```
prryplatypus, line 963: What are all these `# type: ignore` silencing?

prryplatypus (approved):

sanic/app.py
ahopkins: Once as `True`, once as `False`. Correct?

sanic/app.py
ahopkins: `ReportStream`. I want to get rid of it, so I was not overly concerned. But you are right. Will fix properly.

PR #2555, opened 2022-09-28
Remove deprecated items

### LINKED ISSUE #2303: Remove usage of ctx in routing
## TLDR

Remove all Sanic reliance upon `ctx` values in the `routing` module.

## Rationale

There are several locations throughout Sanic where the developer is provided with a `ctx` object to inject arbitrary values. The two most visible locations are [`Sanic.ctx`](https://sanicframework.org/en/guide/basics/app.html#application-context) and [`Request.ctx`](https://sanicframework.org/en/guide/basics/request.html#body). However, this also exists for the router, routes, blueprints, and connections.

Only in routing does Sanic actually use the object for storing internal state. It will be much more consistent if Sanic _never_ touches these objects, and therefore they can serve their original purpose: injection of arbitrary values without the worry of conflict.

---

## Implementation

The reason that Sanic has used these values is so that `sanic-routing` can exist independently of `sanic`. So, from the perspective of `sanic-routing`, these objects are fulfilling their purpose and not being overburdened with values. Ultimately, however, this is somewhat moot as far as the developer is concerned as they need to be careful not to overwrite these values.

Therefore, the simplest approach is probably to add a `state` object into `sanic-routing` that looks and functions the same was as `ctx`, but is specifically intended for arbitrary "stateful" information coming from `Sanic`.

See #2300 and #2302

ahopkins:
Closes #2303 

Removals:

- `route.ctx` not set by Sanic, therefore...
    - `route.ctx.ignore_body` >> `route.extra.ignore_body`
    - `route.ctx.stream` >> `route.extra.stream`
    - `route.ctx.hosts` >> `route.extra.hosts`
    - `route.ctx.static` >> `route.extra.static`
    - `route.ctx.error_format` >> `route.extra.error_format`
    - `route.ctx.websocket` >> `route.extra.websocket`
- `app.debug` is READ-ONLY
- `app.is_running` removed
- `app.is_stopping` removed
- Sanic._uvloop_setting removed
- Prefixed environment variables will be ignored if not uppercase

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -47,7 +47,7 @@ from sanic_routing.exceptions import FinalizationError, NotFound
 from sanic_routing.route import Route
 
 from sanic.application.ext import setup_ext
-from sanic.application.state import ApplicationState, Mode, ServerStage
+from sanic.application.state import ApplicationState, ServerStage
 from sanic.asgi import ASGIApp
 from sanic.base.root import BaseSanic
 from sanic.blueprint_group import BlueprintGroup
@@ -394,8 +394,8 @@ class Sanic(BaseSanic, StartupMixin, metaclass=TouchUpMeta):
             routes = [routes]
 
         for r in routes:
-            r.ctx.websocket = websocket
-            r.ctx.static = params.get("static", False)
+            r.extra.websocket = websocket
+            r.extra.static = params.get("static", False)
             r.ctx.__dict__.update(ctx)
 
         return routes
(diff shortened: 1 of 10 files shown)


prryplatypus (approved):
Happy with these changes if/when the tests pass!

ahopkins:
> Happy with these changes if/when the tests pass!

Coming shortly. There will be some failures because we have been testing the deprecations.

prryplatypus (approved):

sjsadowski:
Well I'm finally here and @prryplatypus already got to it, but it's got my chop too.

PR #2569, opened 2022-10-12
Add JSONResponse class

prryplatypus:
Adds a new JSONResponse class that's returned for JSON responses which introduces some new JSON utilities, such as accessing the raw JSON body and the ability to modify it directly.


sanic/response/types.py
```diff
@@ -0,0 +1,318 @@
...
+        The response cookies. Cookies should be set and written as follows:
+
+        .. code-block:: python
+
+                response.cookies["test"] = "It worked!"
+                response.cookies["test"]["domain"] = ".yummy-yummy-cookie.com"
+                response.cookies["test"]["httponly"] = True
+
+        `See user guide re: cookies
+        <https://sanicframework.org/guide/basics/cookies.html>`__
```
ahopkins, line 92: While we are at it, can you update the URL here?

sanic/response/types.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def __init__(
+        self,
+        body: Optional[Any] = None,
+        status: int = 200,
+        headers: Optional[Union[Header, Dict[str, str]]] = None,
+        content_type: Optional[str] = None,
+        dumps: Optional[Callable[..., str]] = None,
+        **kwargs: Any,
+    ):
```
ahopkins, line 213: Add a docstring for API docs.

sanic/response/types.py
```diff
@@ -0,0 +1,318 @@
...
+        self.set_json(value)
+
+    def set_json(
+        self,
+        new_json: Any,
+        dumps: Optional[Callable[..., str]] = None,
+        **kwargs: Any,
+    ):
+        dumps_ = dumps or self._dumps_method
+        kwargs_ = kwargs if kwargs else self._dumps_kwargs
```
ahopkins, line 241: I wonder if we should expose `_dumps_kwargs` as a property?

```python
response = json(...)
kwargs = {**response.dumps_kwargs, ...}
response.set_json(..., **kwargs)
```

This would make it possible to extend more easily. Thoughts?

sanic/response/types.py
```diff
@@ -0,0 +1,318 @@
...
+    @raw_body.setter
+    def raw_body(self, value: Any):
+        self.set_json(value)
+
+    def set_json(
+        self,
+        new_json: Any,
+        dumps: Optional[Callable[..., str]] = None,
+        **kwargs: Any,
+    ):
```
ahopkins, line 239: Do you also think it is worth having some common methods for `dict` and `list` types? Specifically thinking of `response.update({"foo": "bar"})`, `response.append(1234)`, `response.extend([1,2,3,4])`.

ahopkins (changes requested):

tests/test_response.py
```diff
@@ -185,11 +186,22 @@ async def delete_handler(request: Request):
...
+        response.set_json(NEW_RESP)
+        assert response.raw_body == NEW_RESP
+        assert response.body == json_dumps(NEW_RESP).encode("utf-8")
 
     request, response = json_app.test_client.get("/")
     assert response.status == 200
-    assert response.text == json_dumps(JSON_DATA)
-    assert response.json == JSON_DATA
+    assert response.text == json_dumps(NEW_RESP)
+    assert response.json == NEW_RESP
```
ahopkins, line 204: You should leave the existing test as is, and just make a new one.

sanic/response/types.py
prryplatypus: Agreed! Do you think they should be overwritten if new ones are passed when calling `set_json`?

sanic/response/types.py
prryplatypus: I thought about it, just couldn't figure out what the methods should do if they were called with other types.

prryplatypus:
Just thought about this... what if someone does `response.raw_body.update(...)`? that wouldn't work as expected... any suggestions?

sanic/response/types.py
ahopkins: Probably not. That would be an odd side effect.

sanic/response/types.py
ahopkins: `TypeError`

prryplatypus:
Alright, I've approached this differently. Main difference is that now the body doesn't get set when the raw_body is set, and instead it is generated "on-the-fly" when accessing the attribute.

This leaves us with the following outcome:
1. It is backwards-compatible. You can set `body` at any point throughout the response's lifetime.
1. The new `raw_body` property can be set at any point throughout the response's lifetime (as long as the `body` property has not been set manually).
1. Common JSON object methods can be used directly on the `raw_body` without us having to implement any of the methods, and the `body` output will remain consistent. For example `resp.raw_body.update(...)` or `resp.raw_body.append(...)`.

Only "drawback" is that `dumps` / `dumps_kwargs` can't be changed once the response has already been created, but this shouldn't be an issue, as that's the way it currently works.

Let me know your thoughts @ahopkins and if you're happy with it I'll write the tests for it.

prryplatypus:
Another potentially breaking change: Currently a `TypeError` is raised from within the constructor if a type can't be serialized by the json encoder. With this change mentioned above it would only be raised when calling `resp.body`. As a workaround we can just add `self.body` somewhere in the constructor, but this is not optimal as it would make every response be serialized twice.

However, we could potentially leave this in only for a deprecation period and then remove it. Thoughts welcome.

ahopkins:
Remind me... where did we settle on the API for this?

prryplatypus:
You suggested doing a `json.loads` if the body was set manually, but I disagreed because it feels like a cross of domains. Therefore, I think we were both on the same page with documenting that `raw_body` should not be mutated at all, unless after mutating it you called `.set_json(raw_body)` or did `resp.raw_body = resp.raw_body` (which would trigger a `dumps` call with the new raw body).

That said, we could potentially add the "common methods" that you mentioned previously, such as `append` or `update`, if so you wish.

prryplatypus:
@ahopkins API now should be as agreed 😎

sanic/response/types.py
```diff
@@ -0,0 +1,442 @@
...
+
+        self._check_body_not_manually_set()
+
+        if not isinstance(self._raw_body, dict):
+            raise SanicException("Cannot update a non-dict object.")
+
+        self._raw_body.update(*args, **kwargs)
+        self.raw_body = self._raw_body
+
+    def pop(self, key: Any) -> Any:
```
ahopkins, line 352: Pop should have a default second argument.

ahopkins (changes requested):
The `json` convenience method needs its return type updated.

prryplatypus:
> The `json` convenience method needs its return type updated.

It is updated, I believe? https://github.com/sanic-org/sanic/pull/2569/files#diff-9e98d5a6f92def2f06f426094980026882610a5e8d8215b9b291b58620f0f15fR50-R57

sanic/response/types.py
prryplatypus: Added a default argument that will always be passed to the underlying object if it is not the default argument. This will cause an exception to be raised if the argument is passed when raw body is a list (since lists don't support a default argument in `pop` for some reason), but that is expected & intended so the error remains as similar as possible to what would be raised if the user was calling the underlying's object method with those arguments directly.

prryplatypus:
Ahh the typehint!

ahopkins (approved):

PR #2578, opened 2022-10-19
Add GenericCreator for loading SSL certs in processes

### LINKED ISSUE #2577: Certificates not created with `mkcert` or `trustme` raise a RuntimeError
The `CertLoader` class in `sanic-org/sanic/sanic/worker/loader.py` checks the creator of the certificate. If the creator is not `mkcert` or `trustme` then it raises a `RuntimeError`. This will prevent Sanic from running with certificates from any other sources.

ahopkins:
Resolves #2577


sanic/worker/loader.py
```diff
@@ -109,18 +110,28 @@ class CertLoader:
...
         creator_name = ssl_data.get("creator")
-        if creator_name not in ("mkcert", "trustme"):
-            raise RuntimeError(f"Unknown certificate creator: {creator_name}")
-        elif creator_name == "mkcert":
-            self._creator_class = MkcertCreator
-        elif creator_name == "trustme":
-            self._creator_class = TrustmeCreator
+        if not creator_name:
+            return
+        else:
```
Tronic, line 118: Don't use `else` if the `if` returns/throws.

sanic/worker/loader.py
```diff
@@ -109,18 +110,28 @@ class CertLoader:
...
+            if creator_name not in ("mkcert", "trustme"):
+                raise RuntimeError(
+                    f"Unknown certificate creator: {creator_name}"
+                )
+            elif creator_name == "mkcert":
+                self._creator_class = MkcertCreator
+            elif creator_name == "trustme":
+                self._creator_class = TrustmeCreator
```
Tronic, line 126: Maybe use `dict(mkcert=MkcertCreator, trustme=TrustmeCreator).get(creator_name)` instead of this whole thing? Less code, no repetition.

sanic/worker/loader.py
```diff
@@ -109,18 +110,28 @@ class CertLoader:
...
 
         self._key = ssl_data["key"]
         self._cert = ssl_data["cert"]
         self._localhost = cast(str, ssl_data["localhost"])
 
     def load(self, app: SanicApp):
+        if not hasattr(self, "_creator_class"):
```
Tronic, line 133: Might prefer member variables set to `None` instead of using `hasattr`.

sanic/worker/loader.py
```diff
@@ -109,18 +110,28 @@ class CertLoader:
...
+            return process_to_context(self._ssl_data)
```
Tronic, line 134: I didn't look into the sensibility of this line. It is late and I have to sleep.

sanic/worker/loader.py
ahopkins: Was thinking about that, but was keeping my code changes minimal.

Tronic (dismissed):

ahopkins:

https://community.sanicframework.org/t/unknown-certificate-creator-error/1090/2

PR #2585, opened 2022-10-22
improve error message if no apps found in registry

### LINKED ISSUE #2579: In multiworker mode better error message if no apps
If a developer does not define their app so that a worker process can access it, they will receive a `KeyError` when Sanic tries to access it from the registry.

For example, that would happen if an app only was instantiated on the main process:

```python
if __name__ == "__main__":
    app = Sanic(...)
    app.run()
```

The solution to this is usually either to move the instantiation outside that block, or to use an [`AppLoader`](https://sanic.dev/en/guide/deployment/app-loader.html).

We should provide a better error message from within a worker process if an app cannot be found in a registry.

scardozos:
Fixes #2579


ahopkins:
Thanks, but I think the way I was planning to handle this was to switch the order of socket config, in which case the existing message should handle this.

sanic/app.py
```diff
@@ -1453,7 +1453,12 @@ def get_app(
                 return cls.get_app("__mp_main__", force_create=force_create)
             if force_create:
                 return cls(name)
-            raise SanicException(f'Sanic app name "{name}" not found.')
+            raise SanicException(
+                f"Sanic app name '{name}' not found.\n"
+                "App instantiation must occur outside "
+                "if __name__ == '__main__' "
+                "block or by using an AppLoader."
```
ahopkins, line 1460: We should add a message to prompt the user to go here: https://sanic.dev/en/guide/deployment/app-loader.html

scardozos: Cool! I'll edit this and update the test tomorrow :)

ahopkins (approved):

PR #2590, opened 2022-10-26
Corrected `Colors` enum under Python 3.11

LiraNuna:
Python 3.11 have made some changes to how mixed-in Enum formatting works. From the changelog:

> Changed [`Enum.__format__()`](https://docs.python.org/3.11/library/enum.html#enum.Enum.__format__) (the default for [format()](https://docs.python.org/3.11/library/functions.html#format), [str.format()](https://docs.python.org/3.11/library/stdtypes.html#str.format) and [f-string](https://docs.python.org/3.11/glossary.html#term-f-string)s) of enums with mixed-in types (e.g. [int](https://docs.python.org/3.11/library/functions.html#int), [str](https://docs.python.org/3.11/library/stdtypes.html#str)) to also include the class name in the output, not just the member’s key. This matches the existing behavior of [`enum.Enum.__str__()`](https://docs.python.org/3.11/library/enum.html#enum.Enum.__str__), returning e.g. 'AnEnum.MEMBER' for an enum AnEnum(str, Enum) instead of just 'MEMBER'.

This PR fixes this issue by using the new `StrEnum` from Python 3.11 if available, substituting it for a subclass otherwise.

NOTE: I would recommend adding python 3.11 to unit tests before merging this one.

Fixes https://github.com/sanic-org/sanic/issues/2589

### DIFF
--- a/sanic/log.py
+++ b/sanic/log.py
@@ -2,10 +2,21 @@ import logging
 import sys
 
 from enum import Enum
-from typing import Any, Dict
+from typing import TYPE_CHECKING, Any, Dict
 from warnings import warn
 
 from sanic.compat import is_atty
+
+
+# Python 3.11 changed the way Enum formatting works for mixed-in types.
+if sys.version_info < (3, 11, 0):
+
+    class StrEnum(str, Enum):
+        pass
+
+else:
+    if not TYPE_CHECKING:
+        from enum import StrEnum
 
 
 LOGGING_CONFIG_DEFAULTS: Dict[str, Any] = dict(  # no cov
@@ -68,7 +79,7 @@ Defult logging configuration
 """
 
 
-class Colors(str, Enum):  # no cov
+class Colors(StrEnum):  # no cov
     END = "\033[0m"
     BOLD = "\033[1m"
     BLUE = "\033[34m"
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -10,6 +10,7 @@ import pytest
 import sanic
 
 from sanic import Sanic
+from sanic.log import Colors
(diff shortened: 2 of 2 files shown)


LiraNuna:
It seems that mypy does not like the conditional import. 

```
  sanic/log.py:14: error: Module "enum" has no attribute "StrEnum"
  sanic/log.py:14: error: Name "StrEnum" already defined on line 11
```
 
I'll take care of the formatting changes.

ahopkins:
> NOTE: I would recommend adding python 3.11 to unit tests before merging this one.

Yup, that is a tomorrow task :sunglasses:

sanic/log.py
```diff
@@ -2,12 +2,23 @@
...
 
 
+# Python 3.11 changed the way Enum formatting works for mixed-in types.
+if sys.version_info < (3, 11, 0):
+
+    class StrEnum(str, Enum):
+        pass
+
+else:
+    if not TYPE_CHECKING:
```
ahopkins, line 18: Why do we care about `TYPE_CHECKING` here?

LiraNuna: Without this, `mypy` errors that `StrEnum` is previously declared.

LiraNuna: @ahopkins what's the status of this?

ahopkins: If you are on 3.11, you need `StrEnum` whether type checking or not. Am I missing something?

LiraNuna: `mypy` does not seem to care about the previous `if` checking for the version and will complain about `StrEnum` being declared twice.

ahopkins (approved):

PR #2592, opened 2022-10-28
Resolve edge case in nested BP Groups

ahopkins:
There is an edge case where nested BP groups where the inner groups have no `url_prefix` behave unexpected.

```python
app = Sanic("TestApp")
app.config.AUTO_EXTEND = False

one = Blueprint("one", url_prefix="/one")


@one.get("/func")
async def func(req):
    ...


two = Blueprint.group(one)
three = Blueprint.group(two, url_prefix="three")
app.blueprint(three)
```

In this situation, the outer group prefix would not be unexpectedly ignored.


prryplatypus (approved):
Might want to add a test to avoid regressions

ahopkins:
I plan to. Meant for this to be in draft mode.

prryplatypus:
Marked as draft for you :P

prryplatypus (approved):

PR #2605, opened 2022-11-18
Add broadcast feature to websocket

kijk2869:
Adds a `broadcast()` function like [websockets](https://websockets.readthedocs.io/en/stable/reference/utilities.html#websockets.broadcast) to WebsocketImplProtocol to broadcast a message to several websocket connections efficiently.
This pushes the message synchronously to all connections without backpressure.

example
```python
from sanic import Websocket

Websocket.boradcast(a_list_of_websocket_connections, a_message_to_send)
```

To implement this, `send_sync` method, data writer without backpressure control, is added to SanicProtocol.

### DIFF
--- a/sanic/server/protocols/base_protocol.py
+++ b/sanic/server/protocols/base_protocol.py
@@ -74,6 +74,13 @@ class SanicProtocol(asyncio.Protocol):
         self.transport.write(data)
         self._time = current_time()
 
+    def send_sync(self, data):
+        """
+        Generic data write implementation without backpressure control.
+        """
+        self.transport.write(data)
+        self._time = current_time()
+
     async def receive_more(self):
         """
         Wait until more data is received into the Server protocol's buffer
--- a/sanic/server/websockets/impl.py
+++ b/sanic/server/websockets/impl.py
@@ -727,6 +727,9 @@ class WebsocketImplProtocol:
             # Catch a common mistake -- passing a dict to send().
             raise TypeError("data is a dict-like object")
 
+        else:
+            raise TypeError("Websocket data must be bytes, str.")
+
         for websocket in websockets:
             if websocket.connection.state is not OPEN:
                 continue
--- a/sanic/server/websockets/impl.py
+++ b/sanic/server/websockets/impl.py
@@ -12,7 +12,7 @@ from typing import (
     Union,
 )
 
-from websockets.connection import CLOSED, CLOSING, OPEN, Event
+from websockets.connection import CLIENT, CLOSED, CLOSING, OPEN, Event
 from websockets.exceptions import (
     ConnectionClosed,
     ConnectionClosedError,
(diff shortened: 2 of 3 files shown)


Tronic:
Sanic should not implement sending without backpressure control. It is known to cause security flaws and plenty of other problems.

kijk2869:
> Sanic should not implement sending without backpressure control. It is known to cause security flaws and plenty of other problems.
> 
> Instead, you can send separately to each socket but `await asyncio.gather(...)`.



> Sanic should not implement sending without backpressure control. It is known to cause security flaws and plenty of other problems.
> 
> Instead, you can send separately to each socket but `await asyncio.gather(...)`.

Then, is there really no safe way to bypass backpressure?

Cost of creating `task` / `future` per connection is too high when mass connection.
> In fact, I'm currently using asyncio.gather, and the majority of the blocking I'm experiencing happens here
> This can be also proven with a simple test.

In websockets docs about broadcasting, https://websockets.readthedocs.io/en/stable/topics/broadcast.html
Several designs were tested, but in the end the most effective was to release the backpressure.

kijk2869:
### Mass task performance test

#### Output
```sh
creating 1000 tasks...
Executing <Task pending name='Task-1' coro=<main() running at /mnt/e/asdfsafdfasd/task.py:17> wait_for=<_GatheringFuture pending cb=[Task.task_wakeup()] created at /usr/lib/python3.10/asyncio/tasks.py:665> cb=[_run_until_complete_cb() at /usr/lib/python3.10/asyncio/base_events.py:184] created at /usr/lib/python3.10/asyncio/tasks.py:636> took 1.239 seconds
elapsed: 4.06 sec
```

#### Test code
```python
import asyncio
import time

import uvloop

loop = uvloop.new_event_loop()
loop.set_debug(True)


async def test():
    await asyncio.sleep(1)


async def main():
    print("creating 1000 tasks...")
    targets = [test() for _ in range(1000)]

    base = time.perf_counter()
    await asyncio.gather(*targets)
    print(f"elapsed: {time.perf_counter() - base:.2f} sec")


loop.run_until_complete(main())
```

Tronic:
If you have a massive number of clients and your network slows down, transmit queues grow and you likely run out of RAM. This hangs your server for a good while before the OOM killer kicks in. Even in normal operation, latency might grow if the upstream is throttled, so you will not even see good performance. How many websockets are you feeding and how often? Perhaps you should have long-lived tasks for them (most naturally within the handlers where they were created). Perhaps implement your own queue prior to broadcasting, such that queued messages only use the memory once and not per each connection.

kijk2869:
> If you have a massive number of clients and your network slows down, transmit queues grow and you likely run out of RAM. This hangs your server for a good while before the OOM killer kicks in. Even in normal operation, latency might grow if the upstream is throttled, so you will not even see good performance. How many websockets are you feeding and how often? Perhaps you should have long-lived tasks for them (most naturally within the handlers where they were created). Perhaps implement your own queue prior to broadcasting, such that queued messages only use the memory once and not per each connection.

I'm trying to keep fewer than 700 connections per worker because of the issues mentioned above. And due to the nature of the service, it is necessary to broadcast to the entire connection in a short period of time. Thanks for your advice. It may be tricky because the handler already has a receiver, but I'll try integrating it. I will close this pr and continue in discord if I have any further questions. Thanks.

PR #2606, opened 2022-11-23
Decode headers as UTF-8 also in ASGI

### LINKED ISSUE #2604: HTTP 1 request headers decoded using default encoding instead of ISO-8859-1

### Describe the bug

headers are decoded here without specifying their encoding:

https://github.com/sanic-org/sanic/blob/ad4e526c775fc3ce950503d6476d9d344492b0dd/sanic/http/http1.py#L205

On my system (osx using python 3.10.8 installed via homebrew) this causes bytes that are valid characters in ISO-8859-1 but not in UTF-8 to be decoded as surrogate escape characters, e.g. `b"\x80"` becomes `"\udf80"` instead of `"\x80"`

### Code snippet

_No response_

### Expected Behavior

headers encoded as ISO-8859-1 with no MIME type to be decoded correctly without using UTF-8 surrogate escape characters.

### How do you run Sanic?

### Operating System

linux

### Sanic Version

22.9.1

### Additional context

this used to work as expected in Sanic<=20.12.7

ChihweiLHBird:
Closes #2604


ahopkins (changes requested):
On hold until v23

Tronic:
This should probably continued. Needs some practical testing with ASGI servers to see how they behave, but the likely outcome is that we need to use UTF-8 on that side as well, the same as with the integrated server.

Two points of interest:
- Header encoding (values in particular, keys should stay ASCII).
- Path encoding, and with that also how clients handle it.

I can help run those tests once I get my other PRs sorted out, but @ChihweiLHBird feel free to do your own tests if you can. If it looks like that UTF-8 is going to work identically with ASGI as it does with the Sanic server, we are all good to implement UTF-8 on ASGI as well.

ChihweiLHBird:
@Tronic Sure, I can try to test it.

Tronic:
@ChihweiLHBird Could you get this finished? We had a discussion and want to have this in the upcoming 23.3 release.

ChihweiLHBird:
@Tronic Sorry, I was busy last few days. I am on it right now.

ChihweiLHBird:
@Tronic Just done some manual testing, and it seems for the test cases I tried, there is not an issue for header decoding. Like if it uses UTF-8, it will successfully decode emojis and other UTF-8 characters but failed to decode some characters in latin-1 which isn't in UTF-8, like `b"\x80"`. Similar result to path. And I think this is the expected results, right? I am not sure what are the more comprehensive test cases I should try.

sanic/asgi.py
```diff
@@ -134,7 +134,7 @@ async def create(
 
         headers = Header(
             [
-                (key.decode("latin-1"), value.decode("latin-1"))
+                (key.decode(), value.decode())
```
Tronic, line 137: Add `errors="surrogateescape"` to match what we do in http1.py, this will handle invalid UTF-8 as well.

Tronic: For key, use `.decode("ASCII")` as per RFC 7230, which defines field-name as *token*, which can according to this RFC contain only visible US-ASCII characters (VCHAR) sans certain separator characters (or by other definition, certain punctuation and DIGIT and ALPHA, where ALPHA is not clearly defined but apparently only refers to ASCII letters).

For value, `.decode(errors="surrogateescape")` as stated, because they can contain opaque bytes other than UTF-8.

Tronic: For reference
https://www.rfc-editor.org/rfc/rfc7230#section-1.2  - VCHAR, ALPHA and DIGIT
https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6  - token (field-name)
https://www.rfc-editor.org/rfc/rfc7230#section-3.2.4  - obs-text (in field-content)
https://www.rfc-editor.org/rfc/rfc7230#section-3.2  - field-name, field-content

Tronic:
Note on further changes: (not necessarily in this PR)
- http1 needs to use ASCII for url_bytes encoding too (and raise that same exception)
- http3 alike needs to encode in ASCII, and also needs ASCII/surrogateescape field name/value as here.

After these changes all three modes should behave more closely to each other, and be stricter in RFC compliance.

Tronic:
PR #2710 now implements charsets in URL handling, so that should not be addressed here. But implement error handling for headers as instructed above.

sanic/asgi.py
ChihweiLHBird: Added

ChihweiLHBird:
> PR #2710 now implements charsets in URL handling, so that should not be addressed here. But implement error handling for headers as instructed above.

What about http1 headers?
https://github.com/sanic-org/sanic/blob/88c918e72f496ed22405b8e3a0fff672553ea2ff/sanic/http/http1.py#L205

Tronic:
> What about http1 headers?

It will stay laxer with header names. The entire header is decoded at once because that is faster than decoding each field separately, and thus we cannot check that they are ASCII (without performance penalty).

sanic/asgi.py
```diff
@@ -132,12 +132,18 @@ async def create(
...
+                [
+                    (
+                        key.decode("ASCII"),
+                        value.decode(errors="surrogateescape"),
+                    )
+                    for key, value in scope.get("headers", [])
+                ]
+            )
+        except UnicodeDecodeError:
+            BadRequest("Header names can only contain US-ASCII characters")
```
Tronic, line 146: Missing `raise`

Tronic: Also needs a test that hits this error.

sanic/http/http3.py
```diff
@@ -333,7 +338,15 @@ def get_receiver_by_stream_id(self, stream_id: int) -> Receiver:
...
-        headers = Header(((k.decode(), v.decode()) for k, v in event.headers))
+        try:
+            headers = Header(
+                (
+                    (k.decode("ASCII"), v.decode(errors="surrogateescape"))
+                    for k, v in event.headers
+                )
+            )
+        except UnicodeDecodeError:
+            BadRequest("Header names can only contain US-ASCII characters")
```
Tronic, line 349: Please revert this. It's done in the other PR #2710

sanic/asgi.py
ChihweiLHBird: Added `raise`. Will add the test later today.

sanic/http/http3.py
ChihweiLHBird: fixed

sanic/asgi.py
ChihweiLHBird: In my experiment, raising a `BadRequest` in `asgi.py` will cause the ASGI server return 500 error, instead of 400.

Tronic (dismissed):
Good work :)

ChihweiLHBird:
@ahopkins May I have your review?

sanic/asgi.py
ahopkins: @ChihweiLHBird You are correct. I'll look into this probably next week.

sanic/asgi.py
```diff
@@ -132,12 +132,20 @@ async def create(
...
+                        key.decode("ASCII"),
+                        value.decode(errors="surrogateescape"),
+                    )
+                    for key, value in scope.get("headers", [])
+                ]
+            )
+        except UnicodeDecodeError:
+            raise BadRequest(
+                "Header names can only contain US-ASCII characters"
+            )
```
ahopkins, line 148: There's no way for anyone who receives this error to know what went wrong. Should we log the header names?

ChihweiLHBird: Is it safe to directly log the header names from the client? I am worried about some very large header name for flood attack or something like injection. And if we decide to log it later, would it be an error logged by `error_logger`?

ChihweiLHBird:
```suggestion

```

ahopkins: You are right. Let's not.

ahopkins (approved):

PR #2608, opened 2022-11-27
Deprecate Conditions and Triggers Saved in `handler` Callable; Save Condition in `signal.extra` Instead

### LINKED ISSUE #2581: A signal handler cannot be a method on an object
**Describe the bug**

If I try and register a method on an object as a signal handler, an exception is thrown because it is not possible to set an attribute on a bound method
```
    self.add_signal(self.after_routing, 'http.routing.after')
  File "sanic/lib/python3.7/site-packages/sanic/mixins/signals.py", line 77, in add_signal
    handler
  File "sanic/lib/python3.7/site-packages/sanic/mixins/signals.py", line 57, in decorator
    self._apply_signal(future_signal)
  File "sanic/lib/python3.7/site-packages/sanic/app.py", line 420, in _apply_signal
    return self.signal_router.add(*signal)
  File "sanic/lib/python3.7/site-packages/sanic/signals.py", line 222, in add
    handler.__requirements__ = condition  # type: ignore
AttributeError: 'method' object has no attribute '__requirements__'
```

**Code snippet**

```python
import sanic

class MyApp(sanic.Sanic):
  def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)
    self.add_signal(self.after_routing, 'http.routing.after')

  def after_routing(self, request, route, kwargs, handler):
    pass

app = MyApp(name='method_signal_handler')
```

**Expected behavior**

I would expect the bound method to be registered, and then called when the signal is dispatched.

 - OS: MacOS
 - Sanic Version: 22.9

**Additional context**

ChihweiLHBird:
closes #2581

### DIFF
--- a/sanic/signals.py
+++ b/sanic/signals.py
@@ -218,7 +218,8 @@ class SignalRouter(BaseRouter):
             event = ".".join([*parts[:2], "<__trigger__>"])
 
         try:
-            # Attaching __requirements__ and __trigger__ to the handler is deprecated and will be removed in v23.6.
+            # Attaching __requirements__ and __trigger__ to the handler
+            # is deprecated and will be removed in v23.6.
             handler.__requirements__ = condition  # type: ignore
             handler.__trigger__ = trigger  # type: ignore
         except AttributeError:
--- a/sanic/signals.py
+++ b/sanic/signals.py
@@ -86,6 +86,8 @@ class SignalRouter(BaseRouter):
         )
         self.allow_fail_builtin = True
         self.ctx.loop = None
+        self.triggers: Dict[SignalHandler, str] = dict()
+        self.conditions: Dict[SignalHandler, Optional[dict]] = dict()
 
     def get(  # type: ignore
         self,
@@ -154,9 +156,7 @@ class SignalRouter(BaseRouter):
         try:
             for signal in signals:
                 params.pop("__trigger__", None)
-                requirements = getattr(
-                    signal.handler, "__requirements__", None
-                )
+                requirements = self.conditions.get(signal.handler, None)
                 if (
                     (condition is None and signal.ctx.exclusive is False)
                     or (condition is None and not requirements)
@@ -219,8 +219,8 @@ class SignalRouter(BaseRouter):
         if not trigger:
             event = ".".join([*parts[:2], "<__trigger__>"])
 
-        handler.__requirements__ = condition  # type: ignore
-        handler.__trigger__ = trigger  # type: ignore
+        self.conditions[handler] = condition
+        self.triggers[handler] = trigger
 
         signal = super().add(
             event,
--- a/sanic/signals.py
+++ b/sanic/signals.py
@@ -86,8 +86,6 @@ class SignalRouter(BaseRouter):
         )
         self.allow_fail_builtin = True
         self.ctx.loop = None
-        self.triggers: Dict[SignalHandler, str] = dict()
-        self.conditions: Dict[SignalHandler, Optional[dict]] = dict()
 
     def get(  # type: ignore
         self,
@@ -156,7 +154,7 @@ class SignalRouter(BaseRouter):
         try:
             for signal in signals:
                 params.pop("__trigger__", None)
-                requirements = self.conditions.get(signal.handler, None)
+                requirements = signal.extra.requirements
                 if (
                     (condition is None and signal.ctx.exclusive is False)
                     or (condition is None and not requirements)
@@ -219,8 +217,11 @@ class SignalRouter(BaseRouter):
         if not trigger:
             event = ".".join([*parts[:2], "<__trigger__>"])
 
-        self.conditions[handler] = condition
-        self.triggers[handler] = trigger
+        try:
+            handler.__requirements__ = condition # type: ignore
+            handler.__trigger__ = trigger # type: ignore
+        except AttributeError:
+            pass
 
         signal = super().add(
             event,
@@ -232,6 +233,7 @@ class SignalRouter(BaseRouter):
         signal.ctx.exclusive = exclusive
         signal.ctx.trigger = trigger
         signal.ctx.definition = event_definition
+        signal.extra.requirements = condition
 
         return cast(Signal, signal)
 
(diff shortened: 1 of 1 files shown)


sanic/signals.py
```diff
@@ -86,6 +86,8 @@ def __init__(self) -> None:
         )
         self.allow_fail_builtin = True
         self.ctx.loop = None
+        self.triggers: Dict[SignalHandler, str] = dict()
+        self.conditions: Dict[SignalHandler, Optional[dict]] = dict()
```
todddialpad, line 90: This is potentially problematic because we would be using a `SignalHandler` as the key in a dictionary. `SignalHandler` is supposed to be `Callable`, so that means it can be something other than just an instance method. There are callable things that are not hashable, so cannot be keys in a dictionary. For example, any user-defined class that implements an `__eq__` method but not a `__hash__` method is not hashable. So this class:
```
class NotHashable:
  def __init__(self, value):
    self.value = value
  def __eq__(self, other):
    return self.vlaue == other.value
  def __call__(self):
    return self.value
```
Instances of `NotHashable` are in theory valid signal handlers, but they are not hashable (have an `__eq__` but do not have a `__hash__`) so would fail here.

```
a = NotHashable(1)
a() # returns 1
{a: 1} # raises TypeError: Unhashable type 'A'
```

I'm not sure of the full range of callables that are not hashable, whether that's a worthwhile tradeoff, but this implementation would not allow the use of such patterns. I can't seem to find a reference for what is the `hash` of "normal" callables.

ChihweiLHBird: Any suggestions for an alternative solution? I think a work around is to wrap  the passed in `Callable` with another function so that would make it hashable or even variable injectable (for the original implementation).

todddialpad: Could the stored routes be 3-tuples of (handler, conditions, trigger), rather than just the handler? Then when iterating over routes (signals), the conditions and trigger would be simultaneously available.

ChihweiLHBird: How about hashing the handler object's [id](https://docs.python.org/3/library/functions.html#id)? This seems simpler than storing and iterating the list of tuples.

But the problem is "two objects with non-overlapping lifetimes may have the same [id()](https://docs.python.org/3/library/functions.html#id) value."

I think adding `__name__` to a tuple with id, `(handler.__name__, id(handler))` can possibly get rid of the duplicated id issue.

ahopkins: I think we are complicating this further than it has to. I am pretty sure that methodology preexisting the existence of `Route` (in this case a `Signal`) having both a `ctx` (which should really not be used except by the end user) and `extra`.

`trigger` has already moved over to this context. We can probably do the same. How about just do this:


```diff
diff --git a/sanic/signals.py b/sanic/signals.py
index 3b66268..a95de83 100644
--- a/sanic/signals.py
+++ b/sanic/signals.py
@@ -154,9 +154,7 @@ class SignalRouter(BaseRouter):
         try:
             for signal in signals:
                 params.pop("__trigger__", None)
-                requirements = getattr(
-                    signal.handler, "__requirements__", None
-                )
+                requirements = signal.extra.requirements
                 if (
                     (condition is None and signal.ctx.exclusive is False)
                     or (condition is None and not requirements)
@@ -219,8 +217,11 @@ class SignalRouter(BaseRouter):
         if not trigger:
             event = ".".join([*parts[:2], "<__trigger__>"])
 
-        handler.__requirements__ = condition  # type: ignore
-        handler.__trigger__ = trigger  # type: ignore
+        try:
+            handler.__requirements__ = condition  # type: ignore
+            handler.__trigger__ = trigger  # type: ignore
+        except AttributeError:
+            ...
 
         signal = super().add(
             event,
@@ -232,6 +233,7 @@ class SignalRouter(BaseRouter):
         signal.ctx.exclusive = exclusive
         signal.ctx.trigger = trigger
         signal.ctx.definition = event_definition
+        signal.extra.requirements = condition
 
         return cast(Signal, signal)
```

On a completely different note, I must say I find this very frustrating:

```python
>>> class Foo:
...  def bar(self):
...   ...
... 
>>> Foo.bar.z = 1
>>> foo = Foo()
>>> foo.bar.z
1
>>> foo.bar.z = 3
AttributeError: 'method' object has no attribute 'z'
```

ChihweiLHBird: That sounds a good idea. But if `trigger` and `condition` are going to be in `ctx` and/or `extra`, why we still need this?
```python
try:
    handler.__requirements__ = condition  # type: ignore
    handler.__trigger__ = trigger  # type: ignore
except AttributeError:
    ...
```

ChihweiLHBird: @ahopkins May I apply your code to this PR?

ahopkins: :+1:

ChihweiLHBird: @ahopkins 
I just did a small experiment and found `property` doesn't work as I expect if it is injected in to `function` object. 

```
>>> def handler():
...     return
...
>>> attr = 1
>>> def fget():
...     return attr
...
>>> def fset(new_val):
...     attr = new_val
...
>>> handler.attr = property(fget=fget, fset=fset)
>>> handler.attr
<property object at 0x000001CB8E290720>
```

I thought `handler.attr` would invoke `fget` but it doesn't... I can't think of another wat to add deprecation warning. Can we just remove it (in whether next version or few versions after it) and state this change in the release note since this was always an internal API? Or any idea to add the warning?

ahopkins: Yeah. We'll document the deprecation/removal. Leave it for now though.

sanic/signals.py
```diff
@@ -219,8 +217,11 @@ def add(  # type: ignore
         if not trigger:
             event = ".".join([*parts[:2], "<__trigger__>"])
 
-        handler.__requirements__ = condition  # type: ignore
-        handler.__trigger__ = trigger  # type: ignore
+        try:
```
ahopkins, line 220: Can you throw in a code comment here about this being deprecated also? If nothing else, it will make it easier for me to locate it in the future.

```
Attaching __requirements__ and __trigger__ to the handler is deprecated and will be removed in v23.6.
```

ChihweiLHBird: Done

ahopkins (dismissed):

ahopkins (approved):

ChihweiLHBird:
I forgot the test case... Will add it in another PR.

PR #2609, opened 2022-11-27
Add compatibility with websockets 11.0.

aaugustin:
I failed at naming things in the Sans-I/O layer of websockets. I ended up with `Connection` and `Protocol` swapped vs. what one would expect: `Protocol` was managing the network and `Connection` didn't know about the network.

I am about to rename in https://github.com/aaugustin/websockets/pull/1269. Since Sanic was an early adopter — actually, the first mainstream library to adopt it — it will need an update. I'm sorry.

This PR provides compatibility both with websockets 10.0 (without deprecation warnings) and with the upcoming release websockets 11.0.

----

sanic currently imports private APIs for which websockets won't provide backwards-compatibility. Specifically, it imports `OPEN`, `CLOSING`, `CLOSED` directly rather than the `State` enum. To avoid breaking Sanic users, I am planning to keep websockets 11.0 on hold until you release Sanic 22.12, with this PR included.

This is why I have to support both versions of websockets: I cannot simply release websockets 11.0 then update Sanic (bumping the dependency to websockets >= 11.0 to avoid deprecation warnings).

----

In detail:

* The first commit switches to the correct import paths for websockets 11.0, with fallback to import paths for websockets 10.x.
* The second commit renames a variable for consistency with the new name in websockets. It is optional; I can remove it if you don't want it.
* The third commit ensures that Sanic only uses public APIs of websockets, for which backwards-compatibility is guaranteed a minimum of 5 years.

### DIFF
--- a/sanic/server/protocols/websocket_protocol.py
+++ b/sanic/server/protocols/websocket_protocol.py
@@ -80,7 +80,7 @@ class WebSocketProtocol(HttpProtocol):
         # Called by Sanic Server when shutting down
         # If we've upgraded to websocket, shut it down
         if self.websocket is not None:
-            if self.websocket.connection.state in (CLOSING, CLOSED):
+            if self.websocket.ws_proto.state in (CLOSING, CLOSED):
                 return True
             elif self.websocket.loop is not None:
                 self.websocket.loop.create_task(self.websocket.close(1001))
--- a/sanic/server/protocols/websocket_protocol.py
+++ b/sanic/server/protocols/websocket_protocol.py
@@ -5,8 +5,8 @@ try:  # websockets < 11.0
     from websockets.connection import State
     from websockets.server import ServerConnection as ServerProtocol
 except ImportError:  # websockets >= 11.0
-    from websockets.protocol import State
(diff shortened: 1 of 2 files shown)


aaugustin:
Actually, I have to use the websockets 10.x imports and fallback to the websockets 11.0 imports, or else mypy gets confused.

It will get confused again when I release websockets 11.0. Fortunately, at that point, we can remove the fallback and bump the dependency to websockets >= 11.0.

aaugustin:
CI failures look unrelated to my changes. Other PR suffer from the same failures. I am planning to wait until this is cleared and then rebase this PR.

ahopkins:
I really have no clue what happened to those TLS tests. They run fine locally. Some package somewhere must have changed. Regardless, I have disabled them for now so we can get the ball rolling again.

@aaugustin as always, your continued support and efforts with `websockets` and assistance with `sanic` is greatly appreciated.

ahopkins (commented):
Before we merge this, I want to make one tweak to add a `property` for `connection` that will handle any backwards compat. Will push to your branch.

ahopkins (approved):

PR #2610, opened 2022-11-29
Kill server early on worker error

ahopkins:
This will be to kill the worker process early on an error and to allow the deadlock restriction to be bumped to a higher value.

### DIFF
--- a/sanic/worker/manager.py
+++ b/sanic/worker/manager.py
@@ -137,10 +137,11 @@ class WorkerManager:
             "___."
         )
         while not self._all_workers_ack():
-            if (
-                self.monitor_subscriber.poll(0.1)
-                and self.monitor_subscriber.recv() == "__TERMINATE__"
-            ):
+            if self.monitor_subscriber.poll(0.1):
+                monitor_msg = self.monitor_subscriber.recv()
+                if monitor_msg != "__TERMINATE_EARLY__":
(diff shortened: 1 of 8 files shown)


sjsadowski (commented):
Couple of minor things, I don't think they're blockers for merging though.

tests/test_unix_socket.py
```diff
@@ -179,6 +179,7 @@ async def client(app: Sanic, loop: AbstractEventLoop):
             assert r.status_code == 200
             assert r.text == os.path.abspath(SOCKPATH)
     finally:
+        await sleep(0.2)
```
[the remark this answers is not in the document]
ahopkins: Yeah, since this test is using the async client (so that it does both sides on the same loop), if there is not a delay it proceeds to `stop` before the request handler ever executes. `0.1` probably is fine too, but I figure there is little to no difference for us humans, and that `0.1` might solve some headaches in CI.

sanic/worker/manager.py
```diff
@@ -18,7 +17,7 @@
 
 
 class WorkerManager:
-    THRESHOLD = 50
+    THRESHOLD = 300  # == 30 seconds
```
ahopkins, line 20: Thoughts on this interval?

sjsadowski: I think it's fine. 30 seconds seems reasonable.

PR #2612, opened 2022-11-29
Test for 3.11 support

ahopkins:


ahopkins:
I think there are some tweaks we need to make to the GH files. I honestly want to do bare minimum so we can rip them out next year.

.github/workflows/pr-linter.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         config:
-          - { python-version: 3.8, tox-env: lint}
+          - { python-version: 3.10, tox-env: lint}
```
ChihweiLHBird, line 19: Making `3.10` a `str` object, `"3.10"` may fix this PR.

ahopkins: Ahh, nice catch. Feel free to make the change and push it if you want.

ChihweiLHBird: Thank you, @ahopkins

ChihweiLHBird:
> I think there are some tweaks we need to make to the GH files. I honestly want to do bare minimum so we can rip them out next year.

Why we will be able to remove them next year? Is there any successor of them?

ahopkins:
Something much simpler like what sanic-ext uses. Plus I've been playing with getting rid of codecov and just using coverage.py and posting back the html.

PR #2618, opened 2022-12-07
Inspector by host and port

ahopkins:
This allows the inspector to be host and port and not just path to app:

```
$ sanic localhost:6457 --inspect
```


ahopkins:
Closed in favor of https://github.com/sanic-org/sanic/pull/2626

PR #2620, opened 2022-12-08
Allow for TLS on inspector

ahopkins:
This will allow someone to run the inspector with a TLS cert making it one step closer to being more secure to run against remote hosted applications.

I would still not advise this yet since there is no authentication mechanism yet.


ahopkins:
Closed in favor of https://github.com/sanic-org/sanic/pull/2626

PR #2621, opened 2022-12-08
Add a SIGKILL to second ctrl+c

ahopkins:
Hitting `ctrl+c` sends SIGINT to do a graceful shutdown on each worker process. Subsequent `ctrl+c` currently do nothing.

This PR adds functionality so that the second `ctrl+c` will send a SIGKILL to the processes for immediate shutdown.

### DIFF
--- a/sanic/worker/manager.py
+++ b/sanic/worker/manager.py
@@ -1,5 +1,6 @@
 import os
 
+from contextlib import suppress
 from signal import SIGINT, SIGTERM, Signals
 from signal import signal as signal_func
 from typing import List, Optional
@@ -183,6 +184,11 @@ class WorkerManager:
         raise ServerKilled
 
     def shutdown_signal(self, signal, frame):
+        if self.terminated:
+            logger.info("Shutdown interrupted. Killing.")
+            with suppress(ServerKilled):
+                self.kill()
+
         logger.info("Received signal %s. Shutting down.", Signals(signal).name)
         self.monitor_publisher.send(None)
         self.shutdown()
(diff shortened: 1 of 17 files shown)


prryplatypus (approved):

sanic/worker/manager.py
```diff
@@ -183,6 +183,11 @@ def kill(self):
         raise ServerKilled
 
     def shutdown_signal(self, signal, frame):
+        if self.terminated:
+            logger.info("Shutdown interrupted. Killing.")
+            with suppress(ServerKilled):
+                self.kill()
+
```
prryplatypus, line 190: Do we want a final `return` at the end of this if statement?

Tronic:
Did you check how this behaves on Windows? IIRC, terminate already works the same as kill, and also KeyboardInterrupt may be received by any of the processes. Also, might consider only handling second Ctrl-C after a slight delay, as it is quite easy to accidentally hit it twice, but this is also not of that much importance (I don't mind if there is no delay required).

ahopkins:
> Did you check how this behaves on Windows? IIRC, terminate already works the same as kill, and also KeyboardInterrupt may be received by any of the processes. Also, might consider only handling second Ctrl-C after a slight delay, as it is quite easy to accidentally hit it twice, but this is also not of that much importance (I don't mind if there is no delay required).

Nope, will check.

sanic/worker/manager.py
ahopkins: Not needed. `self.kill` ends with a `raise`

PR #2623, opened 2022-12-08
Wait for new process to ACK before termination of old on restart

### LINKED ISSUE #2619: restart workers online (graceful restart) to hot reload, in production environment.

Condition:  keep service available any time.  

# ---- client code----

```py
async def run():                                                                                        
    while 1:
        await post('http://127.0.0.1:8000/') 
```

when called app.m.restart("__ALL_PROCESSES__") in a worker, sanic crashed.

# ---- server code ----
```py
@app.post("/")
async def handler(request):
    app.m.restart('__ALL_PROCESSES__')
    return response.text('ok')


if __name__ == "__main__":
    app.run(debug=True, workers=2)
```

graceful restarting and reduce the effect when restarting.
my messy describe:
1. graceful restart workers; restart all workers will not crash,  if only 1 worker, block a little while (if worker not started yet) is ok.
2. a way to graceful restart worker one by one, code eg:
    woker_names = tuple(app.m.workers.keys())
    for woker_name in worker_names:
        ret_val = app.m.restart(worker_name)
        # here, the worker has been graceful restarted, ret_val is meaningful
3. may combine the above 2, when restarting all workers,  50% workers restarting, 50% old workers keep serving

### Additional context

simplify the api,

```py
app.m.restart('__ALL_PROCESSES__')   => app.m.restart_all()
```

thanks.

ahopkins:
Current logic terminates a process (sends SIGINT), then starts a new process. This PR reverses the order so that it starts the new process, waits for ACK and then sends the SIGINT.

Closes #2619

### DIFF
--- a/sanic/application/ext.py
+++ b/sanic/application/ext.py
@@ -33,7 +33,7 @@ def setup_ext(app: Sanic, *, fail: bool = False, **kwargs):
         return
 
     if not getattr(app, "_ext", None):
-        Ext: Extend = getattr(sanic_ext, "Extend")
+        Ext = getattr(sanic_ext, "Extend")
         app._ext = Ext(app, **kwargs)
 
         return app.ext
--- a/sanic/application/ext.py
+++ b/sanic/application/ext.py
@@ -8,11 +8,6 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from sanic import Sanic
 
-    try:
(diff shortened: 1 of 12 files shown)


ahopkins:
Going to put this on hold until #2626 is in. Using the new pattern there, we can remove this from a config setting and do it as a command feature.

PR #2626, opened 2022-12-13
Move to HTTP Inspector

ahopkins:
Replace #2620 and #2618 

This PR does the following:

1. Moves the `Inspector` to a Sanic app from a simple TCP socket with a custom protocol
2. Deprecates the `--inspect` commands so that we can nest them: `sanic inspect ...`
3. Adds the ability to ping the inspector by host/port (remote access now possible)
4. Adds TLS support
5. Adds authentication
6. Adds `inspector_class` to the `Sanic` instance similar to some other classes to allow for simple extendability

Because there is a simple pattern to extend, it is possible to add your own custom commands:

```python
class CustomInspector(Inspector):
    async def foo(self, one):
        return "This value will be output back to terminal"


app = Sanic("TestApp", inspector=True, inspector_class=CustomInspector)
```
```
sanic inspect foo --one=1
```

### DIFF
--- a/sanic/cli/app.py
+++ b/sanic/cli/app.py
@@ -98,16 +98,32 @@ Or, a path to a directory to run as a simple HTTP server:
         except ValueError as e:
             error_logger.exception(f"Failed to run app: {e}")
         else:
-            if self.args.inspect or self.args.inspect_raw or self.args.trigger:
+            if (
+                self.args.inspect
+                or self.args.inspect_raw
+                or self.args.trigger
+                or self.args.scale is not None
+            ):
                 os.environ["SANIC_IGNORE_PRODUCTION_WARNING"] = "true"
             else:
                 for http_version in self.args.http:
                     app.prepare(**kwargs, version=http_version)
 
-            if self.args.inspect or self.args.inspect_raw or self.args.trigger:
-                action = self.args.trigger or (
-                    "raw" if self.args.inspect_raw else "pretty"
-                )
+            if (
+                self.args.inspect
+                or self.args.inspect_raw
+                or self.args.trigger
+                or self.args.scale is not None
+            ):
+                if self.args.scale is not None:
+                    if self.args.scale <= 0:
+                        error_logger.error("There must be at least 1 worker")
+                        sys.exit(1)
+                    action = f"scale={self.args.scale}"
+                else:
+                    action = self.args.trigger or (
+                        "raw" if self.args.inspect_raw else "pretty"
+                    )
                 inspect(
                     app.config.INSPECTOR_HOST,
                     app.config.INSPECTOR_PORT,
--- a/sanic/cli/app.py
+++ b/sanic/cli/app.py
@@ -13,8 +13,8 @@ from sanic.application.logo import get_logo
 from sanic.cli.arguments import Group
 from sanic.cli.base import SanicArgumentParser, SanicHelpFormatter
 from sanic.cli.inspector import make_inspector_parser
+from sanic.cli.inspector_client import InspectorClient
 from sanic.log import Colors, error_logger
-from sanic.worker.inspector import InspectorClient
 from sanic.worker.loader import AppLoader
 
 
@@ -85,7 +85,7 @@ Or, a path to a directory to run as a simple HTTP server:
         elif parse_args == ["-v"]:
             parse_args = ["--version"]
 
-        if not legacy_version and not self.inspecting:
+        if not legacy_version:
             parsed, unknown = self.parser.parse_known_args(args=parse_args)
             if unknown and parsed.factory:
                 for arg in unknown:
@@ -149,7 +149,7 @@ Or, a path to a directory to run as a simple HTTP server:
             for arg in unknown:
                 if arg.startswith("--"):
                     key, value = arg.split("=")
-                    setattr(self.args, key.strip("-"), value)
+                    setattr(self.args, key.lstrip("-"), value)
 
         kwargs = {**self.args.__dict__}
         host = kwargs.pop("host")
--- a/sanic/cli/app.py
+++ b/sanic/cli/app.py
@@ -3,7 +3,6 @@ import os
 import shutil
 import sys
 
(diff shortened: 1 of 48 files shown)


sanic/cli/app.py
```diff
@@ -58,16 +56,27 @@ def __init__(self) -> None:
...
         self.groups: List[Group] = []
+        self.inspecting = False
 
     def attach(self):
+        if sys.argv[1] == "inspect":
+            self.inspecting = True
+            self.parser.description = get_logo(True)
+            make_inspector_parser(self.parser)
+            return
+
```
prryplatypus, line 69: Can't we do this with subcommand groups instead?

sanic/cli/app.py
```diff
@@ -76,7 +85,7 @@ def run(self, parse_args=None):
         elif parse_args == ["-v"]:
             parse_args = ["--version"]
 
-        if not legacy_version:
+        if not legacy_version and not self.inspecting:
```
prryplatypus, line 88: This addition shouldn't be needed. We `return` above `if self.inspecting`

sanic/cli/app.py
```diff
@@ -140,6 +118,53 @@ def run(self, parse_args=None):
...
+        )
+
+    def _inspector(self):
+        args = sys.argv[2:]
+        self.args, unknown = self.parser.parse_known_args(args=args)
+        if unknown:
+            for arg in unknown:
+                if arg.startswith("--"):
+                    key, value = arg.split("=")
+                    setattr(self.args, key.strip("-"), value)
```
prryplatypus, line 152: Maybe better to use `lstrip` for consistency with the `startswith` check? Won't make much difference other than that though.

sanic/cli/base.py
```diff
@@ -0,0 +1,35 @@
...
+            return
+        super()._check_value(action, value)
+
+
+class SanicHelpFormatter(RawTextHelpFormatter):
+    def add_usage(self, usage, actions, groups, prefix=None):
+        if not usage:
+            usage = SUPPRESS
+            # Add one linebreak, but not two
+            self.add_text("\x1b[1A'")
```
prryplatypus, line 23: I assume this is the equivalent of `\n`? Does a simple `\n` not work?

sanic/worker/inspector.py
```diff
@@ -25,111 +34,156 @@ def __init__(
...
-    with socket(AF_INET, SOCK_STREAM) as sock:
-        try:
-            sock.connect((host, port))
-        except ConnectionRefusedError:
-            error_logger.error(
-                f"{Colors.RED}Could not connect to inspector at: "
-                f"{Colors.YELLOW}{(host, port)}{Colors.END}\n"
-                "Either the application is not running, or it did not start "
-                "an inspector instance."
+class InspectorClient:
```
prryplatypus, line 138: This class would probably be better in a module other than `worker`. Maybe `client` or similar.

tests/conftest.py
```diff
@@ -221,3 +221,14 @@ def sanic_ext(ext_instance):  # noqa
...
+
+@pytest.fixture
+def urlopen():
+    urlopen = Mock()
+    urlopen.return_value = urlopen
+    urlopen.__enter__ = Mock(return_value=urlopen)
+    urlopen.__exit__ = Mock()
+    urlopen.read = Mock()
+    with patch("sanic.worker.inspector.urlopen", urlopen):
+        yield urlopen
```
prryplatypus, line 234: Since this will only work for the `sanic.worker` namespace, probably better to move it inside the `tests/worker` folder.

sanic/cli/inspector.py
```diff
@@ -0,0 +1,47 @@
...
+        action="store_true",
+        help="Whether to output the raw response information",
+    )
+
+    subparsers = parser.add_subparsers(
+        action=SanicSubParsersAction,
+        dest="action",
+        description=(
+            "Run one of the below subcommands. If you have created a custom "
+            "Inspector instance, then you can run custom commands.\nSee ___ "
```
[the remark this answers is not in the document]
ahopkins: Yup, I do this intentionally since I will not write the docs and have a link until next week. Then I make a final PR before release with changelog info and fill in what is missing.

sanic/cli/app.py
ahopkins: I thought about this and doing a similar hack like this to allow a static and dynamic value in the same location. But I do not like what it does to the help docs, and it makes the whole CLI app bleed a bit too much. In this location I think it works okay. When we are talking about the module, not so much. By keeping the parser attachments separate, it is cleaner to maintain.

image

sanic/cli/app.py
ahopkins: Yup. Legacy. TY.

sanic/cli/base.py
ahopkins: No, it removes a line in the terminal. Argparse is adding `\n\n`, which is quite annoying. It makes a HUGE gap here.

image

If we do nothing though, it looks bad without a gap. The alternative approach is to just add the padding to the logo itself, but that is not ideal either. Not tied to this approach, and I am not sure how it will work on Windows (other ANSI escape codes seem to work well enough). Care to try?

sanic/worker/inspector.py
ahopkins: I will move it to `./cli`.

tests/conftest.py
ahopkins: `tests_cli.py` uses it too

sanic/cli/base.py
prryplatypus: Will give this a try tomorrow!

prryplatypus: This is how it looks on Windows:
imagen

prryplatypus: Similar in Windows (CMD):
imagen

ahopkins: perfect :ok_hand:

prryplatypus (approved):

PR #2628, opened 2022-12-16
Replace deprecated distutils.strtobool

### LINKED ISSUE #2388: distutils.strtobool deprecation

kijk2869:
Fix #2388 

Because setup.py also uses the function, code duplication inevitably occurred.
If there is a better solution, please suggest.

### DIFF
--- a/sanic/server/loop.py
+++ b/sanic/server/loop.py
@@ -1,11 +1,11 @@
 import asyncio
 import sys
 
-from distutils.util import strtobool
 from os import getenv
 
 from sanic.compat import OS_IS_WINDOWS
 from sanic.log import error_logger
+from sanic.utils import str_to_bool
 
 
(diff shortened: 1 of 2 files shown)


ahopkins (approved):
Duplication is 👌

PR #2636, opened 2022-12-18
Add priority to register_middleware method

### LINKED ISSUE #2596: Missing priority functionality in app-wide middlware

### Describe the bug

In v22.9 the priority feature was added, but unfortunately it appears it was overlooked and there is priority parameter to `register_middleware()` for `Sanic`.

I am looking to use this feature because I have a Blueprint-specific middleware which appears to run before the application-wide middleware. In my case the application-wide middleware does authorization and updates `request.ctx` which is then used in the Blueprint-specific middleware.

### Code snippet

_No response_

### Expected Behavior

_No response_

### How do you run Sanic?

ASGI

### Operating System

Windows

### Sanic Version

Sanic 22.9.1; Routing 22.8.0

### Additional context

_No response_

ahopkins:
Closes #2596

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -61,7 +61,7 @@ from sanic.exceptions import (
     URLBuildError,
 )
 from sanic.handlers import ErrorHandler
-from sanic.helpers import Default
+from sanic.helpers import Default, _default
 from sanic.http import Stage
 from sanic.log import (
     LOGGING_CONFIG_DEFAULTS,
@@ -299,7 +299,7 @@ class Sanic(BaseSanic, StartupMixin, metaclass=TouchUpMeta):
         middleware: Union[MiddlewareType, Middleware],
         attach_to: str = "request",
         *,
-        priority=0,
+        priority: Union[Default, int] = _default,
     ) -> Union[MiddlewareType, Middleware]:
         """
         Register an application level middleware that will be attached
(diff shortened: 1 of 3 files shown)


sanic/app.py
```diff
@@ -311,19 +316,37 @@ def register_middleware(
...
+        retval = middleware
+        location = MiddlewareLocation[attach_to.upper()]
+
+        if not isinstance(middleware, Middleware):
+            middleware = Middleware(
+                middleware,
+                location=location,
+                priority=priority if isinstance(priority, int) else 0,
+            )
+        elif middleware.priority != priority and isinstance(priority, int):
```
prryplatypus, line 328: Might be worth reversing these statements to make it clear we're comparing two `int`s.
```suggestion
        elif isinstance(priority, int) and middleware.priority != priority:
```

sanic/app.py
```diff
@@ -311,19 +316,37 @@ def register_middleware(
             **response** - Invoke before the response is returned back
         :return: decorated method
         """
-        if attach_to == "request":
+        retval = middleware
```
prryplatypus, line 319: Slightly tangential to this PR, but what's the point of returning the exact same object that we're receiving (even if we end up changing it/creating a new one)?

prryplatypus (approved):

sanic/app.py
ahopkins: Not necessary.

sanic/app.py
ahopkins: TBH, there is no reason IMO to return anything. But we do. If we are, it certainly should not be modified (which we need to do with `priority` ordering) since that would lead to some very hard to discover bugs.

PR #2640, opened 2022-12-25
ASGI websocket recv text or bytes

### LINKED ISSUE #2638: There is an obvious bug in ASGI WebsocketConnection of Sanic

### Describe the bug

I started my sanic app with UvicornWorker. The original websocket will become WebsocketConnection. When I call
the ws.recv function will report an error if bytes data is received at this time.
`KeyError：‘text’`
[https://github.com/sanic-org/sanic/blob/main/sanic/server/websockets/connection.py](url)
`    async def recv(self, *args, **kwargs) -> Optional[str]:
        message = await self._receive()

        if message["type"] == "websocket.receive":
            return message["text"]
        elif message["type"] == "websocket.disconnect":
            pass

        return None`
There is no data of bytes type processed here.

### Code snippet

_No response_

### Expected Behavior

_No response_

### How do you run Sanic?

ASGI

### Operating System

ubuntu

### Sanic Version

22.3

### Additional context

_No response_

ahopkins:
Closes #2638

### DIFF
--- a/sanic/server/websockets/connection.py
+++ b/sanic/server/websockets/connection.py
@@ -9,8 +9,10 @@ from typing import (
     Union,
 )
 
+from sanic.exceptions import InvalidUsage
 
-ASIMessage = MutableMapping[str, Any]
+
+ASGIMessage = MutableMapping[str, Any]
 
 
 class WebSocketConnection:
@@ -25,8 +27,8 @@ class WebSocketConnection:
 
     def __init__(
         self,
-        send: Callable[[ASIMessage], Awaitable[None]],
-        receive: Callable[[], Awaitable[ASIMessage]],
+        send: Callable[[ASGIMessage], Awaitable[None]],
+        receive: Callable[[], Awaitable[ASGIMessage]],
         subprotocols: Optional[List[str]] = None,
     ) -> None:
         self._send = send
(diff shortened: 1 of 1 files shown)


sanic/server/websockets/connection.py
```diff
@@ -47,7 +49,13 @@ async def recv(self, *args, **kwargs) -> Optional[str]:
         message = await self._receive()
 
         if message["type"] == "websocket.receive":
-            return message["text"]
+            try:
+                return message["text"]
+            except KeyError:
+                try:
+                    return message["bytes"].decode()
```
Tronic, line 56: Shouldn't bytes be received as is, without decoding? This is what the built-in server does, to support binary data.

ahopkins: I wanted to receive it for now without breaking the existing contract. I think we need to revisit this.

Tronic: Yes, we need to. I would think this being necessary for 23.12.1 release because we cannot have LTS processing it wrong like that. Apparently not many use ASGI mode with binary WebSockets because this wasn't caught earlier.

The binary format is specifically meant for binary data, for which my own use cases include MsgPack, audio streaming and other raw data that will never be able to be decoded into text (if not for a freak accident of chance that still breaks the software). Never JSON which can and should be sent in text mode instead.

ahopkins: Yeah, probably sooner than a year from now and back port it.

PR #2646, opened 2023-01-01
Separate ASGI request and lifespan callables

ahopkins:
I noticed some breaks in ASGI implementation. This should largely have no impact on how Sanic performs with ASGI, but creates a cleaner separation between the `ASGIApp` (which is now exclusively for request handling) and `Lifespan` (which now has top level treatment in the `Sanic.__call__` method).


sanic/asgi.py
```diff
@@ -77,35 +76,33 @@ async def shutdown(self) -> None:
...
+                    await self.send(
+                        {"type": "lifespan.startup.failed", "message": str(e)}
+                    )
+                else:
+                    await self.send({"type": "lifespan.startup.complete"})
+            elif message["type"] == "lifespan.shutdown":
+                try:
+                    await self.shutdown()
+                except Exception as e:
+                    error_logger.exception(e)
```
Kludex, line 99: Doesn't this shows the error message twice?

Kludex (approved):
:+1:

sanic/asgi.py
ahopkins: Well, yeah, kinda. It sends the message back to whatever the server logger is using. But this sends to Sanic's internal error logger including the traceback. There is an API to configure Sanic's loggers independent of whatever ASGI server is running, which may be logging to different locations.

PR #2662, opened 2023-01-24
Establish basic file browser and index fallback

### LINKED ISSUE #2661: Optional file browser

_No response_

An auto-generated file browser and `index.html` fallback.

1. When enabled, if serving a static dir check for existence of `index.html` and serve that
2. If it does not exist, and autoindex is also enabled, show a basic file browser

### Additional context

See #893

ahopkins:
Closes #2661 

This adds a new opt in pattern when a static directory is being served.

```py
app.static(..., autoindex=True)
app.static(..., index_name="index.html")

# also available within a handler...

@app.get("/somepath")
async def handler(_):
    return await file(..., autoindex=True, index_name="index.html")
```

These may be used together or separately.

- `autoindex` = Whether to show a basic HTML file browser
- `index_name` = The name of a fallback file in that directory to show, if any

The logic is as follows:
- if serving a dir:
    - if `index_name` set and exists:
        - return index file
    - else if `autoindex`:
        - show file browser

image
image


ahopkins:
@Tronic This is the fallback browser solution that I told you I had. LMK is this is what you had in mind.

sanic/response/convenience.py
```diff
@@ -226,16 +230,25 @@ async def file(
...
-        if _range:
-            await f.seek(_range.start)
-            out_stream = await f.read(_range.size)
-            headers[
-                "Content-Range"
-            ] = f"bytes {_range.start}-{_range.end}/{_range.total}"
-            status = 206
-        else:
-            out_stream = await f.read()
+    try:
```
Tronic, line 233: This function is extremely long. Might be a good time to refactor into smaller functions before growing it even more.

Tronic:
The file browser formats fairly lengthy HTML documents by string.format (if I read the code correctly). Maybe we at this time should consider a minimal templating engine, something like [html5tagger](https://github.com/Tronic/html5tagger) (my own minimalistic module for creating HTML from Python code), or some more sensible way of handling these?

As a bare minimum, filenames and such strings require escaping.

ahopkins:
This is very minimal stuff we're doing. It does not warrant a dependency IMO.

ahopkins:
Let's keep this in hold until we resolve the Discord conversation. In short, we are talking about potentially including a minimal lib for html rendering.

sanic/response/convenience.py
ahopkins: Nice idea. I'll make a separate PR for introducing `FileResponse` similar to our recent `JSONResponse`.

Tronic:
Another somewhat off-topic comment (sorry about my wandering mind). With a file browser Sanic is getting very close to something like `python -m http.server` (but with support for TLS and everything). Not saying that Sanic CLI should have it but 💭

ahopkins:
> Another somewhat off-topic comment (sorry about my wandering mind). With a file browser Sanic is getting very close to something like `python -m http.server` (but with support for TLS and everything). Not saying that Sanic CLI should have it but thought_balloon

It already does have this:

```
sanic /path/to/some/dir --simple
```

sanic/response/convenience.py
```diff
@@ -331,3 +346,94 @@ async def _streaming_fn(response):
...
+    def render(self) -> str:
+        doc = Document(title=self.TITLE, lang="en")
+        self._head(doc)
+        with doc.main:
+            self._headline(doc)
+            self._file_table(doc)
+        return str(doc)
+
+    def _head(self, doc: Document):
+        doc.head.title(self.TITLE).style(self.STYLE)
```
Tronic, line 395: The `Document` function already adds a title, so the only thing needed in `_head` is the inline style (notice that the `head` tag is also not needed in HTML5). Possibly refactor not to have it broken down to that many methods, only a few bigger parts (that one might wish to override I suppose).

For purposes of typing, you need `html5tagger.Builder`, which is what using `Document` or `E` returns. Possibly use `self.doc` to avoid having to pass it around?

sanic/response/convenience.py
```diff
@@ -331,3 +346,94 @@ async def _streaming_fn(response):
...
+        file_access: str,
+        file_size: str,
+    ):
+        first = E.span(icon, class_="icon").a(file_name, href=file_name)
+        doc.tr.td(first).td(file_access).td(file_size)
+
+    def _prepare_file(self, path: Path) -> Dict[str, Union[int, str]]:
+        stat = path.stat()
+        modified = datetime.fromtimestamp(stat.st_mtime)
+        is_dir = S_ISDIR(stat.st_mode)
```
Tronic, line 424: pathlib.Path has `path.is_dir()`

sanic/response/convenience.py
```diff
@@ -331,3 +346,94 @@ async def _streaming_fn(response):
         headers=headers,
         content_type=mime_type,
     )
+
+
+class AutoIndex:
+    STYLE = """
+        html { font-family: sans-serif; }
+        main { padding: 1rem; }
```
Tronic, line 354:
```python
    STYLE = dedent("""\
        html { font-family: sans-serif; }
        main { padding: 1rem; }
```

Tronic:
It's looking good, didn't have much to comment on code but I didn't try actually running it yet.

Tronic:
At the rate Sanic is evolving with your patches, I wouldn't be surprised if this soon also supported multiple return formats, and then offered a JS frontend that fetches file lists in JSON to avoid page reloads while streaming video files to browser for quick playback like [Droppy](https://drop.zi.fi/) can 😆

ahopkins:
> It's looking good, didn't have much to comment on code but I didn't try actually running it yet.

Yeah. I think we should take your idea a step further and make the styles for this and the error pages common. Probably it's own PR though. It's a nice touch.

sanic/response/convenience.py
ahopkins: I figured since we already are grabbing the stat flags 🤷‍♂️
I did not benchmark the difference though, but it has to be at least something more. I'm pretty sure pathlib would do the same lookup.

sanic/response/convenience.py
ahopkins: I added it explicitly because it was not inside head. When I put the generated code into a validation tool it yelled at me for not having it.

Tronic:
I think the file browser and autoindex serve very different functions, and I don't want anyone on the Internet browsing the files of my webapps (that simply have a static index.html at root to serve their frontend), so keep them separated.

Regarding filename encoding, that could get really tricky with Unicode (the mentioned Droppy allows uploads with Unicode NFD and saves them on disk, but can then no longer load/touch those files because filenames from URLs become NFC, but that might be some NodeJS internals and not what browsers do). I will look into this once I get to run this [an extremely busy day].

ahopkins:
I did some refactoring to break apart the autoindexing and allow for a new common module for our growing list of handlers and pages.

One thing I noticed, and I am not sure if this is new to this PR, but I am getting tracebacks in logs when I would not expect it.

sanic/response/convenience.py
ahopkins: image


Yup, under the hood, it is doing the same thing. Although, we may need to look at OS compat.

Tronic:
> I did some refactoring to break apart the autoindexing and allow for a new common module for our growing list of handlers and pages.

Great. I was also thinking of that. And it could also have a small Sanic header/footer with the S logo in inline SVG, if we want to push the branding a bit more.

ahopkins:
> Great. I was also thinking of that. And it could also have a small Sanic header/footer with the S logo in inline SVG, if we want to push the branding a bit more.

Thus far I have been reluctant to have anything super identifying it as Sanic. For example, we do not set a `Server` header. Granted, our use of emojis is a bit distinctive by itself and I have seen Sanic error pages in the wild that I could identify like that.

As for branding the autoindex and error pages, I would certainly like to hear opinions from more people.

Tronic:
There's that for testing. Logo SVG optimised very small (Inkscape, minifiers, manual editing).

sanic/worker/container.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def prepare(self, *apps) -> None:
+        for app in apps:
+            app.prepare(**app._early_prepare)
+
+    def serve(self) -> None:
+        from sanic import Sanic
+
+        primary = self.loader.load()
+        self.prepare(primary)
```
ahopkins, line 16: Oops, this file was the beginning of a different idea I had. Doesn't belong in this PR. I'll remove.

sanic/handlers/error.py
```diff
@@ -23,16 +20,21 @@ class ErrorHandler:
...
     def __init__(
         self,
         base: Type[BaseRenderer] = TextRenderer,
     ):
         self.cached_handlers: Dict[
             Tuple[Type[BaseException], Optional[str]], Optional[RouteHandler]
-        ] = {}
+        ] = {
+            **self.DEFAULT_HANDLERS  # type: ignore
+        }
```
ahopkins, line 37: I'm thinking rather than this being the pattern, we should register in the app instance so that:
1. custom error handlers still have this feature 
2. we can make the directory handler a passed arg like error handler is for overriding

Tronic:
Regarding Unicode issues, which was discussed earlier, I've run my tests. No problems what-so-ever. Sanic works both with NFC and NFD, and also with emoji filenames without raising a sweat. URLs display same bytes as filename on disk, and while downloading the browser converts and saves everything in NFD on a Mac. Possibly they also handle that for downloads on Windows and Linux (but to NFC as is used on those systems).

TLDR: Everything Just Works.

sanic/pages/base.py
```diff
@@ -1,25 +1,25 @@
...
 
 class BasePage(ABC):
     BASE_STYLE = dedent(
         """
-        body { margin: 0; font: 1.2rem sans-serif; }
+        body { margin: 0; font: 16px sans-serif; }
```
Tronic, line 13: The default size of 16px (i.e. 1.0rem) is a bit small. Did you need to make it smaller to make something fit?

ahopkins: Matching the user guide. Base style should itself be in px.

Tronic: There is "the style guide" for HTML5? Anyway, that would apply to the `html` element that the `rem` unit is then based on. `body` can use em/rem units just fine (so can html, but maybe it is bad style).

Are you okay with keeping the slightly larger size?

ahopkins: I think the bigger style looks clunky and amateurish. But, I am no designer. I am happy to be proven wrong. We don't have a `html { }` style def. Wherever we put the root size I would like to be in px and use the relative sizes thereafter.

Tronic: For a professional look it needs a better font (that we won't supply because it'd need extra files), amateur sites would be using the default I guess. Font size should depend more on filling the screen (but this obviously depends on user device and zoom setting). The niceback traceback viewer was using a smaller font, so is Github for similar reasons.

It may be a personal issue, because despite using a 165cm (65") screen I need to zoom in to easily read small text when not using glasses but also the full width layout left a lot of empty space between filenames and sizes, and emoji icons hard to see. The Droppy file browser I mentioned uses 1.5rem (24px) to avoid this.

As for px vs. rem/em I find it better to avoid the use of pixels entirely because they as a unit of measurement have lost their relevance with mobile devices, desktop scaling, HiDPI and browser zoom function. A layout using **only** rem/em can be scaled simply by adjusting the html font size (noting that body is not document root even if html is not being styled).

TLDR: After discussion I suggest sticking with rem, but leaving the actual size open for testing with different situations (does the header and h1 still fit on one line on all devices, in typical situations).

ahopkins: If you feel strongly about that, I am not going to disagree. I did a quick check of a couple sites

- Dropbox: 14px on both the body and the line item
- Google Drive: 13px on the item
- [musicForProgramming](https://musicforprogramming.net) - based upon vw at different breakpoints

To address your whitespace concern, what about just making it more narrow then?

image

Tronic: Yeah, so why waste nearly all of your screen space to add huge black borders?

Tronic: Not feeling strongly, definitely not going to argue it further (some bikeshedding here, LOL). Just seems like a sensible thing to increase the size instead. And like I said, prefer 1rem for the font size on body element if you wish to keep the default size (of define font-family only).

ahopkins: I am resolving this. I reverted to something similar to your original, but also made it somewhat scalable in the table. I tested it on a bunch of sizes, and am calling it good enough for now.

ahopkins:
@Tronic I think this is pretty much ready. Just needs some tests.

sanic/mixins/routes.py
```diff
@@ -702,6 +712,8 @@ def static(
         content_type: Optional[bool] = None,
         apply: bool = True,
         resource_type: Optional[str] = None,
+        autoindex: bool = False,
+        index_name: Union[str, Default] = _default,
```
Tronic, line 716: I've been thinking of how those arguments could be better. I don't associate autoindex with file browser but I see that Apache and Nginx have such named modules, where Apache also includes index.html while Nginx apparently does not.

If we take the two features being closely related, perhaps overload a single argument (autoindex) for both? This also makes it easier to pass it around (I see that the two appear in five methods' arguments in this PR).

```python
autoindex=False                 # Disable both
autoindex=True                  # Enable file browser only
autoindex="index.html"          # Only index.html, no file browser
autoindex=(True, "index.html")  # Enable both, be explicit on index.html
```

The rule then would be to take a list/tuple of arguments by type:
* bool for file browser
* str for filename(s) to look for

For convenience allow leaving out the tuple when there is only one argument. Being explicit on index.html handling with the file manager avoids accidentally enabling them where only file listings are intended (like with the nginx module). This is more lightweight than the `ssl=` argument of `app.run` that is similarly overloaded for many things and such a list should be more than sufficient for this (using a dict instead seems overkill, bad for syntax).

Perhaps you have other idea of simplification, but the current two separate arguments and how they function are itching me in a way that says that it probably isn't quite right.

ahopkins: Interesting. I do get the concept, and think that `autoindex` is the right term since (as you pointed out) we are not inventing it. I also am not entirely happy with the `index_name` argument. It does certainly need to be an implementation detail. Such overloading may be confusing, but if we type annotate it properly with `@overload` and put sufficient docs in for it, maybe not.

ahopkins: The problem that I have is that:

```py
autoindex=("index.html", True)
```

That is really what we want to "normal" usecase to be when implementing this. File browser should not be enabled, except when explicit. And, enabling both the browser and `index.html` fallback is a bit cumbersome.

ahopkins: Hmmm.... I don't know. This feels like a complication. Nginx is nice and explicit on both arguments: [autoindex](https://nginx.org/en/docs/http/ngx_http_autoindex_module.html) and [index](https://nginx.org/en/docs/http/ngx_http_index_module.html#index).

ahopkins: What about making it two args, but you only need two in certain circumstances

```python
autoindex=False                 # Disable both, DEFAULT
autoindex=True                  # Enable both
autoindex="index.html"          # Enable both, more explicit form of previous
autoindex=("index.html", "index.txt")
                                # Try in order (browser still enabled as last resort)
autoindex=True, browser=False   # Index only, explicit disable browser
```

Tronic: I don't like enabling file browser "by default". `--simple` handles that use case naturally (enabling both), but I think it would rarely be needed in webapps.

Considering modern webapps, that serve a static front (something generated by React/VueJS/etc, or written by hand) at site root and only make API calls to Sanic, the typical use would look like:

```python
app = Sanic(__name__)
app.static("/", staticdir, autoindex="index.html")

@app.websocket("/ws")  # or some JSON methods, or a Blueprint for the whole API
...
```

This webapp should not show any file listings even if the user happens to find a static directory URL other than the root (assuming there is index.html only at root).

Being explicit about both the file browser `True` and the index filename(s) keeps things secure and unsurprising. If both are needed, they are still easily enabled manually (that's why a tuple rather than dict or two kwargs). Agreed, giving the name "index.html" is a bit superfluous because it is always called that, but again it is self-documenting and has a clear meaning.

ahopkins: Explicit APIs that are self-documenting are better than implicit. The tuple method can lead to confusion, and/or feel somewhat "magical" in that it is not clear on the surface what it is doing.

What if we just rename the arguments?
- `autoindex => browser` - Whether to fallback to browser
- `index_name => index` - Single or list to be checked in order

Tronic: That works for me. But I do prefer only keeping `autoindex`, for the reasons in my previous message and in the initial proposal (first message in this thread).

sanic/handlers/directory.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def handle(
+        self, directory: Path, autoindex: bool, index_name: str, url: str
+    ):
+        index_file = directory / index_name
+        if autoindex and (not index_file.exists() or not index_name):
+            return self.index(directory, url)
+
+        if index_name:
+            return file(index_file)
```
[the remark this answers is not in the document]
Tronic: Ideally the parsing would be done during startup and not in handler, mostly to catch errors early, but of course Gotta Go Fast!

Also, using `file_stream` would be safer as some files can be very large (unless I am mistaken and `file` nowadays does streaming too, haven't followed the development that closely). The same would go for static files in general, not only this snippet, and is probably something for another PR.

sanic/handlers/directory.py
```diff
@@ -0,0 +1,71 @@
...
+    ):
+        for file_name in index:
+            index_file = directory / file_name
+            if index_file.is_file():
+                return await file(index_file)
+
+        if autoindex:
+            return self.index(directory, url)
+
+        raise NotFound(f"{str(directory)} is not found")
```
Tronic, line 35:
```suggestion
        raise NotFound(f"{directory.as_posix()} is not found")
```
Forward slashes also on Windows

sanic/mixins/routes.py
```diff
@@ -702,6 +704,8 @@ def static(
         content_type: Optional[bool] = None,
         apply: bool = True,
         resource_type: Optional[str] = None,
+        autoindex: bool = False,
+        index: Optional[Union[str, Sequence[str]]] = None,
```
Tronic, line 708: If we are going with these two separate arguments, better name autoindex to something that does NOT get mixed with the other argument. Now one would assume that `autoindex=True` is needed for `index` to have an effect.

Maybe the whole thing could be named FolderViewer, FileListing, FileExplorer or something alike, because it doesn't actually view any files and isn't a (web) browser, and again that sort of naming would be fully self-explanatory.

Out of those that I could think of, I might prefer `FileListing` for the class, `response.file_listing` and simply `listings` as the kwarg here.

If you wish to avoid the renaming in other places, at least change the argument to `browser`, to match `response.file_browser` and to avoid the aforementioned confusion between the two kwargs (maybe you already suggested this name in an earlier thread).

The possibility for multiple index files is very much unnecessary (everyone only ever want index.html, if anything). I added it as a side product of the single argument already requiring a tuple. Removing the sequence option would simplify typing and documentation, even if it only saves a couple of lines in implementation.

sanic/handlers/directory.py
```diff
@@ -0,0 +1,71 @@
...
+
+from sanic.exceptions import NotFound
+from sanic.pages.autoindex import AutoIndex, FileInfo
+from sanic.request import Request
+from sanic.response import file, html, redirect
+
+
+class DirectoryHandler:
+    def __init__(self, debug: bool) -> None:
+        self.debug = debug
```
Tronic, line 17: Why, anyway, is debug argument copied all around Sanic, instead of always accessing it from config?

Passing around variables should generally be avoided, so this (wide) practice of Sanic has always been a mystery to me. This PR seems to add about four more such variables (the two we've discussed elsewhere, debug, and the `directory_handler` argument). There's gotta be better ways to do this but while a bigger refactoring is not a nice idea, I would to a highest degree resist adding even more.

sanic/mixins/routes.py
ahopkins: > Now one would assume that autoindex=True is needed for index to have an effect.
It's not. And, if the current implementation requires it then that's a mistake that will be fixed in testing.

As for naming, autoindex is quite literally what's in use. I don't disagree that it is misleading in the larger scheme though. Rather, it's parity with nginx that drove that naming.

As for single v multiple, it seems like a familiar option for a couple lines of code. It totally feels like something that someone would ask for. Not tied to keeping it, but I also don't see it as a burden to maintain.

sanic/handlers/directory.py
Tronic: FWIW, Java coders apparently invented Singleton Pattern to covertly use global variables (shh, don't tell anyone), and webfront devs are all over a shiny new thing "state management" now because they too were passing (too many) things by arguments from component to another just like Sanic is now, and a state manager conveniently provides the whole application access to them without any arguments being passed through long chains of components.

I am pretty sure that we could very well be using a module-level storage somewhere to get access to app or other relevant objects from which the config, routes and other such things can be extracted.

sanic/app.py
```diff
@@ -213,6 +217,9 @@ def __init__(
         self.blueprints: Dict[str, Blueprint] = {}
         self.configure_logging: bool = configure_logging
         self.ctx: Any = ctx or SimpleNamespace()
+        self.directory_handler: DirectoryHandler = (
+            directory_handler or DirectoryHandler(self.debug)
+        )
```
Tronic, line 222: Suggest leave out this option entirely, keep it hard-coded at least for now, such that easier to change later.

Tronic: As mentioned on Discord, it would make a lot more sense to me to instantiate for each `app.static()` also handling the autoindex/index args there directly, rather than at runtime when a request is being handled.

ahopkins: That just feels very inconsistent.

Tronic:
Ideally this would be called in a normal context rather than at exception handling, such that if the directory listing crashes, it still lands on a pretty error page. I believe this is due to sanic/mixins/route.py not checking whether a path is a file before trying to serve it as a file first.

I am not quite sure whether that would be fixable by moving the call there, and in any case it might best be left for another PR, as @ahopkins might have suggested on discord (I took that specifically for the `raise ... from None` thing but thinking now it could've referred to this too).

sanic/handlers/directory.py
ahopkins: That is not possible because we _expressly_ allow for multiple applications. Therefore, while we have `Sanic.get_app()` that allows app developers this convenience, it is not something that Sanic internals can make use of.

sanic/app.py
Tronic: Inconsistent against what?

sanic/mixins/routes.py
Tronic: I mean it is misleading **locally**, with app.static having two adjacent arguments index arguments that intuitively *seem* like the first one enables enables the second one.

Since for that reason `autoindex` needs to be changed, we might as well be consistent with how the file browser feature is called all around Sanic, and only use one name for it (suggestions above, but file browser is also OK, now that autoindex is not available).

sanic/handlers/directory.py
Tronic: Okay, in that case we should be passing the app (or request from where one can get the app) around more often than a list of too many kwargs.

I didn't properly look at the whole architecture yet, but is there a chance of removing this entire DirectoryHandler class? To me the current implementation looks like it only needs the handle function whose code should be part of the static handler, and then returning `response.file_browser` in some cases.

Tronic: I.e. probably what its handle function currently does (returns index file or returns browser, a few lines of code) should go into `sanic.mixins.routes._static_request_handler` instead.

sanic/mixins/routes.py
ahopkins: How about `directory_view=True`. I would like to avoid "folder" because I do not think we use that term anywhere.

sanic/handlers/directory.py
ahopkins: Well, now that it has been removed from `file(...)` it does change the architecture a bit.

sanic/app.py
ahopkins: Our existing use of "Handler" objects.

sanic/mixins/routes.py
Tronic: It's up to you, possibly also dirview or other abbreviation, but rename the response and other parts accordingly (there is a project-wide rename function in VS Code that helps but it misses docstrings/comments and those chains of passing the same thing through other functions, so global find/replace could also be helpful).

TLDR: same name for all.

sanic/app.py
Tronic: Okay... I wasn't aware of other than ErrorHandler and still not quite sure what the pattern is. But I am going to mark this resolved, in hopes for the whole class to eliminated as discussed in another thread, and to keep this discussion in that thread.

sanic/mixins/static.py
```diff
@@ -0,0 +1,351 @@
...
+            If this is an integer, this represents the threshold size to
+            switch to :func:`StreamingHTTPResponse.file_stream`
+        :param name: user defined name used for url_for
+        :param host: Host IP or FQDN for the service to use
+        :param strict_slashes: Instruct :class:`Sanic` to check if the request
+            URLs need to terminate with a */*
+        :param content_type: user defined content type for header
+        :return: routes registered on the router
+        :rtype: List[sanic.router.Route]
+        """
```
ahopkins, line 74: Need to add here...

sanic/pages/base.py
```diff
@@ -0,0 +1,46 @@
...
+                self._sanic_logo()
+            if self.debug:
+                self.doc.div(f"Version {VERSION}")
+
+    @abstractmethod
+    def _body(self) -> None:
+        ...
+
+    def _sanic_logo(self) -> None:
+        self.doc.a(HTML(SVG_LOGO), href="https://sanic.dev", target="_blank")
```
Tronic, line 46:
```suggestion
        self.doc.a(HTML(SVG_LOGO), href="https://sanic.dev", target="_blank", referrerpolicy="no-referrer")
```
Avoid tracking of sites using Sanic (for privacy).

Tronic: Needs a run of black to format.

sanic/pages/styles/DirectoryPage.css
```diff
@@ -0,0 +1,67 @@
...
+
+table.autoindex td:last-child {
+    text-align: right;
+}
+
+@media (min-width: 1143px) {
+    table.autoindex {
+        font-size: 1.25rem;
+    }
+}
```
Tronic, line 61: Should ideally do these only on base page and via html element font size (that then scales everything). Responsive design done from multiple places gets tricky with potentially conflicting rules. No change request for now, I need to look at this in browser first, and need this comment as a reminder.

sanic/pages/styles/DirectoryPage.css
```diff
@@ -0,0 +1,67 @@
...
+    font-family: monospace;
+}
+
+table.autoindex tr {
+    display: flex;
+}
+
+table.autoindex tr:hover {
+    background-color: #ddd;
+}
```
Tronic, line 39: Consider using #8884 (transparency) and avoid the media selector altogether. Otherwise, keep the default and the selector adjacent in the CSS code whenever conveniently possible.

sanic/mixins/static.py
```diff
@@ -0,0 +1,351 @@
...
+        stream_large_files: Union[bool, int] = False,
+        name: str = "static",
+        host: Optional[str] = None,
+        strict_slashes: Optional[bool] = None,
+        content_type: Optional[str] = None,
+        apply: bool = True,
+        resource_type: Optional[str] = None,
+        directory_view: bool = False,
+        directory_index: Optional[Union[str, Sequence[str]]] = None,
+        directory_handler: Optional[DirectoryHandler] = None,
```
Tronic, line 47:
```suggestion
        index: Optional[Union[str, Sequence[str]]] = None,
        directory_view: bool = False,
        directory_handler: Optional[DirectoryHandler] = None,
```
The index parameter needs to be differently named to avoid the original confusion I was referring to, and also because I need to type it in all my apps. So, keeping it as `index`, and reordered so that it doesn't get left between the two directory ones.

sanic/pages/base.py
ahopkins: I meant to ask you how to add that.

sanic/mixins/static.py
ahopkins: Code completion helps here :wink:

Tronic: I need to fit my webapp in a Tweet along with installation instructions of the framework and how to run the server (I have done that with Sanic).

Tronic: Regarding autocompletion, I got directory_handler when I wanted directory_view, so that doesn't exactly save you either.

sanic/mixins/static.py
```diff
@@ -0,0 +1,358 @@
...
+
+                    if not stats:
+                        stats = await stat_async(file_path)
+                    if stats.st_size >= threshold:
+                        return await file_stream(
+                            file_path, headers=headers, _range=_range
+                        )
+                return await file(file_path, headers=headers, _range=_range)
+        except IsADirectoryError:
+            return await directory_handler.handle(request, request.path)
```
Tronic, line 316:
```suggestion
        except (IsADirectoryError, PermissionError):  # Windows gives the latter trying to open a folder
            return await directory_handler.handle(request, request.path)
```
May cause issues with actual permission errors...

sanic/pages/styles/DirectoryPage.css
ahopkins: Removed media query as no longer necessary

sanic/mixins/static.py
ahopkins: GTK

sanic/pages/css.py
```diff
@@ -0,0 +1,35 @@
...
+
+def _extract_style(maybe_style: Optional[str], name: str) -> str:
+    if maybe_style is not None:
+        maybe_path = Path(maybe_style)
+        if maybe_path.exists():
+            return maybe_path.read_text()
+        return maybe_style
+    maybe_path = CURRENT_DIR / "styles" / f"{name}.css"
+    if maybe_path.exists():
+        return maybe_path.read_text()
```
Tronic, line 17:
```suggestion
            return maybe_path.read_text(encoding="UTF-8")  # Charset required on Windows
        return maybe_style
    maybe_path = CURRENT_DIR / "styles" / f"{name}.css"
    if maybe_path.exists():
        return maybe_path.read_text(encoding="UTF-8")
```

sanic/pages/directory_page.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def _headline(self):
+        """Implement a heading with the current path, combined with
+        breadcrumb links"""
+        with self.doc.h1(id="breadcrumbs"):
+            p = self.url.split("/")[:-1]
+
+            for i, part in enumerate(p):
+                if i == 0:
+                    part = "🏠"
```
Tronic, line 42: This doesn't belong here, it belongs to CSS.

sanic/pages/directory_page.py
```diff
@@ -0,0 +1,59 @@
...
+                path = "/".join(p[: i + 1]) + "/"
+                self.doc.a(part, href=path)("/")
```
Tronic, line 44:
```suggestion
                with self.doc.a(href=path):
                    self.doc.span(part, class_="dir").span("/", class_="sep")
```
You should be able to then style it in CSS the way you want it, but slashes are clickable.

Tronic (dismissed):
LGTM

PR #2663, opened 2023-01-24
Additional accept functionality

ahopkins:
This adds some convenience to accept parsing:

- An `Accept` object has both `has_wildcard` (at least one) and `is_wildcard` (both parts)
- Add `AcceptContainer.match_explicit` (no wildcard matching)
- Add `AcceptContainer.find_first` (see below)

```python
header = "application/json;q=0.8,text/csv;q=0.9,*/*;q=0.5"
acceptable = parse_accept(header)
priority = acceptable.find_first(["application/json", "text/csv"])
assert priority.match("text/csv")
```

---

There is some discussion that started on Discord server about what should the `in` operator do: `match` or `match_explicit`. TBD.


Tronic:
The API is all different from before and from the original post on this PR, even type names were refactored. See the documentation in code for why there need to be three different ways to do the matching.

The biggest remaining problem, it would seem, is the handling of missing accept header, which leads to repeating a pattern like
```python
# Check that either the client didn't specify accept header or that it matches
is_acceptable = not request.accept or request.accept.match("some/type")
```
This would be solved by defaulting `*/*;q=0` as the header value, rather than an empty string, where the client supplies no header. But I can see why Sanic injecting its own ideas of headers into public accessors could be problematic (even if well documented).

Tronic:
I tested browsers and found out that they all give q=0.9 for xml, and q=0.8 for `*/*`. This allows for an alternative implementation of this accept logic that has no special treatment for the wildcard but where the `choose` function out of equally q-valued options in the order of preference of Sanic (first the route format if known, then all supported formats).

* Browsers would still always get html because they support nothing else (other than via wildcard).
* Other clients, incl. curl, requests, `fetch()` and XHR send by default `*/*` (q=1) and would get Sanic's preferred format (route, ...)

Neither of those cases would change. Only with a client that explicitly lists multiple supported formats would then not get the first/best one that it lists but the route format, provided that the route format is of highest possible q as well.

I will look into this for feasibility (correctness, simplicity of implementation/docs, etc).

sanic/headers.py
```diff
@@ -35,141 +35,78 @@
...
-            return method(self.qvalue, other.qvalue)
-        except (AttributeError, TypeError):
-            return NotImplemented
-
-    @parse_arg_as_accept
-    def __lt__(self, other: Union[str, Accept]):
-        return self._compare(other, lambda s, o: s < o)
+        self.q = float(params.get("q", "1.0"))
+        self.params = params
+        self.str = f"{type_}/{subtype}"
```
ahopkins, line 57: `self.str` to `self.value`

More explicit.

sanic/headers.py
```diff
@@ -35,141 +35,78 @@
...
-            if allow_subtype_wildcard
-            else (
-                self.subtype.match(other.subtype)
-                and not self.subtype.is_wildcard
-                and not other.subtype.is_wildcard
-            )
-        )
-
-        return type_match and subtype_match
+        media_type: str,
```
ahopkins, line 68: This should accept `str` and `MediaType`

sanic/headers.py
```diff
@@ -35,141 +35,78 @@
...
+        @param media_type: A type/subtype string to match.
+        @return `self` if the media types are compatible, else `None`
+        """
+        mt = MediaType._parse(media_type)
+        return self if (
+            # Subtype match
+            (self.subtype in (mt.subtype, "*") or mt.subtype == "*")
+            # Type match
+            and (self.type_ in (mt.type_, "*") or mt.type_ == "*")
+            and (allow_type_wildcard or self.type_ != "*" and mt.type_ != "*")
```
ahopkins, line 88:
```suggestion
            and (allow_type_wildcard or (self.type_ != "*" and mt.type_ != "*"))
```
More explicit

sanic/headers.py
```diff
@@ -35,141 +35,78 @@
...
+            and (allow_subtype_wildcard or self.subtype != "*" and mt.subtype != "*")
```
ahopkins, line 89:
```suggestion
            and (allow_subtype_wildcard or (self.subtype != "*" and mt.subtype != "*"))
```
More explicit

sanic/headers.py
```diff
@@ -178,28 +115,77 @@ def parse(cls, raw: str) -> Accept:
...
+class AcceptList(list):
+    """A list of media types, as used in the Accept header.
+
+    The Accept header entries are listed in order of preference, starting
+    with the most preferred. This class is a list of `MediaType` objects,
+    that encapsulate also the q value or any other parameters.
+
+    Two separate methods are provided for searching the list:
+    - 'match' for finding the most preferred match (wildcards supported)
+    -  operator 'in' for checking explicit matches (wildcards as literals)
```
ahopkins, line 142: Where is this being done? Is it ignoring q values?

ahopkins:
There are a lot of changes in the tests. I still need to dig into them to understand the scope of this change.

sanic/headers.py
Tronic: This intentionally does not accept MediaTypes, but only mime strings (that have no parameters such as q). But I implemented other types for equality operator because it expects that.

sanic/headers.py
Tronic: In the class that this docstring is for? Operator `in` is the one from `list`.

Tronic: > In the class that this docstring is for? Operator `in` is the one from `list`. Check the docs of `AcceptList.match` for how the order is determined (it considers q and order of both lists).

Tronic: I didn't check how the API docs render, but the class tries to give an overview, with the match function going to specifics of that rather complex matching logic. Hopefully that shows up decently.

Tronic:
> There are a lot of changes in the tests. I still need to dig into them to understand the scope of this change.

I suggest looking at `repr(req.accept)` and the debug log entries shown in `guess_mime` (error page logic), as well as the new test that takes advantage of this. Half of parametrized cases on some tests had to be removed due to removed functionality, IIRC.

sanic/headers.py
ahopkins: This is breaking then.

Tronic:
Regarding tests, the big thing missing is routes having `error_format` set or automatically determined. The test client would always give me only empty string. It is not directly related to this PR though, but such tests would have made the `empty()` PR just merged fail big time (and it passed without any changes to tests).

sanic/headers.py
Tronic: Very obviously. This whole thing depended on `request.accept` never being documented, and getting to .3 release where things can break. This substantially changes the semantics of that handler, not only the innermost type.

Tronic: The most breaking change is probably that of the `in` operator, but `match` gives somewhat different results now. Both were very "broken", and hopefully these changes automatically fix more than they break anything for those users who already discovered the accessor and made use of it.

A particular "oops" moment was with my own app, that had been checking `"text/html" in request.headers.get("accept", "")` (the check itself is a bit b0rked but that's what I had working all these years), and responding in HTML in that case, otherwise plain text (mainly fetch calls from the front, but also some curl users).

I changed that to use `request.accept` instead, hearing of it being added, and suddenly *every client* started matching: fetch and curl accidentally got HTML.

Tronic:
NOTE: This branch is OUT OF DATE. I had been doing further development on error-format-redux, to which today's change requests are implemented to, now. It has better tests and is generally better in terms on `request.accept` too.

Closing this in favor of reviewing and merging #2668 to main, containing also everything done on this branch.

sanic/headers.py
ahopkins: I know that we are breaking, but what I thought the extent of the agreed upon change was to allow for a differentiation between exact matching and wildcard matching.

Meaning, that `in` would behave differently than `match`. But it seems that also `match` is changing.

ahopkins:
> NOTE: This branch is OUT OF DATE. I had been doing further development on error-format-redux, to which today's change requests are implemented to, now. It has better tests and is generally better in terms on `request.accept` too.
> 
> Closing this in favor of reviewing and merging #2668 to main, containing also everything done on this branch.

This is not good practice. We need a single PR that does the accept changes independent of the error formatting.

ahopkins:
It still needs to be its own. You can branch from a feature branch.

Tronic:
The other PR touches one property of Request and one internal function of errorpages. It is not by any means a large one. I stick to that due to practical reasons both are better kept in the same PR, due to the two things being deeply intertwined. If it were a fork of this, it would have needed a merge to this, to the same end result, but conflicts in tests and other trouble.

End of discussion on my part here. This is a thing I will eventually have to take to others if you refuse to review that PR at all. Continue discussion of what needs to be fixed there, if you wish.

PR #2667, opened 2023-01-27
Nicer traceback formatting

Tronic:
Shinier error pages for Sanic, using the sanic-org/tracerite module for traceback formatting and the new BasePage layout merged in #2662.

Any exception in an app, while running Sanic in debug mode gives a formatted traceback:
Screenshot

An attempt is made to show only the relevant parts of call chains and exception chains, pointing to where the error is in user code instead, but other details are available under tab headers (e.g. Sanic.handle_request in the above screenshot - but the current version would hide that entirely and the style has also changed since).


Tronic:
Hmm, not possible to make draft PR to another PR? Anyway, all be reminded that this is very explosive draft code build on top of another draft PR.

ahopkins:
Not sure about adding two deps for this right now. Also, changing travebacks is a fairly opinionated item. That to me feels more like something that should be in sanic-ext, with this focusing more on adding the styling changes.

Tracyca209 (approved):

Tracyca209 (approved):

Tronic:
Yes, definitely not going to push it, but I might leave it floating around for a bit later should there be interest. I don't see it living in sanic-ext because either you get better exception handling in default install or you never get one (I have many years of prior experience with niceback, trying to get Jupyter Notebook, Google Colab and others to use it but not even regular data scientists do).

I made some cleanup today. The last night thing was about 15 minutes quick adoption (after some hours porting other parts of the old error page). Instead of spending more time porting the TB formatter, using niceback was essentially just one function call.

Long term plan: few minor improvements on niceback (that I've done today, maybe a bit more needed), get that moved to sanic-org as well, and change the name. Turns out that all Internet search engines search for Nickelback when you want Niceback, and even when forced give you physiotherapy instead, so the name clearly needs to be changed.

Tronic:
With the cleanup done today, you don't get the two white tabs as shown in the screenshot above, in that situation. Also the Request object is hidden in variable inspector because it provided no information not already known. Thus, for a simple bug you get very simplified output that directly points out what happened and where.

Screenshot

For more complicated cases it still prints tabs for the full call chain (with variable inspector on each tab), and "the above exception occurred after catching ..." kind of thing as well (try causing a 404 error with `sanic --simple --dev` and you get some aiofiles/thread internals displayed that way).

ahopkins:
Real quick thought I've had, then I'll respond in more detail when I'm at my computer.

Let's keep the formatting in, but only conditionally. We have a note that says you can have nicer travebacks if you install this. Like how we deal with Jinja. You are not required to download it, but we support it if you do.

Tronic:
@ahopkins What's your take on the common pages, is that something you want in 23.3? It won't be unreasonably hard to port the old traceback formatting to this as well.

I am fine with leaving the common page format entirely for the use of the file browser for now, and keep this floating.

If people need to install the niceback package manually, it is even worse than being in sanic-ext. Then literally no-one will use it. Some might actually use the ext package and accidentally get better tracebacks then. Overall, I feel like this needs to be in the core, but needs more time to make sure everything is OK, and to get the BasePage stuff settled.

Niceback separates traceback parsing and HTML rendering, passing an info dict with the information for the renderer. I don't quite recall whether that structure is JSON-compatible, but potentially it could also be sent to client as additional info by the JSONRenderer as well.

ahopkins:
> @ahopkins What's your take on the common pages, is that something you want in 23.3? 
I think I want to get it all in now. Including using `niceback` for displaying tracebacks. I am going back on my earlier statements. I still want to see a more "real world" example of a deeply nested exception. And, by "going back", this also includes my earlier stance regarding error page rendering. While previously I held the belief that unless specified the handler intent controlled, if we use `niceback`, than browsers really must _always_ show HTML content for errors.


> Overall, I feel like this needs to be in the core, but needs more time to make sure everything is OK, and to get the BasePage stuff settled.
As stated above: :+1: I think the `BasePage` is probably settled in the other PR. I merged it into a local version of this PR and I think it works well in both.

> Niceback separates traceback parsing and HTML rendering, passing an info dict with the information for the renderer. I don't quite recall whether that structure is JSON-compatible, but potentially it could also be sent to client as additional info by the JSONRenderer as well.

Let's def look into this. Sounds like a nice abstraction then.

---

One thing to point out, the error page as it currently sits does not display the exception `extra` (debug mode only, if exists) and `context` (always, if exists) objects.

Tronic:
> I still want to see a more "real world" example of a deeply nested exception.

Enable static files, trigger 404 on it. You'll get two exceptions and one of them with a call chain (but because it's not a 500 error, details are hidden by default). Styling depended on Jupyter Notebook default CSS to a degree, needs more tuning to look clean on Sanic, and also was only designed for white background. I've already done fixes on the TraceRite project so that it can work with either dark or light background reasonably well without actually supporting each (the same as the initially gray title bar of BasePage).

Tronic:
> One thing to point out, the error page as it currently sits does not display the exception `extra` (debug mode only, if exists) and `context` (always, if exists) objects.

I couldn't find any uses of these within Sanic's own code (if not for the TeaPotError test). Could I have some practical examples to work on? As the UI stands now with the details object, it is a bit cumbersome to separate context and details that way, and I might prefer displaying both on the same table, perhaps on the upper right corner (but still hiding extra if not in debug mode).

ahopkins:
> Could I have some practical examples to work on?

[See docs](https://sanic.dev/en/guide/best-practices/exceptions.html#contextual-exceptions)

Here's another though...

```py
from functools import wraps

from sanic import Request, Sanic
from sanic.exceptions import Forbidden

app = Sanic("TestApp")


class CustomForbidden(Forbidden):
    quiet = False

    @property
    def message(self) -> str:
        return f"Sorry {self.extra['user']['name']}"


def authorize(required_permission):
    def decorator(f):
        @wraps(f)
        async def decorated_function(request, *args, **kwargs):
            user = {"name": "Dave"}

            # Simulate something bad happening
            raise CustomForbidden(
                extra={"user": user},  # not be displayed in PROD
                context={  # always displayed
                    "required_permission": required_permission
                },
            )

        return decorated_function

    return decorator


@app.get("/", error_format="html")
@authorize("some:scope")
    ...
```

In PROD

image


In DEBUG

image

They are additional dicts that need to be dumped. In debug mode: both. In prod mode: just context.

Tronic:
> They are additional dicts that need to be dumped. In debug mode: both. In prod mode: just context.

Implemented yesterday! (roughly)

ahopkins:
Happy to rake a swing at it for you.

Tronic:
@ahopkins Already on it. Looks like it's just take most files from main (because a lot got auto-merged wrong), and manually handle the few files this actually touches, then rewrite all CSS rules by experiment.

ahopkins:
image

Let's bump the vertical margin and make this more noticeable. Also, we should explain the order of exceptions being caught left-to-right.

Tronic:
> image
> 
> Let's bump the vertical margin and make this more noticeable. Also, we should explain the order of exceptions being caught left-to-right.

The yellow arrows explain left to right order, also easily seen from what the code calls (marked in yellow). Bomb means exception thrown and a warning sign is something in your code that calls library code that eventually raises (i.e. the bug might be at the warning sign). KeyboardInterrupt and other BaseException show a stop sign instead of the bomb.

It'll need to suppress a lot more of Sanic's internals, but I haven't touched that yet because those deep tracebacks provide useful test material (just try 404 with no matching route, not even static handler), shows even uvloop internal route function because that is where the exception is thrown (otherwise non-Python frames are suppressed from call chains).

ahopkins:
This can be made way more obvious and user friendly. A little formatting and a line to describe the order will go a long way. What's obvious to you will kot be obvious to everyone. This is a big change, let's ease into it.

Tronic:
> This can be made way more obvious and user friendly. A little formatting and a line to describe the order will go a long way. What's obvious to you will not be obvious to everyone. This is a big change, let's ease into it.

Styling fixed in tracerite repo, that element was missing its class but had a style definition from before. Good catch!

ahopkins:
Would be awesome if there was a little tooltip on ➤, :bomb:, and :warning: that explained what they meant.

ahopkins:
image

Maybe this style formatting for extra and context?

image
Currently, they are not identified with a header. The text is larger than the headers and therefore seems inconsistent. Plus, there could be some more vertical space because it is very busy.

ahopkins:
Just noticed we need to include two versions of the logo
image

Tronic:
> Just noticed we need to include two versions of the logo, or take the easy way out and make footer a dark background when in light mode.

```css
#logo { filter: drop-shadow(0.05em 0.05em 0.1em #000a) }
@media (prefers-color-scheme: dark) {
  #logo { filter: none }
}
```

sanic/pages/error.py
```diff
@@ -0,0 +1,101 @@
...
+            if context:
+                self._key_value_table(
+                    "Issue context", "exception-context", context
+                )
+
+            if not debug:
+                return
+            # Show additional details in debug mode,
+            # open by default for 500 errors
+            with self.doc.details(open=self.full, class_="smalltext"):
```
ahopkins, line 63: If I am reading this right, I think we may be using `full` here incorrectly. If this is meant to be full/minimal from the error page renderers, then it should be doing what you are doing with debug here and not controlling whether the details are open or not.

ahopkins: https://sanic.dev/en/guide/best-practices/exceptions.html#html

Tronic: The intent is different than debug. The details are supposed to be open by default for 500 errors and not for 404 (roughly speaking). `full` accomplishes that, but if you have a better variable that does the same, let me know.

ahopkins: Right, my point is that its misusing something already. You shouldn't read debug from the app instance and should use full for that purpose.

For deciding if it's open or not, I would use `getattr(exc, "quiet", False)`.

ahopkins:
image
We should either figure out how to strip the tags going into `<title>` or just skip using `<strong>` altogether.

sanic/pages/error.py
```diff
@@ -0,0 +1,104 @@
+from typing import Any, Mapping
+
+import tracerite.html  # type: ignore
```
ahopkins, line 3: We should not be ignoring this. If there is an import issue we should handle it.

sanic/pages/error.py
```diff
@@ -0,0 +1,104 @@
...
+
+from sanic.request import Request
+
+from .base import BasePage
+
+
+# Avoid showing the request in the traceback variable inspectors
+inspector.blacklist_types += (Request,)
+
+ENDUSER_TEXT = """We're sorry, but it looks like something went wrong. Please try refreshing the page or navigating back to the homepage. If the issue persists, our technical team is working to resolve it as soon as possible. We apologize for the inconvenience and appreciate your patience."""  # noqa: E501
```
ahopkins, line 16: One super long line with a `noqa` makes this difficult to maintain. We can break this up and make it easier to read using `dedent` if needed, etc.

sanic/pages/error.py
```diff
@@ -0,0 +1,104 @@
...
+from html5tagger import E
+from tracerite import html_traceback, inspector  # type: ignore
+
+from sanic.request import Request
+
+from .base import BasePage
+
+
+# Avoid showing the request in the traceback variable inspectors
+inspector.blacklist_types += (Request,)
```
ahopkins, line 14: Will this work on subclasses of `Request` also?

sanic/pages/error.py
```diff
@@ -0,0 +1,104 @@
...
+        self.details_open = not getattr(exc, "quiet", False)
+
+    def _head(self) -> None:
+        self.doc._script(tracerite.html.javascript)
+        super()._head()
+
+    def _body(self) -> None:
+        debug = self.request.app.debug
+        try:
+            route_name = self.request.route.name  # type: ignore
```
ahopkins, line 54: Code smell. Can be handled nicer.

```py
self.request.name or "[route not found]"
```

sanic/pages/error.py
```diff
@@ -0,0 +1,104 @@
...
+        text: str,
+        request: Request,
+        exc: Exception,
+    ) -> None:
+        super().__init__()
+        # Internal server errors come with the text of the exception,
+        # which we don't want to show to the user.
+        # FIXME: Needs to be done in a better way, elsewhere
+        if "Internal Server Error" in title:
+            text = "The application encountered an unexpected error and could not continue."  # noqa: E501
```
ahopkins, line 34: We don't need this. It is already handled here: https://github.com/sanic-org/sanic/blob/niceback-error-handling/sanic/errorpages.py#L75

sanic/pages/error.py
```diff
@@ -0,0 +1,104 @@
...
+        name = request.app.name.replace("_", " ").strip()
+        if name.islower():
+            name = name.title()
```
ahopkins, line 37: I don't think we should be changing the user's app name. Also, we probably should just use a generic term when `debug=False` because we do not explicitly tell our users that app names might be public. For example, I use internal names for applications all the time that are not the equivalent to what an end user might know it as.

sanic/app.py
```diff
@@ -875,6 +875,8 @@ async def handle_request(self, request: Request):  # no cov
         :param request: HTTP Request object
         :return: Nothing
         """
+        __tracebackhide__ = True
```
ahopkins, line 878: What's this?

sanic/pages/error.py
ahopkins: Maybe out of scope here, but I would not be opposed to the idea of returning a default value (configurable of course) for `Request.name` when there is no route rather than an empty string.

sanic/pages/error.py
```diff
@@ -0,0 +1,104 @@
...
+    STYLE_APPEND = tracerite.html.style
+
+    def __init__(
+        self,
+        title: str,
+        text: str,
+        request: Request,
+        exc: Exception,
+    ) -> None:
+        super().__init__()
```
ahopkins, line 29: This does not handle `debug` param.

ahopkins (changes requested):
Looks good, I am excited to get this in. Just a few questions. Also, I am proposing a simpler color scheme here: https://github.com/sanic-org/sanic/pull/2703. Also it corrects a couple items around debug mode, and adds a few helpful links when debug=True.

sanic/app.py
Tronic: Tracerite feature (also shared by some other backtrace formatters, as well as IPython internals) to hide that frame (tab in tracerite). This is used for a few parts of Sanic internals that would otherwise very often appear, to make tidier and only show user code.

It probably would get hidden anyway when Sanic is installed on a system but we really need to hide this for `pip install -e` as well, even though in that case we want to show other Sanic internals.

sanic/pages/error.py
Tronic: I would like to see that handled as public end user string, but taking that it is normally a Python identifier, it needs that pretty printing. It is shown on page header.

sanic/pages/error.py
Tronic: Yes, it uses instanceof.

sanic/pages/error.py
Tronic: It becoming public needs to be mentioned in release notes but I don't see how revealing a module name, "main", or such, even an internal codename, would be dangerous even if someone missed that happening, and can be easily fixed to follow product branding if needed.

ahopkins: Okay, we can leave for now. But, I am wondering if maybe we should add a public name. Not required, but would get used in place of this if present.

```python
app = Sanic("ProjectName", public_name="Adam's Awesome API")
```

Let's tackle that outside of this context though.

ahopkins (approved):

PR #2668, opened 2023-01-29
Error page rendering format selection

### LINKED ISSUE #2658: FALLBACK_ERROR_FORMAT does not work with empty()

### Describe the bug

Today I tested a handler which always returns `empty()` (see example below). 
I was quite confused by the format of the errors, as they were all in HTML format, even though the fallback was set to json.
It took me a while to figure out that this was because of the `empty()` statement returned by the function, which somehow hijacks the error format, so that the errors are converted to html. 

I could use the `error_format` from now on for handlers returning `empty()`, but was wondering if this is a feature or a bug.

### Code snippet

```python
from sanic import Sanic, empty
from sanic.exceptions import Unauthorized


app = Sanic(__name__)

# Returns error in html format
@app.post("/empty")
async def empty_(request):
    raise Unauthorized("bli bla blub")

    return empty()

# Returns error in json format
@app.post("/nothing")
async def nothing(request):
    raise Unauthorized("bli bla blub")


if __name__ == '__main__':
    app.run("127.0.0.1", port=3000, auto_reload=True, debug=True, dev=True)
```

### How do you run Sanic?

As a script

### Operating System

Ubuntu

### Sanic Version

22.9.1

Tronic:
## Error format selection

This PR simplifies the error handling logic in Sanic by making it more dependent on the `accept` header when available. The goal is to produce more suitable responses for real-world use cases, such as always returning HTML on pages displayed in a browser, even if the handler would normally return `text()`.

Sanic's preferred format is determined by route `error_format` (set by app, or determined from handler source code if "auto", the default) and by `app.config.FALLBACK_ERROR_FORMAT`.

If the fallback is set to "auto" (default), all Sanic's supported formats are offered such that client preference takes priority. If the route also had no format determined, Sanic uses heuristics for request JSON format: is it verbatim in accept header or content-type, or does the body parse as JSON (the latter is being DEPRECATED in this PR). If no JSON is found either, text becomes the first preferred format of any remaining formats.

Sanic logs show which `accept` part matched the format and from where it came (route name, content-type, request.json, FALLBACK, any), or if there was no match with accept at all, in which case the base renderer is used.

Just as before, the base renderer is **always text** unless set via undocumented functionality `Sanic(error_handler=ErrorHandler(SomeOtherRenderer))`. It was previously used in many cases but now it is only reached if a client specifically asks only for unsupported format(s), e.g. `accept: application/xml`.

## Typical use cases

* Viewing in browser
   - If fallback format is auto, always gets html because `text/html` is the only q=1 match
   - Otherwise, uses Sanic's preferred format from route or fallback (matching `*/*;q=0.8`)
* Browsers loading img or other resources only match the final `*/*;q=0.8` and get Sanic's preferred format
* Curl, requests, fetch(), XHR etc. default to `accept: */*` and get Sanic's preferred format
* Clients setting `accept` to a specific format get that or text

## Background

In Sanic 22.12:
* Text/JSON are returned when viewed in browser (affects `text("hello world")` type apps)
* HTML or text is returned to applications expecting JSON but returning `empty()`
* The fallback setting has little effect (using route formats or hardcoded text format instead)

There are breaking changes needing careful review, but the PR is still planned for release in 23.3. It addresses the above urgent issues by respecting the q values given in `accept` and reducing the use of the base renderer. Those who previously received JSON mostly still receive it, and can (as before) force it by accept header or by setting `error_format="json"` on the route.

The previous large rewrite was in #2162 (patched in #2310), originally stemming from #1937 where text and json were added as alternatives to html error pages. There is useful background and explanation of the existing logic in these issues.

PR #2659 is related to this and fixed some of the issues mentioned above. This PR is based on the new request.accept functionality merged in #2687.


Tronic:
* The client sends `accept: application/xml`
* The app sets `FALLBACK_ERROR_FORMAT = "json"`, and
* The handler returns `html(...)` but raises an exception.

## Which of text/json/html should be used on the error response?

Oh, but **text** of course, because html and json don't fit the accept header, I'll just send text instead!

This PR sends **json** but I am not quite sure there is a good answer. Presumably someone could supply XMLRenderer as an alternative base renderer? But this is a niche case, as usually all formats are accepted or at least the right format is.

Tronic:
TODO: Refactor `_guess_renderer` to return a string rather than a renderer object, to make it usable for other things as well (file browser API?). Do the Renderer lookups instead in `exception_response` that calls the other function.

Tronic:
Seems to pass tests (headers and errorpages at least).

JSON header check now does accept (for literal application/json) as well as content-type (when route and fallback are auto and not found). This only affects the case when multiple supported formats are accepted, where this gives JSON the higher priority regardless of q values (backwards compatibility, hopefully avoids breaking someone's API calls).

Tronic:
Now it seems more or less ready for actual review. I have not yet backported the latest changes to accept-enhance that need to be there.

ahopkins (changes requested):
Accept parsing changes should not be in this PR. We need a PR for just these changes.

sanic/errorpages.py
```diff
@@ -435,98 +433,72 @@ def exception_response(
...
+    name = ""
+    # Route error_format (by magic from handler code if auto, the default)
+    try:
+        if req.route:
+            name = req.route.name
+            f = req.route.extra.error_format
+            if f in MIME_BY_CONFIG:
+                formats[f] = name
+    except AttributeError:
+        pass
```
Tronic, line 458: @ahopkins Coverage tests are not hitting the AttributeError. Do we need this try-except or can it be removed? I suppose it was there to catch a missing `error_format` attribute on `route.extra`. The `if req.route` check was needed for something else (cannot remember anymore) and both it and the try-except were in the original code.

Tronic:
All tests passing, a few lines missed in coverage still.

The MediaType.match function has kwargs to disable wildcard matches but those are not really needed and should probably be removed. I left them in to satisfy existing tests that use them (the linter error is an overlong comment on this).

The is_wildcard and has_wildcard properties are a bit unnecessary as well but presumably one could have some use for them and they are also used in existing tests.

Tronic:
Note that the MediaType class is not specific to Accept headers, and could potentially be used for other things dealing with mime types and parameters (content-type).

Tronic:
@ahopkins 100 % patch coverage with all tests passing, more comprehensive testing than previously.

sjsadowski:
Note: I'm tagging myself in here to review and understand the changes

ahopkins:
@Tronic I resolved the merge conflicts in this PR into this branch: https://github.com/sanic-org/sanic/pull/2691

sanic/errorpages.py
```diff
@@ -435,98 +434,68 @@ def exception_response(
...
+            formats[f] = name
+
+    if not formats and fallback in MIME_BY_CONFIG:
+        formats[fallback] = "FALLBACK_ERROR_FORMAT"
+
+    # If still not known, check for the request for clues of JSON
+    if not formats and req.accept.match(JSON):
+        if JSON in req.accept:  # Literally, not wildcard
+            formats["json"] = "request.accept"
+        elif req.headers.getone("content-type", "").split(";", 1)[0] == JSON:
```
ahopkins, line 465:
```suggestion
        elif JSON in req.headers.getone("content-type", ""):
```

This should be faster.

ahopkins (dismissed):

sanic/errorpages.py
```diff
@@ -435,94 +434,67 @@ def exception_response(
...
+                    f'  error_format="json" on route {name}, by\n'
+                    '  FALLBACK_ERROR_FORMAT = "json", or by adding header\n'
+                    "  accept: application/json to your requests.",
+                    24.3,
+                )
+
+    # Any other supported formats
+    for k in MIME_BY_CONFIG:
+        if k not in formats:
+            formats[k] = "any"
```
ahopkins, line 488: I feel like we should skip this test if the format was explicitly set:

```python
@app.get("/", error_format="json")
    1 / 0
```

This PR does a much better job giving `FALLBACK_ERROR_FORMAT` its intended impact: the format to use when we are not sure what else to use. But, one thing that I feel we are still not covering is when the developer explicitly declares the format.

In the above scenario, we should only ever use JSON. Taking this step then begs the next: an API to explicitly declare a definitive format to use for everything. Right now to do this requires a mess of `@app.exception` or building a custom `ErrorHandler`.

Happy to take that up in a subsequent PR.

Tronic: Agreed, leave out "any" if explicitly set on route.

Tronic: I suppose this is to force JSON response to browser viewing a page on such route? Despite my previous comment, I am not entirely convinced of that being preferable even when manually defined on the route. I'll leave that up to you, but I prefer simplicity in documentation and code if adding additional logic doesn't serve a clear purpose.

ahopkins: Fine with that for now. But let's do that skip for explicitly set, and then I think this is ready.

Tronic: Used fallback format for that instead. If set to auto, client preference is respected (via any format) as before, but apps can set that manually to get JSON responses while viewing in browser. Seems like the right way to handle this.

ahopkins (approved):

PR #2670, opened 2023-02-01
Increase KEEP_ALIVE_TIMEOUT default to 120 seconds

### LINKED ISSUE #2681: Send keep-alive header and increase the default timeout

Nginx doesn't notice Sanic closing its keep-alive connection by timeout, and then another request will fail for no good reason. High latency connections suffer of having to do frequent reconnections when a client or proxy loads additional resources 5+ seconds later, despite keeping such connections open being practically free with the async server.

Sanic http1 server should send `keep-alive: timeout=<KEEP_ALIVE_TIMEOUT>`, to inform the client of the duration it allows. Also Sanic should not limit max number of requests per pipeline (no reason to) and use a higher default `KEEP_ALIVE_TIMEOUT` setting of at least 70 seconds to exceed that of Nginx to avoid connection problems.

Sanic should also investigate ways of closing the connection so that Nginx or another client gets the notice of connection RST and won't attempt to send another request once those are no longer accepted. Could be problematic with TLS but this is also broken in plain HTTP in current version.

### Additional context

HTTP/3 does not use keep-alive header. On ASGI this might be up to the ASGI server, needs investigation.

Tronic:
Sanic has used 5 second timeout to quickly close idle connections, for no particular reason, as idle connections use very little resources. This PR increases it to 80 seconds, slightly larger than Nginx default of 75 s, to avoid a problem with a request failing when Sanic has closed the connection but Nginx hasn't noticed. This should also improve practical performance on long latency connections, where reconnecting is expensive, and better fits typical user flow browsing pages with longer-than-5-second intervals.

The root cause of the next request failing (i.e. connection not being properly terminated) should be looked at in more detail, but meanwhile this should be a useful workaround to keep the defaults working fluently, as well as to improve performance for the aforementioned reasons.

Fix #2681


ahopkins:
Tackle timeout overhaul in June?

Tronic:
Yes, sounds good. This could be included as a workaround in 23.3 but probably better to avoid the hassle (docs update, release notes etc) and simply do it in 23.6 in one go with the simpler timeouts.

Tronic:
Reopening for inclusion in 23.6 because we didn't get to do the bigger timeout revamping yet.

sanic/config.py
```diff
@@ -50,7 +50,7 @@
     "INSPECTOR_TLS_KEY": _default,
     "INSPECTOR_TLS_CERT": _default,
     "INSPECTOR_API_KEY": "",
-    "KEEP_ALIVE_TIMEOUT": 5,  # 5 seconds
+    "KEEP_ALIVE_TIMEOUT": 120,  # Two minutes matches Caddy, exceeds Nginx default
```
ahopkins, line 53: Please take out this comment. It does not add anything here.

ahopkins (approved):

PR #2672, opened 2023-02-06
Error handling style suggestions

ahopkins:
This PR is meant for @Tronic as some style suggestions for #2667 

1. Standardize table generation
2. Put context above the debug collapse and extra inside the debug collapse
3. Use some Sanic accent color
4. More space around tables to make them stand out
5. Tooltip hover effect over :bomb: (and should be other special symbols as well)

image


sanic/pages/error.py
```diff
@@ -35,26 +49,53 @@ def _body(self) -> None:
...
+        with self.doc.table(id=table_id, class_="key-value-table"):
+            self.doc.caption(title)
+            for key, value in data.items():
+                try:
+                    self.doc.tr.td(key, class_="nobr key").td(value)
+                # Printing values may cause a new exception, so suppress it
+                except Exception:
+                    self.doc.tr.td(key, class_="nobr key").td(
+                        E.em("Unable to display value")
+                    )
```
Tronic, line 101: You could simply `.td.em(...)` avoiding the use of E. This works (and doesn't go "outside" of td) because with html5(tagger) td have no closing tags, and the module is designed to take advantage of that (vs. DOM which cannot).

sanic/pages/styles/ErrorPage.css
```diff
@@ -1,2 +1,52 @@
...
-summary { color: #888; }
+summary {
+    color: #888;
+}
+
+.tracerite {
+    /* --color-var: var(--accent);
+    --color-val: #ff0; */
+    --color-type: var(--accent);
+}
```
Tronic, line 10: Nice touch, I assume this is the Sanic logo red color?

Note: needs a change in tracerite as well (move from :root to .tracerite and rename from --tracerite-type back to --color-type).

sanic/pages/styles/ErrorPage.css
```diff
@@ -1,2 +1,52 @@
...
+    position: absolute;
+    top: -1.25em;
+    right: 0;
+    padding: 0.25em;
+    background-color: var(--accent);
+}
+
+mark[data-symbol="💣"]:hover::before {
+    display: block;
+}
```
Tronic, line 52: Please make generally useful changes to tracerite project style.css instead.

sanic/pages/styles/ErrorPage.css
```diff
@@ -1,2 +1,52 @@
...
+    padding: 0.25rem;
+}
+
+.key-value-table .key {
+    color: var(--accent)
+}
+
+mark {
+    position: relative;
+}
```
Tronic, line 38: What's with the position hackery? I generally avoid that whenever possible, and have done this with negative margin instead. Did that not work with the tooltip?

sanic/pages/error.py
```diff
@@ -35,26 +49,53 @@ def _body(self) -> None:
...
+            # Show context details if available on the exception
+            context = getattr(self.exc, "context", None)
             if context:
-                # Printing values may easily cause a new exception, so suppress it
-                with self.doc.table(id="exception-context"), suppress(Exception):
-                    for k, v in context.items():
-                        self.doc.tr.td(k).td(v)
+                self._key_value_table(
+                    "Exception context", "exception-context", context
+                )
```
Tronic, line 57: This table is for end users, maybe "Context of the issue" instead.

Tronic:
All good changes, will merge once comments are addressed.

sanic/pages/error.py
ahopkins: Sure. It was all meant as a proof-of-concept for you to take from this what you will into your own. Whether it is by merging this and fixing, or otherwise I leave that to you.

sanic/pages/styles/ErrorPage.css
ahopkins: Yes, accent being the base Sanic color. We really only have the one, but if ever there was a time to create a bit of a palette, now is the chance.

sanic/pages/styles/ErrorPage.css
ahopkins: Sure, I can make a PR into that. Again, I only meant this to play with some thoughts/ideas.

ahopkins:
> All good changes, will merge once comments are addressed.

Okay, that answers that then. I will make the PR into `tracerite`, and then remove all table styles.

ahopkins:
@Tronic Incorporated your comments. Ready for you to take this from here. I am making a PR into tracerite now to add more CSS variables and also add the tooltips.

sanic/pages/styles/ErrorPage.css
ahopkins: @Tronic 

1. No need to move the variables in tracerite from `:root`. We should still be able to override them okay in `.tracerite` in this project. I suggest they remain there.
2. I also think tracerite should continue to  namespace its variables as `--tracerite-*`. No need to convert back to `--color-*`.

PR #2673, opened 2023-02-06
Some Sanic branding to error page

ahopkins:
- Swap Sanic light/dark logo
- Apply Sanic colors to tracerite (see https://github.com/sanic-org/tracerite/pull/2)

image
image image


Tronic:
An interesting hack. Of course inline SVG is styleable :)

Tronic (approved):

PR #2676, opened 2023-02-09
Color tweaks for Error Handling. 

smlbiobot:
What’s done:

- Lower contrast in dark theme
- Color tweaks
- Added variables for the following so they can be set via variables:
  - text
  - background
  - tab text
  - tab background
  - tab shadow (this is mainly for hover state in light theme)
  - highlight text
  - highlight background (these two properties allow better control of highlight in light / dark theme — subsequently I have decided to keep the same values for now)
- Moved all the color variable config to the top of the styles under `:root {}` and the dark theme config immediately follows. For me this makes color management easily to deal with.
- Added main container padding so it’s not so claustrophobic

## CSS for the error page 

I had to add some `!important` styles. But as mentioned on Discord, it would be better if the container on the page would have an ID so that CSS can just use `#sanic-error` to take precedence over the styles from `tracerite`.

Right now, these `!important` CSS values are mainly used to override the styles from `tracerite`.

## Light Theme Screenshot

sanic-error-light

## Dark Theme Screenshot

sanic-error-dark

## Further Notes

I’m not sure how far I am allowed to tweak these so I kept it to colors only and some minor typographic issues.

### DIFF
--- a/sanic/pages/styles/BasePage.css
+++ b/sanic/pages/styles/BasePage.css
@@ -2,10 +2,10 @@
 
 :root {
     --sanic: #ff0d68;
-    --sanic-blue: #0369a1;
-    --sanic-yellow: #ffde41;
-    --sanic-purple: #7e22ce;
-    --sanic-green: #4d7c0f;
+    --sanic-blue: #0092FF;
+    --sanic-yellow: #FFE900;
+    --sanic-purple: #833FE3;
+    --sanic-green: #37ae6f;
     --sanic-background: #f1f5f9;
     --sanic-text: #1f2937;
     --sanic-tab-background: #fff;
@@ -18,15 +18,10 @@
 
 @media (prefers-color-scheme: dark) {
     :root {
-        --sanic: #ff0d68;
-        --sanic-blue: #f08d49;
-        --sanic-yellow: #ffde41;
-        --sanic-purple: #cc99cd;
-        --sanic-green: #7ec699;
-        --sanic-background: hsl(0, 0%, 10%);
-        /*--sanic-background: #2d2d2d;*/
-        /*--sanic-background: #111;*/
-        --sanic-text: rgba(255, 255, 255, 0.9);
+        --sanic-purple: #D246DE;
+        --sanic-green: #16DB93;
+        --sanic-background: #111;
+        --sanic-text: #e7e7e7;
         --sanic-tab-background: #484848;
         --sanic-tab-text: #e1e1e1;
         --sanic-tab-shadow: #000;
--- a/sanic/pages/styles/BasePage.css
+++ b/sanic/pages/styles/BasePage.css
@@ -13,6 +13,7 @@
     --sanic-tab-shadow: #adadad;
     --sanic-highlight-background: var(--sanic-yellow);
     --sanic-highlight-text: var(--sanic-text);
+    --sanic-header-background: #000;
 }
 
 @media (prefers-color-scheme: dark) {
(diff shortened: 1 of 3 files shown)


ahopkins:
> I’m not sure how far I am allowed to tweak these so I kept it to colors only and some minor typographic issues.

Tweak away. We'll comment if we think you've gone too far. I much appreciate this and we can probably work to reduce the !important by bringing some changes to tracerite.

At first glance, my biggest concern though is the need for more branding colors. Mainly, the background color differs from the docs background color in dark mode, and I don't see the Sanic color represented.

sanic/pages/styles/BasePage.css
```diff
@@ -1,28 +1,57 @@
...
+    --sanic-tab-background: #fff;
+    --sanic-tab-text: #0f172a;
+    --sanic-tab-shadow: #adadad;
+    --sanic-highlight-background: var(--sanic-yellow);
+    --sanic-highlight-text: var(--sanic-text);
+}
+
+@media (prefers-color-scheme: dark) {
+    :root {
+        --sanic: #ef4444;
```
ahopkins, line 20: this value is meant to be Sanic's on brand color. It should not differ in dark mode.

smlbiobot: > this value is meant to be Sanic's on brand color. It should not differ in dark mode.

Sure I can fix this.

Tronic:
Okay, thanks for the work. I have to say I didn't have time to read the design discussion you had in full, but this looks very colorful to me, a lot of bright colors.

I kind of prefer bringing emphasis to the relevant details by not making everything bright. The combination of mostly white and grey accomplished this, making the exception stand up in Sanic red (brand color) and highlighting the relevant code lines with the yellow marking. A little bit of color in variable inspectors makes them visually distinct of code, but not all those three parts of it need to be colored. I initially tried passive and similar colors (greenish-blueish) to keep it distinct but not grab too much attention, not appear like a rainbow.

These are of course also matters of opinion, and perhaps you have discussed this, but just my points at this phase.

Also, some of the styling changes should ideally be done on Tracerite, with only those specifically needed for Sanic left to be done on Sanic side (especially the use of that Sanic red). But bear in mind that the module also needs to work on Jupyter notebooks where themes come in all colors (but don't use prefers-color-scheme, unfortunately), and the module is purposefully made to use default text and background colors in the majority of things (gray text is a fairly safe choice that stands out from either dark or light backgrounds).

sanic/pages/styles/BasePage.css
```diff
@@ -1,28 +1,56 @@
...
+@media (prefers-color-scheme: dark) {
+    :root {
+        --sanic-purple: #D246DE;
+        --sanic-green: #16DB93;
+        --sanic-background: #111;
+        --sanic-text: #e7e7e7;
+        --sanic-tab-background: #484848;
+        --sanic-tab-text: #e1e1e1;
+        --sanic-tab-shadow: #000;
+        --sanic-highlight-background: var(--sanic-yellow);
```
Tronic, line 28: The yellow can be a lot darker (and less disruptive) on dark mode. Yellow is a strange color in that only seems yellow if it is brighter or similar brightness to its surroundings, which is why on black background it can be quite dark while on white background it needs to be close to #ff0. Yellow darker than its surroundings is seen as (greenish) brown, a different color and not simply a darker shade.

ahopkins:
FWIW I put some of my own tweeks of your PR (admittedly before your last change I believe) into this branch: https://github.com/sanic-org/sanic/tree/sml-changes.

I definitely do not want to bikeshed on colors and spend a month bumping a color 1% or 10%. If this means dropping all colors except for background, foreground, and Sanic-accent: fine. Let's get the feature done and we can iterate on it. But I am hoping to not let colors be what holds back the PR getting merged.

Tronic:
I've been using tracerite 1.0.1 on VSCode Jupyter, Jupyter Notebook and Jupyterlab, seems to work great in all of those with its current styles.

Tronic:
@smlbiobot Are you doing further changes to this or should we merge now?

ahopkins:
I agree. Let's get this done and then come back to coloring. I like the idea of maybe just handling it with various shades. I suggest we continue the coloring discussion and come back before release.

My only question though is do we need a corresponding PR into `niceback`?

ahopkins (approved):

PR #2682, opened 2023-02-12
Sanic on pypy

deounix:
A small patch for the `os` module to make Sanic works with PyPy

The problem is that `aiofiles` uses `os.readlink` which doesn't exists on PyPy.


ahopkins:
Nice find. Thanks.

sanic/compat.py
```diff
@@ -73,6 +75,11 @@ def enable_windows_color_support():
     kernel.SetConsoleMode(kernel.GetStdHandle(-11), 7)
 
 
+def patch_os() -> None:
+    module = sys.modules["os"]
+    module.readlink = os.path.realpath
```
Tronic, line 80: This needs a comment explaining why it needs to monkeypatch the os module.

sanic/compat.py
```diff
@@ -75,7 +76,20 @@ def enable_windows_color_support():
...
+
+    This monkeypatch replace the `os.readlink` with `os.path.realpath`,
+    Both have the same functionality.
+    """
+    if hasattr(os, "readlink"):
+        error_logger.warning(
+            "PyPy: Skip patching the os module "
+            "it seems they added the readlink function"
+        )
+        return
```
Tronic, line 91:
```suggestion
def pypy_os_module_patch() -> None:
    """
    The PyPy os module is missing the 'readlink' function, which causes issues with
    aiofiles. This workaround replaces the missing 'readlink' function with 
    'os.path.realpath', which serves the same purpose.
    """
    if hasattr(os, "readlink"):
        error_logger.warning(
            "PyPy: Skipping patching of the os module as it appears the "
            "'readlink' function has been added."
        )
        return
```
I asked ChatGPT for better language and to wrap to line length 79.

deounix:
@Tronic
@ahopkins 

Updated. Thank you guys for the good work ( Sanic )

sanic/compat.py
```diff
@@ -10,6 +11,7 @@
 from multidict import CIMultiDict  # type: ignore
 
 from sanic.helpers import Default
+from sanic.log import error_logger
```
ahopkins, line 14: This is causing a circular import issue. We _could_ go the option of a larger refactor. However, given the nature of this PR and the context in which it is called, I think we should opt for the shortcut and just put this import inside the `pypy_os_module_patch` function. This seems find enough for me. If you want to tackle a more elegant way, LMK.

deounix: Now there's a new error I'm trying to run the examples\blueprints.py

```
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```

To get it to work I added

```
if __name__ == '__main__':
    from multiprocessing import freeze_support
    freeze_support()

    app.run(host="0.0.0.0", port=9999, debug=True)
```

I think there's a problems with forking

deounix: also the printing on terminal is a mess

Screenshot 2023-04-17 060827

deounix: a workaround is force motd to be False on pypy + windows. maybe ?

Tronic: Pypy probably needs to switch their terminal handling to use wchar i.e. UTF-16 output instead of 8 bit char that causes the problem you are seeing. This is something they really need to fix, and I don't see much point in Sanic (and all other Unicode-enabled Python software out there) working around the issue.

deounix: @Tronic Fixed 🥳 Sanic with PyPy ready to run on all platforms

Screenshot 2023-04-17 222146

deounix:
> Oh right, you can have UTF-8 codepage nowadays too. Still, the fix belongs upstream in Pypy, can you get it there?

I'm sorry I can't. If you could do it.

Tronic:
> > Oh right, you can have UTF-8 codepage nowadays too. Still, the fix belongs upstream in Pypy, can you get it there?
> 
> I'm sorry I can't. If you could do it.

Nope, I never even used Pypy. I figured since you found the fix and solved the problem could make a PR or issue report to them as well.

sanic/helpers.py
```diff
@@ -157,6 +159,10 @@ def import_string(module_name, package=None):
     return obj()
 
 
+def is_atty() -> bool:
+    return bool(sys.stdout and sys.stdout.isatty())
+
```
ahopkins, line 164: Why are we moving this?

deounix: I found that the best place to put this function inside helper.py, since placing it inside compact would cause a circular import issue.

ahopkins (approved):

PR #2687, opened 2023-02-16
Refresh Request.accept functionality

ahopkins:
This falls on the heels of #2663 and #2668 and builds of the work that @Tronic did in the latter of the two PRs. This closely tracks his changes with some additions to make the pattern more compatible with the existing implementation. The following are the changes from the implementation in #2668:

- Rename `Matched` >> `Accept`
- `Accept.__eq__` operator also checks `q` value
- Add all comparison operators to `Accept`
- Add `match` to Accept
- `MediaType` parts to have wildcard
- `MediaType` to match using `str` or `MediaType` as input

The main breaking changes from `main` are that the `in` operator is not longer equivalent to `match`. Also, the params to use wildcards has been simplified in `match` to a single flag.

### DIFF
--- a/sanic/headers.py
+++ b/sanic/headers.py
@@ -231,7 +231,7 @@ class AcceptList(list):
         @return A match object with the mime string and the MediaType object.
         """
         a = sorted(
-            (acc.q, i, j, mime, acc)
+            (-acc.q, i, j, mime, acc)
             for j, acc in enumerate(self)
             if accept_wildcards or not acc.has_wildcard
             for i, mime in enumerate(mimes)
--- a/sanic/headers.py
+++ b/sanic/headers.py
@@ -7,6 +7,7 @@ from urllib.parse import unquote
 
 from sanic.exceptions import InvalidHeader
 from sanic.helpers import STATUS_CODES
+from sanic.log import deprecation
 
 
 # TODO:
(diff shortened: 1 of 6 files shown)


Tronic:
Had a brief look at the code but not properly yet, commenting based on the summary you made:

> * Rename `Matched` >> `Accept`

OK. The original implementation had something else named Accept, so I kept the match object named differently to avoid confusion.

> * `Accept.__eq__` operator also checks `q` value
> * Add all comparison operators to `Accept`

These are problematic. I suggest not making any of the things behave like strings to avoid possible confusion especially with non-standard comparisons

> * Add `match` to Accept

What is this needed for? This feature is already bloated, would rather keep it smaller (as stated already in #2200).

> * `MediaType` parts to have wildcard

Keeping the parts separate custom types is bad for performance. I believe this was the primary reason why my PR is twice faster. 

> * `MediaType` to match using `str` or `MediaType` as input

I will look at that more closer later. My implementation intentionally kept header and what they are matched against asymmetric in some cases, not pretending that MediaType (of header) and MIME str (match argument) are the same. Note that it also allows matching against a parameter but a parameter on header does not cause match to fail.

sanic/headers.py
```diff
@@ -33,175 +34,310 @@
...
-            for item in self
+
+class AcceptContainer(AcceptList):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        deprecation(
+            "The AcceptContainer has been renamed AcceptList. Please update "
+            "your code before v23.9 when AcceptContainer will be removed.",
+            23.9,
         )
```
Tronic, line 307: I don't believe this will be ever hit in app code. Users don't create their own objects, and there was no mention of such construction in docs either, so this is clearly internal functionality that doesn't need a deprecation path.

tests/test_headers.py
```diff
@@ -365,6 +339,117 @@ def test_value_not_in_accept(value):
...
+
+def test_accept_misc():
+    header = (
+        "foo/bar;q=0.0, */plain;param=123, text/plain, text/*, foo/bar;q=0.5"
+    )
+    a = headers.parse_accept(header)
+    assert repr(a) == (
+        "[*/plain;param=123, text/plain, text/*, "
+        "foo/bar;q=0.5, foo/bar;q=0.0]"
+    )  # noqa: E501
```
Tronic, line 418: Split lines like this are no longer visible on single line pytest summary and are much harder to read in code, suggest to keep asserts oneliners if possible, or at least the string together in one literal.

sanic/headers.py
```diff
@@ -33,175 +34,310 @@
...
 
 
+def _sort_accept_value(accept: MediaType):
+    return (
+        accept.q,
+        len(accept.params),
+        accept.subtype != "*",
+        accept.type_ != "*",
+    )
```
Tronic, line 316: Okay, I see in RFC this is technically more correct. Should probably use same rules in AcceptList.match then as well, adding to its sort factors all these. I wonder how much performance effect all this has. If it could go significantly faster by matching q only, I would ignore the criteria for "most specific" as not really needed in real world.

In any case, inverting the values -q, -len(...) etc. is probably helpful to have sorting work without reversal (ascending order for highest precedence first, keeping stable sorting of equally valued originals).

Tronic: Suggest storing sort key in MediaType as property during construction to avoid recalculation, and potentially inverting the values here to avoid reverse=True on sorting phase.

sanic/headers.py
```diff
@@ -33,175 +34,310 @@
...
-class MediaType(str):
-    def __new__(cls, value: str):
-        return str.__new__(cls, value)
+    def __init__(
+        self,
+        type_: str,
+        subtype: str,
+        **params: str,
+    ):
+        self.type_ = MediaTypePart(type_)
```
Tronic, line 59: Property name doesn't need underscore

sanic/headers.py
ahopkins: You have no knowledge of that. Breaking changes like this are a constant source of frustration to users. If you want to rename an object, this is a way to do that.

tests/test_headers.py
ahopkins: Take it up with black. If you want to make this change and ignore this rule, we have a lot of refractor work ahead to make this consistent. For now, this simply follows established patterns.

sanic/headers.py
ahopkins: I'm fine storing the keys.

sanic/headers.py
ahopkins: Pretty sure that is a copy/paste change.

sanic/headers.py
Tronic: Actually I do know that. It is only created by the `parse_accept` function, and is internal implementation that is now unreachable. Keeping this deprecation layer (presumably to satisfy tests - didn't look at that yet) is useless bloat.

sanic/headers.py
Tronic: The sorting of course is "breaking", as previously only q values were used IIRC. I am fine with strict RFC compliance but would like to see the benchmark of how much we are paying for that.

tests/test_headers.py
Tronic: Black doesn't break string literals the last I checked, also noqa: E501 avoids errors on too long lines (althought AFAIK linter is not enforced on tests).

sanic/headers.py
ahopkins: This is the exact same function as is in `main`. The only difference is to change the param value from `qvalue` to `q`.

ahopkins:
I have run some benchmarks on the various implementations.

| branch | s / 100,000 parses | RFC compliant
|---|---|---|
| `main` (current implementation) | 1.587453 | :heavy_check_mark: |
| #2663 (my first PR) | 1.583373 | :heavy_check_mark: |
| #2668 (@Tronic  PR) | 0.927016 | :x: |
| #2687 (my second PR) | 1.105352 | :heavy_check_mark: |
| #2687 (my second PR, using only `str`) | 1.081902 | :heavy_check_mark: |

So changing from `MediaTypePart` (a subclass of `str` to plain `str` saves `0.02344999999999997` over 100,000 runs, or an average of `0.234499μs` per run. I am happy to make this change as I am not tied to `MediaTypePart`. It only served the purpose of abstracting away `== '*'`.

Testing further, going to a cached key for sorting is `1.064842`/100k runs or `.405099μs`. It should be noted that these are certainly not meant to be statistically significant as they are super sensitive to background noise on my machine. I run these multiple times and attempt to normalize to gather trends. So at best the numbers are a guideline and not an exact rule.

Tronic:
Good changes. I still suggest leaving out the comparison operators of match objects (btw, I do find the name Accept of that still a bit confusing). The intended use case with my PR is checking which argument matched:

```python
if request.accept.match("application/json", "text/html") == "text/html":
    # respond HTML
else:
    # respond JSON
```

Granted, there are good alternative ways to do that e.g. by accessing the mime property, and many possible semantics for equality comparison of the match objects. Still, having multiple match objects compare by their quality is certainly stretching that and it is hard to imagine real use for that in applications, when the match function already performs selection of best format.

Also, the PR still mixes comparisons by plain q vs. by the sort key, and should be consistent one way or the other in all its operations.

ahopkins:
> I do find the name Accept of that still a bit confusing

OK

> Also, the PR still mixes comparisons by plain q vs. by the sort key, and should be consistent one way or the other in all its operations.

This is certainly a question for open debate.

Consider the following change in `AcceptList.match`:

```python
# from ...
        a = sorted(
            (-acc.q, i, j, mime, acc)
            for j, acc in enumerate(self)
            if accept_wildcards or not acc.has_wildcard
            for i, mime in enumerate(mimes)
            if acc.match(mime)
        )

# to ...
        a = sorted(
            (*acc.key, i, j, mime, acc)
            for j, acc in enumerate(self)
            if accept_wildcards or not acc.has_wildcard
            for i, mime in enumerate(mimes)
            if acc.match(mime)
        )
```

What would you expect the result of this to be?

```python
accept = parse_accept("text/*, text/plain, text/plain;format=flowed, */*")
accept.match("text/csv", "text/plain")
```

Should it be "try and give me `text/csv` if possible" or "give me `text/csv` if explicitly allowed, otherwise fallback to `text/plain`"?

I think it would be more intuitive if the result was `text/csv`. @Tronic?

sanic/headers.py
```diff
@@ -33,173 +33,237 @@
...
+        **params: str,
+    ):
+        self.type = type_
+        self.subtype = subtype
+        self.q = float(params.get("q", "1.0"))
+        self.params = params
+        self.mime = f"{type_}/{subtype}"
+        self.key = (
+            -1 * self.q,
+            -1 * len(self.params),
```
Tronic, line 52: Could use simple negation rather than multiplication (I think Python optimizes but for human readers...)

Tronic (approved):
I guess we've had enough review of this. In principle all looks good to me and it certainly is an improvement over what we had. If you have any finishing touches, feel free to do them but this is LGTM as is or with changes.

Tronic:
Moved documentation from other PR (may not be fully up to date but archiving here anyway):

## Accept header

Sanic has had a helper for parsing the Accept header in HTTP requests since version 21.9, PR #2200. This feature was never documented, but it provides two methods for matching: `"text/html" in request.accept` and `request.accept.match("text/html")`. Both are identical and always produce True if the header includes the wildcard `*/*`, which is included by all clients by default. The latter method has optional kwargs to skip wildcard matches to make it more useful in such cases.

Additionally, `request.accept` is a list that apps can traverse to do their own matching against each item of the header, or print its repr() for debugging purposes. The items themselves are derived from str, but with equality and comparisons implemented by q values only, thus `item == "text/html"` does not do what one might expect.

This PR rewrites the entire handling, changing its semantics and inner function, making it behave in a more practical, less surprising way, and with additional and removed functionality. A default value of `accept: */*` is used if the header is missing, as is required by the RFC. Conversion `str(request.accept)` reformats the parsed and sorted header as an Accept header value.

The match function can now take multiple MIME types as arguments and return the best match based on both the client's and the app's preferences. It returns a `Matched` object that it is truthy alike the earlier bool return value, but tells which argument matched which header item. This is mostly compatible with the old version but the two kwargs are replaced by one that has different semantics: it skips any wildcard entries on the header, but still allows the application to provide wildcard types that will match non-wildcard header items.

The `in` matching uses items' equality comparison and behaves mostly identically to that against the `request.headers["accept"]` string. The item equality comparison is now by literal MIME alone, i.e. wildcards only match identical wildcards.

A deprecation warning will be needed for version 22.12 LTS users who would be affected by these changes, with alternatives that can work with old and new versions.

PR #2697, opened 2023-02-26
Fix comparison from wrong formats with datetime objects

stricaud:
Make sure datetime objects are used in their timestamp format to avoid the exception "TypeError: can't compare offset-naive and offset-aware datetime" with Python 3.11


stricaud:
Any update on this?

sjsadowski:
Hi there, looks good. Can you please add a test to prevent a regression?

stricaud:
This is my first time contributing to Sanic, could you please show me how to add test?

Tronic:
> This is my first time contributing to Sanic, could you please show me how to add test?

Sure. Sanic uses pytest for testing (you can find guides for that online) and you can run the tests on your own machine by
```
# Replace py311 with your Python version (here 3.11)
tox -e py311 -- tests/test_response.py
```
Or alternatively directly with `pytest`, without `tox`. If you leave out the double hyphen and test filename, it will run all tests which will take some time. Before changing anything, see that the tests pass OK.

Since your patch is in sanic/request/, the appropriate test file is probably that mentioned in the example above. It might already have a test function that tests the `validate_file` function. You might be able to modify an existing test function, or simply add your own function named `test_validate_file_datetime`, where you call the function in a way that without your patch would raise an exception.

When doing bugfixes and tests for them, it is a good practice to then try running your test without your patch, to see that it actually fails, and then apply the patch and see that it passes.

Please ask if you have any further questions, we'll be happy to help getting started with that!

Tronic:
@stricaud Hey, how are you doing? Let me know if you need help with the test.

stricaud:
I am trying to write the test, however it has been many days and I don't see how to reproduce this problem. 
I do not have much time. I will see if I can live with a non patched version and see the problem coming again to be able to write the test.

ahopkins:
Looking real quick, was it perhaps a naive v tz-aware issue?

ahopkins:
@stricaud I think I see what you are after. Do you mind if I make some pushes to your branch?

stricaud:
I don't mind! Very welcome

Tronic:
I'd like to make sure that the fix proposed is the right one: what causes tz-aware and unaware timestamps being mixed here, can we make both either tz-aware or not aware? If not, is there a chance that one of them would be in local tz, and produce wrong timestamp value?

ahopkins:
> If not, is there a chance that one of them would be in local tz, and produce wrong timestamp value?

Yup. I am not sure how best to avoid this without going too deep and over-engineering this. My vote is as the implementation I pushed: handle when they are both aware and both naive. If there is a mismatch, add the missing tz as UTC and throw up a warning.

ahopkins (approved):

PR #2700, opened 2023-03-01
Simpler CLI targets

ahopkins:
This simplifies the CLI experience, but should continue to work as it currently does.

Currently, to serve a directory:

```
sanic /some/path --simple
```

And, a factory:

```
sanic path.to.server:create_app --factory
```

Without the `--simple` or `--factory` flags it would fail. This PR will look at the targets passed. If it is a directory, then launch simple server. If it is a function, try calling it.

These are now all possible:

```
sanic path.to.server:app
```
```
sanic /some/path
```
```
sanic path.to.server:create_app
```

The main motivation behind this is the ongoing discussion to encourage the CLI usage more and more.


Tronic:
I have a feeling that this is too ambiguous. Many things could be either module or path, e.g.
```
c:\directory   # unless it checks that after colon only certain characters are used
module.app  # unless it checks that module.app is not a file or directory
```

Maybe with stronger heuristics for the corner cases, combined with Path(...).exists(). For improving usability, it might also be useful to check by default for `app` in the module without having to specify that, i.e. `sanic mymodule` to run mymodule:app.

ahopkins:
Yes, it already does these things. 

> For improving usability, it might also be useful to check by default for app in the module without having to specify that, i.e. sanic mymodule to run mymodule:app.

This would assume people use `app`. Seems a step far for my taste, but I'm not opposed. It would be a bigger change though TBH. I was trying to keep this a small, yet meaningful improvement.

ahopkins:
FTR, passing a file will not work. Only a directory.

ahopkins:
@Tronic 

1. The implied paths _should_ only be used to serve a directory (i) if it is an existing directory, and (ii) has at least one path seperator in it. This should help to differentiate between a directory that happens to be named as a module. Also, no `:` splitting if the input startswith `C:\`. This should help clear up some ambiguity.
2. `sanic somemodule` will look for `app` that is a Sanic instance as per your suggestion.

sanic/worker/loader.py
```diff
@@ -35,7 +39,9 @@ def __init__(
 
         if module_input:
             delimiter = ":" if ":" in module_input else "."
-            if module_input.count(delimiter):
+            if module_input.count(
+                delimiter
+            ) and not module_input.upper().startswith("C:\\"):
```
Tronic, line 44: Suggest `delimeter in module_input and ":\\" not in module_input` (more Pythonic, accounting also for other drive letter than C).

Possibly something different than that to check for paths because on Windows `d:/directory` (i.e. using forward slash, different drive) should be a valid path too.

Tronic:
Doing a full review later. The logic is quite complicated and needs some more thought to see that it handles all use cases correctly.

sanic/worker/loader.py
```diff
@@ -55,21 +63,29 @@ def load(self) -> SanicApp:
             from sanic.app import Sanic
             from sanic.simple import create_simple_server
 
-            if self.as_simple:
-                path = Path(self.module_input)
-                app = create_simple_server(path)
+            maybe_path = Path(self.module_input)
+            if self.as_simple or (
+                maybe_path.is_dir() and os.sep in self.module_input
```
Tronic, line 68: This still only handles one kind of separator on Windows. It should be safe to check for forward and backward slashes here regardless of OS because a module name cannot contain either of those anyway.

Also, possibly would be cleaner to do this either in load or init, but not split/duplicated between the two functions as it is now.

ahopkins: It is doing two different things in those locations.

ahopkins: I am going to get this in now and we can refactor the whole pattern later for cleanliness later.

Tronic (approved):

PR #2701, opened 2023-03-01
API to define a number of workers

ahopkins:
This PR adds a relatively small change to add a `workers` param to the manager so that it is easier to spin up multiple custom workers.

```py
@app.main_process_ready
async def ready(app: Sanic, _):
    app.manager.manage(..., workers=3)
```


Tronic (approved):
LGTM

PR #2703, opened 2023-03-02
Sanic pages coloring

ahopkins:
This is a PR for the color discussion. As changes are applied I will try to keep these screenshots up to date.

image
image
image
image

### DIFF
--- a/sanic/pages/styles/BasePage.css
+++ b/sanic/pages/styles/BasePage.css
@@ -1,8 +1,40 @@
 /** BasePage **/
+
+:root {
+    --sanic: #ff0d68;
+    --sanic-blue: #0092FF;
+    --sanic-yellow: #FFE900;
+    --sanic-purple: #833FE3;
+    --sanic-green: #37ae6f;
+    --sanic-background: #f1f5f9;
+    --sanic-text: #1f2937;
+    --sanic-tab-background: #fff;
+    --sanic-tab-text: #0f172a;
+    --sanic-tab-shadow: #adadad;
+    --sanic-highlight-background: var(--sanic-yellow);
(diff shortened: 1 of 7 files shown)


sanic/pages/styles/BasePage.css
```diff
@@ -2,32 +2,40 @@
...
-    --sanic-blue: #0092FF;
     --sanic-yellow: #FFE900;
-    --sanic-purple: #833FE3;
-    --sanic-green: #37ae6f;
-    --sanic-background: #f1f5f9;
-    --sanic-text: #1f2937;
-    --sanic-tab-background: #fff;
-    --sanic-tab-text: #0f172a;
-    --sanic-tab-shadow: #adadad;
+    --sanic-background: #fff;
```
Tronic, line 6: I would suggest either a slightly darker white or even some color here. A combination of 000 and fff is too highly contrasted on modern displays, and fff should really only be used for highlights. I do realize that a lot of web design is still based on those default colors but it really helps to eyes and makes text more readable if there is a little less brightness. Using colors other than shades of gray would be also nice for the overall look of web but it is harder to pick color combinations that work well.

ahopkins: image
How about like this?

Tronic: Looks better. Probably also should use inherited/same background on details list, and remove padding of the wrapping div so that it aligns with the rest (now it still seems to use fff background and is slightly indented compared to the rest of the page, and in particular to other h2 headings that all should look the same and be aligned the same.

ahopkins: Fixed. All contents now aligned.

Tronic (approved):

PR #2705, opened 2023-03-04
FIX: check whether object is bytes for shared_ctx

### LINKED ISSUE #2690: Don't show warning on bytes in shared_ctx

_No response_

When sending `bytes` over `shared_ctx` we should not display this:

```
Unsafe object num with type <class 'bytes'> was added to shared_ctx. It may not not function as intended. Consider using the regular ctx.
```

### Additional context

_No response_

sjsadowski:
Don't warn if bytes object is added to shared_ctx. Fixes #2690

### DIFF
--- a/sanic/types/shared_ctx.py
+++ b/sanic/types/shared_ctx.py
@@ -38,7 +38,7 @@ class SharedContext(SimpleNamespace):
         if not any(
             module.startswith(prefix)
             for prefix in ("multiprocessing", "ctypes")
-        ):
+        ) and not type(value) is bytes:
             error_logger.warning(
                 f"{Colors.YELLOW}Unsafe object {Colors.PURPLE}{name} "
                 f"{Colors.YELLOW}with type {Colors.PURPLE}{type(value)} "
--- a/sanic/types/shared_ctx.py
+++ b/sanic/types/shared_ctx.py
@@ -40,7 +40,7 @@ class SharedContext(SimpleNamespace):
                 module.startswith(prefix)
                 for prefix in ("multiprocessing", "ctypes")
             )
-            and not type(value) is bytes
+            and not isinstance(value, bytes)
         ):
             error_logger.warning(
                 f"{Colors.YELLOW}Unsafe object {Colors.PURPLE}{name} "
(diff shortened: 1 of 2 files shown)


sanic/types/shared_ctx.py
```diff
@@ -35,9 +35,12 @@ def _check(self, name: str, value: Any) -> None:
...
             module = ""
-        if not any(
-            module.startswith(prefix)
-            for prefix in ("multiprocessing", "ctypes")
+        if (
+            not any(
+                module.startswith(prefix)
+                for prefix in ("multiprocessing", "ctypes")
+            )
+            and not type(value) is bytes
```
ahopkins, line 43: [why not `isinstance`?](https://stackoverflow.com/a/1549854)

sjsadowski: just didn't think about it while I was drinking my coffee - I'll update.

ahopkins:
I am going back on #2690 and am not sure that it is true we should be skipping this warning. While sharing bytes is okay, mutating them is not.

sjsadowski:
Should we check for mutations in the shared context then? Or alternately, make the values immutable?

ahopkins:
Well, there certainly should be mutable objects because the standard library specifically provides for this. But not all objects can be mutable. I suppose the question becomes how to safely do that. I suppose we could subclass standard types and convert them to immutable versions of themselves. 🤔 Open to other thoughts.

Tronic:
I don't quite follow what you are talking of because `bytes` is in fact immutable. Perhaps you meant

```python
s = b"foo"
s += b"bar"  # This is actually a shorthand for s = s + b"bar", original b"foo" object is not mutated
```

ahopkins:
> I don't quite follow what you are talking of because `bytes` is in fact immutable. Perhaps you meant
> 
> ```python
> s = b"foo"
> s += b"bar"  # This is actually a shorthand for s = s + b"bar", original b"foo" object is not mutated
> ```

Sure, that's fine. But that is all happening within the process. I mean objects that are multiprocessing mutable. As in, you apply a mutation to the object and it propogates to the other processes. I am closing this as there is no reason to pass `bytes` to `shared_ctx`. They can be passed to regular `ctx`. Changing the property value will not propogate and sync with other processes and should not be used for this purpose.

PR #2710, opened 2023-03-09
Stricter charset handling and escaping of request URLs

Tronic:
- HTTP1 and HTTP3 built-in servers now enforce ASCII (%-quotes should be used for UTF-8).
- ASGI uses `raw_path` to avoid quote/encode cycle and unquoting of `%2F` into `/` early on.
- ASGI no longer prepends `root_path` to path used in Sanic.
- Using backslash escapes in logging and BadURL messages to print invalid characters.

### DIFF
--- a/sanic/asgi.py
+++ b/sanic/asgi.py
@@ -143,14 +143,9 @@ class ASGIApp:
         if scope["type"] == "lifespan":
             await instance.lifespan(scope, receive, send)
         else:
-            path = (
-                scope["path"][1:]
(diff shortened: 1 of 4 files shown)


Tronic:
Note: earlier discussion in https://github.com/sanic-org/sanic/pull/1837#issuecomment-621674418

ahopkins:
Can we add some tests to catch regressions?

ahopkins (approved):

PR #2716, opened 2023-03-19
Adding allow route overwrite option in blueprint

### LINKED ISSUE #2554: Support overwriting a route of blueprint from copy
**Feture Description**
I want to overwrite the implement of route partially when they belong to different version blueprint, and it raises sanic_routing.exceptions.RouteExists.
**Sample Code**
```python
from sanic import Blueprint
from sanic import Sanic

app = Sanic('test')

bpv1 = Blueprint('bpv1', version=1)

@bpv1.route('/hello')
async def root(request):
    return json('hello v1')

app.blueprint(bpv1)

bpv2 = bpv1.copy('bpv2', version=2)

@bpv2.route('/hello')
async def root(request):
    return json('hello v2')

app.blueprint(bpv2)
```
**Current Solution**
I get this solution from forum and i was encouraged to post new a feature request based on it~
```python
bpv2 = bpv1.copy("bpv2", version=2)

bpv2._future_routes = {
    route for route in bpv2._future_routes if route.uri != "/hello"
}


@bpv2.route("/hello")
async def root2(request):
    return json("hello v2")
```

**Web Link**
https://community.sanicframework.org/t/how-to-overwrite-a-route-when-using-blueprint-copy/1067

ChihweiLHBird:
closes #2554


ahopkins (commented):
Don't forget a test 😎

ahopkins:
@ChihweiLHBird Don't bother trying to keep it up to date and merging in latest (unless part of a local commit). It will likely change 10 more times before this is ready.

ChihweiLHBird:
@ahopkins Thanks! I added one test case for overwriting route after bp copy. It probably needs another test case for a non-copied blueprint to allow overwriting a route handler. Do you plan to include this PR in this release?

1.py
ahopkins, line 1: What's this?

tests/test_blueprint_copy.py
```diff
@@ -66,3 +66,29 @@ def handle_request(request):
 
     _, response = app.test_client.get("/version6/page")
     assert "Hello world!" in response.text
+
+
+def test_bp_copy_with_route_overwriting(app: Sanic):
+    bpv1 = Blueprint("bp_v1", version=1)
```
ahopkins, line 72: I'd throw something like `url_prefix` on the test just so that it shows some constant value is being copied and still applied.

ChihweiLHBird:
GitHub seems having a delay to sync branch with PR... I try initializing a bp instance with overwriting route option set to `True`, and it seems it is not functioning well. I would prefer to postpone this to the next release instead of rushing it out.

1.py
ChihweiLHBird: Sorry, some test code accidentally in the commit. Removed.

ChihweiLHBird:
@ahopkins Thanks!

tests/test_blueprint_copy.py
```diff
@@ -74,3 +74,42 @@ def handle_request(request):
     assert "test_bp_copy.test_bp4.handle_request" in route_names
     assert "test_bp_copy.test_bp5.handle_request" in route_names
     assert "test_bp_copy.test_bp6.handle_request" in route_names
+
+
+def test_bp_copy_with_route_overwriting(app: Sanic):
```
ahopkins, line 79: Can you also add a `test_bp_copy_without_route_overwriting` test that specifcally is set to `False` and catches/asserts on the exception?

tests/test_blueprints.py
```diff
@@ -1112,3 +1112,21 @@ async def index(_):
...
+    def handler1(request):
+        return text("Hello")
+
+    @bp.route("/")
+    def handler2(request):
+        return text("Overwritten Hello")
+
+    app.blueprint(bp)
+    _, response = app.test_client.get("/")
+
```
ahopkins, line 1130: This pattern strikes me as dangerous. What would the use case be when we would want to allow this? On `copy` it makes sense, but why here?

ChihweiLHBird: Because the flag in at the `Blueprint` object attribute level so it allows this when the flag is set to `True` when the object created. If this is not desired and looks dangerous, I can remove the flag parameter from the `Blueprint` object constructor to make it safer. How does that sound?

tests/test_blueprint_copy.py
ChihweiLHBird: Added

tests/test_blueprint_copy.py
ChihweiLHBird: added

ChihweiLHBird:
We can consider to proceed with [this PR](https://github.com/sanic-org/sanic/pull/2746) if that's a better way to implement it.

tests/test_blueprints.py
ahopkins: Yeah, I think we should not expose it. If if needs to be a flag on the Blueprint, let's make it internal `_allow_route_overwrite` and not expose it "publicly".

ChihweiLHBird: Updated

ahopkins (approved):

PR #2717, opened 2023-03-19
v23.3 Deprecation Removal

ahopkins:
- duplicate names
- duplicate exception handler definition
- inspector CLI with flags
- legacy server (including `sanic.server.serve_single` and `sanic.server.serve_multiple`)
- serving directory with bytes string
- `Request.request_middleware_started`
- `Websocket.connection`


Tronic (dismissed):
Seems correct but I didn't dig into all the removed server code.

PR #2728, opened 2023-03-30
Fix traversals for intended results

### LINKED ISSUE #2726: Static routes containing ".." don't work

### Describe the bug

Sanic does serve static files if the server file path contains "..", instead it returns file not found errors.

### Code snippet

The following doesn't work:
```
from sanic import Sanic

app = Sanic("Example")
app.static("/", "./static/../static/")
if __name__ == "__main__":
```
Attempt to access any files in the static directory will get the error:
`[2023-03-27 04:05:31 -0700] [60813] [ERROR] File not found: path=static/../static, relative_url=test.txt`
Removing the redundant `../static/` will result in the files being correctly served.

### Expected Behavior

The above example should serve all files in the static directory and not return file not found errors.

### How do you run Sanic?

Sanic CLI

### Operating System

linux

### Sanic Version

22.9.1

### Additional context

This issue appears to have been introduced by #2506 or #2508.
The fix might look something like the following:
```
diff --git a/sanic/mixins/static.py b/sanic/mixins/static.py
index bcffbc8..28effc4 100644
--- a/sanic/mixins/static.py
+++ b/sanic/mixins/static.py
@@ -323,7 +323,7 @@ class StaticHandleMixin(metaclass=SanicMeta):
             # python from herping a derp and treating the uri as an
             # absolute path
-            file_path_raw = Path(file_or_directory, unquoted_file_uri)
+            file_path_raw = Path(root_path, unquoted_file_uri)
             file_path = file_path_raw.resolve()
             if (
                 file_path < root_path and not file_path_raw.is_symlink()
```

It looks like the ".." was originally being done on the URI part only but with the #2506 change this is now done on the combination of the root path *and* relative path. The above change instead uses the resolved form of the root path (which won't contain "..").

However, there are likely other options and having more context about what the checks in this area are for would be crucial.

ahopkins:
Closes #2726


tests/test_static.py
```diff
@@ -101,6 +101,31 @@ def test_static_file_pathlib(app, static_file_directory, file_name):
...
+)
+def test_static_file_pathlib_relative_path_traversal(
+    app, static_file_directory, file_name
+):
+    """Get the current working directory and check if it ends with "sanic" """
+    cwd = Path.cwd()
+    if not str(cwd).endswith("sanic"):
+        pytest.skip("Current working directory does not end with 'sanic'")
+
+    file_path = "./tests/static/../static/"
```
chuckds, line 122: Would it be possible to remove the dependency on cwd (and the directory structure) by using setting file_path to:
```
str(
        Path(static_file_directory) / ".." / Path(static_file_directory).name
    )
```

ahopkins: Possibly. I will take a look at it again. It was being a PITA and I just wanted to get something in for the test to run.

ChihweiLHBird (dismissed):

tests/test_static.py
ChihweiLHBird: Maybe we can try python [tempfile module](https://docs.python.org/3/library/tempfile.html) for the testing files.

PR #2729, opened 2023-03-30
Handle case when `headers` argument of `ResponseStream` constructor is `None`

liamcoatman:
The default value for the `headers` argument is `None`. This leads to an Exception when `headers` is passed to the `Header` constructor, which is expecting a Mapping type.


liamcoatman:
@ahopkins I've added a test, which was failing prior to my fix and is now passing. I've probably put it in the wrong place, so let me know if you want me to move it.

ahopkins:
It's a long file, but I'd dump it here: https://github.com/sanic-org/sanic/blob/main/tests/test_response.py

liamcoatman:
In case you missed it, I moved the test to `test_response.py` as suggested

QylinFly:
ResponseStream is not stream,bug?

ahopkins:
> ResponseStream is not stream,bug?

Sorry, I don't understand your question.

ahopkins (approved):

PR #2741, opened 2023-04-17
fix examples for freeze_support() issue on windows

deounix:
Sanic on windows should always run inside:

`if __name__ == "__main__": app.run()`

Check this: https://docs.python.org/2/library/multiprocessing.html#the-process-class

`For an explanation of why (on Windows) the if __name__ == '__main__' part is necessary, see` [[Programming guidelines]](https://docs.python.org/2/library/multiprocessing.html#multiprocessing-programming)

Without the `__name__ == "__main__"` it throw this error on windows:

```
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```

On Linux, Mac it's run normally.
I think we should update the docs to notify the users about this issue.

### DIFF
--- a/examples/amending_request_object.py
+++ b/examples/amending_request_object.py
@@ -25,5 +25,5 @@ def key_exist_handler(request):
 
     return text("num does not exist in request")
 
-
-app.run(host="0.0.0.0", port=8000, debug=True)
+if __name__ == "__main__":
+    app.run(host="0.0.0.0", port=8000, debug=True)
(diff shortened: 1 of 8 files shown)


SaidBySolo (approved):
Good catch

ahopkins (approved):

PR #2742, opened 2023-04-17
Fix typing problems and add # type: ignore to ignore platform related errors

deounix:


deounix:
I have tested the code with mypy and there were no typing errors.

`Success: no issues found in 114 source files`

sanic/mixins/startup.py
```diff
@@ -308,7 +308,7 @@ def prepare(
         if fast:
             self.state.fast = True
             try:
-                workers = len(os.sched_getaffinity(0))
+                workers = len(os.sched_getaffinity(0)) # type: ignore
```
ahopkins, line 311: Couldn't we just solve that above with:

```py
workers: int
...
```

sanic/models/asgi.py
```diff
@@ -27,9 +27,9 @@ def __init__(self, transport: "MockTransport", loop):
             self._not_paused.set()
             self._complete = asyncio.Event()
         else:
-            self._not_paused = asyncio.Event(loop=loop)
+            self._not_paused = asyncio.Event(loop=loop) # type: ignore
             self._not_paused.set()
-            self._complete = asyncio.Event(loop=loop)
+            self._complete = asyncio.Event(loop=loop) # type: ignore
```
ahopkins, line 32: What is the type error? I would rather fix the typing than ignore it.

sanic/response/types.py
```diff
@@ -521,7 +521,7 @@ def __init__(
         content_type: Optional[str] = None,
     ):
         if not isinstance(headers, Header):
-            headers = Header(headers)
+            headers = Header(headers) # type: ignore
```
ahopkins, line 524: Same.

sanic/models/asgi.py
deounix: `Python 3.11.3`: error: Unexpected keyword argument "loop" for "Event"

sanic/response/types.py
deounix: `sanic\response\types.py:524: error: Argument 1 to "Header" has incompatible type "Optional[Dict[str, str]]"; expected "Union[Mapping[str, Any], Mapping[istr, Any], Dict[str, Any], Dict[istr, Any], MultiMapping[Any], Iterable[Tuple[str, Any]], Iterable[Tuple[istr, Any]]]"`

sanic/mixins/startup.py
deounix: `sanic\mixins\startup.py:311: error: Module has no attribute "sched_getaffinity"`

sanic/exceptions.py
```diff
@@ -59,7 +59,7 @@ class SanicException(Exception):
     :type headers: Optional[Dict[str, Any]], optional
     """
 
-    status_code: int = 500
+    status_code: Union[int, Any, None] = 500
```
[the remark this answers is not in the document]
deounix: updated.

```
      status_code = status_code or getattr(
          self.__class__, "status_code", None
      )
```

The getattr function returning: `(Any | None)`

ahopkins:
I think this is old and conflicted enough that I am closing. Please feel free to reopen again against the codebase and we will merge.

PR #2745, opened 2023-04-25
sanic/app: When streaming, check REQUEST_MAX_SIZE if length is known

ozancaglayan:
Following the [discussion](https://community.sanicframework.org/t/streaming-handling-improvements/1192), this PR adds a check for the streaming case to early quit request processing in case request has a content length larger than the `REQUEST_MAX_SIZE`. Tested locally and it seems to work.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -57,6 +57,7 @@ from sanic.compat import OS_IS_WINDOWS, enable_windows_color_support
 from sanic.config import SANIC_PREFIX, Config
 from sanic.exceptions import (
     BadRequest,
+    PayloadTooLarge,
     SanicException,
     ServerError,
     URLBuildError,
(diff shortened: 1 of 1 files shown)


Tronic:
This appears to change functionality such that streaming requests are no longer of infinite size (which is intended functionality). Can you accomplish the same if you raise that exception in your handler? Also, which client are you testing with? Curl and possibly httpx (the test client) behave differently of browsers because they use `expect: 100-continue` mechanism that allows rejecting the request without receiving it; possibly this is what you accomplish with this PR, but it browsers don't send that header and thus won't be "fixed" if that is the case.

My comments are a bit fuzzy - sorry about that - because I don't have time to dig into the code and run my own tests on this right now.

ahopkins:
I thought this would only be something to take into account if there is a known length, and not just any streaming request.

PR #2746, opened 2023-05-03
Parameter [overwrite] is supported when adding a route. 

haok228:
- addding a route in blueprints
``` python
from sanic import Blueprint

v1 = Blueprint(name="Version1", version=1)

@v1.route("/something")
def something(request):
    return json({"my": "something1"})

v2 = v1.copy("Version2", version=2)

# sanic_routing.exceptions.RouteExists: Route already registered: v2/something [GET]
@v2.route("/something", overwrite=True)
def something(request):
    return json({"my": "something2"})
```


ahopkins:
Thanks for the effort on this PR. With the change in #2716, is this still necessary? Are you trying to achieve a different outcome?

```py
bpv2 = bpv1.copy("bp_v2", version=2, allow_route_overwrite=False)
```

ahopkins:
I'm closing this because I do not think we need it anymore. If I am wrong, please LMK.

PR #2761, opened 2023-06-12
custom json library

hamedsh:
ability to use custom json loads and dumps libraries (orjson or rapidjson) like newer Sanic libraries.


ahopkins (commented):
Took me a minute to realize this was on v20. Took some head scratching because the code looked super old, and I was positive that this API already existed. Thanks, will take a look.

One thing that is missing for sure are some unit tests.~

PR #2775, opened 2023-07-09
Start and restart arbitrary processes

ahopkins:
The idea is to provide a better API to allow restarting of arbitrary processes. Also, it is meant to provide a `manage` method to the multiplexer so that endpoints can also start arbitrary workers.

```py
def task(n: int = 10, **kwargs):
    print("TASK STARTED", kwargs)
    for i in range(n):
        print(f"Running task - Step {i+1} of {n}")
        sleep(1)

@app.get("/restart")
async def restart_handler(request: Request):
    request.app.m.restart("Sanic-TEST-0")
    return json({"foo": request.app.m.name})


@app.get("/start")
async def start_handler(request: Request):
    request.app.m.manage("NEW", task, kwargs={"n": 7}, workers=2)
    return json({"foo": request.app.m.name})

@app.main_process_ready
def start_process(app: Sanic):
    app.manager.manage("TEST", task, kwargs={"n": 3}, restartable=True)
```

TODO:
- [x] Cleanup the API (passing messages is a bit messy)
- [x] Finalize the API (`restartable`, `tracked`, etc)
- [ ] ~Possibly add a worker pool option?~
	- Future enhancement. Not this go-around.


ahopkins:
Ready for review, I'm still working on tests, but I want to get codecov to run to see what it thinks I'm missing for coverage.

PR #2781, opened 2023-07-10
Initial build

ahopkins:
This is an initial POC build of the User Guide build using Sanic, html5tagger, and htmx. It is meant to be a thought experiment with building the documentation from the main repository on our own platform. We then would be able to:

1. Have an example application running in production to point people towards that is running Sanic
2. Bring the writing of documentation closer to PRs so that we can keep them aligned as changes are made

image


Tronic:
It's a very large diff. Clearly this does something other than html5tagger and htmx? Should also cover msgspec in docs.

Staying unopionated, JS fronts like htmx (with html5tagger) and vue (with msgspec) should probably be left to their separate tutorial pages, rather than discussed in the main guide. The same goes for databases (urgently needed in practice, still with poor async support).

guide/content/en/guide/getting-started.md
```diff
@@ -0,0 +1,96 @@
...
+
+
+
+
+.. info:: Tip
+
+    If you are coming from Flask or another framework, there are a few important things to point out. Remember, Sanic aims for **performance**, **flexibility**, and **ease of use**. These guiding principles have tangible impact on the API and how it works.
+    
+    - Sanic does not use a global context. You do not need to import a `request` object from somewhere else. Instead, that object is passed to your handler.
+    - Route handlers must return a response object. This is a very important distinction. Other frameworks allow you to return a string or a dictionary and they will try to figure out what you meant. Sanic requires you to be explicit about what you want to return, therefore saving valuable time in the request/response cycle.
```
Tronic, line 38: This is sort of a FAQ and a turndown for some. I don't think the performance argument is relevant because the check is quite fast. Suggested clarified formatting:
```suggestion
    - Handlers must return a response object rather than bare data. This distinguishes our approach from other frameworks that let you return a string or dictionary, which they convert into html or json with a 200 OK status. By insisting on a response object, we avoid assuming just those two formats and allow for custom headers and status codes like 201 Created, and also enable the support of other formats such as msgpack. To make the process more streamlined for your API, a response middleware can be used to convert any data your handlers return into an appropriate response before it reaches Sanic.
```

The middleware approach works but hasn't ever been quite officially supported. I think we should make it so. On a blueprint level this can come in handy.

ahopkins:
> It's a very large diff. Clearly this does something other than html5tagger and htmx? Should also cover msgspec in docs.

@Tronic I think you misunderstood what this branch is. It is an attempt to drop `sanic-guide` and vuepress in favor of dogfooding Sanic. To achieve templating it uses html5tagger. To get a similar crisp navigation feel, it uses htmx. Not sure msgspec plays into this at all.

> Staying unopionated, JS fronts like htmx (with html5tagger) and vue (with msgspec) should probably be left to their separate tutorial pages, rather than discussed in the main guide. 

No disagreement. Sanic does not take an opinion. But, clearly in building its own site it needs to make some decisions. In the past it was was Sphinx with RTD. Then vuepress with GH. I am proposing a POC for how we can serve https://sanic.dev using Sanic.

> This is sort of a FAQ and a turndown for some. I don't think the performance argument is relevant because the check is quite fast.

Make the suggestion and discussion here: https://github.com/sanic-org/sanic-guide. That is somewhat irrelevant since it is a copy/past with changes only to handle a difference in Markdown extras.

> The middleware approach works but hasn't ever been quite officially supported. I think we should make it so. On a blueprint level this can come in handy.

Same.

guide/content/en/guide/getting-started.md
ahopkins: I just reread what you wrote...

We do have an official pattern for this: https://sanic.dev/en/plugins/sanic-ext/convenience.html#custom-serializer

```python
def message(retval, request, action, status):
    return json(
        {
            "request_id": str(request.id),
            "action": action,
            "message": retval,
        },
        status=status,
    )


@app.post("/<action>")
@serializer(message)
async def do_action(request, action: str):
    return "This is a message"
```

However, I suspect this would cause some typing annotation issues that need to be solved.

PR #2785, opened 2023-07-11
Add custom typing to config and ctx

ahopkins:
One thing I have always found slightly annoying is that the reliance upon `app.ctx` and `app.config` lead to only partially typed objects. This PR attempts to resolve that by adding some generics into `sanic.Sanic` and `sanic.Request`:

```python
app: Sanic[CustomConfig, CustomCtx]
```

This can also be achieved by implication just through instantiation:

```python
from sanic import Sanic
app = Sanic("test")
reveal_type(app)  # N: Revealed type is "sanic.app.Sanic[sanic.config.Config, types.SimpleNamespace]"
```
```python
from sanic import Sanic
from sanic.config import Config

class CustomConfig(Config):
    pass

app = Sanic("test", config=CustomConfig())
reveal_type(app)  # N: Revealed type is "sanic.app.Sanic[main.CustomConfig, types.SimpleNamespace]"
```
```python
from sanic import Sanic

class Foo:
    pass

app = Sanic("test", ctx=Foo())
reveal_type(app)  # N: Revealed type is "sanic.app.Sanic[sanic.config.Config, main.Foo]"
```
```python
from sanic import Sanic
from sanic.config import Config

class CustomConfig(Config):
    pass

class Foo:
    pass

app = Sanic("test", config=CustomConfig(), ctx=Foo())
reveal_type(app)  # N: Revealed type is "sanic.app.Sanic[main.CustomConfig, main.Foo]"
```

The goal of this PR is a nicer dev experience to be able to fully type out the ctx and config objects.

Given a simple handler with no custom typing:
```python
@app.get("/")
    reveal_type(request.app)
```
image

Or, a fully custom typed:

image

A side benefit is that `sanic.Request.ctx` is now a lazy object and will not be instantiated until used. This is accomplished by introducing a new staticmethod on the `Request` object. It can be used to instantiate a custom ctx object as shown in the above screenshot. This is the only functional change in this PR: moving `self.ctx = SimpleNamespace()` into the lazy `make_context` pattern.

## TODO:
- [x] Add generics also to `Request.ctx`
- [x] Add generics also to `Request` so that it can pass the ref in a handler for example


sanic/server/websockets/frame.py
```diff
@@ -96,6 +96,7 @@ async def get(self, timeout: Optional[float] = None) -> Optional[Data]:
         If ``timeout`` is set and elapses before a complete message is
         received, :meth:`get` returns ``None``.
         """
+        completed: bool
```
prryplatypus, line 99: Is this related to this PR at all?

sanic/server/websockets/impl.py
```diff
@@ -21,7 +21,7 @@
 
 
 try:  # websockets < 11.0
-    from websockets.connection import Event, State
+    from websockets.connection import Event, State  # type: ignore
```
prryplatypus, line 24: Why's this needed now and it wasn't before?

tests/typing/test_typing.py
```diff
@@ -0,0 +1,127 @@
...
+    path = CURRENT_DIR / path_location
+    command = [mypy_path, path.resolve().as_posix()]
+
+    process = subprocess.run(
+        command,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        universal_newlines=True,
+    )
+    output = process.stdout + process.stderr
```
prryplatypus, line 28: Do we need to check if stuff is in stderr too for any specific reason?

sanic/server/websockets/frame.py
ahopkins: It was a typing error that was caught when I undid the upper limit on mypy.

sanic/server/websockets/impl.py
ahopkins: Because it is not importable from that location (which is precisely why we have the try/except). mypy upgrade now complains and is not smart enough to capture the try/except logic.

sanic/server/websockets/frame.py
prryplatypus: I see

tests/typing/test_typing.py
ahopkins: If there is a typing error, we don't really care. It does not need to cause this test to fail. Here, we only care about the notice of the `reveal_type`. We will let `tox -e type-checking` report on the actual issue. Conceivably, `mypy` could fail, but all these tests pass.

prryplatypus (approved):

PR #2792, opened 2023-07-17
Add convenience method for exception reporting

ahopkins:
#2724 adds a new signal to trigger on all exceptions: `"server.lifecycle.exception"`

This PR adds a convenience for setting up a task for capturing this signal to do something with it. The obvious use case is for error reporting that does not need to be in the way of returning a response.

```python
@app.report_exception
async def catch_any_exception(app: Sanic, exception: Exception):
    print("Caught exception:", exception)
```

Also, since it is still an unreleased signal, this PR renames it to `"server.exception.report"`, which is I think is a less confusing name since it is not strictly tied to server lifecycle.

Lastly, in exploring this earlier PR, I realized that there was a bug (double dispatch on some built-in signals that happened inline in a request), and I also realized that we were not capturing exceptions in background tasks. This PR handles these scenarios as well.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -613,7 +613,7 @@ class Sanic(
             await handler(self, exception)
 
         self.add_signal(
-            handler=report, event=Event.SERVER_GLOBAL_EXCEPTION.value
+            handler=report, event=Event.SERVER_EXCEPTION_REPORT.value
         )
 
         return report
@@ -890,7 +890,7 @@ class Sanic(
         """
         response = None
         await self.dispatch(
-            "server.lifecycle.exception",
+            "server.exception.report",
             context={"exception": exception},
         )
         await self.dispatch(
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -890,6 +890,7 @@ class Sanic(
         """
         response = None
         if not getattr(exception, "__dispatched__", False):
+            ...  # DO NOT REMOVE THIS LINE. IT IS NEEDED FOR TOUCHUP.
             await self.dispatch(
                 "server.exception.report",
                 context={"exception": exception},
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -613,7 +613,7 @@ class Sanic(
             await handler(self, exception)
 
         self.add_signal(
-            handler=report, event=Event.SERVER_LIFECYCLE_EXCEPTION.value
+            handler=report, event=Event.SERVER_GLOBAL_EXCEPTION.value
         )
 
         return report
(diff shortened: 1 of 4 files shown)


sanic/app.py
```diff
@@ -1310,13 +1325,28 @@ def _prep_task(
...
+                error_logger.warning(
+                    f"Task {task} was cancelled before it completed."
+                )
+                raise
+            except Exception as e:
+                await app.dispatch(
+                    "server.exception.report",
+                    context={"exception": e},
+                )
+                raise
```
ahopkins, line 1348: Codecov, you lie. Both of those are covered by tests :unamused:

sanic/app.py
```diff
@@ -876,10 +889,12 @@ async def handle_exception(
         :raises ServerError: response 500
         """
         response = None
-        await self.dispatch(
-            "server.lifecycle.exception",
-            context={"exception": exception},
-        )
+        if not getattr(exception, "__dispatched__", False):
+            ...  # DO NOT REMOVE THIS LINE. IT IS NEEDED FOR TOUCHUP.
```
prryplatypus, line 893: For anyone curious (thanks @ahopkins for the explanation):

> At startup, there are a few performance critical methods that are introspected.
> We generate an AST and walk thru it applying each of a few different operaitons
> One of them looks for a `dispatch` method, then compares the event to the list of registered events in the signal router. If there is no handler, then we remove the `dispatch` from the AST, and then finally recompile the source.
> This way, we do not even call an empty `dispatch` if there will never be a signal on the other end to receive it.
> This has the potential to leave an empty block though. The quick and simple solution is to put an ellipsis or pass in the block. I had been planning to do this dynamically, but never got back to it.
> 
> TL;DR: Ok so the ellipsis is just there in case the dispatch gets removed during touchup because there's no listeners for it.

prryplatypus (approved):

sanic/signals.py
```diff
@@ -174,11 +174,12 @@ async def _dispatch(
...
                 error_logger.exception(e)
 
-            if event != Event.SERVER_LIFECYCLE_EXCEPTION.value:
+            if event != Event.SERVER_EXCEPTION_REPORT.value:
                 await self.dispatch(
-                    Event.SERVER_LIFECYCLE_EXCEPTION.value,
+                    Event.SERVER_EXCEPTION_REPORT.value,
                     context={"exception": e},
                 )
+                setattr(e, "__dispatched__", True)
```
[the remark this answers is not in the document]
ahopkins: That will induce a type error, and therefore need a type ignore.

PR #2796, opened 2023-07-24
Refactoring unit test cases

freddiewanah:
Hello, first-time submitting PR here. 

This PR improves the readability of the assert statements in the unit tests. These changes improve code quality (avoid test smells) by:

- Using isinstance() is the more Pythonic way to check types, as it supports polymorphism, and also consistent with other test files within the project.

- Simplifying code management. For example, instead of simply saying "application/octet-stream", use the parameter name(DEFAULT_HTTP_CONTENT_TYPE) so that developers could only update the value in one place.

Happy to update more related issues in the test code if needed :)

### DIFF
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -20,6 +20,7 @@ from pytest import LogCaptureFixture
 
 from sanic import Request, Sanic
 from sanic.compat import Header
+from sanic.constants import DEFAULT_HTTP_CONTENT_TYPE
 from sanic.cookies import CookieJar
 from sanic.response import (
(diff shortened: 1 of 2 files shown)


ahopkins (approved):

prryplatypus (approved):

ChihweiLHBird (approved):

ahopkins:
@freddiewanah Thanks for submitting this. Any time we get an opportunity to improve tests it is a huge win. I'd be super happy to have your help working more on tests or anywhere else that you would be interested in contributing.

PR #2801, opened 2023-08-17
Fix test_fast when there's only one CPU

tjni:
I ran into this when running the tests in a VM that only had one CPU.


ahopkins (approved):

SaidBySolo (commented):
Rather than an if statement, it would be better to write one more test function when there is only one worker.

PR #2808, opened 2023-08-23
Refactor GitHub Actions

prryplatypus:
- Simplifies all Tox actions by using a `simple-tox` action that does all the initial test setup transparently.
- Merges all PR actions into a single `pr-tests.yml` file
- Improves release flow merging PyPi and Docker release flow into `publish-release.yml`:
  - Package version gets injected into the package automatically when publishing to PyPi
  - Docker images get built after package is pushed to PyPi with the following tags (using Python 3.9 as an example):
    - `3.9-latest` if the "Set as the latest release" checkbox is marked when creating a release
    - `3.9-YY.MM` every time
    - `3.9-LTS` if the release `MM` is `12`


prryplatypus:
Actions are failing because of #2807

ahopkins:
One thing I've also been thinking about doing is dropping codecov for a number of reasons, and simplifying to just use the coverage.py results. CI can create the HTML results and stash them as an artifact to display with maybe posting back a comment on the PR. WDYT?

prryplatypus:
> One thing I've also been thinking about doing is dropping codecov for a number of reasons, and simplifying to just use the coverage.py results. CI can create the HTML results and stash them as an artifact to display with maybe posting back a comment on the PR. WDYT?

Could look into it. For a future PR maybe?

.github/workflows/coverage.yml
```diff
@@ -12,29 +12,23 @@ on:
...
-        run: tox -e coverage
-        continue-on-error: true
-      - uses: codecov/codecov-action@v2
+        uses: prryplatypus/simple-tox@v1
+        with:
+          python-version: "3.11"
+          tox-env: coverage
+          ignore-errors: true
+      - name: Run Codecov
+        uses: codecov/codecov-action@v2
```
ChihweiLHBird, line 31: Can we use the latest v3 of this action?

ChihweiLHBird:
This refactor looks so great! Thanks for the effort!

.github/workflows/coverage.yml
prryplatypus: Yes, we definitely can! I didn't know there was a v3 (just copy-pasted the old one), but I'll update it!

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,174 @@
...
+        run: |
+          latest_tag=$(
+            curl -L \
+              -H "Accept: application/vnd.github+json" \
+              -H "Authorization: Bearer ${{ github.token }}" \
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              https://api.github.com/repos/${{ github.repository }}/releases/latest \
+              | jq -r '.tag_name'
+          )
+          echo "latest_tag=$latest_tag" >> $GITHUB_OUTPUT
```
ChihweiLHBird, line 58: Using the official API action might make this easier and more readable. And I also think it would be fine to use CURL, lol.

https://github.com/octokit/request-action

ChihweiLHBird (approved):
Looks good to me! But I think it would be nicer if we can test the release in a fork of the repo.

.github/workflows/coverage.yml
```diff
@@ -12,29 +12,23 @@ on:
...
-
-      - name: Install dependencies 🔨
-        run: |
-          python -m pip install --upgrade pip
-          pip install tox
       - name: Run coverage
-        run: tox -e coverage
-        continue-on-error: true
-      - uses: codecov/codecov-action@v2
+        uses: prryplatypus/simple-tox@v1
```
ahopkins, line 25: :suspect: 

Let's not replace one custom action with another. If we cannot do it without, then it should at least be something owned by the SCO.

ahopkins (changes requested):

prryplatypus:
> Looks good to me! But I think it would be nicer if we can test the release in a fork of the repo.

We did! There's a bunch of test releases in test pypi.org and in my Docker account (prryplatypus). :)

.github/workflows/publish-release.yml
prryplatypus: I've looked at this, and I think I'm gonna go for cURL for the time being. Main reason being because the output the action returns is in string format, and then you have to manually convert it to JSON by calling the `toJSON` function whenever you want to use it, which isn't too nice imo. Here's the diff if ur curious :D

imagen

.github/workflows/coverage.yml
ahopkins: it's days are numbered

ahopkins (approved):

PR #2812, opened 2023-08-29
Suppress task cancel traceback

ahopkins:
If you have an open websocket when the server shutsdown (or restarts), there is an annoying traceback for the `recv`. This PR properly cancels that task so that it does not clutter logs.


prryplatypus (approved):
E

PR #2817, opened 2023-09-05
Improved URL display on MOTD

Tronic:
Sanic would use any host setting for URL construction. This PR fixes the common cases of normal localhost and wildcard addresses, for which we instead display localhost. Similarly for UNIX sockets, as that format works e.g. with curl. URL to wildcard doesn't work at all, and connecting localhost by IP rather than domain name can be problematic with TLS if not otherwise.

Additionally, hides the port number if standard ports are used.

### DIFF
--- a/sanic/mixins/startup.py
+++ b/sanic/mixins/startup.py
@@ -657,10 +657,9 @@ class StartupMixin(metaclass=SanicMeta):
     def get_server_location(
         server_settings: Optional[Dict[str, Any]] = None
     ) -> str:
-        serve_location = ""
         proto = "http"
         if not server_settings:
-            return serve_location
+            return ""
 
         host = server_settings["host"]
         port = server_settings["port"]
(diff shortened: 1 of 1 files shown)


Tronic:
Setting `host` to localhost instead of IP probably needs to be done in server startup rather than MOTD only, to make `url_for` work as expected without extra configuration. This is tricky, leaving in draft mode for now.

ahopkins:
> This is tricky, leaving in draft mode for now.

Maybe also time to handle the whole server name issue. :wink:

Tronic:
> Maybe also time to handle the whole server name issue. 😉

That's a can of worms, for sure. Would be useful to first identify all the problems that we are solving here, and things affected by them.

- External URL forming (btw, should use `app.url_for` in MOTD too).
   * Scheme and host and port may be different than locally seen
   * Currently by `SERVER_NAME`, what are the concerns about this (URL vs. host only, host:port)
   * If apps can be externally mounted also at some path other than `/`, internal URLs are also affected
- Host (request header) fallback
   * Local or remote values here? Start giving 400 Bad Request to those without valid header?
- Listening address(es) to bind (IPs, unix path)
   * Might be displayed separately on MOTD, should always be displayed inside the box?
- CLI and `app.run` interfaces for specifying these things?
- What is `--host` used for?
   * Binding or hostname, both?
   * Changing the default to `localhost` would make sense and avoid some special cases
- Are we adding a `--listen` shorthand for doing all at once and/or listening to more than one
   * Could specify multiple to bind, but the server can only have one host:port with `app.url_for`
   * Listen to IPv4 and IPv6 wildcards if `:8000` (nothing before colon) is specified (matches Caddy)
   * Default to 80/443 if plain hostname/ip is given?
   * ACME client / localhost CA, background server on 80? (redirect/ACME, matches Caddy)

What'd I miss?

PR #2818, opened 2023-09-05
Better handling of port bind failures

Tronic:
`sanic --port 80` was printing very ugly error messages.

### DIFF
--- a/sanic/mixins/startup.py
+++ b/sanic/mixins/startup.py
@@ -49,7 +49,7 @@ from sanic.application.motd import MOTD
 from sanic.application.state import ApplicationServerInfo, Mode, ServerStage
 from sanic.base.meta import SanicMeta
 from sanic.compat import OS_IS_WINDOWS, StartMethod
-from sanic.exceptions import ServerKilled
+from sanic.exceptions import SanicException, ServerKilled
 from sanic.helpers import Default, _default, is_atty
 from sanic.http.constants import HTTP
 from sanic.http.tls import get_ssl_context, process_to_context
@@ -880,12 +880,17 @@ class StartupMixin(metaclass=SanicMeta):
             manager.run()
         except ServerKilled:
             exit_code = 1
+        except SanicException as e:
+            kwargs = {}
+            if e.quiet:
+                error_logger.error(str(e))
+            else:
+                raise
         except BaseException:
             kwargs = primary_server_info.settings
             error_logger.exception(
                 "Experienced exception while trying to serve"
             )
-            raise
         finally:
             logger.info("Server Stopped")
             for app in apps:
--- a/sanic/server/socket.py
+++ b/sanic/server/socket.py
@@ -112,6 +112,16 @@ def configure_socket(
                 server_settings["port"],
                 backlog=backlog,
             )
+        except PermissionError as e:
+            p = server_settings["port"]
+            if p and server_settings["port"] < 1024:
+                addr = f"{server_settings['host']}:{p}"
+                e = ServerError(
+                    f"Permission denied binding to {addr}.\n\n"
+                    "Use `sudo sanic` to run on a privileged port.\n"
+                )
+                e.quiet = True
+            raise e from None
         except OSError as e:  # no cov
             error = ServerError(
                 f"Sanic server could not start: {e}.\n\n"


sanic/server/socket.py
```diff
@@ -112,6 +112,17 @@ def configure_socket(
...
                 backlog=backlog,
             )
+        except PermissionError:
+            p = server_settings["port"]
+            if not p or p >= 1024:
+                raise
+            addr = f"{server_settings['host']}:{p}"
+            error = ServerError(
+                f"Permission denied binding to {addr}.\n\n"
+                "Use `sudo sanic` to run on a privileged port.\n"
```
ahopkins, line 122: Rather than assuming the command was `sanic` we can pull from `sys` to try and give a more friendly reproduction of what they used.

sanic/server/socket.py
```diff
@@ -112,6 +112,17 @@ def configure_socket(
                 server_settings["port"],
                 backlog=backlog,
             )
+        except PermissionError:
+            p = server_settings["port"]
+            if not p or p >= 1024:
+                raise
+            addr = f"{server_settings['host']}:{p}"
```
ahopkins, line 119: Do we care about ipv6 brackets here?

sanic/server/socket.py
```diff
@@ -112,6 +112,17 @@ def configure_socket(
...
+            error = ServerError(
+                f"Permission denied binding to {addr}.\n\n"
+                "Use `sudo sanic` to run on a privileged port.\n"
+            )
+            error.quiet = True
```
ahopkins, line 124: Will this do anything here?

sanic/mixins/startup.py
```diff
@@ -880,12 +880,17 @@ def serve(
...
+            if e.quiet:
+                error_logger.error(str(e))
+            else:
+                raise
         except BaseException:
             kwargs = primary_server_info.settings
             error_logger.exception(
                 "Experienced exception while trying to serve"
             )
-            raise
```
ahopkins, line 888: Why? With this change, won't it exit with 0?

sanic/server/socket.py
Tronic: Seems a bit overkill but I can have a look if there are no corner cases that need handling other than prepending `sudo` to produce the full command (I recall Python does some magic with argv so it isn't really what the system sent).

One possible concern is also with the OSError handler below, which I presume is meant for EADDRINUSE (98), but with diagnostic for the WorkerManager FAQ. We can also see various other errors not really handled here, in particular with unix sockets. I could add a few more handlers for the common problems with better diagnostics, and some tests.

sanic/server/socket.py
Tronic: I should probably add a function in sanic/helpers for that and port numbers as these are getting repeated in many places.

sanic/mixins/startup.py
ahopkins: That is what raising would achieve. But I get your point assuming you have console logging on. exit here is fine as long as it is `exit(1)`.

Tronic: This is what `raise` causes (on sanic main):

```
[2023-09-07 14:41:52 +0100] [481] [ERROR] Experienced exception while trying to serve
  File "/home/user/sanic/sanic/server/socket.py", line 110, in configure_socket
    sock = bind_socket(
  File "/home/user/sanic/sanic/server/socket.py", line 33, in bind_socket
    sock.bind(location)
PermissionError: [Errno 13] Permission denied

  File "/home/user/sanic/sanic/mixins/startup.py", line 1033, in serve
    for sock in [
  File "/home/user/sanic/sanic/mixins/startup.py", line 1034, in <listcomp>
    configure_socket(server_info.settings)
  File "/home/user/sanic/sanic/server/socket.py", line 125, in configure_socket
    raise error
sanic.exceptions.ServerError: Sanic server could not start: [Errno 13] Permission denied.

This may have happened if you are running Sanic in the global scope and not inside of a `if __name__ == "__main__"` block.

See more information: https://sanic.dev/en/guide/deployment/manager.html#how-sanic-server-starts-processes

[2023-09-07 14:41:52 +0100] [481] [INFO] Server Stopped
  File "/home/user/sanic/sanic/server/socket.py", line 110, in configure_socket
    sock = bind_socket(
  File "/home/user/sanic/sanic/server/socket.py", line 33, in bind_socket
    sock.bind(location)
PermissionError: [Errno 13] Permission denied

  File "/usr/local/bin/sanic", line 33, in <module>
    sys.exit(load_entry_point('sanic', 'console_scripts', 'sanic')())
  File "/home/user/sanic/sanic/__main__.py", line 12, in main
    cli.run(args)
  File "/home/user/sanic/sanic/cli/app.py", line 117, in run
    serve(app)
  File "/home/user/sanic/sanic/mixins/startup.py", line 1033, in serve
    for sock in [
  File "/home/user/sanic/sanic/mixins/startup.py", line 1034, in <listcomp>
    configure_socket(server_info.settings)
  File "/home/user/sanic/sanic/server/socket.py", line 125, in configure_socket
    raise error
sanic.exceptions.ServerError: Sanic server could not start: [Errno 13] Permission denied.

This may have happened if you are running Sanic in the global scope and not inside of a `if __name__ == "__main__"` block.

See more information: https://sanic.dev/en/guide/deployment/manager.html#how-sanic-server-starts-processes
```

Notice how the full traceback is printed twice? First by `error_logger.exception`, then there is `raise` and then CPython catches Sanic crashing and prints another traceback (without the logger).

Using `sys.exit(1)` instead of `raise` avoids this.

sanic/server/socket.py
ahopkins: Are you planning to do that here, or should we make it a new issue to handle on its own?

PR #2824, opened 2023-09-20
Do not strip entity-headers with HTTP status 304 or 412

### LINKED ISSUE #2823: The Sanic built-in server blocks for exactly 90 seconds on status code 412

### Describe the bug

When responding to any request in the Sanic development server, it usually responds immediately. However, if the response has status code 412, it takes is very slow and takes exactly 90 seconds to respond, after the handler finishes.

This behavior does not happen when running Sanic with uvicorn. Only the official Sanic server.

### Code snippet

```python
import sanic

app = sanic.Sanic(__name__)

@app.get("/")
async def root(req: sanic.Request):
    status = int(req.args.get("status", "200"))
    return sanic.json({"message": "Hello World"}, status=status)
```

### Expected Behavior

```bash
sanic main:app --port 8051
```

Then:

```sh-session
$ curl http://localhost:8051
{"message":"Hello World"}
$ curl http://localhost:8051/?status=400  # fine
{"message":"Hello World"}
$ curl http://localhost:8051/?status=411  # fine
{"message":"Hello World"}
$ curl http://localhost:8051/?status=413  # fine
{"message":"Hello World"}
$ curl http://localhost:8051/?status=412
# stalls with no response for 90 seconds
```

### How do you run Sanic?

Sanic CLI

### Operating System

Linux

### Sanic Version

Sanic v23.6.0

### Additional context

I have reproduced this on both Linux and macOS. I have also reproduced this using both the Sanic CLI and the `Sanic.serve()` function programmatically.

Tronic:
Fix #2823


sjsadowski (dismissed):
This accurately removes the check for entity headers and adds a test to prevent reversion for HTTP response code 412.

sanic/response/types.py
```diff
@@ -24,7 +24,6 @@
     Default,
     _default,
     has_message_body,
-    remove_entity_headers,
```
ahopkins, line 27: Do we still need the function?

sanic/response/types.py
```diff
@@ -106,9 +105,6 @@ def processed_headers(self) -> Iterator[Tuple[bytes, bytes]]:
         Returns:
             Iterator[Tuple[bytes, bytes]]: A list of header tuples encoded in bytes for sending
         """  # noqa: E501
-        # TODO: Make a blacklist set of header names and then filter with that
-        if self.status in (304, 412):  # Not Modified, Precondition Failed
-            self.headers = remove_entity_headers(self.headers)
```
ahopkins, line 111: I feel like rather than removing this we need to more closely follow the standard and limit when we are removing the headers.

ahopkins: > If the conditional GET used a strong cache validator (see [section](https://datatracker.ietf.org/doc/html/rfc2616#section-13.3.3)
   [13.3.3](https://datatracker.ietf.org/doc/html/rfc2616#section-13.3.3)), the response SHOULD NOT include other entity-headers.
   Otherwise (i.e., the conditional GET used a weak validator), the
   response MUST NOT include other entity-headers; this prevents
   inconsistencies between cached entity-bodies and updated headers.

Tronic: I have no interest to work further on this. Feel free to fork if you do.

Tronic: Failing test removed, passes OK now.

ahopkins (approved):

PR #2830, opened 2023-09-27
Add `loop` argument to signal router `finalize` method to finalize without having a running loop

### LINKED ISSUE #2829: Allow the signal router to finalize without a running event loop

I am running some tasks by awaiting the `Sanic.run_delayed_task` interface. All is well, except for the fact that I cannot dispatch signals.

The error raised when dispatching a signal is `TypeError: 'NoneType' object is not callable` and originates from a non-finalized signal router.

The signal router cannot be finalized without a running event loop.

Add a `loop` keyword argument to the signal router's `finalize` method. 

```python
def finalize(self, do_compile: bool = True, do_optimize: bool = False, loop: asyncio.AbstractEventLoop = None):
    self.add(_blank, "sanic.__signal__.__init__")

    try:
        self.ctx.loop = loop or asyncio.get_running_loop()
    except RuntimeError:
        raise RuntimeError("Cannot finalize signals outside of event loop")

    for signal in self.routes:
        signal.ctx.event = asyncio.Event()

    return super().finalize(do_compile=do_compile, do_optimize=do_optimize)
```

This will allow me to inject a loop instance into the finalize method, circumventing the runtime error caused by `asyncio.get_running_loop()` in the absence of a running loop.

### Additional context

_No response_

eelkevdbos:
Closes #2829

### DIFF
--- a/sanic/signals.py
+++ b/sanic/signals.py
@@ -277,7 +277,7 @@ class SignalRouter(BaseRouter):
 
         return cast(Signal, signal)
 
-    def finalize(self, do_compile: bool = True, do_optimize: bool = False):
+    def finalize(self, do_compile: bool = True, do_optimize: bool = False, loop: asyncio.AbstractEventLoop = None):
         """Finalize the router and compile the routes
 
         Args:
@@ -293,7 +293,7 @@ class SignalRouter(BaseRouter):
         self.add(_blank, "sanic.__signal__.__init__")
 
         try:
-            self.ctx.loop = asyncio.get_running_loop()
+            self.ctx.loop = loop or asyncio.get_running_loop()
         except RuntimeError:
             raise RuntimeError("Cannot finalize signals outside of event loop")
 


ahopkins (dismissed):
:sunglasses: Nice. Thanks for raising the issue and making the fix.

I did not look at the reason for the failing tests, but I would guess it is a formatting thing. 

There should be a command in the Makefile to run black across the repo.

eelkevdbos:
@ahopkins I've updated the formatting and doc blocks, but some seemingly unrelated test is failing. Is this test perhaps also failing on main or do I need to dig deeper into the issue?

Tronic:
I gather you should use asyncio to set the running loop or otherwise execute within its normal bounds, rather than passing loop arguments. It is worth noting that asyncio itself just went though a removal of loop argument passing.

eelkevdbos:
> I gather you should use asyncio to set the loop or otherwise execute within its normal bounds, rather than passing loop arguments. It is worth noting that asyncio itself just went though a removal of loop argument passing.

@Tronic Do you object merging this and suggest injecting the loop differently? I'm happy to go a different route, but provided PR at least works fine in my use case (https://github.com/sanic-org/sanic/issues/2829). The failing test seems an artifact from a failing main branch test (https://github.com/sanic-org/sanic/actions/runs/6109680961/job/16624825220).

@ahopkins Just as a sanity check, I can't see any uses of the `self.ctx.loop` when doing a search for references. Does the finalize method use the loop somewhere?

Tronic:
Can you use `asyncio.set_event_loop(loop)` for that?

PR #2831, opened 2023-10-04
Reducing memory consumption

gluhar2006:
I have found that the more requests Sanic processes, the more memory consumption increases. This behavior, as far as I could understand, is relevant only with a sufficiently large number of requests.
After some research, I took an idea from [here](https://github.com/denis-ryzhkov/mem_top) to make sure that no unnecessary objects remain in memory after processing requests.

<details>
  <summary>Application I used</summary>
 
  ```python
import gc
import tracemalloc

from sanic import Sanic
import psutil
import os

tracemalloc.start()


request_cnt = 0

def mem_top():
    gc.collect()
    objs = gc.get_objects()

    return "\n".join((
        '',
        _top(10, 100, "\n", "{num} {obj}", (
            (len(gc.get_referents(obj)), obj) for obj in objs
        )),
    ))


def _top(limit, width, sep, fmt, nums_and_objs):
    return sep.join(
        fmt.format(num=num, type=type(obj), obj=repr(obj)[:width])
        for num, obj in sorted(nums_and_objs, key=lambda num_obj: -num_obj[0])[:limit]
    )


@app.post("/")
async def hello_world(request):
    global request_cnt
    request_cnt += 1
    print(round(psutil.Process(os.getpid()).memory_info().rss / 2 ** 30, 6))
    if request_cnt == 10000:
        print(str(mem_top()))
    return text("qq")


if __name__ == "__main__":
    app.run(port=5021)

  ```
  
</details>

<details><summary>I used ab to make requests</summary>

  (it doesn't matter what's in the request body)
  ```bash

    ab -p data -T "application/x-www-form-urlencoded" -c 50 -n 10000 "http://0.0.0.0:5021/"

  ```

</details>

So i discovered thousands of `TimerHandle HttpProtocol.check_timeouts` and `abort` from `SanicProtocol` (required much more than 10000 requests).

My changes solves the memory freeing question, but it also has another side effect: here is rps from ab results before and after my changed - 7273 (main branch) vs 8558 (this branch). 
Yes, I know that this is not the most relevant way to measure performance using such requests, but I don't have time to properly test the framework. I can only say that the applications that I develop and use Sanic at least did not slow down after my changes.


sanic/server/protocols/base_protocol.py
```diff
@@ -90,8 +90,9 @@ def close(self, timeout: Optional[float] = None):
         if self.transport:
             self.transport.close()
             if timeout is None:
-                timeout = self.app.config.GRACEFUL_SHUTDOWN_TIMEOUT
-            self.loop.call_later(timeout, self.abort)
+                self.abort()
```
sjsadowski, line 93: This may have inadvertent consequences by immediately closing the connection instead of waiting for the default timeout.

gluhar2006: absolutely agree. however, I was unable to simulate such a situation, but it was easy to simulate a situation when there were a bunch of objects left that the garbage collector did nothing with

gluhar2006: ok, maybe this really isn't a good change

Tronic:
Should also look at clearing cyclic links once done with request/connection, so that Sanic itself would not leak memory when used without GC. Some of them are already cleared but surely some loops remain too.

sanic/server/protocols/base_protocol.py
gluhar2006: any updates? I've removed this diff

gluhar2006:
@Tronic I didn't find any other issues of such type or memory leaks

Tronic:
> @Tronic I didn't find any other issues of such type or memory leaks

But you still had GC enabled? Sanic takes care to clear to avoid cyclic links but I am sure it ain't 100 % yet. This is only relevant once GC is disabled.

gluhar2006:
> But you still had GC enabled? Sanic takes care to clear to avoid cyclic links but I am sure it ain't 100 % yet. This is only relevant once GC is disabled.

Yes, GC is enabled. I don't think that the basic recommendation of sanic usage is to disable GC

gluhar2006:
> > @Tronic
> > > But you still had GC enabled? Sanic takes care to clear to avoid cyclic links but I am sure it ain't 100 % yet. This is only relevant once GC is disabled.
> > 
> > 
> > Yes, GC is enabled. I don't think that the basic recommendation of sanic usage is to disable GC
> 
> Definitely not a basic recommendation. It takes a lot of care to avoid/break cycles in all code, user apps included. But for maximal performance you would want it disabled, can make quite a difference at times.

Memory consumption with disabled gc is low and stable

ahopkins (approved):

PR #2840, opened 2023-10-18
Pathlib update to TLS

Tronic:
Allows using pathlib paths for TLS certificate directories and cert/key filenames. Previously only `str` was accepted on `app.run(ssl=...)`.


ChihweiLHBird (approved):
lgtm!

ahopkins (commented):
Needs some tests.

ChihweiLHBird (approved):

PR #2841, opened 2023-10-20
Add websocket.handler.before/after/exception signals

Tronic:
Adding ability to implement middleware functionality on WebSocket handlers by signals that receive `websocket` parameter.

The flow is:
```
http.handler.before (request) -> websocket.handler.before (request, websocket) -> handler
 ____________________________________________________________________________________/
|
| -> clean exit: websocket.handler.after (request, websocket)
| -> except Exception: websocket.handler.exception (request, websocket, exception)
| -> cancelled/closed: no signal
                             |
                      any of the above -> http.handler.after(request)
```


Tronic:
Diagram created by ChatGPT based on the ASCII sketch in OP.

sanic/app.py
```diff
@@ -1369,6 +1369,12 @@ async def _websocket_handler(
             protocol = request.transport.get_protocol()
             ws = await protocol.websocket_handshake(request, subprotocols)
 
+        await self.dispatch(
+            "websocket.handler.before",
+            inline=True,
+            context={"request": request, "websocket": ws},
+            fail_not_found=False,
```
ahopkins, line 1376: I don't think we need this.

ahopkins (approved):

sanic/app.py
Tronic: This is the only place to access the WS prior to the handler, and as such is very much needed.

PR #2852, opened 2023-11-16
Logging improvements

Tronic:
Reduces useless noise in logs in particular with WebSockets for which a new logger sanic.websockets is added in place of previously used sanic.root. Exchanges various messages with ones that are more useful. Adds WebSocket access log messages, moves some HTTP connection messages to access log as well.


Tronic:
Screenshot 2023-11-16 at 15 33 58

Tronic:
Screenshot 2023-11-16 at 15 51 35

ahopkins:
I did something similar with a simplified logging format in snx
image

A few thoughts:

- Maybe we only add the PID when there are multiple workers
- I like the access log info you have with the alignment. We probably can collapse the URL, and we probably also need some way to handle overflows.
- If we can keep access logs and regular logs all somewhat aligned and uniform (like in my example) that would be ideal
- We probably should have default logging formats for DEV and PROD

PR #2853, opened 2023-11-17
Make error page renderers set correct status instead of 200

Tronic:


prryplatypus:
The `BaseRenderer` is doing this in line 113 already by the looks of it

Tronic:
> The `BaseRenderer` is doing this in line 113 already by the looks of it

A good catch. render() sets status and headers on the response object. I was not hitting this due to calling full() directly, and that was indeed quite confusing.

Tronic:
This PR is not needed.

PR #2859, opened 2023-11-26
Add a REPL to the CLI

ahopkins:
This comes from a request in the forums: https://community.sanicframework.org/t/sanic-implement-shell-like-flask-shell/1234

The actual implementation adds a `--repl` flag to the CLI. By default, `--dev` will also enable it.

image

The way it works is by opening a Python REPL in the main process. It will also enable a client for easy access to calling the app. But, one thing that it might confuse people is that altering the `app` instance in the main will obviously not make it available in the HTTP client. So there probably needs to be some kind of warnings. Also there should be some safety re: TTY, etc.


ahopkins:

https://community.sanicframework.org/t/sanic-implement-shell-like-flask-shell/1234/5

prryplatypus (approved):

PR #2843, opened 2023-10-24
Adopt Ruff linter/formatter

Tronic:
Gotta go fast, so trying the fastest Python formatter as well (replaces isort, black, flake8 and bandit).

PR #2865, opened 2023-12-04
Add ruff to toolchain

ahopkins:
Replaces #2844 and #2843 

It is a more mild approach than #2844 in that it does not implement any large changes to the code base right now, and keeps `bandit`. I think both approaches in #2844 can be implemented over time, but I am in favor of a more gentle approach starting now, and moving towards that after the LTS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,9 +14,6 @@ help:
 	@echo "pretty"
 	@echo "		Analyze and fix linting and format using ruff"
 	@echo ""
-	@echo "beautify [sort_imports=1] [include_tests=1]"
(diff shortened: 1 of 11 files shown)


iAndriy:
TIL about ruff, looks like great tool.
Any sense in applying ruff for example and scripts folders ?

ahopkins:
> Any sense in applying ruff for example and scripts folders ?

Absolutely, and I think we should and will get there. But, I wanted to make this a slower introduction first and avoid any sort of major diff heading into the LTS.

iAndriy:
@ahopkins Followed up with the following ticket to get around those two folders: https://github.com/sanic-org/sanic/issues/2867

Issue #2871, opened 2023-12-08
Introduce ruff execution on the tests folder


iAndriy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

The tests folder isn't formatted with ruff

### Describe the solution you'd like

The ruff formatter would get applied to tests folder.

PR #2872, opened 2023-12-08
Introduce ruff execution on the tests folder #2871

iAndriy:


Makefile
```diff
@@ -1,3 +1,4 @@
+RUFF_FORMATTED_FOLDERS = sanic examples scripts tests
```
ahopkins, line 1: Want to add `./guide` while you're at it?

iAndriy: @ahopkins thanks for suggestion, this makes sense, though in a spirit of smaller iterations, I would rather prefer to address it separate MR, which I can definitely do.

.github/workflows/tests.yml
```diff
@@ -34,15 +34,15 @@ jobs:
           - { python-version: "3.9",  tox-env: type-checking }
           - { python-version: "3.10", tox-env: type-checking }
           - { python-version: "3.11", tox-env: type-checking }
-          - { python-version: "3.8",  tox-env: py38,          max-attempts: 3 }
-          - { python-version: "3.8",  tox-env: py38-no-ext,   max-attempts: 3 }
+          - { python-version: "3.8",  tox-env: py38,          max-attempts: 1, timeout-minutes: 20}
```
iAndriy, line 37: Note: This and other adjustments to tests.yml are going to be deleted once we have passing tests.

tests/typing/test_typing.py
iAndriy, line 1: @ahopkins This specific test started to fail, but all changes are backward compatible, I do see that parametrised test has hardcoded numbers of lines.
Do you have any take on how this test shall be addressed and how we can proceed with this ?

Makefile
ahopkins: 👍 sounds like a plan

ahopkins (approved):

iAndriy:
@ahopkins all comments resolved and thanks for your review.

PR #2880, opened 2023-12-17
Use importlib and packaging for websockets version validation

### LINKED ISSUE #2873: (websockets 12.0) DeprecationWarning: websockets.connection was renamed to websockets.protocol and Connection was renamed to Protocol

### Describe the bug

The DeprecationWarning is thrown here:

https://github.com/sanic-org/sanic/blob/82bf46bea3b55e6b1371107cccda280add8e70eb/sanic/server/protocols/websocket_protocol.py#L4-L9

With websockets 12 the try block would run successfully with the warning while the catch block does not have chance to be run.

### Code snippet

_No response_

### Expected Behavior

The catch block is being run instead.

### How do you run Sanic?

Sanic CLI

### Operating System

MacOS

### Sanic Version

Sanic 23.6.0; Routing 23.6.0

### Additional context

_No response_

ChihweiLHBird:
Fixes #2873

One concern I have is that by this way, we will have one more dependency ([`packaging`](https://packaging.pypa.io/en/stable/)) for parsing the package version.

Alternatively, we can do it (as code snippet below) without [`packaging`](https://packaging.pypa.io/en/stable/) but might be less reliable against some invalid version numbers, which I think should be fine too.

```python
from importlib.metadata import version

int(version("websockets").split(".")[0])
```

Or maybe we can just keep the current way since it won't have any issue when the deprecated WebSocket classes are removed (at the time they will be caught by the `except` block), lol. This is actually my preferred solution, and I can close this PR if we decide to go this way. But I would like to hear other opinions.


ahopkins:
I definitely do not want to add a dependency for this. I'd prefer to keep it as try/except where we can.

PR #2882, opened 2023-12-20
Introduce dynamic port fixture for test_app and helper function for tests which require port within a loop.

iAndriy:


ahopkins (approved):
Nice. TY

ahopkins (approved):

PR #2884, opened 2023-12-25
DOCS: sync/async guide for sanic

sjsadowski:
Much needed guide for explaining what happens when blocking calls are used in Sanic

### DIFF
--- a/guide/content/en/guide/basics/sync_async.md
+++ b/guide/content/en/guide/basics/sync_async.md
@@ -1,0 +1,125 @@
+# Sync and Async in Sanic
+
+Sanic is an **asynchronous** framework - it is designed to be responsive to requests, which allows it to theoretically serve more requests, more quickly, than a synchronous framework. It shares this functionality with other asynchronous frameworks like [starlette](https://www.starlette.io/) and a more full-featured, batteries included framework that is in the public spotlight, [FastAPI](https://fastapi.tiangolo.com/) (which uses starlette under the hood).
+
+There are preventable situations where this is not the case. Using synchronous functions in asynchronous code is the biggest hurdle for many Sanic adopters.
+
+## Introduction
+Understanding asynchronicity is sometimes baffling even to experienced developers. The point of this mini-guide is to help understand what happens in Sanic, why certain things might not work the way you expect, and will help walk you through some of the impact of synchronous and asynchronous code may have on your Sanic application.
+
+There are many guides to async/await available as of this writing (Dec 2023) so we will not cover everything specifically, instead exploring and reviewing what happens inside Sanic when you use non-blocking (asynchronous) and blocking (synchronous) code.
+
+## Synchronous Execution
+Let’s start with as simple an explanation as we can produce and an example. Most python code is synchronous, that is to say when something happens, it must be done in order and must complete before the python interpreter can move on. This is called **blocking** because the process is **blocked** while it waits for each statement to finish.
+
+Our first example, below, is a standard program. The ***execution flow*** can be understood by simply reading the code.
+
+```py test1.py
+#!/usr/bin/env python3.11
+import time
+from random import randrange
+
+def sync_print(text: str) -> None:
+    rand_wait: int = randrange(0,5)
+    time.sleep(rand_wait)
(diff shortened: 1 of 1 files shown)


guide/content/en/guide/basics/sync_async.md
```diff
@@ -0,0 +1,175 @@
...
+Ex. 1
+```py test1.py
+#!/usr/bin/env python3.11
+import time
+from random import randrange
+
+def sync_print(text: str) -> None:
+    rand_wait: int = randrange(0,5)
+    time.sleep(rand_wait)
+    print(f'{text} ({rand_wait}s)')
```
Tronic, line 30: I suggest to take the delay as argument, rather than random which only adds to complexity here.

sjsadowski: I'll be using it for consistency when I get to demonstrating the async/await versions. It adds complexity, but it's also a relatively simple path to show the impact of asynchronous code. I'm putting a pin in it, but I'd like your opinion when I've got the section complete.

Tronic: Should work with fixed delays similarly for both async and sync. I hope you have read https://trio.readthedocs.io/en/stable/tutorial.html which is a particularly good introduction to async issues (even when not using Trio).

guide/content/en/guide/basics/sync_async.md
```diff
@@ -0,0 +1,175 @@
+---
+title: Sync and Async in Sanic
+---
+
+# Sync and Async in Sanic
+
+Sanic is an **asynchronous** framework - it is designed to be responsive to requests, which allows it to theoretically serve more requests, more quickly, than a synchronous framework. Other python asynchronous frameworks also share the same limitations when calling synchronous code, so the guide below will hopefully serve you if you end up in a situation using something other than Sanic in the future.
```
ahopkins, line 7:
```suggestion
Sanic is an **asynchronous** framework—it is designed to be responsive to requests, which allows it to theoretically serve more requests, more quickly, than a synchronous framework. Other python asynchronous frameworks also share the same limitations when calling synchronous code, so the guide below will hopefully serve you if you end up in a situation using something other than Sanic in the future.
```

ahopkins: Sorry, the typographer in me cannot overlook the mistake of a hyphen for a dash.

guide/content/en/guide/basics/sync_async.md
sjsadowski: @Tronic after sleeping on it, I'll revise to use the delay as an argument as you suggest. I don't think the complexity is inherently a problem with using random sleep times, but given how many people seem confused about async/await removing that from the process makes sense.

ahopkins:
Can we get this finalized soon?

guide/content/en/guide/basics/sync_async.md
```diff
@@ -0,0 +1,175 @@
...
+
+However this guide is not going to dig into the material impact of scheduling, but while analogies in describing async/await are always suspect, consider this example scenario:
+
+You wish to brew a cup of tea.
+
+Using an electric kettle, you fill it with water, and start the kettle.
+
+Most human beings are not that interested in standing at the kettle, doing nothing until it boils, but with regular functions, that's what happens: Fill kettle. Boil. Use water to make drink. The operations happen in order, and you cannot do anything else until they complete.
+
+For someone making a drink, that seems like a waste of time. A person does not need to stand around and watch the kettle until it boils, instead, other things can be done. So the kettle is left to do its task, which by filling it and turning it on, we can consider to be scheduled, and when it is done, we can come back to it and make our hot drink.
```
slaupster, line 144: I know this is trying to talk to a wide audience but I feel its too dumbed down to be of use to someone that actually gets to the point of reading it. i think that most people that would be reading this would definitely have an understanding of `asyncio` (though the motivation here could be something I've just not witnessed!)

I think you could deal with the concept in far fewer lines and do away with a noddy example, and defer more detail to https://docs.python.org/3/library/asyncio.html and go on to deal with `sanic` specific detail instead ?

PR #2887, opened 2024-01-01
Fix Docker publish

prryplatypus:


ahopkins (dismissed):

.github/workflows/publish-release.yml
```diff
@@ -3,7 +3,6 @@ name: Publish release
 on:
   release:
     types: [created]
-  workflow_dispatch:
```
ahopkins, line 6: You don't think we want this on demand? How do you want to release the LTS images?

prryplatypus: > You don't think we want this on demand?
The workflow currently will error when triggered manually. LTS patches can be created by creating the corresponding release and NOT marking it as `latest` in the release page.

> How do you want to release the LTS images?
For this version I'll upload them manually.

prryplatypus:
Don't merge yet, there's another issue

prryplatypus:
Images for v23.12.0 have been uploaded to Docker Hub. And fixes another couple of issues along the way 🙂

ahopkins (approved):

PR #2892, opened 2024-01-07
2882 Apply dynamic port fixture  for tests of cli, config and server …

iAndriy:


Tronic:
There is a race condition in `get_port` because the socket opened is closed when returning the function so another process running at the same time could accidentally get the same port. If it instead returned the socket, that was then used with server, there would be no such issue, but that cannot work with CLI tests where no socket can be passed.

Alternatively, simply picking a random int would possibly be faster and still only very rarely have collisions. Presumably the kernel essentially picks a random number each time we bind to port 0, so this might be close to current behavior.

Issue #2912, opened 2024-02-01
Unexpected behavior with bp.middleware() using classes


nicolaipre:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

I was playing around with middleware registration directly on a blueprint, since I want to register it directly without decorators. I therefore used bp.middleware(middleware) and noticed something quite odd in the process where middleware does not seem to get added if you are using a callable class with `__call__()`

### Code snippet

```python
#!/usr/bin/env python3

#sanic==23.12.1
#sanic-ext==23.12.0
#sanic-routing==23.12.0

from sanic import Sanic
from sanic.response import text
from sanic.request import Request
from sanic.response import json
from sanic import Blueprint
from inspect import isfunction

bp = Blueprint("my_blueprint")

@bp.route("/")
async def bp_root(request: Request):
    print(request.method)
    print(request.body)
    print(request.head)
    return json({"my": "blueprint"})


# Middlewares
def middleware_as_def(request: Request):
    print("middleware_as_def")


class MiddlewareAsClass:
    something = "lol"
    def __call__(self, request: Request):
        print("MiddlewareAsClass called")
        return text("failed")


# Check if callable
print(f"middleware_as_def: IsCallable: { callable(middleware_as_def) }")
print(f"MiddlewareAsClass: IsCallable: { callable(MiddlewareAsClass) }")

# middleware function
bp.middleware(middleware_as_def)

# middleware class (callable __call__() )
mw = MiddlewareAsClass()
bp.middleware(mw) # works, but only if another middleware function is registered first
print("Is function:", isfunction(mw))

# in blueprints.py (line 62 & 63), as_decorator never turns False
# in blueprints.py (line 47): set as_decorator=False as default, it will then work with classes.

# Main
app = Sanic("MiddlewareTestApp")
app.blueprint(bp)

if __name__ == "__main__":
    app.run(host="127.0.0.1", port=8000, dev=True)
```

### Expected Behavior

I noticed that middleware did not work if I used a callable class, unless I registered a middleware that was a function first, i.e:
```python
bp.middleware(middleware_as_def)    # must be added here for the one below to work
bp.middleware(MiddlewareAsClass())  # if this is the only Middleware, nothing is registered
```

[...]

PR #2923, opened 2024-02-29
Allows classes as middleware

nicolaipre:
Tiny fix that seemingly solves #2912

### DIFF
--- a/sanic/blueprints.py
+++ b/sanic/blueprints.py
@@ -7,7 +7,6 @@ from collections import defaultdict
 from collections.abc import MutableSequence
 from copy import deepcopy
 from functools import partial, wraps
-from inspect import isfunction
 from itertools import chain
 from types import SimpleNamespace
 from typing import (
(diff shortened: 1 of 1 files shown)


ahopkins (changes requested):
:sunglasses:  Thanks
Can you add a unit test?

nicolaipre:
> 😎 Thanks Can you add a unit test?

Will do. Currently working on it but experienced some weird errors that I am not quite sure how to solve.

Some of the MiddlewareMixin Overloads seem to cause some confusion. My tests work when running locally, but fails with pytest for some reason. 

I could push the test as it is right now if you would be willing to look at it to see if you are able to spot any obvious mistakes?

nicolaipre:
I have decided to close this PR as the change would introduce confusion since decorators are not intended to be used on Classes anyways.

The main goal was to allow users to register routes and middleware without using decorators as an alternative way of using Sanic, but this fix crashes with the way Sanic is intended to be used and is therefore no point adding in.

PR #2926, opened 2024-03-14
fix DeprecationWarning: Attribute s is deprecated and will be removed…

hamedsh:
… in Python 3.14; use value instead

The attribute check within the 'ode.py' script has been updated. Previously, it checked whether the event has an attribute 's', however, this has been modified to check for the 'value' attribute instead. This ensures the correct attribute is being checked for the ongoing processes.


ahopkins (changes requested):
This should be conditional check so that we are backwards compat on all supported versions.

ahopkins (approved):

ahopkins:
Thanks for the contribution. I went a slightly different route: #2959

PR #2928, opened 2024-03-22
Auto generate route name

hamedsh:
The ability to automatically generate names for routes has been added to Sanic. This feature is controlled by the 'generate_name' parameter available in the application, blueprint, and route definitions. Additionally, comprehensive tests have been added to ensure the correct functionality of this new feature.

a parameter added to the Sanic and Blueprint constructor
`app = Sanic(name='hello', unique_route_name_generate=True)`

in recent versions of Sanic, it's some kind of essential to assign names fo routes (it makes migrating to a new version of Sanic, very time-consuming). For example this little example: 
```python
app = Sanic(name='hello')

@app.route('/')
async def handler(request):
    await asyncio.sleep(2)
    print('Still running')
    return response.text('OK')


@app.route('/err')
async def handler(request):
    await asyncio.sleep(0.1)
    return response.text('OK')
```
by using this new argument, its becomes mandatory to assign names and be more backward compatible also naming can be ignored in most cases.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -223,7 +223,7 @@ class Sanic(
         "strict_slashes",
         "websocket_enabled",
         "websocket_tasks",
-        "generate_name",
+        "unique_route_name_generate",
     )
 
     _app_registry: ClassVar[Dict[str, "Sanic"]] = {}
@@ -248,7 +248,7 @@ class Sanic(
         inspector: bool = False,
         inspector_class: Optional[Type[Inspector]] = None,
         certloader_class: Optional[Type[CertLoader]] = None,
-        generate_name: bool = False,
+        unique_route_name_generate: bool = False,
     ) -> None:
         ...
 
@@ -271,7 +271,7 @@ class Sanic(
         inspector: bool = False,
         inspector_class: Optional[Type[Inspector]] = None,
         certloader_class: Optional[Type[CertLoader]] = None,
-        generate_name: bool = False,
+        unique_route_name_generate: bool = False,
     ) -> None:
         ...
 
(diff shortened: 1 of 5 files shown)


tests/test_blueprints.py
```diff
@@ -308,6 +308,34 @@ def handler2(request):
...
+
+    @bp.route("/route_with_name", methods=["GET", "POST"], name="handler3")
+    def handler3(request):
+        return text("Hello subdomain!")
+
+    app.blueprint(bp)
+
+    route_names = [r.name for r in app.router.routes]
+    assert "test_bp_with_auto_name_generate.test_bp_host.handler1_GET_/" in route_names
+    assert "test_bp_with_auto_name_generate.test_bp_host.handler2_GET_/" in route_names
```
ahopkins, line 335: We do not want slashes in route names.

sanic/app.py
```diff
@@ -247,6 +248,7 @@ def __init__(
         inspector: bool = False,
         inspector_class: Optional[Type[Inspector]] = None,
         certloader_class: Optional[Type[CertLoader]] = None,
+        generate_name: bool = False,
```
ahopkins, line 251: I guess I am a little confused by this because we _do_ generate route names.

ahopkins (commented):
Can you explain a little the use case for this? I am trying to understand how it will be useful other than the two existing options:

1. Use the handler function's name, or
2. Explicit naming with `name` param.

I guess I can see possible a use case like:

```python
class BaseMixin:
    def generate_name(self) -> str:
        return self._generate_name()
```

This would then expose it as an easily overridable function.

tests/test_blueprints.py
hamedsh: Its replaced with 'S'
also, can you please merge this one? on Python 3.12, tests will fail in other case: https://github.com/sanic-org/sanic-testing/pull/74

ahopkins:
Thanks for the work on this. I decided to rework the pattern in #2961.

PR #2931, opened 2024-03-31
Logging improvements

ahopkins:
This replaces the nice work that @Tronic did on #2852 to include debug and prod style logging, as well as a legacy formatter to allow for backwards compat logging for some period of time.

## Debug logging

image

## Prod logging (w/ access logs explicitly on)

image

---

As you can see the main difference is just in the usage of the timestamp. We skip it in Debug (except for access logs).


prryplatypus:
What's the reason for removing the timestamp? I personally find it useful while developing - helps spotting slow bits of code and correlating events between systems/apps (for example if you have a webserver and a background worker).

ahopkins:
@prryplatypus Muddies up the logs more than necessary if not needed. If it is needed, it is not so difficult to get the timestamps:

```python
from sanic.log import LOGGING_CONFIG_DEFAULTS

LOGGING_CONFIG_DEFAULTS['formatters']['generic']['class'] = 'sanic.logging.formatter.ProdFormatter'
```

Perhaps we could make it all even simpler with config values on the Sanic app. Something like:

```python
app.config.LOG_FORMATTER = "sanic.logging.formatter.ProdFormatter"
```

ahopkins:
image

FWIW, even with the time in dev (no date, just time) we still pick up a lot of real estate and cleanliness.

prryplatypus:
I'd be happy with just time 👀

PR #2947, opened 2024-05-06
Make the `.message` field on exceptions non-empty

ekzhang:
This allows subclasses of SanicException to access their message via the `message` attribute. This makes it consistent with the `status_code`, `quiet`, `headers` attributes that were previously present on the class.

Currently, the message attribute is present on the class but not on the instance, so accessing SanicException("Failed!").message will return an empty string "" instead of the actual message "Failed!", which you can get by calling the `__str__()` method or `str()` function.

This is a bit surprising, since the .message attribute shows up in autocomplete and type-checking. It also happens for the other exceptions, like BadRequest().message == "" as well.

I set the message attribute on instances of SanicException and added tests for this new behavior.

### Doc reference

https://sanic.dev/en/guide/best-practices/exceptions.html#exception-properties

### DIFF
--- a/sanic/exceptions.py
+++ b/sanic/exceptions.py
@@ -87,7 +87,7 @@ class SanicException(Exception):
         self.status_code = status_code or self.status_code
         self.quiet = quiet
         self.headers = headers
-        self.message = message
+        self.message = message  # type: ignore
 
 
 class HTTPException(SanicException):
--- a/sanic/exceptions.py
+++ b/sanic/exceptions.py
@@ -65,7 +65,7 @@ class SanicException(Exception):
         quiet: Optional[bool] = None,
         context: Optional[Dict[str, Any]] = None,
         extra: Optional[Dict[str, Any]] = None,
-        headers: Optional[Dict[str, Any]] = None,
+        headers: Optional[Dict[str, str]] = None,
     ) -> None:
         self.context = context
         self.extra = extra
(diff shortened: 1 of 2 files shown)


sanic/exceptions.py
```diff
@@ -86,6 +87,7 @@ def __init__(
         self.status_code = status_code or self.status_code
         self.quiet = quiet
         self.headers = headers
+        self.message = message  # type: ignore
```
ekzhang, line 90: There is a type error here — it looks like `message` the argument of `__init__` can take a `bytes` object, but the class attribute can only be a `str`.

Tronic: I suspect the option for it to be `bytes` might be because of `STATUS_CODES` messages being in bytes (assuming someone passed one of those to init), and this may be necessary to keep for compatibility.

Can you get away with typing without the ignore if you add some logic to decode the argument before storing to `self.message`, if `bytes` was received?

Tronic (changes requested):

sanic/exceptions.py
ekzhang: Ah, that makes sense! I'll update my PR to do that :)

ekzhang:
Hi @Tronic, I updated the constructor to decode bytes arguments to str. It is a bit more verbose but works without disabling the type-checker on that line now.

I also updated the test!

---

The current CI failures appear to be from an unrelated error to this changeset.

```
sanic/request/types.py:89: error: "ctx_type" cannot appear after "sanic_type" in type parameter list because it has no default type  [misc]
```

Tronic (dismissed):
LGTM, if not barring minor cleanup (e.g. removal of "utf8" argument on decode, perhaps some restructuring). I will assume @ahopkins to do another review before merging; from my point this is approved as is or with changes. Thanks for the contribution!

ekzhang:
> if not barring minor cleanup (e.g. removal of "utf8" argument on decode, perhaps some restructuring).

Thanks for reviewing so quickly! I followed the pattern of the existing code to use `.decode("utf8")` but I'm also happy to amend that call or restructure anything.

ahopkins:
@ekzhang I made some changes to the PR to keep the logic closer to the original intent to handle some of the border cases that your implementation may have broken. Thanks for the PR, and I certainly appreciate the additional tests you added.

ahopkins (approved):

ekzhang:
That's amazing, thank you!!

PR #2949, opened 2024-06-01
remove depricated attributes

dyrkuwu:
DeprecationWarning: Attribute s is deprecated and will be removed in Python 3.14; use value instead


prryplatypus:
I believe this is a dupe of #2926

klausmyrseth:
> I believe this is a dupe of #2926

Yes it is, but the code in this pull request is better to be honest so this one should take priority

ahopkins:
Thanks for the contribution. I went a slightly different route: #2959

PR #2951, opened 2024-06-04
Not to remove entity headers for 412 responses

### LINKED ISSUE #2575: Raising a 412 on a Sanic handler causes a wait period and maybe a ReadTimeout exception on clients
**Describe the bug**

When raising a 412, clients are left waiting to read until a timeout is triggered and a ReadTimeout exception may occur.

**Code snippet**

The issue can be reproduced with code for an app as minimal as:
```python
# Server Side

from sanic import Sanic
from sanic import exceptions

@app.get('/hello')
async def hello_world(request, *args, **kwargs):
  raise exceptions.SanicException('Failed Precondition', status_code=412)


if __name__ == '__main__':
  app.run()
```

The client can be Postman, bash curl, or even a Python fetch:

```python
# Client Side

import asyncio
import time

import httpx


async def async_fetch():
  async with httpx.AsyncClient() as client:
    return await client.get('http://localhost:8000/hello')


if __name__ == '__main__':
  now = time.time()
  try:
    resp = asyncio.get_event_loop().run_until_complete(async_fetch())
  except Exception as e:
    print('Exception:', type(e))
  else:
    print('Response:', resp)
  finally:
    print('Time:', time.time() - now)
``` 

The above client code will print:
```
Exception: <class 'httpx.ReadTimeout'>
Time: 5.057372808456421
```

**Expected behavior**

The client should quickly receive the 412 response.

 - OS: macOS Monterey with M1 chip
 - Sanic Version:  22.9.0

**Additional context**

What I think is happening is that [Sanic strips the headers on 412 responses](https://github.com/sanic-org/sanic/blob/main/sanic/response.py#L127-L128), including the `content-type` header. However, and contrary to what happens for a 304 code, Sanic will still send the 412 response body to the client, which causes the latter to wait for more data, without knowing the actual size of the response, until it times out.

ChihweiLHBird:
Resolves #2575

There seems to be not a reason to remove the entity headers for 412 responses, and the removal is causing the client-side timeout error.

### DIFF
--- a/sanic/response/types.py
+++ b/sanic/response/types.py
@@ -104,8 +104,7 @@ class BaseHTTPResponse:
         Returns:
             Iterator[Tuple[bytes, bytes]]: A list of header tuples encoded in bytes for sending
         """  # noqa: E501
-        # TODO: Make a blacklist set of header names and then filter with that
-        if self.status in (304, 412):  # Not Modified, Precondition Failed
+        if self.status == 304:  # Not Modified
             self.headers = remove_entity_headers(self.headers)
         if has_message_body(self.status):
             self.headers.setdefault("content-type", self.content_type)
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -18,7 +18,7 @@ import pytest
 from aiofiles import os as async_os
 from pytest import LogCaptureFixture
 
-from sanic import Request, Sanic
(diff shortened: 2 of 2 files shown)


ahopkins:
We made this change in the LTS already, so I pulled that commit into `main` in #2962.

Issue #2890, opened 2024-01-05
Error: aborted in sanic 23.12


vincentdtt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When i call get method to get list data from server, with paging data (< 1000 object) sometime connection was lost and show **Error: aborted** but if i set paging > 1000 all request get **Error: aborted**. This error is gone when i downgrade to old version 23.6. i'm using haproxy for loading balance.

![image](https://github.com/sanic-org/sanic/assets/38010234/cdf3baee-b923-4372-a743-2ca8bd0000a8)

### How do you run Sanic?

Sanic CLI

### Operating System

Linux

### Sanic Version

23.12

PushUpek:
I have the same problem with Nginx reverse proxy. After day of checking commits from Sanic 23.12 changelog I found that this change #2831 is responsible for this problem. When I removed this method:
https://github.com/sanic-org/sanic/blob/acb29c9dc4d6ba3a453a18e30f0664ba6772a9b4/sanic/server/protocols/http_protocol.py#L231C5-L242C29
problem not occures any more.

ahopkins:
Thanks for digging into this. Looking into it now .

Borsalinoo:
> I have the same problem with Nginx reverse proxy. After day of checking commits from Sanic 23.12 changelog I found that this change #2831 is responsible for this problem. When I removed this method: https://github.com/sanic-org/sanic/blob/main/sanic/server/protocols/http_protocol.py#L231-L242 problem not occures any more.

Great job, buddy! This issue has been troubling me for several days, very grateful for your answer!

ahopkins:
Anyone have some good minimal reproducible cases to add to test suite?

PR #2966, opened 2024-06-23
Call abort later

### LINKED ISSUE #2921: Sanic drops part of HTTP response data

### Describe the bug

The bug is that Sanic closes the connection before the final transfer of all data (see example below)

From my point of view this is a very critical bug. For a reason unknown to me, it repeats only when starting the Sanic server inside the docker container (i.e., perhaps it is somehow related to the networkmode or delays during the work of the docker).

After analyzing the commits, we managed to understand that the bug was introduced in https://github.com/sanic-org/sanic/commit/13106842a4303d119551f89deda7325975cc67f5 and not repeated on https://github.com/sanic-org/sanic/commit/d1fc867a4efc027860da2fc31412389ac651ee6d

### Code snippet

So we have really simple server returning some json demo.py:
```python
from sanic import Sanic

app = Sanic("Demo")

@app.route("/")
async def handle(request):
    return json({'some_val': 'A'*743000})

if __name__ == "__main__":
    app.run(host="0.0.0.0")
```

Running in docker container:
```Dockerfile
FROM python:3.11

COPY demo.py demo.py
RUN pip install sanic==23.12.1
ENTRYPOINT python3 demo.py
```

```bash
$ docker build -t demo .
$ docker run -p 127.0.0.1:8000:8000 --rm -it demo
```

and the client.py:
```python
import socket
import sys


def getsock(ip, port, req):
    s = socket.socket(2, 1, 6)
    s.connect((ip, port))
    s.send(req)
    return s


REQ = (
    b'GET / HTTP/1.1\r\n'
    b'User-Agent: Mozilla/5.0\r\n'
    b'Connection: close\r\n'
    b'\r\n'
)

s = getsock('127.0.0.1', 8000, REQ)

headers = s.recv(94).decode()

print(f'Headers: {headers!r}')

total = 0
by_len = int(sys.argv[1])
while (data := s.recv(by_len)):
    total += len(data)

print(f'Total length: {total}')

s.close()
```
I was not able to reproduce by curl, may be it's read too fast... But in real case it's repeats with Nginx proxy and Sanic as upstream.

so now if you will run hundred times you will get something like this:
```
$ for i in {1..100}; do python3 client.py 4096; done
...
Headers: 'HTTP/1.1 200 OK\r\ncontent-length: 743015\r\nconnection: close\r\ncontent-type: application/json\r\n\r\n'
Total length: 586346
Headers: 'HTTP/1.1 200 OK\r\ncontent-length: 743015\r\nconnection: close\r\ncontent-type: application/json\r\n\r\n'
Total length: 586346
Headers: 'HTTP/1.1 200 OK\r\ncontent-length: 743015\r\nconnection: close\r\ncontent-type: application/json\r\n\r\n'
Total length: 652954
Headers: 'HTTP/1.1 200 OK\r\ncontent-length: 743015\r\nconnection: close\r\ncontent-type: application/json\r\n\r\n'
Total length: 586346
Headers: 'HTTP/1.1 200 OK\r\ncontent-length: 743015\r\nconnection: close\r\ncontent-type: application/json\r\n\r\n'
Total length: 650058
Headers: 'HTTP/1.1 200 OK\r\ncontent-length: 743015\r\nconnection: close\r\ncontent-type: application/json\r\n\r\n'
Total length: 586346
Headers: 'HTTP/1.1 200 OK\r\ncontent-length: 743015\r\nconnection: close\r\ncontent-type: application/json\r\n\r\n'
Total length: 652954
Headers: 'HTTP/1.1 200 OK\r\ncontent-length: 743015\r\nconnection: close\r\ncontent-type: application/json\r\n\r\n'
Total length: 586346
...
```
So length should be 743015 but sanic returns only 586346-652954.

client.py must be runner outside the docker container, e.g on host. If you will run inside the docker it will not reproduce.

### Expected Behavior

Return all the data from response.

### How do you run Sanic?

Sanic CLI

### Operating System

Linux

### Sanic Version

v23.12.1

### Additional context

_No response_

ahopkins:
Fixes #2921 #2890

### DIFF
--- a/sanic/server/protocols/http_protocol.py
+++ b/sanic/server/protocols/http_protocol.py
@@ -261,7 +261,7 @@ class HttpProtocol(HttpProtocolMixin, SanicProtocol, metaclass=TouchUpMeta):
             self._callback_check_timeouts.cancel()
             if self.transport:
                 self.transport.close()
-                self.abort()
+                self.loop.call_later(0.1, self.abort)
 
     async def send(self, data):  # no cov
         """


ahopkins:
> Approving because it fixes the issue, but not too convinced because -as you mentioned in the issue- it feels more like a workaround than like a fix.

100% was a workaround and not a fix.

PR #2973, opened 2024-06-26
Better fix for scheduling logic for transport close() and abort()

### LINKED ISSUE #2921: Sanic drops part of HTTP response data

ashleysommer:
Fixes #2921 #2890
Replaces #2966 

Rather than simply calling `abort()` at a set interval after calling `close()`, we implement a mechanism to detect if any data is still in the transport write buffer after close is called, and schedules an async close operation to wait until the transport has finished, or finally calls abort after a reasonable timeout.

For more information, see my recent in-depth comments on https://github.com/sanic-org/sanic/issues/2921

This does not modify any existing Protocol transport logic (apart from the ill-placed `abort()`), does not introduce any breaking changes, and all tests pass without modification.

### DIFF
--- a/sanic/server/protocols/base_protocol.py
+++ b/sanic/server/protocols/base_protocol.py
@@ -114,8 +114,8 @@ class SanicProtocol(asyncio.Protocol):
                 # This close timeout needs to be less than the graceful
                 # shutdown timeout. The graceful shutdown _could_ be waiting
                 # for this transport to close before shutting down the app.
-                timeout = self.app.config.GRACEFUL_SHUTDOWN_TIMEOUT / 2.0
-                # This is 15seconds/2 = 7.5seconds by default.
+                timeout = self.app.config.GRACEFUL_TCP_CLOSE_TIMEOUT
+                # This is 5s by default.
         else:
             # Schedule the async close checker but with no timeout,
             # this will ensure abort() is called if required.
--- a/sanic/server/protocols/base_protocol.py
+++ b/sanic/server/protocols/base_protocol.py
@@ -86,12 +86,47 @@ class SanicProtocol(asyncio.Protocol):
         """
         Attempt close the connection.
         """
-        # Cause a call to connection_lost where further cleanup occurs
-        if self.transport:
-            self.transport.close()
+        if self.transport is None or self.transport.is_closing():
+            # do not attempt to close again, already aborted or closing
+            return
+
+        # Check if write is already paused _before_ close() is called.
+        write_was_paused = not self._can_write.is_set()
+        # Trigger the UVLoop Stream Transport Close routine
+        # Causes a call to connection_lost where further cleanup occurs
+        # Close may fully close the connection now, but if there is still
+        # data in the libuv buffer, then close becomes an async operation
+        self.transport.close()
+        try:
+            # Check write-buffer data left _after_ close is called.
+            # in UVLoop, get the data in the libuv transport write-buffer
+            data_left = self.transport.get_write_buffer_size()
+        # Some asyncio implementations don't support get_write_buffer_size
+        except (AttributeError, NotImplementedError):
+            data_left = 0
+        if write_was_paused or data_left > 0:
+            # don't call resume_writing here, it gets called by the transport
+            # to unpause the protocol when it is ready for more data
+
+            # Schedule the async close checker, to close the connection
+            # after the transport is done, and clean everything up.
             if timeout is None:
-                timeout = self.app.config.GRACEFUL_SHUTDOWN_TIMEOUT
-            self.loop.call_later(timeout, self.abort)
+                # This close timeout needs to be less than the graceful
+                # shutdown timeout. The graceful shutdown _could_ be waiting
+                # for this transport to close before shutting down the app.
+                timeout = self.app.config.GRACEFUL_SHUTDOWN_TIMEOUT / 2.0
+                # This is 15seconds/2 = 7.5seconds by default.
+        else:
+            # Schedule the async close checker but with no timeout,
+            # this will ensure abort() is called if required.
+            if timeout is None:
+                timeout = 0
+        self.loop.call_soon(
+            _async_protocol_transport_close,
+            self,
+            self.loop,
+            timeout,
+        )
 
     def abort(self):
         """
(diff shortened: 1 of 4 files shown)


ashleysommer:
I think there needs to be additional communication (in the docs?) about ideally using a Sanic Streaming Response to chunk the response data when a response payload is larger than approx 16kb, rather than writing the whole payload in a single HTTP Response body. That would prevent the UVLoop UVStreamTransport from having situations where it cannot send the whole buffer in one go, so preventing this kind of error occurring that we were seeing here.

Though I know often its not possible to know the size of the response, eg, a dict passed to `json()` response could easily be a couple hundred KB without the user realizing it.

ahopkins:
Beautiful. Thanks for taking this on @ashleysommer. I wonder if we should add a configurable option to automatically kick over to streaming responses. The benefit to the dev is obviously not having to worry about sizing. 

I'll give this a look later tonight and get it merged. We can backport to the LTS also.

robd003:
Thank you so much for taking the time to debug this @ashleysommer You're a rock star!

sanic/server/protocols/base_protocol.py
```diff
@@ -86,12 +86,47 @@ def close(self, timeout: Optional[float] = None):
...
+            # don't call resume_writing here, it gets called by the transport
+            # to unpause the protocol when it is ready for more data
+
+            # Schedule the async close checker, to close the connection
+            # after the transport is done, and clean everything up.
+            if timeout is None:
+                # This close timeout needs to be less than the graceful
+                # shutdown timeout. The graceful shutdown _could_ be waiting
+                # for this transport to close before shutting down the app.
+                timeout = self.app.config.GRACEFUL_SHUTDOWN_TIMEOUT / 2.0
```
ahopkins, line 117: I agree, we should just have a different value (like `GRACEFUL_TCP_CLOSE_TIMEOUT` and not reuse this which is really something different).

ahopkins (approved):

ahopkins (approved):

ashleysommer:
I agree with the addition of the `GRACEFUL_TCP_CLOSE_TIMEOUT`. I was going to add that change to this PR, but I didn't want this diff to grow too large and affect other parts of sanic (I know adding new tuneables in the settings can be hard to get right). Thanks for adding it.

One further thing I would suggest, on startup during sanity-checks we could verify that GRACEFUL_TCP_CLOSE_TIMEOUT is less than GRACEFUL_SHUTDOWN_TIMEOUT, and issue a warning if it is not.
Or force:
```python
GRACEFUL_TCP_CLOSE_TIMEOUT=min(GRACEFUL_SHUTDOWN_TIMEOUT, GRACEFUL_TCP_CLOSE_TIMEOUT)
```

ahopkins:
Forcing that would be awkward because it wouldn't behave as expected. Warning or error makes it clear. But I doubt it will be a value anyone will have any need to configure, so it's very unlikely we'd need this. Startup checks are however a good idea.

PR #2976, opened 2024-06-28
Fix deleting a cookie that was created with secure=False

### LINKED ISSUE #2972: Cannot delete cookies due to missing `secure` parameter

### Describe the bug

The `delete_cookie` method in `CookieJar` lacks the `secure` parameter, and this parameter in [its internal call to `self.add_cookie`](https://github.com/sanic-org/sanic/blob/39d4ea49761bb2221cd6cb970a41e9099990273f/sanic/cookies/response.py#L408) default to True.

Therefore, attempts to delete a cookie in an insecure (plain HTTP) origin will be always blocked by the browser because it will set the `;Secure` flag:

image

### Code snippet

```python
from sanic import Sanic
from sanic.response import text, redirect

app = Sanic("test")
app.config.AUTO_EXTEND = False

@app.get("/")
async def index(request):
    return text(f'logged in: {"YES" if request.cookies.get("session", None) else "NO"}')

@app.get("/login")
async def login(request):
    res = redirect('/')
    res.cookies.add_cookie('session', 'foobar', secure=False)
    return res
    
@app.get("/logout")
async def logout(request):
    res = redirect('/')
    res.cookies.delete_cookie('session')
    return res

if __name__ == '__main__':
    app.run(host='0.0.0.0', port=4321, debug=True)
```

1. Visit `http://YOUR_IP:4321/` and the page says `logged in: NO`
2. Visit `http://YOUR_IP:4321/login` and the page says `logged in: YES`
3. Visit `http://YOUR_IP:4321/logout` and the page still says `logged in: YES`

Note: it is important to use the real IP address (instead of 127.0.0.1) to reproduce this bug, since 127.0.0.1 is a secure origin.

### Expected Behavior

Sanic should add an optional `secure` parameter to `delete_cookie` method, so we can write something like `res.cookies.delete_cookie('session', secure=False)`.

Then the page should say `logged in: NO` after visiting `/logout`.

### How do you run Sanic?

### Operating System

Linux

### Sanic Version

v23.12.1

### Additional context

_No response_

ashleysommer:
Add ability to create a non-secure delete-cookie.
Also re-use existing cookie properties where possible if deleting a cookie that is already known in the response headers. Sanity-check and correct deletion cookie's `secure_prefix` and `host_prefix` based on the value of `secure` bool.

Add tests for these three things. (they were failing before the fix, passing after the fix).
Fixes #2972

### DIFF
--- a/sanic/cookies/response.py
+++ b/sanic/cookies/response.py
@@ -392,10 +392,6 @@ class CookieJar(dict):
             This requires that secure=True, defaults to False
         :type secure_prefix: bool
         """
-
-        # Host prefix can only be used on a cookie where
-        # path=="/", domain==None, and secure==True
-
         if host_prefix and not (secure and path == "/" and domain is None):
             raise ServerError(
                 "Cannot set host_prefix on a cookie without "
@@ -406,7 +402,6 @@ class CookieJar(dict):
                 "Cannot set secure_prefix on a cookie without secure=True"
             )
 
-        # remove it from header
         cookies: List[Cookie] = self.headers.popall(self.HEADER_KEY, [])
         existing_cookie = None
         for cookie in cookies:
@@ -417,8 +412,6 @@ class CookieJar(dict):
             ):
                 self.headers.add(self.HEADER_KEY, cookie)
             elif existing_cookie is None:
-                # Keep a snapshot of the cookie-to-be-deleted
-                # for reference when creating the deletion cookie
                 existing_cookie = cookie
         # This should be removed in v24.3
         try:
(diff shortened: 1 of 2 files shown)


sanic/cookies/response.py
```diff
@@ -389,32 +392,66 @@ def delete_cookie(
...
+
+        # Host prefix can only be used on a cookie where
+        # path=="/", domain==None, and secure==True
+        host_prefix = (
+            False
+            if not (secure and path == "/" and domain is None)
+            else host_prefix
+        )
+        # Secure prefix can only be used on a secure cookie
+        secure_prefix = False if not secure else secure_prefix
```
ahopkins, line 404: Hiding safety rails leads to hard to debug code. Better off failing explicitly with an error.

PR #3000, opened 2024-09-27
chore: support `JSONRepsonse` `dumps` callable return type `bytes`

imnotjames:
the `dumps` parameter is used to serialize the `JSONResponse` to text.  this is often done with python's built in `json.dumps` but other times via `ujson.dumps` and `orjson.dumps`

however, to use `orjson.dumps` [as suggested in the docs](https://sanic.dev/en/guide/basics/response.html#json), you have to ignore the suggested type (`str`) because `orjson.dumps` returns a `bytes`

to correct this, the typing for the dumps callable has been switched to be an `Callable[..., AnyStr]` so either `bytes` or `str` returning callable can be passed

this was previously done in https://github.com/sanic-org/sanic/pull/2193 but for some reason not brought over to `JSONResponse`?

### DIFF
--- a/sanic/response/types.py
+++ b/sanic/response/types.py
@@ -59,7 +59,7 @@ class BaseHTTPResponse:
         class_name = self.__class__.__name__
         return f"<{class_name}: {self.status} {self.content_type}>"
 
-    def _encode_body(self, data: Optional[AnyStr]):
+    def _encode_body(self, data: Optional[str | bytes]):
         if data is None:
             return b""
         return data.encode() if hasattr(data, "encode") else data  # type: ignore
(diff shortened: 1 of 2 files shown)


sanic/response/types.py
```diff
@@ -59,7 +59,7 @@ def __repr__(self):
         class_name = self.__class__.__name__
         return f"<{class_name}: {self.status} {self.content_type}>"
 
-    def _encode_body(self, data: Optional[AnyStr]):
+    def _encode_body(self, data: Optional[str | bytes]):
```
imnotjames, line 62: This was an odd change - without this, everything complained that it should be `Sequence[object]`?

```
sanic/response/types.py:334: error: Value of type variable "AnyStr" of "_encode_body" of "BaseHTTPResponse" cannot be "Sequence[object]"  [type-var]
```

Unfortunately, not familiar enough with mypy for this.

imnotjames:
For some reason the type checking is acting up - but it's on things that I don't think are related.  Happy to fix as needed with a bit of direction, though.

ahopkins (approved):

PR #3009, opened 2024-11-15
Fix SanicException `quiet` attribute handling when set to False

C5H12O5:
Fixes an issue where the `quiet` attribute in `SanicException` would default to the class attribute when set to `False`. This occurs when the subclass `quiet` attribute is `True`. The fix ensures only `None` triggers the fallback.


ahopkins (approved):

ahopkins:
Excellent. Thanks for the test also!

PR #3016, opened 2024-12-18
Fix incompatible type annotation of `get` method in the `HTTPMethodView` class

### LINKED ISSUE #2916: mypy error type for async get function in HTTPMethodView

### Describe the bug

Mypy throws an error overriding the signature of the get method, as it is described only for synchronous execution.

```
error: Signature of "get" incompatible with supertype "HTTPMethodView"  [override]
note:      Superclass:
note:          Callable[..., Any] | None
note:      Subclass:
note:          def get(self, request: Request[Any, Any]) -> Coroutine[Any, Any, HTTPResponse]
```

I suggest changing the description of the get method /sanic/views.py to:

```
get: Union[Callable[..., Any], Coroutine[..., Any], None]
```

### Code snippet

_No response_

### Expected Behavior

_No response_

### How do you run Sanic?

### Operating System

Linux

### Sanic Version

23.12.0

### Additional context

_No response_

ChihweiLHBird:
Resolves #2916

Use `getattr` to try getting the `get` handler to avoid the typing annotation incompatibility issue.

### DIFF
--- a/sanic/views.py
+++ b/sanic/views.py
@@ -145,7 +145,7 @@ class HTTPMethodView:
     def dispatch_request(self, request: Request, *args, **kwargs):
         """Dispatch request to appropriate handler method."""
         method = request.method.lower()
-        handler = getattr(self, method)
+        handler = getattr(self, method, None)
 
         if not handler and method == "head":
             handler = getattr(self, "get")
--- a/sanic/views.py
+++ b/sanic/views.py
@@ -115,8 +115,6 @@ class HTTPMethodView:
         to `"/v"`.
     """
 
-    get: Optional[Callable[..., Any]]
-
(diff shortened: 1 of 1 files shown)


sanic/views.py
```diff
@@ -146,9 +144,11 @@ def __init_subclass__(
 
     def dispatch_request(self, request: Request, *args, **kwargs):
         """Dispatch request to appropriate handler method."""
-        handler = getattr(self, request.method.lower(), None)
-        if not handler and request.method == "HEAD":
-            handler = self.get
+        method = request.method.lower()
+        handler = getattr(self, method)
```
ahopkins, line 148:
```suggestion
        handler = getattr(self, method, None)
```

ahopkins (approved):

Issue #3018, opened 2024-12-24
A more customisable CLI and REPL


goodki-d:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the solution you'd like

Sanic CLI currently has options for different run configurations.
**Requested:** 
- Ability to add custom commands for application related workflows. With application context available to the command.

Usage can be similar to how inspector commands are created currently. 

Sanic REPL currently provides 4 objects by default In the repl context.
**Requested:**
- Ability to add/inject additional objects to the repl context.
- Ability to prepare REPL context. (e.g. pre importing modules, models)

Thanks

goodki-d:
REPL was one of the best feature added for developer experience imo. Being able to have some models/classes pre-imported or adding custom utility objects would be great. I'd love to contribute too.

ahopkins:
The CLI bit I was planning for a long time. The how of it was always the concern, but I figured that out. Stay tuned. 

I like the REPL injection idea also.

ahopkins:
See #3019 

If the REPL idea is something you'd like to  work on @goodki-d, LMK.

goodki-d:
Yes! I'd love to work on it. I have some approaches in mind

ahopkins:
Excellent. What was your thought? It's going to be a bit of a challenge as you are diving deep into the process manager internals. Maybe ping me on the Discord server and we can brainstorm?

goodki-d:
Sure, lets connect on discord!

I was not sure about the process manager here. I assumed REPL was running on the main process.
https://github.com/sanic-org/sanic/blob/a575f5c8858248680653c5a5b9565be4dcc56f9a/sanic/cli/app.py#L183-L186

---

I was thinking If we could merge the user supplied variables with the currently available locals
```diff
- super().__init__(locals=locals_available)
+ super().__init__(locals={**locals_available, **user_locals})
```

and generating the variable description to show dynamically. 

For the user API, I was thinking of something like
```python
from sanic import Sanic
app = Sanic(...)

# variable == module/class/object/callable
app.repl.add_local(variable, "<name>", "desc")
```

ahopkins:
Merged in #3042

PR #3026, opened 2025-01-09
Add REPL context

goodki-d:
Introduces REPL context to allow users customise their REPL experience.

Please refer to #3018

**API**
```python
# variable == module/class/object/callable
app.repl_ctx.add(variable, "<name>", "desc")
```

Renders in console as:

    <name> <type> : <desc>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,7 +58,7 @@ Sanic | Build fast. Run fast.
 
 .. end-badges
 
-Sanic is a **Python 3.8+** web server and web framework that's written to go fast. It allows the usage of the ``async/await`` syntax added in Python 3.5, which makes your code non-blocking and speedy.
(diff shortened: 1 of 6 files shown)


ahopkins:
I'm wondering if we should get a little *smart* with this and allow assignment so that in the simplest use case there is just this:

```python
def foo():
    """Some docstring"""


app.repl_ctx.foo = foo
```

The `.add(...)` method is just when you want the more fine-grained control.

goodki-d:
That'd be somewhat simpler to implement also a simpler API. We'll need to do formatting (like truncating doctoring to fit console) on render time, that's fine.

Should we use an internal state (like currently `_locals` set) or just iterate the attributes?

ahopkins (approved):

PR #3067, opened 2025-06-03
feat: add SERVER_NAME to passthru config if specified in app config

dhensen:
This does not work as expected:

```python
@app.main_process_start
async def update_server_name(app, loop):
    some_url = some_reason()
    if some_url:
        app.update_config({"SERVER_NAME": some_url})
```

## Problem
I would have expected my SERVER_NAME update to have propagated to workers, especially since doing this on `main_process_start`, but this is not the case. Because I have SANIC_SERVER_NAME env var set to some different value, all my workers still take that value, even though I just updated it with a new value.

## Solution
With this PR, code above now works, and I can optionally change my SERVER_NAME at `main_process_start` and all my workers will have the correct SERVER_NAME.

### DIFF
--- a/sanic/mixins/startup.py
+++ b/sanic/mixins/startup.py
@@ -1078,7 +1078,10 @@ class StartupMixin(metaclass=SanicMeta):
                 "shared_ctx": app.shared_ctx.__dict__,
             }
             if app.config.SERVER_NAME:
-                kwargs["passthru"]["config"]["SERVER_NAME"] = app.config.SERVER_NAME
+                if app.config.SERVER_NAME:
+                    kwargs["passthru"]["config"]["SERVER_NAME"] = (
+                        app.config.SERVER_NAME
+                    )
             for app in apps:
                 kwargs["server_info"][app.name] = []
                 for server_info in app.state.server_info:
--- a/sanic/mixins/startup.py
+++ b/sanic/mixins/startup.py
@@ -1077,6 +1077,8 @@ class StartupMixin(metaclass=SanicMeta):
                 },
                 "shared_ctx": app.shared_ctx.__dict__,
             }
+            if app.config.SERVER_NAME:
+                kwargs["passthru"]["config"]["SERVER_NAME"] = app.config.SERVER_NAME
             for app in apps:
                 kwargs["server_info"][app.name] = []
                 for server_info in app.state.server_info:


dhensen:
Converted to draft: when `auto_reload=True` this method does not update SERVER_NAME.

BTW: any idea why all those checks are failing? I see lint errors on files I did not touch.

DSeaStar:
Three notes on this one — it has had no human review since June 2025.

### 1. The `if` is duplicated

```python
if app.config.SERVER_NAME:
    if app.config.SERVER_NAME:
        kwargs["passthru"]["config"]["SERVER_NAME"] = (
            app.config.SERVER_NAME
        )
```

The inner test repeats the outer one; collapsing to a single `if` would be enough.

### 2. Why all those checks failed (not your change)

GitHub no longer retains any artifacts for head `d89f1faf944767cb15977ecb68210d0b4d269ead` — `actions/runs?head_sha=...` returns nothing and `check-runs` is empty — so the red X's can't be re-read. On current `main` (`5ffc7b37`) there are two known failures that hit *every* PR:

- **`tox -e type-checking`** — `sanic/pages/error.py:43: Incompatible types in assignment (Expression[Any] / None)`. `setup.py:143` declares `"mypy"` with no version pin and `[testenv:type-checking]` in `tox.ini` has no `deps`, so CI always installs the newest mypy, and a newer mypy is what started flagging this. Fixed in #3190.
- **`py311-no-ext`** — `cannot import name 'ServerProtocol' from 'websockets.server'`. `uvicorn >= 0.50` needs `websockets >= 13`, but `pytest-sanic 1.9.1` pins `websockets < 11`; the conflict only surfaces at import time. Worked around in #3191.

You are `mergeable=true` (branch is behind), so a rebase onto current `main` plus a re-run should give you a readable result.

### 3. One design question worth pinning down with a test

`passthru["config"]` exists to carry values that only exist in the *parent* process: the two keys it already carries, `ACCESS_LOG` and `NOISY_EXCEPTIONS`, are set from CLI flags at startup (`sanic/mixins/startup.py:1093-1094`). `SERVER_NAME` is normally re-created inside the worker, because `AppLoader.load()` re-imports the application module (`sanic/worker/loader.py:92`), so a `SERVER_NAME` assigned in your source will already be present there.

The case where your line genuinely matters is when the value is only assigned **at runtime in the parent** — e.g. from a `main_start` / `before_server_start` listener. A regression test that sets `app.config.SERVER_NAME` at runtime in the parent and then asserts the worker sees it (via `request.url_for(..., _external=True)` or `request.server_name`, both of which read `config.SERVER_NAME` in `sanic/request/types.py:977,1004`) would both demonstrate the gap and lock the behaviour in. Without that test, this reads as defensive and a reviewer may ask for it anyway.

PR #3071, opened 2025-06-25
feat: Add automatic charset=UTF-8 to text content types

### LINKED ISSUE #2987: Content type on static files lacks charset=UTF-8

### Describe the bug

`app.static` should specify UTF-8 for MIME types `text/*`. Currently it returns simply `text/html` for HTML files.

### Code snippet

```python
from sanic import Sanic

app = Sanic("index")
app.static("/", ".", index="index.html")
```

### Expected Behavior

```
content-type: text/html; charset=UTF-8
```

### How do you run Sanic?

Sanic CLI

### Operating System

Linux

### Sanic Version

24.6.0

### Additional context

_No response_

Tronic:
Refactor content type handling to automatically append charset=UTF-8 to text/* MIME types when serving static files and file responses.

- Add new guess_content_type() utility function that wraps mimetypes.guess_type()
- Automatically append '; charset=UTF-8' to text content types
- Replace direct mimetypes.guess_type() usage with new utility
- Update static file serving and file() response functions to use new utility

Fix #2987

### DIFF
--- a/sanic/mixins/static.py
+++ b/sanic/mixins/static.py
@@ -1,7 +1,6 @@
 from collections.abc import Sequence
 from email.utils import formatdate
 from functools import partial, wraps
-from mimetypes import guess_type
 from os import PathLike, path
 from pathlib import Path, PurePath
 from typing import Optional, Union
@@ -11,7 +10,6 @@ from sanic_routing.route import Route
 
 from sanic.base.meta import SanicMeta
 from sanic.compat import stat_async
-from sanic.constants import DEFAULT_HTTP_CONTENT_TYPE
 from sanic.exceptions import FileNotFound, HeaderNotFound, RangeNotSatisfiable
 from sanic.handlers import ContentRangeHandler
 from sanic.handlers.directory import DirectoryHandler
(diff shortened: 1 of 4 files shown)


Tronic:
Looks like tests are (still) broken. I cannot run them on my own box either without plenty of (unrelated) failures. Further work may be required on the PR, in particular with any existing tests that could've been broken by the change.

Tronic:
I manually ran only the relevant tests because so many things fail with Sanic test server "address already in use" making the full test results unmanageable. Fixed a few issues and optimized the tests run a bit faster avoiding overlap.

I note that `file()` should probably default to HTTP default (i.e. application/octet-stream), rather than `text/plain` as it was hardcoded to do. In particular this affects if some arbitrary format data files are sent (e.g. test.dat would be responded as plain text), although one can see the reasoning to send a bare `README` or such (legacy) files as text so it can be displayed in browsers rather than downloaded. This however is in conflict with how Sanic behaves for the static file handler.

For the sake of maintaining compatibility with prior versions, this PR does not change the `file()` fallback type, other than by appending to the charset to it as well. `app.static()` just as before defaults to HTTP default, which has no charset.

Issue #3074, opened 2025-06-27
Replace setup.py entirely with pyproject.toml and modern toolchain


Tronic:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Sanic still uses setup.py to make sure that uvloop and ujson are installed as part of the default install (package `sanic`), as they increase performance significantly -- but while at the same time allowing one to disable each of them (e.g. to maintain pure Python implementation).

Unfortunately the extras system does not accommodate to this unless Sanic takes a similar approach as FastAPI who in their documentation tell people to install `fastapi[standard]` rather than the minimal package without extras. For this reason, Sanic has stuck with the setup.py system and environment variables to control the build, until now.

### Describe the solution you'd like

The simple solution would be to make the `sanic` package require those optional deps. This will cause the packages be installed (can be disabled for specific platforms where they are simply not available). Additionally, implement runtime controls to disable the use of these.

Alternative solutions that I could think of is to maintain a separate `sanic-minimal` package without the deps, or to instruct users to install `sanic[standard]` like FastAPI does. Both are rather cumbersome, especially considering the existing userbase that depends on `sanic`.

I favor the solution presented in the first paragraph, as depending on the existing custom hacks on legacy setup.py install is becoming intolerable. E.g. `uv sync` needs the project build to be done via pyproject.toml.

ahopkins:
Coming in 26.3

PR #3092, opened 2025-11-29
Modernise tooling (pyproject.toml, nox, uv)

Tronic:
Implementing #3074

### DIFF
--- a/sanic/server/websockets/impl.py
+++ b/sanic/server/websockets/impl.py
@@ -1,6 +1,5 @@
 import asyncio
-import random
-import struct
+import secrets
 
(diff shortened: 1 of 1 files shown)


Tronic:
Now `uv run nox` runs pretty much everything starting with formatting, and it passes all stages except the tests because many of them are still flaky. However, this platform at least allows me to run tests (that has been broken for a long time with tox & friends).

An open question is whether to replace bandit entirely with ruff. Currently ruff security checks (S) are disabled in config and we run bandit separately. These could all naturally be part of the lint action, along with type checking.

PR #3093, opened 2025-11-29
Fix typing errors

Tronic:
Fix two typing errors, and implement SameSite to always use capitalized values (still allows passing them in lower case).


ahopkins:
No longer needed.

PR #3127, opened 2026-01-06
More gating around the new Daemon for Windows compatability

### LINKED ISSUE #3126: v25.12.0 no longer works on Windows

### Describe the bug

v25.12.0 no longer runs on windows. Seems there are a few issues caused by #3110:
- https://github.com/sanic-org/sanic/blob/785d77f8fe208576a48d339bd81d866021154092/sanic/worker/daemon.py#L6 Windows does not have the `pwd` module.
- https://github.com/sanic-org/sanic/blob/785d77f8fe208576a48d339bd81d866021154092/sanic/worker/daemon.py#L4 Windows does not have the `grp` module.
- https://github.com/sanic-org/sanic/blob/785d77f8fe208576a48d339bd81d866021154092/sanic/cli/app.py#L479 Checks for `args.unix` but this now crashes on windows because the parser doesn't add that argument on windows anymore. https://github.com/sanic-org/sanic/blob/785d77f8fe208576a48d339bd81d866021154092/sanic/cli/arguments.py#L170-L173

Cursory testing indicates everything works after those 3 issues are fixed.

### Code snippet

_No response_

### Expected Behavior

_No response_

### How do you run Sanic?

Sanic CLI

### Operating System

Windows

### Sanic Version

2025.12.0

### Additional context

_No response_

ahopkins:
Fixes #3126


nicoddemus:
Could you improve CI to catch this errors early, before a release?

DSeaStar:
Not a maintainer. Heads-up that #3167 (2026-06-21) makes the same two changes for #3126 — `daemon.py` (`grp`/`pwd` guarded) and `cli/app.py` (`getattr(self.args, "unix", "")`), just split into two `try/except` blocks instead of one.

Two differences worth knowing before either lands:

1. This branch is based on 785d77f8 and GitHub now reports it `behind`. #3167 is based on current `main` (5ffc7b37).
2. The `sanic/__version__.py` hunk here bumps to `25.12.1`, but `main` is already at `25.12.1` — that hunk is now stale and should probably be dropped.

I verified the fix itself on real Windows 11 / Python 3.14.3: on `main`, `python -m sanic --version` and `import sanic.worker.daemon` both fail with `ModuleNotFoundError: No module named 'grp'`; with the guard they work, and the CLI change is separately required (without it the CLI still dies at `cli/app.py:479` on `self.args.unix`). Details in my comment on #3167.

PR #3140, opened 2026-02-18
Fix SSL configuration ignored for non-primary apps in multi-serve

### LINKED ISSUE #3131: SSL configuration ignored for non-primary apps

### Describe the bug

When serving multiple apps, the ssl configuration of non-primary apps is ignored and they're served through http instead of https.

Looked through the source a bit and it seems that ssl configuration is not passed from the startup code to `WorkerManager` on a per-app basis. Once control gets to `worker_serve`, only the primary app's ssl configuration is available.

As an experiment, I passed the primary's ssl configuration to the other apps in `worker_serve` and it seemed to work, so it might just be a matter of moving the configuration through the layers.

### Code snippet

```python
main_app = Sanic("main")
secondary_app = Sanic("secondary")

if __name__ == "__main__":
    main_app.prepare(port=5000, ssl="main_certs")
    secondary_app.prepare(port=5001, ssl="secondary_certs")   # ssl arg ignored, served as http
    Sanic.serve()
```

### Expected Behavior

_No response_

### How do you run Sanic?

### Operating System

Windows

### Sanic Version

25.12

### Additional context

_No response_

worksbyfriday:
## Description

Fixes #3131

When serving multiple apps via `Sanic.serve()`, the SSL configuration for non-primary apps was silently ignored — secondary apps were served over HTTP instead of HTTPS.

### Root Cause

The SSL configuration flow had a gap for non-primary apps:

1. **`Sanic.serve()`** stripped SSL from all apps' `server_info.settings` during serialization for multiprocessing (since `SSLContext` objects can't be pickled)
2. Only the primary app's SSL was passed as a global `kwargs["ssl"]`
3. **`worker_serve()`** only restored SSL to the primary app's server_info

### Fix

**In `startup.py`**: Instead of stripping SSL entirely, convert `SanicSSLContext` to its serializable dict form (`sanic` attribute) and keep it in each app's `server_info.settings`. Raw dict/string SSL values are passed through as-is.

**In `serve.py`**: After the existing primary-app SSL loading, iterate through all apps' server_info and load SSL contexts from any serialized SSL dicts that haven't been converted yet.

This preserves backward compatibility — the primary app still gets its SSL from the global kwarg, while non-primary apps now get theirs from their own server_info.

### DIFF
--- a/sanic/mixins/startup.py
+++ b/sanic/mixins/startup.py
@@ -1098,11 +1098,19 @@ class StartupMixin(metaclass=SanicMeta):
             for app in apps:
                 kwargs["server_info"][app.name] = []
                 for server_info in app.state.server_info:
-                    server_info.settings = {
+                    # Convert SSL context to serializable dict form
+                    # so it can be passed through multiprocessing
+                    app_ssl = server_info.settings.get("ssl")
+                    filtered_settings = {
                         k: v
                         for k, v in server_info.settings.items()
                         if k not in ("main_start", "main_stop", "app", "ssl")
                     }
+                    if isinstance(app_ssl, SanicSSLContext):
+                        filtered_settings["ssl"] = app_ssl.sanic
+                    elif app_ssl is not None:
+                        filtered_settings["ssl"] = app_ssl
+                    server_info.settings = filtered_settings
                     kwargs["server_info"][app.name].append(server_info)
 
             ssl = kwargs.get("ssl")
--- a/sanic/worker/serve.py
+++ b/sanic/worker/serve.py
@@ -83,6 +83,18 @@ def worker_serve(
             for info in app.state.server_info:
                 info.settings["ssl"] = ssl
 
+        # Load per-app SSL for non-primary apps from their server_info settings
+        if server_info:
+            for si_app_name in server_info:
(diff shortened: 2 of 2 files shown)


DSeaStar:
Not a maintainer, just triaging. One concern with the `startup.py` hunk:

```python
elif app_ssl is not None:
    filtered_settings["ssl"] = app_ssl
```

`_helper()` stores `process_to_context(ssl)`, and `process_to_context()` returns a plain `ssl.SSLContext` untouched when the app was started with `app.run(ssl=<SSLContext>)`. That branch then puts a raw `SSLContext` back into `kwargs["server_info"]`, which `WorkerManager` passes to `multiprocessing` as process kwargs (`factory(..., kwargs=self.kwargs)`). Under the spawn/forkserver start methods those kwargs are pickled:

```
TypeError: cannot pickle 'SSLContext' object
```

(confirmed locally on 3.14.3). The existing `k not in (..., "ssl")` filter is there for exactly this reason.

#3154 avoids it by sending `SanicSSLContext.sanic`, which is a plain dict — the same thing `main` already does for the primary just below, at `kwargs["ssl"] = ssl.sanic`. Suggest dropping the `elif`, or routing it through `.sanic` too.

PR #3141, opened 2026-02-20
Assign datetime.replace() return value in validate_file

bysiber:
`validate_file()` calls `datetime.replace(tzinfo=timezone.utc)` to normalize tz-naive datetimes before comparing `last_modified` and `if_modified_since` timestamps. However, `datetime.replace()` returns a **new** datetime — it doesn't mutate in place. Since the return values aren't assigned back, both variables remain tz-naive after the "fix."

The practical effect: on any server not running in UTC, `last_modified.timestamp()` interprets the tz-naive datetime as local time while the tz-aware one is interpreted as UTC. This produces incorrect comparisons and breaks HTTP 304 Not Modified responses — either serving stale content or forcing unnecessary re-downloads depending on the timezone offset direction.

### DIFF
--- a/sanic/response/convenience.py
+++ b/sanic/response/convenience.py
@@ -187,7 +187,7 @@ async def validate_file(
             "Cannot compare tz-aware and tz-naive datetimes. To avoid "
             "this conflict Sanic is converting last_modified to UTC."
         )
-        last_modified.replace(tzinfo=timezone.utc)
+        last_modified = last_modified.replace(tzinfo=timezone.utc)
     elif (
         last_modified.utcoffset() is not None
         and if_modified_since.utcoffset() is None
@@ -196,7 +196,7 @@ async def validate_file(
             "Cannot compare tz-aware and tz-naive datetimes. To avoid "
             "this conflict Sanic is converting if_modified_since to UTC."
         )
-        if_modified_since.replace(tzinfo=timezone.utc)
+        if_modified_since = if_modified_since.replace(tzinfo=timezone.utc)
     if last_modified.timestamp() <= if_modified_since.timestamp():
         return HTTPResponse(status=304)
 


DSeaStar:
Confirmed on `main` (`5ffc7b37`) — this is a real bug with user-visible impact, and I could reproduce it deterministically.

`datetime.replace()` returns a **new** object, so `last_modified.replace(tzinfo=timezone.utc)` (and the `if_modified_since` one) is a no-op: the datetime stays tz-naive, and the very next line `last_modified.timestamp()` interprets it as **local** time. The conversion the log message announces ("Sanic is converting last_modified to UTC") never actually happens, so the 304 decision ends up off by the server's UTC offset.

Repro (Windows, host timezone UTC+8):

```python
from datetime import datetime, timezone
from sanic.compat import Header
from sanic.response.convenience import validate_file

lm = datetime(2024, 1, 1, 12, 0, 0)                                   # naive, meant as 12:00 UTC
hdr = Header({"If-Modified-Since": "Mon, 01 Jan 2024 11:00:00 GMT"})  # aware UTC
asyncio.run(validate_file(hdr, lm))
```

| `last_modified` | `If-Modified-Since` | unpatched | patched |
|---|---|---|---|
| naive `2024-01-01 12:00` | `11:00 GMT` | **304** | `None` (200) |
| aware `2024-01-01 12:00+00:00` | `11:00 GMT` | `None` (200) | `None` (200) |

Both rows describe the *same instant*. Today the naive path answers "not modified" while the aware path answers "modified" — so a client whose cached copy is one hour stale keeps getting `304` and never receives the newer file. After the patch the two agree.

Worth noting why CI never caught it: the runners are UTC, where naive and aware datetimes yield the same `timestamp()`, so this is invisible on any UTC host. It only bites deployments on a non-UTC server (which is most of them).

Regression check: `tests/test_static.py` is **85 passed / 4 failed** with and without the patch (the 4 failures are pre-existing on Windows in this environment).

A few suggestions:

- The branch is based on `785d77f8` and is now behind `main` — a rebase would be good.
- A regression test would be valuable, and the table above is the cleanest form: the same instant expressed naive vs. aware must produce the same verdict. Please construct the naive datetime explicitly rather than reading the host clock, otherwise the test silently passes for the wrong reason on UTC machines.
- Per repo convention this wants a `changelogs/<issue>.bugfix.rst`; I could not find an existing issue for it, so opening one (or the maintainers may want to fold it into a static-file issue) would help.

I have not pushed anything to your branch.

PR #3142, opened 2026-02-20
Guard signal waiter set_result against already-done futures

bysiber:
In `SignalRouter.dispatch()`, `set_result` is called on signal waiter futures without first checking whether the future is already done. If multiple signals in the same dispatch batch match the same waiter, or if the future was cancelled by another task before the loop reaches it, `set_result` raises `asyncio.InvalidStateError`.

This unhandled exception aborts the entire signal dispatch loop, preventing subsequent signal handlers from executing. Adding a `future.done()` guard before `set_result` avoids the crash and ensures all matching signals are dispatched.

### DIFF
--- a/sanic/signals.py
+++ b/sanic/signals.py
@@ -233,7 +233,8 @@ class SignalRouter(BaseRouter):
             for signal in signals:
                 for waiter in signal.ctx.waiters:
                     if waiter.matches(event, condition):
-                        waiter.future.set_result(dict(params))
+                        if not waiter.future.done():
+                            waiter.future.set_result(dict(params))
 
             for signal in signals:
                 requirements = signal.extra.requirements


DSeaStar:
Confirmed on `main` (`5ffc7b37`) — this is a real crash, not just a defensive guard, and it reproduces deterministically.

`SignalWaiter.wait()` de-registers itself in a `finally`, which only runs once the waiting coroutine is *resumed*. `_dispatch()` resolves the future synchronously in its waiter loop and then walks straight into the handlers. If a second dispatch of the same event reaches the waiter loop before the waiting task has been scheduled, `set_result()` is called on an already-done future:

```python
@app.signal("foo.bar.baz")
def sig(): ...                    # non-suspending handler

async def main():
    app.signal_router.finalize()
    t = asyncio.ensure_future(app.event("foo.bar.baz"))
    await asyncio.sleep(0)                              # waiter registered
    await app.dispatch("foo.bar.baz", inline=True)
    await app.dispatch("foo.bar.baz", inline=True)      # <- boom
```

- unpatched: `InvalidStateError: invalid state`
- with this patch: no error, both dispatches complete normally

The blast radius is bigger than one waiter: `_dispatch()` wraps its body in a `try/except` that fires `server.exception.report` and then re-raises, so the exception propagates out of `dispatch()` — a waiter registered by user code (the `while True: await app.event(...)` background-task pattern shown in `app.event()`'s own docstring) can break dispatching for every other listener.

On reachability: the window needs two dispatches of the same event with no suspension point in between. `inline=True` is what makes that achievable, and Sanic itself uses it for all of its lifecycle dispatches (`http.lifecycle.response`, `http.lifecycle.handle`, `http.routing.after`, `http.lifecycle.exception`, ... in `sanic/app.py`). With the default `inline=False` there is an `await asyncio.sleep(0)` in `dispatch()` and the waiter gets to clean up first — which is exactly why this has stayed hidden.

`tests/test_signals.py`: **49 passed** with and without the patch.

Two suggestions:

- The branch is on `785d77f8` and is behind `main`; a rebase would be good.
- A regression test built on the shape above would be genuinely useful here — it reproduces on the first try and does not depend on timing or on scheduling luck.

PR #3143, opened 2026-02-20
Use None check for request.id to preserve falsy IDs

bysiber:
The `request.id` property uses `if not self._id` to decide whether to generate/fetch the ID. This treats any falsy value (including `0`, `""`) as "not yet set," causing the ID to be regenerated on every property access.

A custom `generate_id()` returning `0` or an integer request ID header value of `"0"` (which gets cast to `int(0)`) would trigger re-generation every time `request.id` is read. Switching to `is None` preserves any explicitly-set falsy ID while still triggering lazy initialization when genuinely unset.

### DIFF
--- a/sanic/request/types.py
+++ b/sanic/request/types.py
@@ -521,7 +521,7 @@ class Request(Generic[sanic_type, ctx_type]):
             uuid.UUID | str | int | None: A request ID passed from the
                 client, or generated from the backend.
         """
-        if not self._id:
+        if self._id is None:
             self._id = self.headers.getone(
                 self.app.config.REQUEST_ID_HEADER,
                 self.__class__.generate_id(self),  # type: ignore


themavik (commented):
Using is None for request._id fixes legitimate falsy ids (0, ""). nit: a bytes-like falsy id would still be regenerated—only matters if headers can surface that type.

DSeaStar:
Confirmed on `main` (`5ffc7b37`) — and the broken case is the example printed in the docstring a few lines above this line.

`generate_id()` is documented as returning `uuid.UUID | str | int | None`, and the example in `Request.id`'s docstring uses `itertools.count()`, which starts at **0**. With `if not self._id`, a legitimately falsy id (`0`, or `""`) is indistinguishable from "not generated yet", so every read of `request.id` re-enters the branch:

```python
class IntRequest(Request):
    counter = count()
    def generate_id(self):
        return next(self.counter)

app = Sanic(..., request_class=IntRequest)

@app.get("/")
async def handler(request):
    return json({"first": request.id, "second": request.id, "third": request.id})
```

- unpatched: `{"first": 0, "second": 1, "third": 1}` — three reads, **two different ids**, `generate_id()` called twice
- with this patch: `{"first": 0, "second": 0, "third": 0}` — `generate_id()` called once

Note the third read is `1` again: once the counter passes 0 the value is truthy and the id sticks. That makes this look intermittent and easy to misdiagnose as "my counter is being called twice somewhere". In practice anything that reads `request.id` more than once per request (handler + access log + a signal handler) can record two different ids for the same request, which quietly breaks log correlation.

Regression check: `tests/test_requests.py` is **129 passed / 6 failed** with and without the patch (the same 6 pre-existing Windows failures in this environment).

Two suggestions:

- The branch is on `785d77f8` and is behind `main`; a rebase would be good.
- For a regression test, assert that `generate_id()` is called **exactly once** across repeated `request.id` reads on a request class whose `generate_id()` returns `0`. Asserting only on the final value would pass as soon as the counter moves past 0, which is precisely how the current behavior hides.

PR #3145, opened 2026-03-11
Handle UnpicklingError during worker shutdown

pmontepagano:
When SIGTERM arrives during a `_sync_states()` call, Python's signal handler interrupts the blocking `read()` on the multiprocessing proxy connection. The shutdown handler then accesses the same connection (via `set_state()`), and `_ForkingPickler.loads()` sees corrupt bytes from the interrupted read, raising `_pickle.UnpicklingError`.

This adds `UnpicklingError` to the existing exception handling in:
- `WorkerProcess.terminate()` — alongside `BrokenPipeError`, `ConnectionResetError`, `EOFError`
- `WorkerProcess.exit()` — same tuple, for cleanup consistency
- `WorkerManager._sync_states()` — alongside `KeyError`

These follow the same pattern already used for other connection-related exceptions during shutdown.

Related: https://community.sanicframework.org/t/random-exceptions-with-workermanager/1154

### DIFF
--- a/sanic/worker/manager.py
+++ b/sanic/worker/manager.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import os
-import pickle
+import pickle  # nosec B403
 import signal
 
 from collections.abc import Iterable, MutableMapping
--- a/sanic/worker/manager.py
+++ b/sanic/worker/manager.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import pickle
 import signal
 
 from collections.abc import Iterable, MutableMapping
@@ -442,7 +443,7 @@ class WorkerManager:
         for process in self.processes:
             try:
                 state = self.worker_state[process.name].get("state")
-            except KeyError:
+            except (KeyError, pickle.UnpicklingError):
                 process.set_state(ProcessState.TERMINATED, True)
                 continue
             # Skip state sync if process is restarting to avoid race condition
--- a/sanic/worker/process.py
+++ b/sanic/worker/process.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import pickle
 
 from collections.abc import MutableMapping
 from datetime import datetime, timezone
@@ -99,6 +100,7 @@ class WorkerProcess:
                 ConnectionRefusedError,
                 ConnectionResetError,
                 EOFError,
+                pickle.UnpicklingError,
             ):
                 logger.debug("Monitor process has already exited.")
             except KeyError:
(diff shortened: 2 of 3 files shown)


themavik (commented):
Treating pickle.UnpicklingError like KeyError in _sync_states matches a broken multiprocessing proxy during shutdown. nit: pickle is only imported for the exception type—a short note beside nosec keeps bandit from looking for loads().

DSeaStar:
Not a maintainer, just triaging — this has been open since March with no review, so I checked it against `main`.

**The fix itself is still needed.** On `main` today the three sites are unchanged:
- `sanic/worker/manager.py:445` — `except KeyError:` only
- `sanic/worker/process.py:97-103` — `(BrokenPipeError, ConnectionRefusedError, ConnectionResetError, EOFError)`
- `sanic/worker/process.py:118` — `(BrokenPipeError, ConnectionResetError, EOFError)`

And `UnpicklingError` really isn't covered by any of them:

```
UnpicklingError mro: ['UnpicklingError', 'PickleError', 'Exception', 'BaseException', 'object']
issubclass(UnpicklingError, EOFError)  -> False
issubclass(UnpicklingError, KeyError)  -> False
```

A `multiprocessing.connection` that delivers a complete-but-corrupt payload raises exactly what you catch (verified on CPython 3.14.3):

```
a, b = mpc.Pipe(); b.send_bytes(b"\x00\x01\x02\x03"); a.recv()
-> _pickle.UnpicklingError: invalid load key, '\x00.'
isinstance(e, pickle.UnpicklingError) -> True
```

Worth noting in the PR description: `EOFError` already covers the *truncated* case, so the new catch adds the *complete but garbage* case — the two are complementary, not redundant.

**One of the three new tests can't run as written.** `test_terminate_handles_unpickling_error` builds the manager with four arguments:

```python
manager = WorkerManager(1, fake_serve, {}, worker_state)
```

but `WorkerManager.__init__` takes six:

```
(self, number, serve, server_settings, context, monitor_pubsub, worker_state)
-> TypeError: WorkerManager.__init__() missing 2 required positional arguments: 'monitor_pubsub' and 'worker_state'
```

so it errors at construction and never reaches `terminate()`. The other two tests already use the six-argument form, so it's a one-line copy:

```python
manager = WorkerManager(1, fake_serve, {}, context, (Mock(), Mock()), worker_state)
```

The mocking targets look right to me: `set_state()` reads `self.worker_state[self.name]` before writing, so `__getitem__` is the hook `terminate()` needs, and `exit()` uses `del self.worker_state[self.name]`, so `__delitem__` is right there.

I could not execute `tests/worker/test_manager.py` myself (Windows; the module is `OS_IS_WINDOWS`-guarded), so the above is from reading `main` plus a local CPython 3.14.3 check of the exception semantics — the arity problem is worth confirming on Linux. Base `785d77f8` is one commit behind `main`, so a rebase is cheap.

PR #3146, opened 2026-03-22
fix: suppress Python 3.14 deprecation warnings for asyncio policy API (#3129)

### LINKED ISSUE #3129: Python 3.14 deprecation warnings for asyncio.get_event_loop_policy still present in 25.12.0

### Describe the bug

When running Sanic 25.12.0 on Python 3.14, the following deprecation warnings are still emitted:

```
  /usr/local/lib/python3.14/site-packages/sanic/server/loop.py:46: DeprecationWarning: 'asyncio.get_event_loop_policy' is deprecated and slated for removal in Python 3.16
    if not isinstance(asyncio.get_event_loop_policy(), uvloop.EventLoopPolicy):
  /usr/local/lib/python3.14/site-packages/sanic/server/loop.py:47: DeprecationWarning: 'asyncio.set_event_loop_policy' is deprecated and slated for removal in Python 3.16
  /usr/local/lib/python3.14/site-packages/sanic/server/runners.py:174: DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead
    loop.add_signal_handler(
```

  PR #3115 addressed some Python 3.14 issues but these specific deprecations in loop.py and runners.py were not fixed.

### Code snippet

```
from sanic import Sanic
  from sanic.worker.loader import AppLoader

  def create_app() -> Sanic:
      app = Sanic("test-app")

      @app.get("/health")
      def health(request):
          return text("OK")

      return app

  def main():
      loader = AppLoader(factory=create_app)
      app = loader.load()
      app.prepare(host="0.0.0.0", port=8080)
      Sanic.serve(primary=app, app_loader=loader)

  if __name__ == "__main__":
      main()
```

### Expected Behavior

No deprecation warnings should be emitted when running on Python 3.14.

### How do you run Sanic?

### Operating System

Linux

### Sanic Version

25.12.0

### Additional context

Reproduced on both Linux and macOS

r266-tech:
## Description

Fixes #3129

Python 3.14 deprecates `asyncio.get_event_loop_policy()` and `asyncio.set_event_loop_policy()` with `DeprecationWarning`. These APIs are still the recommended way to configure uvloop and Windows event loop policies, so the fix wraps calls in `warnings.catch_warnings()` on Python >= 3.14 to suppress warnings originating from the asyncio module.

### Changes
- `sanic/server/loop.py`: Wrapped event loop policy checks in `warnings.catch_warnings()` blocks for both `try_use_uvloop()` and `try_windows_loop()` functions

### Testing
On Python 3.14, running Sanic with uvloop should no longer emit:
```
DeprecationWarning: 'asyncio.get_event_loop_policy' is deprecated and slated for removal in Python 3.16
DeprecationWarning: 'asyncio.set_event_loop_policy' is deprecated and slated for removal in Python 3.16
```

The pattern is intentionally kept simple — once Python 3.16 provides a `loop_factory` API, the `warnings.catch_warnings` wrapper can be replaced with direct `set_event_loop_factory()` calls.

### DIFF
--- a/sanic/app.py
+++ b/sanic/app.py
@@ -416,7 +416,7 @@ class Sanic(
         try:
             return get_running_loop()
         except RuntimeError:  # no cov
-            return asyncio.get_event_loop_policy().get_event_loop()
+            return asyncio.get_event_loop()
 
     # -------------------------------------------------------------------- #
     # Registration
--- a/sanic/server/loop.py
+++ b/sanic/server/loop.py
@@ -9,6 +9,15 @@ from sanic.log import error_logger
 from sanic.utils import str_to_bool
 
 
+# Python 3.14 deprecates the asyncio policy API (slated for removal in
+# Python 3.16). Match only the policy-related messages so other
+# DeprecationWarnings raised from asyncio are not hidden.
+_POLICY_DEPRECATION_MESSAGE = (
+    r"'asyncio\.((get|set)_event_loop_policy|\w*EventLoopPolicy)'"
+    r" is deprecated"
+)
+
+
 def try_use_uvloop() -> None:
     """Use uvloop instead of the default asyncio loop."""
     if OS_IS_WINDOWS:
@@ -48,7 +57,9 @@ def try_use_uvloop() -> None:
     with warnings.catch_warnings():
         if sys.version_info >= (3, 14):
             warnings.filterwarnings(
-                "ignore", category=DeprecationWarning, module="asyncio"
+                "ignore",
+                category=DeprecationWarning,
+                message=_POLICY_DEPRECATION_MESSAGE,
             )
         if not isinstance(
             asyncio.get_event_loop_policy(), uvloop.EventLoopPolicy
@@ -70,7 +81,9 @@ def try_windows_loop():
     with warnings.catch_warnings():
         if sys.version_info >= (3, 14):
             warnings.filterwarnings(
-                "ignore", category=DeprecationWarning, module="asyncio"
+                "ignore",
+                category=DeprecationWarning,
+                message=_POLICY_DEPRECATION_MESSAGE,
             )
         if not isinstance(
             asyncio.get_event_loop_policy(),
--- a/sanic/server/loop.py
+++ b/sanic/server/loop.py
@@ -1,4 +1,6 @@
 import asyncio
+import sys
+import warnings
 
 from os import getenv
 
@@ -43,8 +45,15 @@ def try_use_uvloop() -> None:
             "false."
         )
 
-    if not isinstance(asyncio.get_event_loop_policy(), uvloop.EventLoopPolicy):
-        asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
+    with warnings.catch_warnings():
+        if sys.version_info >= (3, 14):
+            warnings.filterwarnings(
+                "ignore", category=DeprecationWarning, module="asyncio"
+            )
+        if not isinstance(
+            asyncio.get_event_loop_policy(), uvloop.EventLoopPolicy
+        ):
+            asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 
 
 def try_windows_loop():
@@ -58,7 +67,15 @@ def try_windows_loop():
         )
         return
 
-    if not isinstance(
-        asyncio.get_event_loop_policy(), asyncio.WindowsSelectorEventLoopPolicy
-    ):
-        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+    with warnings.catch_warnings():
+        if sys.version_info >= (3, 14):
+            warnings.filterwarnings(
+                "ignore", category=DeprecationWarning, module="asyncio"
+            )
+        if not isinstance(
+            asyncio.get_event_loop_policy(),
+            asyncio.WindowsSelectorEventLoopPolicy,
+        ):
+            asyncio.set_event_loop_policy(
+                asyncio.WindowsSelectorEventLoopPolicy()
+            )
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -1,8 +1,10 @@
 import asyncio
 import logging
 import re
+import warnings
 
 from collections import Counter
+from concurrent.futures import ThreadPoolExecutor
 from inspect import isawaitable
 from os import environ
 from unittest.mock import Mock, patch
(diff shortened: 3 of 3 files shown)


themavik (commented):
module=asyncio in warnings.filterwarnings hides every DeprecationWarning from asyncio inside try_use_uvloop and try_windows_loop, not only the policy call. nit: narrow with message= if tests still pass.

r266-tech:
Good catch, thanks. The filter is now narrowed to `message=` so it only matches the policy-API deprecation messages (`get/set_event_loop_policy` and the `*EventLoopPolicy` class aliases) instead of every DeprecationWarning raised from asyncio.

Verified on CPython 3.14.4: the policy warnings are suppressed — including `asyncio.AbstractEventLoopPolicy`, which uvloop's policy instantiation touches — while unrelated asyncio deprecations (e.g. `asyncio.iscoroutinefunction`) still surface. `tests/test_server_loop.py` passes.

Also rebased onto current `main`: the earlier red CI came from a bandit-baseline mismatch in the `security` env (pre-existing B105 findings unrelated to this diff), which is resolved on `main` now.

DSeaStar:
Not a maintainer, just triaging. FWIW this is the narrowest of the three open PRs for #3129 — a message-scoped filter rather than the blanket `simplefilter("ignore", DeprecationWarning)` used by #3173 — and it is the only one based on current `main`.

One gap: #3137 also changes `sanic/app.py:416` from `asyncio.get_event_loop_policy().get_event_loop()` to `asyncio.get_event_loop()`. That is a real part of the fix, not a cosmetic swap. I checked on 3.14.3 and 3.13:

- 3.14.3, old: `DeprecationWarning: 'asyncio.get_event_loop_policy' is deprecated and slated for removal in Python 3.16`, then `RuntimeError: There is no current event loop in thread 'MainThread'`.
- 3.14.3, new: same `RuntimeError`, no warning.
- 3.13: both return a loop.

So folding that line in removes a warning at no behaviour cost. (#3173's loop.py hunk duplicates this PR.)

r266-tech:
Thanks for pointing out the remaining `app.loop` fallback, and for comparing #3137. I have added that change in c9244c88: the fallback now calls `asyncio.get_event_loop()` directly. The message-scoped filtering around policy setup remains as before.

The added tests exercise an existing, non-running loop and an explicitly unset loop, with deprecation warnings treated as errors and loop state isolated in a worker thread. Both failed on the previous head under Python 3.14 and now pass. The complete `test_app.py` and `test_server_loop.py` suites pass on Python 3.13.15 and 3.14.7 (60 passed, 2 skipped on each); Ruff check/format and slotscheck also pass.

DSeaStar:
Re-checked `c9244c88` locally on **Windows 11 / CPython 3.14.3** (worktree on `5ffc7b37`).

**1. The new test fails for the right reason.** With the full patch, `test_app_loop_without_running_loop[True]` and `[False]` both pass. Reverting *only* `sanic/app.py` (leaving the `loop.py` filter in place) makes both fail with:

```
DeprecationWarning: 'asyncio.get_event_loop_policy' is deprecated and slated for removal in Python 3.16
```

so the test is genuinely pinned to the `app.loop` fallback rather than to the filter.

**2. I can confirm the Windows half that a Linux runner cannot.** On 3.14.3, `try_windows_loop()` emits **4** DeprecationWarnings before the patch — `asyncio.get_event_loop_policy`, `asyncio.WindowsSelectorEventLoopPolicy` (twice: the `isinstance` check and the constructor) and `asyncio.set_event_loop_policy` — and **0** after it. Your message regex covers all three spellings, including the class alias.

**3. No behavioural change on 3.14.** With no loop set, `asyncio.get_event_loop()` raises `RuntimeError: There is no current event loop in thread 'MainThread'.` and emits no DeprecationWarning — the same outcome as the old `get_event_loop_policy().get_event_loop()` path. So the `app.py` swap is warning-cleanup only.

**4. Scope check.** `grep -rn "event_loop_policy\|EventLoopPolicy" sanic/` returns only `sanic/server/loop.py:65,67,89,90,92,93`, all inside the two `catch_warnings()` blocks — no policy call site is left outside the filter.

**5. CI.** The only real failure on `c9244c88` is `Linux / Python 3.10 / tox -e type-checking`, and its log has exactly one error line — `sanic/pages/error.py:43: error: Incompatible types in assignment` — which is pre-existing on `main` and untouched here (#3190 fixes it). Every `tox -e py3xx` job is `cancelled` by fail-fast, so **your new tests still have never actually executed in CI**; the local run above is the only evidence either way.

**6. On the three PRs for #3129.** #3146 now looks like the superset: it is based on current `main` (`5ffc7b37`) and includes the `app.py` change that #3137 also makes (#3137 is still on `785d77f8`, 30 commits behind). #3173 uses a blanket `simplefilter("ignore", DeprecationWarning)` instead of a message-scoped filter, and its second file (`handler_types.py`) is the #3159 fix, which duplicates #3161. Merging #3146 + #3161 would close both #3129 and #3159 without overlap.

Forward-looking nit, not a blocker: when the policy API is actually removed in 3.16 the `Sanic.loop` fallback will need something other than `asyncio.get_event_loop()`; probably worth a separate issue.

r266-tech:
Thanks for the Windows verification and the regression check against the previous app.py; that adds coverage I could not exercise locally.

One correction to the CI boundary: the separate [Coverage check run](https://github.com/sanic-org/sanic/actions/runs/34689332080) did execute the suite on Python 3.12 (1916 passed; `test_app.py` passed and `test_server_loop.py` reported two passes/two skips). Its three failures are in `test_asgi.py`: importing `ServerProtocol` from `websockets.server` fails, with the third test then also missing its expected warning. The Tests matrix was indeed cancelled after the `sanic/pages/error.py:43` typing error tracked by #3190. That coverage run does not establish Python 3.14 hosted coverage, so your Windows 3.14 result and the previously reported local 3.13/3.14 runs remain the relevant evidence for the deprecation behavior.

The Python 3.16 policy removal needs a separate migration; this PR only addresses the warnings while preserving current loop behavior.

DSeaStar:
You are right, and thanks for the correction — I pulled the log myself and I am retracting the wrong part of my previous comment.

**Retraction.** "Your new tests still have never actually executed in CI" was wrong as written. It holds only for the `Tests` matrix, where every `tox -e py3xx` job is `cancelled` by fail-fast after `tox -e type-checking` fails. The `Coverage check` workflow does run pytest end-to-end, so the suite was exercised.

**Verified from the log** of run 34689332080 (head `c9244c88`, job `Check coverage` 103541724000):

```
= 3 failed, 1916 passed, 9 skipped, 3 xfailed, 6 xpassed, 1364 warnings in 182.89s (0:03:02) =
tests/test_app.py ...................................................... [  3%]   # 54 passed
tests/test_server_loop.py ss..                                           [ 68%]   # 2 passed, 2 skipped
```

Both statements in your comment reproduce exactly.

**The three remaining failures are environmental, not yours.** All three are one root cause:

```
tests/test_asgi.py:117: ImportError: cannot import name 'ServerProtocol' from 'websockets.server'
    (.tox/coverage/lib/python3.12/site-packages/websockets/server.py)
FAILED tests/test_asgi.py::test_listeners_triggered
FAILED tests/test_asgi.py::test_listeners_triggered_async
FAILED tests/test_asgi.py::test_non_default_uvloop_config_raises_warning
```

The frozen environment in that job resolves `uvicorn==0.52.4`, `websockets==10.4`, `pytest-sanic==1.9.1`: `pytest-sanic` pins `websockets<11` (resolving to 10.4), while uvicorn >= 0.50's sansio implementation needs `websockets>=13`. Nothing conflicts at install time, so it only detonates at import.

**Two open PRs already turn this green, and neither touches your change:**
- **#3190** annotates `BasePage.HEADING` — on its own run 33940654343 all five `tox -e type-checking` jobs are `success`, leaving only the same `py311-no-ext` failure. Merged, it stops the `Tests` matrix from being cancelled at all.
- **#3191** pins `uvicorn<0.50` in the test extras — its `Coverage check` came back `success` on 2026-09-06, the first green result in that workflow's recent history.

So the pieces exist; none of the three failures here need anything from this PR.

**Fully agree with your last point**: that job is Python 3.12 only, so it says nothing about 3.13/3.14. The Windows 3.14 measurement (4 DeprecationWarnings before the patch, 0 after: `get_event_loop_policy`, `WindowsSelectorEventLoopPolicy` twice, `set_event_loop_policy`) and your own 3.13/3.14 runs remain the only evidence for the deprecation behaviour itself. Nothing further from me on this PR.

PR #3149, opened 2026-04-07
Replace security-critical assert statements with explicit exceptions

ChihweiLHBird:
- Replace `assert` with `ValueError` in `headers.py` for proxy count validation and `app.py` for duplicate blueprint detection
- `assert` statements are stripped when Python runs with `-O`, silently disabling these checks in production


Tronic:
The proxies count check should be reritten as `not (count > 0)` to preserve functionality, at least if we care about NaNs, which the check in this PR would not detect (both `NaN > 0` and `NaN <= 0` read False). Other than that, looks like ValueError is the proper handling.

Maybe the blueprint one should be KeyError?

Other than that, looks good to me.

Tronic:
I see `Config.PROXIES_COUNT` is `int | None`, so NaN should not be a concern here. The previous line however checks that it is truthy (not None, not 0), so perhaps simpler to `count and count > 0` on that same if? This would make negative values have the same effect than 0 or if not set, and in this case it is logical that a smaller than zero value still allows no proxy headers.

ChihweiLHBird:
> I see `Config.PROXIES_COUNT` is `int | None`, so NaN should not be a concern here. The previous line however checks that it is truthy (not None, not 0), so perhaps simpler to `count and count > 0` on that same if? This would make negative values have the same effect than 0 or if not set, and in this case it is logical that a smaller than zero value still allows no proxy headers.

So that means negative number that's previously rejected is now accepted and treated as unset?

Tronic:
> > I see `Config.PROXIES_COUNT` is `int | None`, so NaN should not be a concern here. The previous line however checks that it is truthy (not None, not 0), so perhaps simpler to `count and count > 0` on that same if? This would make negative values have the same effect than 0 or if not set, and in this case it is logical that a smaller than zero value still allows no proxy headers.
> 
> So that means negative number that's previously rejected is now accepted and treated as unset?

Rejection of negative values happens in sanic/mixins/startup.py:653. I am not sure we even want to repeat that here, so assert could be most appropriate. In particular, one might find the almost unreachable ValueError with a similar message we print elsewhere just confusing. A negative value in any case should not produce security issues, even if one would set it negative after the startup runs, right?

I note that currently it treats `SANIC_PROXIES_COUNT=""` and 0 as no proxy config, and only rejects explicit negative values with
```
ValueError: PROXIES_COUNT cannot be negative. https://sanic.readthedocs.io/en/latest/sanic/config.html#proxy-configuration
```

While other non-int values cause crash:
```
╭ ⚠️  Uncaught TypeError
│ sanic:10:                                               main()➤
│ sanic/__main__.py:25        main:                       maybe_handle_startup_error(e)➤
│ sanic/startup/errors.py:18  maybe_handle_startup_error: raise exc➤
│ sanic/__main__.py:23        main:                       cli.run(args)➤
│ sanic/cli/app.py:264        SanicCLI.run:               app.prepare(…)➤
│ sanic/mixins/startup.py:454 Sanic.prepare:              self._helper(…)➤
│ sanic/mixins/startup.py:648 Sanic._helper:             
│   """Helper function used by `run` and `create_server`."""
│   if self.config.PROXIES_COUNT and self.config.PROXIES_COUNT < 0: 💣
│       raise ValueError(
│           "PROXIES_COUNT cannot be negative. "
╰ TypeError: '<' not supported between instances of 'str' and 'int'
```

This handling obviously is NOT appropriate, and might be separately addressed (giving similar `ValueError` as above, possibly even using the same message for both to avoid code bloat). Possibly even the config system should directly raise in case the value does not match typing, but this also is another discussion and a far larger matter.

I have reviewed other parts of this PR. Implement whichever solution you deem best for proxies count and I will approve it in full.

ChihweiLHBird:
@Tronic Thanks, and sounds good to me! Does [this](https://github.com/sanic-org/sanic/pull/3149/commits/aa86ccc4778e21dffaec4a0dda767a0da71cd887) work?

Tronic (approved):

DSeaStar:
Not a maintainer, just triaging. This one looks close — `@Tronic` said on 07-12 he'd "approve it in full" once the proxies-count part was settled, and the current head `aa86ccc4` does implement his suggestion. The only thing standing between this and approval is that **the test in this PR currently fails**.

**1. `test_duplicate_blueprint` fails as written**

`app.py:928` raises `KeyError`, but `tests/test_blueprints.py:887` still expects `ValueError`. `KeyError` is not a subclass of `ValueError`, so it escapes:

```
E  KeyError: "A blueprint with the name 'bp' is already registered. Blueprint names must be unique."
sanic/app.py:928: KeyError
```

**2. Switching the test to `pytest.raises(KeyError)` is not enough**

`str(KeyError(msg))` returns the *repr* of the message, so it gains an extra pair of quotes:

```
E  assert '"A blueprint...t be unique."' == 'A blueprint ...st be unique.'
```

Minimal fix that I verified passes (Windows 11 / CPython 3.14.3, worktree on `main` 5ffc7b37):

```python
with pytest.raises(KeyError) as excinfo:
    app.blueprint(bp1)

assert str(excinfo.value.args[0]) == (
    f"A blueprint with the name '{bp_name}' is "
    "already registered. Blueprint names must be unique."
)
```

With that → `1 passed`. (`raise ValueError` instead of `KeyError` and leaving the test alone works equally well, if you prefer the type you originally used.)

**3. The change itself is worth having — verified the `-O` hole**

On `main`, the duplicate-name guard is an `assert`, so it disappears under `python -O`:

```
# main, python -O
NO-ERROR: duplicate name silently accepted
  blueprints keys: ['bp']   _blueprint_order: ['bp']

# with this PR, python -O
RAISED: KeyError -> "A blueprint with the name 'bp' is already registered. ..."
```

**4. No regressions**

`tests/test_blueprints.py` patched vs. `main` baseline: both **42 passed / 4 failed** (the 4 are pre-existing Windows failures). `tests/test_config.py` + `tests/test_requests.py`: both **171 passed / 8 failed**, identical sets.

For the record, the `headers.py` part (`if not addr and proxies_count and proxies_count > 0:`) matches what Tronic asked for; negative `PROXIES_COUNT` is already rejected earlier at `sanic/mixins/startup.py:653`, so the extra guard here is redundant rather than load-bearing.

PR #3151, opened 2026-04-07
refactor: Simplify trivial code patterns

ChihweiLHBird:
- Use `!=` instead of `not x == y` in websocket `connection_lost`
- Use `not expr` instead of `False if expr else True` in CLI arg parsing
- Use tuple literals instead of `tuple([...])` for `CancelledErrors`
- Return boolean expression directly instead of `if ...: return True; return False` in `WorkerProcess._add_config`


Tronic (approved):

DSeaStar:
Not a maintainer, just triaging. This is a pure no-op refactor, and I verified each of the four hunks is behaviour-preserving (Python 3.14.3, Windows; head `a2554d56` applied on `5ffc7b37`):

**1. `cli/app.py`** — `False if arg.startswith("--no-") else True` → `not arg.startswith("--no-")`. Both versions return a strict `bool` (not merely truthy), identical for every arg form I tried (`--no-bar`, `--bar`, `--debug`, `--no-access-log`, `--foo=1`, `--no-x=2`). The strictness matters: the value lands in the dict sent to the inspector as JSON.

**2. `compat.py`** — `tuple([...])` → `(...)`: the result is the same tuple, and the new module-level `CancelledErrors: tuple[type[BaseException], ...]` is an *annotation-only* statement, so it creates no binding (confirmed: `compat.__annotations__["CancelledErrors"]` is populated while the value still comes from the `if/else` branches). I also re-checked the only consumer, `sanic/request/types.py:407`:

```python
except CancelledErrors:   # still catches asyncio.CancelledError, still does NOT swallow KeyError
```

**3. `server/websockets/impl.py`** — `not x == CLOSED` → `x != CLOSED`. Safe here because `ws_proto.state` is `websockets.connection.State`, whose `__ne__` resolves to `int.__ne__`, i.e. exactly the negation of `==` (verified in both the `OPEN` and `CLOSED` cases). It would only differ for a type with a hand-written `__ne__`.

**4. `worker/process.py`** — the `if … return True / return False` collapse of `_add_config`: I compared old vs new over 8 signature shapes (plain params, `**kwargs`, `*args`, no params, `config=None`, keyword-only `config`) — identical results, including the `VAR_KEYWORD` case that drives whether `config` is injected at `process.py:137`.

**One caveat about the green checks.** `a2554d56` sits exactly on today's `main` tip (`5ffc7b37`) and shows **28/28 green — but every run is dated 2026-05-31**. `main` has not moved since, yet recent PRs based on that same tip (e.g. #3146, run 2026-09-12) fail `Linux / Python 3.10 / tox -e type-checking` on `sanic/pages/error.py:43`. The difference is not your patch: `mypy` is unpinned (`setup.py:143` is just `"mypy"`, and `[testenv:type-checking]` in `tox.ini` declares no `deps`), so every CI run installs whatever `mypy` is current and a newer release now flags that assignment (#3190 fixes it). Practical consequence: a re-run today would go red for a reason that has nothing to do with this PR, so don't treat the old green as a fresh signal.

**Merge-order notes** — other open PRs touching the same files, all in different regions so a textual merge should stay clean:
- `sanic/server/websockets/impl.py`: #3184 (hunks at ~810-857) and #3157 (hunks at ~553-636) — this patch is at ~860.
- `sanic/worker/process.py`: #3145.
- `sanic/cli/app.py`: #3167, #3127, #2881, #2339.

Nothing to change from my side; this is ready whenever a maintainer picks it up.

PR #3154, opened 2026-04-08
Fix #3131: SSL configuration respected for non-primary apps

### LINKED ISSUE #3131: SSL configuration ignored for non-primary apps

MuraveyApp:
## Summary
- Fixes SSL configuration being ignored for non-primary apps when serving multiple apps
- Non-primary apps now correctly use their SSL config instead of falling back to HTTP

## Test plan
- Verify multi-app serving with SSL works for all apps
- Existing test suite passes

Generated by [OwlMind](https://owlmind.dev) — 96.67% on SWE-bench Lite

### DIFF
--- a/sanic/mixins/startup.py
+++ b/sanic/mixins/startup.py
@@ -1098,11 +1098,14 @@ class StartupMixin(metaclass=SanicMeta):
             for app in apps:
                 kwargs["server_info"][app.name] = []
                 for server_info in app.state.server_info:
+                    ssl_conf = server_info.settings.get("ssl")
                     server_info.settings = {
                         k: v
                         for k, v in server_info.settings.items()
                         if k not in ("main_start", "main_stop", "app", "ssl")
                     }
+                    if isinstance(ssl_conf, SanicSSLContext):
+                        server_info.settings["ssl"] = ssl_conf.sanic
                     kwargs["server_info"][app.name].append(server_info)
 
             ssl = kwargs.get("ssl")
--- a/sanic/worker/serve.py
+++ b/sanic/worker/serve.py
@@ -75,6 +75,9 @@ def worker_serve(
                     for info in server_info_objects:
                         if not info.settings.get("app"):
                             info.settings["app"] = a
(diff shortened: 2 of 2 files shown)


DSeaStar:
Not a maintainer, just triaging. The `.sanic` approach in `startup.py` looks right — it is what `main` already does for the primary, and a dict survives pickling where a raw `SSLContext` does not (see my note on #3140).

Two gaps:

1. The restore is gated on `isinstance(ssl_conf, SanicSSLContext)`. `_helper()` stores `process_to_context(ssl)`, which returns a plain `ssl.SSLContext` unchanged when the app was started with `app.run(ssl=<SSLContext>)` or `create_context()`. In that case the `k not in (..., "ssl")` filter drops it and nothing restores it, so #3131 would still reproduce for the ordinary explicit-context case.

2. In `worker/serve.py` the new condition also fires when `a.certloader_class is not CertLoader`, even if `info.settings.get("ssl")` is None — so a custom certloader gets invoked with `{}` for every server_info.

Also worth knowing: #3140 targets the same #3131 fix.

PR #3155, opened 2026-04-10
Fix TLS redirect doc example to use `startup()`

### LINKED ISSUE #2832: Outdated docs on http to https redirection

### Describe the bug

I needed to set up http-to-https redirection for my app and came across this:

https://sanic.dev/en/guide/how-to/tls.html#redirect-http-to-https-with-certificate-requests-still-over-http

But it fails because `is_running` isn't a member of `Sanic`. I changed it to instead use `app.state.is_running` and then the second app would start ok, but my catchall route (or any route) could be found. I then replaced the `signalize` and `finalize` lines with:
```py
  await server.startup()
```

and now everything works.

### Code snippet

My final runner function that works looks like this:
```py
        async def _RunServer(app, server):
            app.state.is_running = True
            try:
                await server.startup()
                await server.serve_forever()
            finally:
                app.state.is_running = False
                app.state.is_stopping = True
```

### Expected Behavior

I think the docs need to be updated to something like the above changes, but maybe even this approach is outdated? I found #2348 but am not sure if that's what I should be doing instead. Thanks!

### How do you run Sanic?

Sanic CLI

### Operating System

Linux

### Sanic Version

23.6.0

### Additional context

_No response_

ChihweiLHBird:
Replace manual `app.signalize()`, `app.finalize()`, and `app.state.is_started = True` with `await app_server.startup()` in the HTTP-to-HTTPS redirect documentation

The `create_server` docstring explicitly recommends using `await app_server.startup()` for app initialization. The manual `signalize()`/`finalize()` approach skips steps that `_startup()` performs — extension setup, route duplicate checking, uvloop conflict detection, and touchup optimizations. While the manual approach works for trivial cases, `startup()` is the correct API and safer to recommend in docs where users may adapt the pattern for more complex apps.

Resolves #2832 (remaining work after #2864)

### DIFF
--- a/guide/content/en/guide/how-to/tls.md
+++ b/guide/content/en/guide/how-to/tls.md
@@ -158,9 +158,7 @@ async def stop(app, _):
 async def runner(app, app_server):
     app.state.is_running = True
     try:
-        app.signalize()
-        app.finalize()
-        app.state.is_started = True
+        await app_server.startup()
         await app_server.serve_forever()
     finally:
         app.state.is_running = False


SaidBySolo (approved):

DSeaStar:
I looked at this one — it has been sitting since April with no human review — and I believe the change is correct. I verified it empirically on v25.12.1 (Python 3.14.3 / Windows 11) by running the old and new recipes side by side, one variant per process.

### What the old snippet actually skips

`AsyncioServer.startup()` (`sanic/server/async_server.py:32`) is just `return self.app._startup()` (`sanic/app.py:2361`). The old `signalize()` / `finalize()` / `state.is_started = True` trio covers only 2 of the things `_startup()` does. Measured after each recipe:

| | old snippet | `await app_server.startup()` |
| --- | --- | --- |
| `config.TOUCHUP` | `<Default>` — never resolved | `True` |
| `TouchUp.run()` executed | no | yes |
| `Sanic._check_uvloop_conflict()` (`app.py:2393`) | skipped | run |
| duplicate route-name check (`app.py:2378-2391`) | skipped | run |

`TouchUp.run()` (`app.py:2400`) is what rewrites the nine hot-path methods declared via `__touchup__`: `Sanic.handle_request` / `handle_exception` / `_run_response_middleware` / `_run_request_middleware` (`app.py:172`), `Http1.http1_request_header` / `http1_response_header` / `read` (`sanic/http/http1.py:48`), and `HttpProtocol.send` / `connection_task` (`sanic/server/protocols/http_protocol.py:94`). Without it, every `dispatch()` to an unregistered event stays in place (`sanic/touchup/schemes/ode.py`) and the `alt-svc` branch is never pruned (`altsvc.py`) — i.e. the example silently runs the un-optimized build. That is the practical reason to prefer the new line.

Also worth saying out loud: `app.state.is_started = True` is redundant, `_startup()` sets it itself (`app.py:2402`), and it is exactly the flag `AsyncioServer._serve()` guards on (`async_server.py:81`). So the old snippet does not *avoid* that guard, it satisfies it by hand while skipping the work behind it. For contrast, doing neither raises `SanicException: Cannot run Sanic server without first running await server.startup()`.

For reference, this is also what Sanic itself does: `_run_server()` calls `await server_info.server.startup()` (`sanic/mixins/startup.py:1408`).

### A related doc bug you may want to fold in

`create_server()`'s own docstring tells the reader to use `await app.startup()` (`sanic/mixins/startup.py:494`), but there is no public `Sanic.startup()` — only `AsyncioServer.startup()` (`async_server.py:32`) and the private `Sanic._startup()`. The two doc sources currently contradict each other; with your change `tls.md` becomes the correct one.

### CI

`Tests` is red only because of `Linux / Python 3.11 / tox -e security` (run `24220724015`, 2026-04-10); every other job was cancelled by fail-fast, and `Coverage check` + `CodeQL` are green. That job's log has expired (GitHub returns 410). `tox -e security` is green today on a current PR (e.g. #3146, 2026-09-12) for all five Python versions, so a rebase onto current `main` and a re-run should clear it.

Optionally: if the example should emit lifecycle events too, `AsyncioServer.before_start()` / `after_start()` are the other two calls Sanic makes right after `startup()` (`startup.py:1408-1410`).

PR #3167, opened 2026-06-21
fix: restore Windows compatibility for v25.12.0+

### LINKED ISSUE #3126: v25.12.0 no longer works on Windows

BB0813:
## Fix Windows compatibility for sanic 25.12.0+

Fixes #3126

### Problem

v25.12.0 introduced three issues that prevent Sanic from running on Windows:

1. `sanic/worker/daemon.py`: Top-level `import grp` and `import pwd` crash on Windows (these modules do not exist on Windows)
2. `sanic/cli/app.py`: `self.args.unix` raises `AttributeError` on Windows because the `--unix` argument is conditionally not added on Windows (by design in `arguments.py`)
3. `sanic/cli/arguments.py`: The `--unix` argument is correctly excluded on Windows (no change needed here)

### Changes

- **`sanic/worker/daemon.py`**: Wrap `import grp` and `import pwd` in `if not OS_IS_WINDOWS:` conditional, since these are Unix-only modules and the `Daemon` class already raises `DaemonError` on Windows
- **`sanic/cli/app.py`**: Use `getattr(self.args, "unix", "")` instead of `self.args.unix` to safely handle the missing attribute on Windows

### Testing

- All 60 CLI tests pass on Windows (`pytest tests/test_cli.py`)
- All daemon tests correctly skip on Windows (117 skipped)
- `from sanic.worker.daemon import Daemon` succeeds on Windows (previously crashed)
- `python -m sanic --version` works on Windows (previously crashed)

### DIFF
--- a/sanic/cli/app.py
+++ b/sanic/cli/app.py
@@ -476,7 +476,7 @@ Advanced daemon management:
             "noisy_exceptions": self.args.noisy_exceptions,
             "port": self.args.port,
             "ssl": ssl,
-            "unix": self.args.unix,
+            "unix": getattr(self.args, "unix", ""),
             "verbosity": self.args.verbosity or 0,
             "workers": self.args.workers,
             "auto_tls": self.args.auto_tls,
--- a/sanic/worker/daemon.py
+++ b/sanic/worker/daemon.py
@@ -1,9 +1,7 @@
 from __future__ import annotations
 
 import atexit
-import grp
 import os
-import pwd
 import signal
 import sys
 import time
@@ -14,6 +12,10 @@ from pathlib import Path
 
 from sanic.compat import OS_IS_WINDOWS
 from sanic.log import logger
+
+if not OS_IS_WINDOWS:
+    import grp
+    import pwd
 
 
 try:
--- a/sanic/worker/daemon.py
+++ b/sanic/worker/daemon.py
@@ -13,10 +13,16 @@ from pathlib import Path
 from sanic.compat import OS_IS_WINDOWS
 from sanic.log import logger
 
-if not OS_IS_WINDOWS:
-    import grp
-    import pwd
 
+try:
+    import grp  # type: ignore
+except (ImportError, ModuleNotFoundError):  # pragma: no cover
+    grp = None  # type: ignore
+
+try:
+    import pwd  # type: ignore
+except (ImportError, ModuleNotFoundError):  # pragma: no cover
+    pwd = None  # type: ignore
 
 try:
     import fcntl  # type: ignore


DSeaStar:
Not a maintainer, just triaging. I verified this on real Windows 11 / Python 3.14.3 against `main` (5ffc7b37), and it works.

**Unpatched:** `import sanic.worker.daemon` and `python -m sanic --version` both die with `ModuleNotFoundError: No module named 'grp'`.

**Patched:** both succeed (`Sanic 25.12.1; Routing 23.12.0`), and a real app boots via the CLI.

**Both halves are independently required.** With only the `daemon.py` fix, the CLI still crashes — `sanic/cli/app.py:479` raises `AttributeError` on `self.args.unix`, because `--unix` is never added to the parser on Windows (`cli/arguments.py:169` gates it on `OS_IS_WINDOWS`). So the `getattr(..., "unix", "")` change is not defensive padding, it is the second half of the fix.

**Setting `grp`/`pwd` to `None` is safe here.** `Daemon.__init__` raises `DaemonError` on Windows at `daemon.py:136`, before any use; the only consumers are `pwd.getpwnam` at `:288` and `grp.getgrnam` at `:297`, both on Unix-only paths. `Daemon("probe")` → `DaemonError: Daemon mode is not supported on Windows...` as before.

**No regressions:** `tests/test_cli.py` 60 passed, `tests/test_daemon.py` 117 skipped.

**Duplicate:** #3127 (2026-01-06) makes the same two changes — one combined `try/except` for `grp`/`pwd` instead of two, otherwise equivalent. Two things favor this PR: #3127 is based on 785d77f8 and is now `behind`, and its third hunk bumps `sanic/__version__.py` to `25.12.1`, which `main` already has — that hunk is stale and should be dropped. Whichever one lands should probably also add `changelogs/3126.bugfix.rst`.

For context on impact: #3126 is currently blocking the Windows conda-forge package (conda-forge/sanic-feedstock#75).

PR #3161, opened 2026-05-21
Fix route handler callable typing

puneetdixit200:
Fixes #3159.

### Summary
- widen `RouteHandler` from `Coroutine` to `Awaitable` so non-decorator route registration does not trigger mypy's unused coroutine check
- update the internal handler response annotation to match the widened route handler type
- add a typing sample for registering a bound async method through `app.get(...)(handler)`

### Testing
- `python -m pytest tests\typing\test_typing.py -q`
- `python -m pytest tests\test_handler.py -q`
- `python -m ruff check sanic\app.py sanic\models\handler_types.py tests\typing\test_typing.py tests\typing\samples\route_handler_registration.py`
- `python -m ruff format --check sanic\app.py sanic\models\handler_types.py tests\typing\test_typing.py tests\typing\samples\route_handler_registration.py`
- `git diff --check`

PR #3173, opened 2026-07-20
Fix #3159: Use Awaitable instead of Coroutine for RouteHandler type annotation to avoid mypy unused-coroutine error

### LINKED ISSUE #3159: unused-coroutine mypy error when not using route handlers as a decorator

### Describe the bug

Hello,

When I use the route handlers not as a decorator I get mypy errors being sad that I'm not awaiting my async function

```
class MyClassThatHasRoutes:
    def setup_routes(self, app: sanic.Sanic[sanic.Config, SimpleNamespace]) -> None:
        app.get("/api")(self.api) # error: Value of type "Coroutine[Any, Any, HTTPResponse | None]" must be used  [unused-coroutine]

    async def api(self, request: sanic.Request) -> sanic.response.HTTPResponse:
        return sanic.empty(200)
```

I get that it's designed to be used as a decorator, but it makes sense where I'm doing this that I'm not.

I believe this could be fixed if [RouteHandler](https://github.com/sanic-org/sanic/blob/785d77f8fe208576a48d339bd81d866021154092/sanic/models/handler_types.py#L29) was changed to have a return annotation of `Awaitable[HTTPResponse | None]` rather than Coroutine.

Is that something that could be done or do you see problems with doing that?

### Code snippet

_No response_

### Expected Behavior

_No response_

### How do you run Sanic?

ASGI

### Operating System

Linux

### Sanic Version

25.12.0

### Additional context

_No response_

driphtyio:
This PR fixes the mypy unused-coroutine error that occurs when using route handlers outside of decorator usage.

## Change
Changed  type annotation from  to  in .

## Why
When route handlers are used not as decorators (e.g., `app.get("/api")(self.api)`), mypy reports:
```
error: Value of type "Coroutine[Any, Any, HTTPResponse | None]" must be used  [unused-coroutine]
```

Using `Awaitable` instead of `Coroutine` is more semantically correct since:
- `Coroutine` is a subtype of `Awaitable`
- Python type checkers treat `Awaitable` values as safe to pass to something that will await them, while `Coroutine` triggers unused-coroutine warnings when not awaited immediately

This matches the suggested fix from the issue reporter.

## Testing
All 161 route tests pass.

Fixes #3159

### DIFF
--- a/sanic/models/handler_types.py
+++ b/sanic/models/handler_types.py
@@ -1,5 +1,5 @@
 from asyncio.events import AbstractEventLoop
-from collections.abc import Coroutine
+from collections.abc import Awaitable, Coroutine
 from typing import Any, Callable, TypeVar
 
 import sanic
@@ -26,5 +26,5 @@ ListenerType = (
     Callable[[Sanic], Coroutine[Any, Any, None] | None]
     | Callable[[Sanic, AbstractEventLoop], Coroutine[Any, Any, None] | None]
 )
-RouteHandler = Callable[..., Coroutine[Any, Any, HTTPResponse | None]]
+RouteHandler = Callable[..., Awaitable[HTTPResponse | None]]
 SignalHandler = Callable[..., Coroutine[Any, Any, None]]
--- a/sanic/server/loop.py
+++ b/sanic/server/loop.py
@@ -1,6 +1,7 @@
 import asyncio
 
 from os import getenv
+import sys
 
 from sanic.compat import OS_IS_WINDOWS
 from sanic.log import error_logger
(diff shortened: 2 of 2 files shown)


DSeaStar:
Not a maintainer, just triaging the open PRs here. This one currently duplicates two other open PRs, so it may be worth narrowing or closing:

- The `sanic/models/handler_types.py` hunk (`Coroutine` -> `Awaitable` for `RouteHandler`) is identical to #3161, which additionally fixes `sanic/app.py:1299` and adds the mypy typing test that exercises it.
- The `sanic/server/loop.py` hunk addresses #3129 (asyncio policy API deprecated in 3.14) and overlaps #3146 and #3137. Of the three, #3146 is the narrowest: it filters on the policy messages only, whereas this PR uses `warnings.simplefilter("ignore", DeprecationWarning)`, which also suppresses any unrelated DeprecationWarning raised inside those blocks.

If the typing fix is the goal, #3161 already carries it plus tests.

PR #3177, opened 2026-07-25
Fix ASGI request creation exception responses

### LINKED ISSUE #2718: ASGI exceptions outside handlers should be able to pass status
An exception raised in the lifecycle (but not handler) of a request on an ASGI-served application should be able to pass a status code.

---

>              In my experiment, raising a `BadRequest` in `asgi.py` will cause the ASGI server return 500 error, instead of 400. This seems out of the scope of this PR, but I am wondering whether we should change the raised error to `ServerError` or others before we figure out how to let ASGI server return a 400 error.

_Originally posted by @ChihweiLHBird in https://github.com/sanic-org/sanic/pull/2606#discussion_r1133448078_

deepakganesh78:
## Summary
- route HTTP exceptions raised during ASGI request setup through Sanic's normal exception handler
- preserve status codes from invalid headers, request construction, and `http.lifecycle.request` signals
- use an empty fallback request when construction itself fails, without reflecting malformed headers or paths
- leave WebSocket exception behavior unchanged

Fixes #2718.

## Tests
- `python -m pytest tests/test_asgi.py -q` (26 passed)
- `ruff check sanic/asgi.py tests/test_asgi.py`
- `ruff format sanic/asgi.py tests/test_asgi.py --check`
- `python setup.py check -r -s`

`mypy sanic/asgi.py` was also attempted locally, but the Windows run has existing unrelated failures for missing `types-ujson` stubs and Unix-only APIs such as `AF_UNIX`, `SIGKILL`, and `sched_getaffinity`.

The GitHub Actions type-checking job likewise currently fails in unchanged `sanic/pages/error.py` with mypy 2.3.0; the workflow's fail-fast behavior then cancels the remaining test matrix jobs. Lint and security jobs completed successfully.

### DIFF
--- a/changelogs/2718.bugfix.rst
+++ b/changelogs/2718.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ASGI request creation exceptions to preserve their HTTP status codes.
--- a/sanic/asgi.py
+++ b/sanic/asgi.py
@@ -186,7 +186,7 @@ class ASGIApp:
                 raise
             if instance.exception is None:
                 instance.exception = e
-            instance.request = request_class(
+            instance.request = Request(
                 b"*",
                 Header(),
                 version,
--- a/sanic/asgi.py
+++ b/sanic/asgi.py
@@ -112,6 +112,7 @@ class ASGIApp:
     ws: WebSocketConnection | None
     stage: Stage
     response: BaseHTTPResponse | None
+    exception: Exception | None
 
     @classmethod
     async def create(
@@ -128,6 +129,7 @@ class ASGIApp:
         instance.transport.loop = sanic_app.loop
         instance.stage = Stage.IDLE
         instance.response = None
+        instance.exception = None
         instance.sanic_app.state.is_started = True
         setattr(instance.transport, "add_task", sanic_app.loop.create_task)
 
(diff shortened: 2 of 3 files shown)


DSeaStar:
Not a maintainer. This PR and #3186 both target #2718, one layer apart — I checked locally, they are complementary rather than duplicates.

- Here: the request-creation / `http.lifecycle.request` exception is stored on the instance and re-raised inside `__call__`, so `handle_exception()` builds a real response.
- #3186: only the last-resort branch at `asgi.py:260`, when `handle_exception()` itself raises.

Local A/B (Python 3.14, base 5ffc7b37, pytest-sanic 1.9.1):

- On main, a `SanicException(418)` raised from a `http.lifecycle.request` signal, and a `SanicException(422)` raised by a failing `request_class`, both escape `ASGIApp.create()` entirely — there is no response at all. With this patch they come back as real 418 / 422 responses. `test_asgi_headers_decoding` likewise turns from `pytest.raises(BadRequest)` into a 400.
- This patch alone does **not** cover #3186's case: if `handle_exception()` itself raises, the exception still escapes.
- Applying both on top of each other is clean, no textual conflict; `tests/test_asgi.py` is 26/26 green with both applied. I'd suggest keeping both.

The red CI is not caused by this change: the only failing job is `Linux / Python 3.10 / tox -e type-checking` from 2026-07-25, and both errors are in `sanic/pages/error.py:15` and `:43` — a file this PR never touches. `:43` is the `HEADING` annotation fixed by #3190; `:15` does not reproduce under the current toolchain.

PR #3180, opened 2026-07-29
Fix secondary app signals in spawned workers

### LINKED ISSUE #3179: [Bug] Secondary apps lack built-in HTTP signals (NotFound error) when receiving requests in spawn / dev mode

### Describe the bug

When running multiple Sanic applications using `app.prepare()` and `Sanic.serve(primary=app)` under `dev=True` (or any configuration that forces the `spawn` multiprocessing context), secondary applications fail to register their built-in HTTP signals in the spawned worker process.

When an HTTP request hits the secondary app's port, the worker process crashes immediately with a `sanic_routing.exceptions.NotFound` error because the underlying signal router is completely empty.

### To Reproduce

Here is a Minimal Reproducible Example (MRE). 
I've included a custom managed worker process (via `app.manager.manage`) that automatically sends an HTTP request to the secondary app a few seconds after startup to easily trigger the bug.

```python
import time
from urllib import request
from sanic import Sanic

app = Sanic("PrimaryApp")
app_secondary = Sanic("SecondaryApp")

@app.get("/")
async def primary_handler(req):
    return text("Primary OK")

@app_secondary.get("/")
async def secondary_handler(req):
    return text("Secondary OK")

# A custom background process to automatically trigger the bug
def custom_worker_process():
    print("--- Wait 3 seconds for servers to start...")
    time.sleep(3)
    print("--- Sending request to secondary app...")
    try:
        # This will INSTANTLY trigger the NotFound crash in the server process!
        response = request.urlopen("http://127.0.0.1:8001/")
        html_content = response.read().decode('utf-8')
        print(f"--- 💥 Request SUCCESS! Response content: '{html_content}'")
    except Exception as e:
        print(f"--- ❌ Request failed due to server crash: {e}")

@app.main_process_ready
async def attach_custom_worker(app_: Sanic, _):
    # Spawn the custom process
    app_.manager.manage("trigger-bot", custom_worker_process, {}, workers=1)

if __name__ == "__main__":
    app.prepare(port=8000, dev=True)
    app_secondary.prepare(port=8001, dev=True)
    
    Sanic.serve(primary=app)

```

**Steps:**

1. Run the script above.
2. Wait 3 seconds for the `trigger-bot` process to send the request to port 8001.

**Traceback / Output:**
Instead of a successful response, the background worker reports a 500 error:

```text
--- Wait 3 seconds for servers to start...
--- Sending request to secondary app...
--- ❌ Request failed due to server crash: HTTP Error 500: Internal Server Error

```

And the Sanic server console raises the following exception:

```text
sanic_routing.exceptions.NotFound: Could not find signal http.lifecycle.begin
...
sanic_routing.exceptions.NotFound: Could not find signal http.lifecycle.complete

```

### Root Cause Analysis

I dug into the source code (`sanic/mixins/startup.py` and `sanic/worker/serve.py`) and found the missing link during the worker spawn lifecycle:

1. In `worker_serve`, only the primary app is passed to `_serve_http_1()`, which eventually triggers `app._setup_server()` and `app.register_builtins()`. Thus, the primary app gets its full signal tree.
2. Secondary apps are launched via the `app._start_servers` hook attached to `before_server_start`.
3. Inside `_run_server` (which is called by `_start_servers` for secondary apps), the framework executes `await server_info.server.startup()`, but **it never calls `register_builtins()` or `_setup_server()` for these secondary apps.** As a result, the `signal_router` for non-primary apps remains completely empty in the spawned process.

### Current Workaround

To bypass this issue without patching the framework itself, I currently have to inject dummy handlers into the secondary app in the global scope before `__main__`. This forces the worker process to compile the required signal tree:

```python
def patch_sanic_secondary_app(app: Sanic) -> Sanic:
    """
    Workaround: Pre-register dummy handlers for core HTTP signals
    to prevent NotFound crashes in spawn mode.
    """
    async def _dummy_signal_handler(*args, **kwargs):
        pass

    core_events = [
        "http.lifecycle.begin", "http.lifecycle.read_head", "http.lifecycle.request",
        "http.lifecycle.handle", "http.lifecycle.read_body", "http.lifecycle.response",
        "http.lifecycle.complete", "http.lifecycle.exception", 
        "http.routing.before", "http.routing.after",
        "http.handler.before", "http.handler.after"
    ]
    
    for event in core_events:
        try:
            app.add_signal(_dummy_signal_handler, event)
        except Exception:
            pass

    for attach in ["request", "response"]:
        try:
            app.add_signal(_dummy_signal_handler, "http.middleware.before", condition={"attach_to": attach})
            app.add_signal(_dummy_signal_handler, "http.middleware.after", condition={"attach_to": attach})
        except Exception:
            pass

    return app

# Usage:
# app_secondary = patch_sanic_secondary_app(Sanic("SecondaryApp"))

```

### Expected behavior

Sanic's worker manager should ensure that `register_builtins()` (or the equivalent initialization) is automatically executed for **all** prepared apps assigned to a worker, not just the primary app, so that secondary apps can handle HTTP requests properly in `spawn` mode.

### Environment

* **OS:** Linux / Windows (Any OS using `spawn` multiprocessing context)
* **Sanic Version:** 23.x / 24.x / 25.x (Tested on 25.12.1)

```


### Code snippet

_No response_

### Expected Behavior

_No response_

### How do you run Sanic?

Sanic CLI

### Operating System

Linux

### Sanic Version

25.12.1

### Additional context

_No response_

bm1016bm-svg:
## Summary

- preserve each registered app's runtime state when configuring spawned workers
- refresh secondary apps before their servers start
- verify dev-mode built-in HTTP signals remain available on secondary apps

## Why

In spawn mode, the primary app received its prepared state through `passthru`, but secondary apps were re-imported without their per-app mode and config. A secondary app prepared with `dev=True` could therefore start with production touchups enabled and fail to resolve built-in HTTP lifecycle signals.

## Validation

- `python -m pytest tests/worker/test_worker_serve.py tests/worker/test_startup.py -q` — 19 passed, 1 skipped
- `python -m ruff check sanic/mixins/startup.py sanic/worker/serve.py tests/worker/test_worker_serve.py`
- `python -m ruff format --check sanic/mixins/startup.py sanic/worker/serve.py tests/worker/test_worker_serve.py`
- `git diff --check`

Fixes #3179

### DIFF
--- a/sanic/mixins/startup.py
+++ b/sanic/mixins/startup.py
@@ -1083,18 +1083,26 @@ class StartupMixin(metaclass=SanicMeta):
             kwargs["app_name"] = app.name
             kwargs["app_loader"] = app_loader
             kwargs["server_info"] = {}
-            kwargs["passthru"] = {
-                "auto_reload": app.auto_reload,
-                "state": {
-                    "verbosity": app.state.verbosity,
-                    "mode": app.state.mode,
-                },
-                "config": {
-                    "ACCESS_LOG": app.config.ACCESS_LOG,
-                    "NOISY_EXCEPTIONS": app.config.NOISY_EXCEPTIONS,
-                },
-                "shared_ctx": app.shared_ctx.__dict__,
+
+            def build_passthru(current: Sanic) -> dict[str, Any]:
+                return {
+                    "auto_reload": current.auto_reload,
+                    "state": {
+                        "verbosity": current.state.verbosity,
+                        "mode": current.state.mode,
+                    },
+                    "config": {
+                        "ACCESS_LOG": current.config.ACCESS_LOG,
+                        "NOISY_EXCEPTIONS": current.config.NOISY_EXCEPTIONS,
+                    },
+                    "shared_ctx": current.shared_ctx.__dict__,
+                }
+
+            app_passthru = {
+                current.name: build_passthru(current) for current in apps
             }
+            kwargs["passthru"] = build_passthru(app)
+            kwargs["app_passthru"] = app_passthru
             for app in apps:
                 kwargs["server_info"][app.name] = []
                 for server_info in app.state.server_info:
--- a/sanic/worker/serve.py
+++ b/sanic/worker/serve.py
@@ -48,6 +48,7 @@ def worker_serve(
     version=HTTP.VERSION_1,
     config: bytes | str | dict[str, Any] | Any | None = None,
     passthru: dict[str, Any] | None = None,
+    app_passthru: dict[str, dict[str, Any]] | None = None,
 ):
     try:
         from sanic import Sanic
@@ -99,6 +100,9 @@ def worker_serve(
 
             # Run secondary servers
             apps = list(Sanic._app_registry.values())
+            for secondary in apps:
+                if secondary is not app:
+                    secondary.refresh((app_passthru or {}).get(secondary.name))
             app.before_server_start(partial(app._start_servers, apps=apps))
             for a in apps:
                 a.multiplexer = WorkerMultiplexer(
--- a/tests/worker/test_worker_serve.py
+++ b/tests/worker/test_worker_serve.py
@@ -118,7 +118,8 @@ def test_serve_passes_secondary_app_state(worker_manager: Mock, app: Sanic):
     app.prepare(dev=True)
     secondary.prepare(dev=True)
 
-    Sanic.serve(primary=app)
+    with patch("sanic.mixins.startup.configure_socket"):
+        Sanic.serve(primary=app)
 
     worker_kwargs = worker_manager.call_args.args[2]
     secondary_passthru = worker_kwargs["app_passthru"][secondary.name]
--- a/tests/worker/test_worker_serve.py
+++ b/tests/worker/test_worker_serve.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 
 from os import environ
@@ -6,6 +7,8 @@ from unittest.mock import Mock, patch
 import pytest
 
 from sanic.app import Sanic
+from sanic.application.constants import Mode
+from sanic.signals import Event
 from sanic.worker.loader import AppLoader
 from sanic.worker.multiplexer import WorkerMultiplexer
 from sanic.worker.process import Worker, WorkerProcess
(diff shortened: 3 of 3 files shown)


bm1016bm-svg:
CI failures appear unrelated to this change: the worker regression suite passes locally (11/11), while the failing type check is in sanic/pages/error.py and coverage fails in the uvicorn/websockets test environment. Recent main-branch Tests/Coverage runs are also failing. I cannot rerun the workflows without repository admin permission.

DSeaStar:
I read this against current `main` and the diagnosis holds up. I had been triaging #3179 for weeks without noticing this PR existed — that is my miss, not yours.

**The mechanism, with citations.**

- `sanic/worker/serve.py:60` calls `app.refresh(passthru)` for the **primary app only**; that dict carries `state.mode` and a few config values.
- Secondary apps are collected at `serve.py:101` from `Sanic._app_registry.values()`. In a spawned worker they are re-imported, so they get none of it, and `Sanic.state.mode` defaults to `Mode.PRODUCTION` (`application/state.py:46`). `is_debug` is only a property over `mode` (`state.py:89-96`), so it follows along.
- `app.py:2369-2375`: debug → `TOUCHUP = False`, otherwise `TOUCHUP` defaults to `True`; then `signalize(self.config.TOUCHUP)`, which assigns that to `signal_router.allow_fail_builtin` (`app.py:2354`).
- `signals.py:217-218`: on `NotFound` for a reserved event, it re-raises **only if `allow_fail_builtin` is true**.
- `http_protocol.py:158` dispatches `http.lifecycle.begin` on every connection, unconditionally.

So a secondary app that loses debug mode becomes production-mode, where an unregistered built-in signal is fatal rather than merely warned about. `refresh()` writing `state.mode` is enough to flip it, because `is_debug` is derived.

**Measured on sanic 25.12.1** (the version in the issue), in-process: a production-mode app with no handler for that event raises `NotFound: Could not find signal http.lifecycle.begin` on dispatch; the same app after `refresh({"state": {"mode": Mode.DEBUG}})` does not. That is exactly what `test_secondary_app_receives_worker_passthru` asserts, and asserting both `state.is_debug` and `config.TOUCHUP is False` is the right pair — `TOUCHUP` is the value that actually reaches the router.

**On the red CI** — two separate things, neither caused by this PR:

1. `Tests` fails at the `type-checking` job. Your 08-10 comment placed this in `sanic/pages/error.py`; the declaration that needs changing is one line in **`sanic/pages/base.py`**. On `main` that line is `HEADING = None` with no annotation (`base.py:14`), so mypy infers the type `None`, while `error.py:43` assigns `E("Application ").strong(name)(...)`, which is an `html5tagger.Builder`. mypy 2.3.1 reports `Incompatible types in assignment (expression has type "Builder", variable has type "None")` → `Found 1 error in 1 file (checked 132 source files)`. The matrix is fail-fast, so that one job cancels the rest.

   I have opened **#3190** to add that single annotation. #3188 carried the same line and was closed unmerged on 09-04, which is why nothing on `main` fixes it today. Once #3190 lands, this PR should get past the gate.

2. `Coverage check` — **corrected on 09-06, my previous note here was wrong.** It fails at `actions/upload-artifact@v4` with `No files were found with the provided path: ./coverage.xml`, and that is a knock-on effect, not a fault in the workflow: `tox -e coverage` runs `coverage run --source ./sanic -m pytest tests`, which exits 1 on

   ```text
   ImportError: cannot import name 'ServerProtocol' from 'websockets.server'
   ```

   so the `coverage xml -i` step in the same env never runs. The cause is a dependency conflict in the test extras: `pytest-sanic 1.9.1` declares `websockets>=9.1,<11.0` (so `websockets==10.4` is installed), while `uvicorn` — unpinned — resolves to `0.52.4`, and **uvicorn 0.50.0** switched `protocols/websockets/auto.py` to `websockets_sansio_impl`, which needs `ServerProtocol` from `websockets>=13`. `Coverage check` was green on 07-03 with `uvicorn==0.49.0` and red from 07-12 with `uvicorn==0.51.0`, `websockets==10.4` in both.

   I have opened **#3191** to cap `uvicorn<0.50` in `tests_require`.

   > On 09-05 I wrote that this job had been failing on every PR since 08-15 with `failed to determine base repo: failed to run git: fatal: not a git repository`, and called it a workflow bug. That was wrong: I have now checked nine runs of this workflow between 08-11 and 09-05 and not one contains that text. My apologies for sending you down the wrong path.

Happy to re-run anything here once #3190 is in.

PR #3182, opened 2026-07-29
Fix Windows socket address reuse

### LINKED ISSUE #2660: Sanic should bind host:port exclusively (no SO_REUSEADDR on Windows)

### Describe the bug

On Windows SO_REUSEADDR behaves similar to SO_REUSEADDR & SO_REUSEPORT on other systems, i.e. it allows binding to exact same address:port, provided that whoever bound it before also used this option.

The result is that the new server thinks that it is listening on a port but all requests still go to the old one (by experimentation of about ten page reloads). Only once the first started server exits will the second one start receiving any requests.

I believe MacOS shows similar preference to the first started server, when SO_REUSEPORT is used, i.e. there is absolutely no load sharing, the first one gets all the requests. I have not experimented this on Linux.

Sanic should refrain from using this type of socket sharing, and instead bind exclusively, noting that the flags behave differently on different operating systems, particularly on Windows.

### Code snippet

```sh
sanic one:app&
sanic other:app&
curl http://127.0.0.1:8000/  # Always gets one:app
```

### Expected Behavior

Do not use SO_REUSEPORT on any system, don't use SO_REUSEADDR on Windows.

### How do you run Sanic?

Sanic CLI

### Operating System

Windows

### Sanic Version

22.12

### Additional context

https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ

zyz619963502zyz:
## Summary

- avoid setting `SO_REUSEADDR` when binding TCP server sockets on Windows
- preserve the existing socket option behavior on other platforms
- add a regression test for the Windows branch

On Windows, `SO_REUSEADDR` allows another socket using the same option to bind the exact same address and port. The second server can therefore appear to start successfully while requests continue reaching the first server.

Fixes #2660.

## Validation

- `python -m pytest tests/worker/test_socket.py::test_bind_socket_does_not_reuse_address_on_windows -q` — 1 passed
- `python -m ruff check sanic/server/socket.py tests/worker/test_socket.py` — passed
- `python -m ruff format --check sanic/server/socket.py tests/worker/test_socket.py` — passed
- Windows runtime check: before the change, a second socket using `SO_REUSEADDR` could bind the same `127.0.0.1` port; after the change, the second `bind_socket` call fails with WinError 10048

Running the complete `tests/worker/test_socket.py` file on Windows also executes two pre-existing Unix socket tests. Those fail because this Python build does not expose `socket.AF_UNIX`; the new targeted regression test passes.

### DIFF
--- a/changelogs/2660.bugfix.rst
+++ b/changelogs/2660.bugfix.rst
@@ -1,0 +1,1 @@
+Do not set ``SO_REUSEADDR`` on Windows listener sockets, preventing multiple servers from binding to the same address and port.
--- a/sanic/server/socket.py
+++ b/sanic/server/socket.py
@@ -8,6 +8,7 @@ from ipaddress import ip_address
 from pathlib import Path
 from typing import Any
 
+from sanic.compat import OS_IS_WINDOWS
 from sanic.http.constants import HTTP
 
 
@@ -28,7 +29,8 @@ def bind_socket(host: str, port: int, *, backlog=100) -> socket.socket:
         )
     except ValueError:  # Hostname, may become AF_INET or AF_INET6
         sock = socket.socket()
-    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    if not OS_IS_WINDOWS:
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     sock.bind(location)
     sock.listen(backlog)
     sock.set_inheritable(True)
--- a/tests/worker/test_socket.py
+++ b/tests/worker/test_socket.py
@@ -1,10 +1,24 @@
 from pathlib import Path
+from unittest.mock import Mock, patch
 
 from sanic.server.socket import (
+    bind_socket,
     bind_unix_socket,
     configure_socket,
     remove_unix_socket,
 )
+
+
+def test_bind_socket_does_not_reuse_address_on_windows():
+    sock = Mock()
+    with (
+        patch("sanic.server.socket.OS_IS_WINDOWS", True),
+        patch("sanic.server.socket.socket.socket", return_value=sock),
+    ):
+        bind_socket("127.0.0.1", 8000)
+
+    sock.setsockopt.assert_not_called()
+    sock.bind.assert_called_once_with(("127.0.0.1", 8000))
 
 
 def test_setup_and_teardown_unix():
--- a/tests/worker/test_socket.py
+++ b/tests/worker/test_socket.py
@@ -12,7 +12,7 @@ from sanic.server.socket import (
 def test_bind_socket_does_not_reuse_address_on_windows():
     sock = Mock()
     with (
-        patch("sanic.server.socket.OS_IS_WINDOWS", True),
+        patch("sanic.server.socket.OS_IS_WINDOWS", True, create=True),
         patch("sanic.server.socket.socket.socket", return_value=sock),
     ):
         bind_socket("127.0.0.1", 8000)


zyz619963502zyz:
Hi maintainers — a gentle follow-up in case this slipped through. The patch remains limited to the Windows socket-reuse behavior in `sanic/server/socket.py` and its focused regression test in `tests/worker/test_socket.py`. The two failing workflows are the previously documented baseline failures in untouched ASGI/WebSocket and `sanic/pages/error.py` paths, also reproduced on unrelated PRs; the completed checks outside those failures passed. I would appreciate a review when convenient, and I am happy to address any requested changes. Thank you.

DSeaStar:
I verified this on real Windows (Python 3.14.3) rather than reasoning about it, and the bug is real.

**Current `main`:** two `bind_socket("127.0.0.1", P)` calls on the same port both *succeed*. I then connected and sent data — it was accepted by the **first** listener. So a second server starts "fine" and silently serves nothing, exactly as #2660 describes.

**With this patch:** the second bind is rejected with `OSError winerror=10048` (WSAEADDRINUSE).

**Regression check** (this was my main worry — dropping `SO_REUSEADDR` often breaks fast restarts). All three rebind fine without it:
- close listener, immediately rebind idle port
- restart after a served connection
- restart after a *server-initiated* close, which leaves a server-side TIME_WAIT on that port

A real app still works with the patch: `app.test_client.get("/")` → 200. `tests/worker` gives 148 passed / 6 failed vs. 147 / 6 on unpatched main — the +1 is the new test, and the 6 are pre-existing Windows `AF_UNIX`/reloader failures present on `main` too. No regressions.

Two small suggestions:

1. The new test fails for the *wrong* reason before the fix. On unpatched `main` it raises `AttributeError: module 'sanic.server.socket' does not have the attribute 'OS_IS_WINDOWS'`, because `mock.patch` won't create a name that doesn't exist yet. Adding `create=True` to that one patch makes it fail with the meaningful `AssertionError: Expected 'setsockopt' to not have been called. Called 1 times.` — worth doing so the test documents the actual behavior change.

2. Add `changelogs/2660.bugfix.rst` to match the repo's changelog-fragment convention.

The red CI is the known pre-existing `type-checking` failure in `sanic/pages/error.py` (fixed by #3190) plus the cascading coverage workflow — unrelated to this change.

zyz619963502zyz:
Thank you for the real Windows verification and the precise suggestions. I pushed commit 4d5b3cd1, which adds create=True so the regression fails on the pre-fix setsockopt call, and adds changelogs/2660.bugfix.rst. The focused regression passes locally, and Ruff check/format both pass.

DSeaStar:
Re-verified `4d5b3cd1` on Windows (Python 3.14.3):

- Your new test now fails for the right reason. With the full patch it passes; reverting only `sanic/server/socket.py` gives `AssertionError: Expected 'setsockopt' to not have been called. Called 1 times.` instead of the old `AttributeError`. `create=True` fixed it.
- Real-socket probe on this head: first `bind_socket()` OK, second one raises `[WinError 10048]` (the #2660 symptom), and rebinding right after `close()` still works — so fast restarts are unaffected.
- Scope: `SO_REUSEADDR` is set in exactly one place in `sanic/` (`socket.py:33`), so this covers the only listener path. `changelogs/2660.bugfix.rst` matches the repo convention.

One caution about the red CI: on `4d5b3cd1` the only real failure is `Linux / Python 3.11 / tox -e type-checking`, and its log is a single line — `sanic/pages/error.py:43: Incompatible types in assignment (expression has type "Builder", variable has type "None")` — which is pre-existing on `main` and fixed by #3190. Every `tox -e py3xx` job shows `cancelled` because the workflow fail-fast cancels them when type-checking fails, so your new test has still never actually run in CI. The green I'm reporting above is local only.

PR #3185, opened 2026-08-14
Stop Inspector from serving inherited app sockets after fork

### LINKED ISSUE #2897: INSPECTOR worker may handle HTTP request unexpectly when `Sanic.start_method="fork"` and `INSPECTOR=True`

### Describe the bug

When I set Sanic's `start_method` to `fork` and enable `INSPECTOR`, the HTTP requests I send may **unexpectedly** be processed by the `INSPECTOR` worker process.

Could this be because the webserver socket inherited from the parent process is not closed when forking the child process?

If there are any suggestions, or if this behavior is expected, please let me know. Thank you to all the excellent library developers.

### Code snippet

test server code:
```python
from sanic import Sanic, Request, HTTPResponse, json
import os

Sanic.start_method = "fork"

app = Sanic("test-fork-with-inspector", inspector=True)

@app.get("/test")
async def test_handler(_: Request) -> HTTPResponse:
    return json(body={"pid": os.getpid()})
```

send HTTP request:
```bash
curl -X GET http://localhost:8080/test
{"pid":1199437}  # inspector worker

curl -X GET http://localhost:8080/test
{"pid":1199435}  # nomal worker
```

Socket fd of **inspector** worker:
```bash
lsof -p 1199437 | grep LISTEN
python  1199437 user   12u     IPv4           24318245      0t0      TCP localhost:http-alt (LISTEN)
python  1199437 user   15u     IPv4           24311762      0t0      TCP localhost:6457 (LISTEN)
```

Socket fd of **normal** worker:
```bash
lsof -p 1199435 | grep LISTEN
python  1199435 tusen   12u     IPv4           24318245      0t0      TCP localhost:http-alt (LISTEN)
```

### Expected Behavior

inspector worker should not handle regular HTTP requests

### How do you run Sanic?

Sanic CLI

### Operating System

Linux

### Sanic Version

Sanic 23.12.0; Routing 23.12.0

### Additional context

_No response_

DSeaStar:
## Summary

Fixes #2897.

With `Sanic.start_method = "fork"` and `inspector=True`, the Inspector child inherits the parent `Sanic._app_registry` and the application listener sockets (`set_inheritable(True)`). Inspector then calls `app.run(single_process=True)`, and `serve_single()` starts every registered app — so the Inspector process can accept HTTP on the application port.

Before Inspector serves, it now closes those inherited listener FDs (the parent still holds its own copies) and unregisters the inherited apps so only the Inspector app is served.

## Test plan

- [x] `pytest tests/worker/test_inspector.py` (22 passed)
- [ ] Manual: `Sanic.start_method = "fork"` + `inspector=True`, confirm `/test` is always handled by a worker PID, never the Inspector PID

### DIFF
--- a/sanic/pages/error.py
+++ b/sanic/pages/error.py
@@ -1,5 +1,5 @@
 from collections.abc import Mapping
-from typing import Any
+from typing import Any, cast
 
 import tracerite.html
 
@@ -25,7 +25,7 @@ for the inconvenience and appreciate your patience.\
 class ErrorPage(BasePage):
     """Page for displaying an error."""
 
-    STYLE_APPEND = tracerite.html.style
+    STYLE_APPEND: str = cast(str, tracerite.html.style)
 
     def __init__(
         self,
--- a/sanic/pages/error.py
+++ b/sanic/pages/error.py
@@ -12,7 +12,7 @@ from .base import BasePage
 
 
 # Avoid showing the request in the traceback variable inspectors
-inspector.blacklist_types += (Request,)  # type: ignore[assignment]  # tracerite types this as a fixed-length tuple
+inspector.blacklist_types += (Request,)
 
 ENDUSER_TEXT = """\
 We're sorry, but it looks like something went wrong. Please try refreshing \
--- a/sanic/pages/error.py
+++ b/sanic/pages/error.py
@@ -1,5 +1,5 @@
 from collections.abc import Mapping
-from typing import Any, cast
+from typing import Any
 
 import tracerite.html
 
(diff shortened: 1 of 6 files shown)


sanic/pages/error.py
Tronic, line 1: I was unable to reproduce. Can you confirm if this happens with tracerite 2.6.4 (latest release from 3 days back that changed tuple/struct handling) or also with 2.6.3? It would also be helpful to see what the variable inspector prints exactly, and perhaps a test case if you have one. Likely the correct solution is to fix this in tracerite because Request object really should not register as tuple or struct type.

Tronic: Oh, figure it out: a typing issue from implicit typing of blacklists. Fixed now, upgrade your tracerite to 2.6.5 and remove this typing ignore. One might expect the blacklist types to be a set or list, but isinstance only takes tuple, therefore the odd type that implicit typing got wrong.

Issue #3192, opened 2026-09-18
Request.form and Request.get_form() declare return type that includes None yet cannot return None


davidfstr:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

After upgrading from Sanic 22.12.0 -> 24.12.1, code like the following has a typechecker error under mypy:

```python
@app.post('/code/turn_in_web')
async def code_turn_in_web(request: Request) -> HTTPResponse:
    # ERROR: Item "None" of "RequestParameters | None" has no attribute "get"  [union-attr]
    user_project_id = int(request.form.get('user_project_pk'))
    ...
```

Cause is that `Request.form` (and the related `Request.get_form()`) are declared as potentially returning None:

```python
# sanic/request/types.py
class Request(Generic[sanic_type, ctx_type]):
    ...
    def get_form(
        self, keep_blank_values: bool = False
    ) -> Optional[RequestParameters]:  # 👈 is Optional!
        ...

    @property
    def form(self) -> Optional[RequestParameters]:  # 👈 is Optional!
        if self.parsed_form is None:
            self.get_form()
        return self.parsed_form
```

However if you read the code of `get_form()` you'll see that it will never return None. At worst it will return an empty `RequestParameters()`. 

Similarly if you read the `form` property and recognize that `get_form()` always sets `self.parsed_form` to a non-None value, you'll see that it will never return None.

Issue persists in the latest Sanic v25.12.1 (published 2026-05-31).

### Code snippet

Code that typechecks clean under Sanic 22 but not Sanic 25:

```python
from sanic import Sanic
from sanic.request import Request
from sanic.response import HTTPResponse

app = Sanic(name='http_proxy')

@app.post('/code/turn_in_web')
async def code_turn_in_web(request: Request) -> HTTPResponse:
    # ERROR: Item "None" of "RequestParameters | None" has no attribute "get"  [union-attr]
    user_project_id = int(request.form.get('user_project_pk'))
    ...
```

### Expected Behavior

Typechecks clean.

### How do you run Sanic?

As a script (`app.run` or `Sanic.serve`)

### Operating System

Linux

### Sanic Version

25.12.1

### Additional context

Recommended fix: Alter type declarations of these two `def`s to be `RequestParameters` rather than `Optional[RequestParameters]`.

I can draft a PR if it would be helpful. Let me know.

sachinbisht1:
[...]

PR #3193, opened 2026-09-18
fix: correct request form return types

Luke-salinero:
## Summary

- Correct the public return types for `Request.get_form()` and `Request.form` to `RequestParameters`.
- Add contract tests for both annotations.

## Validation

- `python -m pytest tests/test_request.py tests/test_requests.py -q` (179 passed)
- `python -m mypy --follow-imports=silent --no-incremental -c <focused caller>`
- `python -m mypy --follow-imports=skip --no-incremental sanic/request/types.py`
- `python -m ruff check sanic/request/types.py tests/test_request.py`
- `python -m ruff format --check sanic/request/types.py tests/test_request.py`

The focused caller and contract tests fail with the previous `RequestParameters | None` annotations and pass with this change.

## Scope note

This corrects the type contract for the form container itself: `get_form()` initializes and returns `RequestParameters`, and `form` invokes it when needed. `RequestParameters.get()` intentionally remains nullable because an individual form field can be absent. Callers using `int(request.form.get(...))` should still supply a default or handle a missing value; this PR does not change that runtime behavior.

Related to #3192.


tests/test_request.py
```diff
@@ -8,9 +9,22 @@
...
+def test_request_get_form_return_type():
+    annotations = get_type_hints(Request.get_form)
+
+    assert annotations["return"] is RequestParameters
+
+
+def test_request_form_return_type():
+    annotations = get_type_hints(Request.form.fget)
+
+    assert annotations["return"] is RequestParameters
```
davidfstr, line 25: These tests seem low value. Suggest deleting.

davidfstr (commented):
Implementation seems fine. I left a comment on the tests.

CI seems to be failing. PR contributor to investigate why.

